PR #41, opened 2022-02-04
[DRAFT] Initial proposal for an abstraction of ATE/OTG

robshakir:
This draft PR providea a discussion point (WIP) for the way that we might implement an abstraction specifically for WBB tests that allows for use of both the ATE and OTG APIs whilst we wait for there to be common support for OTG.

At the moment, this is a FYI PR, but as we answer some of the questions (e.g., what the future of `ondatra.ATETopology` is, I'll firm something up into a proposal here.

One large open question is how we want to test this code, to some extent it's going to be tested by running the tests - and tests of this code will generally be assertions that ATE and gosnappi work as expected, so I'm erring towards not writing a lot of change detector tests, and only adding tests where there's real logic in this package.

```
 * (A) absate/absate.go
   - Initial wrapper for simple parts of the ATE API to demonstrate how
     we might run tests against ATE and OTG APIs.
 * (M) go.(mod|sum)
   - Add dependency on gosnappi.
```

### DIFF
--- a/absate/absate.go
+++ b/absate/absate.go
@@ -1,0 +1,162 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


robshakir:
Closing since I think this is now obsolete with other work going on with OTG.

PR #47, opened 2022-02-10
Device config library (MVP)

sthesayi:
This PR contains a minimum device config library that is a dependency for the other feature profile config library. This PR must be merged last in the chain of PRs on config library. Please ignore static analysis errors for now.


feature/device/device_test.go
```diff
@@ -113,10 +111,6 @@ func TestFullReplaceRequest(t *testing.T) {
 		name: "device with basic LLDP and BGP",
```
wenovus, line 111: Remove LLDP from desc?

wenovus (approved):

feature/device/device_test.go
sthesayi: sigh. I will fix in the next PR. Every time I push a change, the LGTMs are removed.

PR #61, opened 2022-02-23
Initial featureprofile for gNMI's `Get` RPC

robshakir:
```
    Initial (basic) test for gNMI Get RPC.
    
     * (M) proto/feature.proto
        - Add support for specifying a method of the gNMI protocol that
          coverage is being implemented for.
     * (M) feature/system/gnmi/feature.textproto
        - Add feature profile for basic gNMI.
     * (M) feature/system/gnmi/get/...
        - Add feature profile for gNMI Get.
        - Add initial functional tests for gNMI Get against the root.
```

### DIFF
--- a/cloudbuild/arista.sh
+++ b/cloudbuild/arista.sh
@@ -28,6 +28,6 @@ password: admin
 topology: ${PWD}/topologies/kne/arista_ceos.textproto
 cli: ${HOME}/go/bin/kne_cli
 EOF
-go test -v feature/system/tests/*.go -config "$PWD"/topologies/kne/testbed.kne.yml -testbed "$PWD"/topologies/dut.testbed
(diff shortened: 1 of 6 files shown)


robshakir:
Note that this fails currently, but is added in as a check to discuss a possible difference in spec interpretation.

robshakir:
This test doesn't pass against the existing targets today, but I'd like to check it in anyway. PTAL.

there's some discussion on the gNMI design for `Get`, I propose to update the test once we've figured that out, since this encodes our current expectations.

bstoll (approved):

sthesayi (approved):

robshakir:
Thanks folks -- merging this in. I'm going to start a few `Subscribe` tests too to give a basis for things that we're discovering need to be tested.

liulk:
Resolved #1367

PR #85, opened 2022-02-24
Update missing paths from go/wbb-oc-paths to ISIS, Interface and BGP …

hariramnat:
…Feature profiles

### DIFF
--- a/feature/bgp/feature.textproto
+++ b/feature/bgp/feature.textproto
@@ -417,6 +417,19 @@ telemetry_path {
   path: "/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/timers/state/keepalive-interval"
 }
 
+# Neighbor state
+telemetry_path {
+    path: "/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/state/last-established"
+}
+
+# Transport
+telemetry_path {
+    path: "/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/transport/state/remote-address"
+}
+telemetry_path {
+    path: "/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/transport/state/remote-port"
+}
+
 gnoi_service {
   service_name: "gnoi.bgp.BGP"
   method_name: "ClearBGPNeighbor"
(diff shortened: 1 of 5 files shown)


dplore (approved):

feature/bgp/feature.textproto
```diff
@@ -17,6 +17,20 @@ id {
   version: 1
 }
 
+# Network Instance config
```
sthesayi, line 20: These paths already existing in NI base. We should just create a dependency on NI base rather than duplicating the paths.

feature/bgp/feature.textproto
```diff
@@ -417,6 +431,19 @@ telemetry_path {
   path: "/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/timers/state/keepalive-interval"
 }
 
+# Neighbor state
+telemetry_path {
```
sthesayi, line 435: We should keep the telemetry paths close to the config path

feature/experimental/lacp_and_base_interface/singleton_interface/feature.textproto
```diff
@@ -152,3 +152,40 @@ telemetry_path {
 telemetry_path {
   path: "/interfaces/interface/subinterfaces/subinterface/ipv6/state/counters/out-pkts"
 }
+telemetry_path {
```
sthesayi, line 155: We should reorder these to keep the config and telemetry paths close and in this order.

feature/isis/feature.textproto
```diff
@@ -60,6 +60,9 @@ telemetry_path {
 }
 
 # Interfaces enabled
+config_path {
```
sthesayi, line 63: how about the telemetry path for this?

feature/isis/feature.textproto
```diff
@@ -157,3 +160,120 @@ telemetry_path {
 telemetry_path {
   path: "/network-instances/network-instance/protocols/protocol/isis/interfaces/interface/levels/level/packet-counters/psnp/state/retransmit"
 }
+
+# Adjacencies
+telemetry_path {
```
sthesayi, line 165: We should keep the state paths close to the config paths

feature/system/feature.textproto
```diff
@@ -59,6 +59,13 @@ telemetry_path {
     path: "/system/clock/state/timezone-name"
 }
 
+config_path {
```
sthesayi, line 62: what about the state paths?

PR #93, opened 2022-03-09
Update bgp restart-timer to match current OpenConfig

rcdomigan:
A [OpenConfig PR](https://github.com/openconfig/public/pull/532) moved
various bgp `prefix-limits/(conf|state)/restart-timers` to
`timers/(conf|state)/restart-time`

### DIFF
--- a/feature/bgp/feature.textproto
+++ b/feature/bgp/feature.textproto
@@ -247,9 +247,6 @@ config_path {
 config_path {
   path: "/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/afi-safis/afi-safi/ipv4-unicast/prefix-limit/config/warning-threshold-pct"
 }
-config_path {
-  path: "/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/afi-safis/afi-safi/ipv4-unicast/prefix-limit/config/restart-timer"
-}
 telemetry_path {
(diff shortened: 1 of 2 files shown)


dplore:
I wanted to note that it may be some time before devices implement this breaking path change.  Since we have tests that touch these leaves, should we include some deviation?  It seems we can choose to allow these tests to break due to lack of compliance with v8.0.0 of the bgp model (no deviations), or we can do something else, such as test 7.0.0 and 8.0.0, where the 7.0.0 version is a deviation and 8.0.0 is the non-deviated test.

Any comments or other options we should consider?

rcdomigan:
We have a failing presubmit, so I'd like to get some fix submitted. Nothing really consumes the paths at this point, so I don't think we need to solve versioning to get this submitted.

It looks like the neighbor.go helper sets this field, but that uses a pinned version of OpenConfig.

None of the tests are currently exercising the restart timers.

dplore:
Ok, thanks for the comment.  Rerunning the featureprofiles-pr check.  If that passes, I'll approve

dplore:
I am not familiar with the check (featureprofiles-pr (disco-idea-817) ) that is failing, is this expected @rcdomigan @bstoll ?

rcdomigan:
I don't have permission to view the featureprofiles-pr page.

rcdomigan:
The check is also failing on the [style-guide PR](https://github.com/openconfig/featureprofiles/pull/94), so I don't think it has much to do with this change.

rcdomigan:
Tests are apparently timing out and @bstoll is taking a look at them. I think this PR should be OK to merge.

PR #94, opened 2022-03-10
Add style guide to contribution guide

dplore:


CONTRIBUTING.md
```diff
@@ -29,30 +31,138 @@ and maintained by Google](https://opensource.google.com/docs/cla/#why), not to
...
+* `feature/` contains definition and tests of feature profiles.
+* `internal/` contains packages used by feature profile tests.
+* `proto/`  contains protobuf files for feature profiles.
+* `tools/` contains code used for CI checks.
+* `topologies/` contains the testbed topology definitions.
+* `yang/` contains autogenerated go structures.
+
+### Test Suite Organization
+
+Under `feature/<featurename>/[<sub-feature>/]tests`, each feature group has its own
```
liulk, line 80: If there are multiple tests for a given subfeature, we should put them into separate subdirectories, otherwise they would belong to the same package and would execute together. Each test can already take 30+ minutes.

So for example, it should be:

* `feature/interfaces/tests/rt_5_1_singleton_test/main_test.go`
* `feature/interfaces/tests/rt_5_2_aggregate_test/main_test.go`

(Test files need to end with `_test.go` otherwise `go test` would not recognize them.)

CONTRIBUTING.md
```diff
@@ -29,30 +31,138 @@ and maintained by Google](https://opensource.google.com/docs/cla/#why), not to
...
+* `feature/lldp/` contains feature profiles for ISIS.
+* `feature/lldp/tests` contains test suite related to LLDP.
+* `internal/deviations` contains code which overrides standard tests where
+    there are known bugs.
+
+Under each feature group, the `*.go` file should be named after an appropriate
+[github featureprofiles project](https://github.com/orgs/openconfig/projects/2/views/1?filterQuery=)
+item. For example:
+
+* `tests/interfaces/rt_5_1_singleton_test.go` implements issue **RT-5.1
```
liulk, line 96: The path should be consistent with the example above.

CONTRIBUTING.md
```diff
@@ -29,30 +31,138 @@ and maintained by Google](https://opensource.google.com/docs/cla/#why), not to
...
+Netblocks used in the test topology should follow IPv4 Address Blocks Reserved
+for Documentation ([RFC 5737]) and IPv6 Address Prefix Reserved for
+Documentation ([RFC 3849]). In particular:
+
+[RFC 5737]: https://datatracker.ietf.org/doc/html/rfc5737
+[RFC 3849]: https://datatracker.ietf.org/doc/html/rfc3849
+
+IPv4:
+
+* `TEST-NET-1`: 192.0.2.0/24
```
liulk, line 143: I propose that we further restrict the assignment to be:

> * `TEST-NET-1` (192.0.2.0/24): control plane addresses split into /30 subnets for each ATE/DUT port pair.
> * `TEST-NET-2` (198.51.100.0/24): data plane source network addresses used for traffic testing; split as needed.
> * `TEST-NET-3` (203.0.113.0/24): data plane destination network addresses used for traffic testing; split as needed.

This is already de-facto in many tests. The rest of the tests will also follow this after they are converted to use the topology helper.

CONTRIBUTING.md
```diff
@@ -29,30 +31,138 @@ and maintained by Google](https://opensource.google.com/docs/cla/#why), not to
...
+* `TEST-NET-2`: 198.51.100.0/24
+* `TEST-NET-3`: 203.0.113.0/24
+
+IPv6:
+
+* 2001:DB8::/32
```
liulk, line 149: Proposed edit:

> 2001:DB8::/32 is a very large space, so we divide them as follows.
>
> * 2001:DB8:0::/64: control plane addresses split into /126 subnets for each ATE/DUT port pair.
> * 2001:DB8:1::/64: data plane addresses used for traffic testing as the source address; split as needed.
> * 2001:DB8:2::/64: data plane addresses used for traffic testing as the destination address; split as needed.

CONTRIBUTING.md
```diff
@@ -29,30 +31,138 @@ and maintained by Google](https://opensource.google.com/docs/cla/#why), not to
...
+
+* 16-bit ASN: 64496 - 64511 (`0xfbf0` - `0xfbff`)
+* 32-bit ASN: 65536 - 65551 (`0x10000` - `0x1000f`)
+
+Both ranges have 16 total numbers each. The hexadecimal notation makes it more
+obvious where the range starts and stops.
+
+## Code Style
+
+All code should follow [Go language style guide](https://github.com/golang/go/wiki/CodeReviewComments) for writing readable Go code.
```
liulk, line 168: `and [Effective Go](https://go.dev/doc/effective_go)`

liulk (changes requested):

CONTRIBUTING.md
```diff
@@ -29,30 +31,138 @@ and maintained by Google](https://opensource.google.com/docs/cla/#why), not to
...
+* `internal/deviations` contains code which overrides standard tests where
+    there are known bugs.
+
+Under each feature group, the `*.go` file should be named after an appropriate
+[github featureprofiles project](https://github.com/orgs/openconfig/projects/2/views/1?filterQuery=)
+item. For example:
+
+* `tests/interfaces/rt_5_1_singleton_test.go` implements issue **RT-5.1
+  Singleton Interface**.
+* `tests/interfaces/rt_5_2_aggregate_test.go` implements issue **RT-5.2
```
bstoll, line 98: I don't think `interfaces` is specific enough for this example.  `tests/aggregate_interfaces/aggregate_interfaces_test.go` makes more sense to me.  The full path would be something like `feature/interfaces/tests/aggregate_interfaces/aggregate_interfaces_test.go`.

CONTRIBUTING.md
```diff
@@ -29,30 +31,138 @@ and maintained by Google](https://opensource.google.com/docs/cla/#why), not to
...
+* `cloudbuild/` contains google cloud build scripts for running virtual
+    routers in containers on [KNE](https://github.com/google/kne)
+* `feature/` contains definition and tests of feature profiles.
+* `internal/` contains packages used by feature profile tests.
+* `proto/`  contains protobuf files for feature profiles.
+* `tools/` contains code used for CI checks.
+* `topologies/` contains the testbed topology definitions.
+* `yang/` contains autogenerated go structures.
+
+### Test Suite Organization
```
bstoll, line 78: I think there should be some references to the `README.md` around this section too.

`feature/interfaces/README.md` contains a description of the BGP feature profile.

`feature/interfaces/tests/singleton_interface/README.md` explains the singleton_interface test of the interfaces feature profile.

CONTRIBUTING.md
bstoll: Also just noticed comment from @sthesayi in another PR - the examples should be singular and not plural.

dplore: Let me know what you think of the latest update.  Do you feel strongly about dropping the test suite identifier (ie: rt_5_1) from the test suite name?

CONTRIBUTING.md
```diff
@@ -29,30 +31,140 @@ and maintained by Google](https://opensource.google.com/docs/cla/#why), not to
...
+* `feature/bgp/addpath/README.md` documents the BGP feature profile.
+* `feature/bgp/addpath/tests` contains test code related to BGP Add Path features.
+* `internal/deviations` contains code which overrides standard tests where
+    there are known issues in a DUT.
+
+Within each test directory, a README.md should document the test suite.  The `*.go` files should be named after an appropriate
+[github featureprofiles project](https://github.com/orgs/openconfig/projects/2/views/1?filterQuery=)
+item. For example:
+
+* `feature/interfaces/tests/rt_5_1_singleton_test/README.md` - documentation for the test suite
```
liulk, line 96: There also seems to be a consensus that I heard that the XX-1.2 designations are going to be removed, so I removed them. e.g. `feature/interfaces/ate_tests/singleton_test`.

CONTRIBUTING.md
```diff
@@ -29,30 +31,140 @@ and maintained by Google](https://opensource.google.com/docs/cla/#why), not to
...
+* `topologies/` contains the testbed topology definitions.
+* `yang/` contains autogenerated go structures.
+
+### Test Suite Organization
+
+Test suites should be placed in subdirectories formatted like `feature/<featurename>/[<sub-feature>/]tests/<testname>/main_test.go`, for example:
+
+* `feature/interfaces/` contains feature profiles for interfaces.
+* `feature/interfaces/README.md` - documents the interface feature profile.
+* `feature/interfaces/tests/` contains the interfaces test suite.
```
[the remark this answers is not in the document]
dplore: Will we have separate tests for ate and otg?  or just one test which has a conditional to select traffic generation based on topology?   Maybe it is not known yet.

CONTRIBUTING.md
dplore: Ok, will leave XX-#.# the identifiers  for the moment.

CONTRIBUTING.md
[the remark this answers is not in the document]
dplore: Ok, made suggested update

liulk (approved):

PR #106, opened 2022-03-23
gribi helpers

mojiiba:
This pull requests adds helpers (initial version) for gribi to simplify writing gribi test cases. Please let me know if you find the helpers useful and you are willing to merge the PR when it is completed.


feature/experimental/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return g
+}
+
+func (g *GRIBIHandler) Close(t testing.TB) {
+	t.Helper()
+	t.Logf("closing GRIBI connection for dut: %s", g.dut.Name())
+	g.fluentC.Stop(t)
+	v := reflect.Indirect(reflect.ValueOf(g.gribiC)).FieldByName("cc")
+	clientConn, ok := (reflect.NewAt(v.Type(), unsafe.Pointer(v.UnsafeAddr())).Elem().Interface()).(*grpc.ClientConn)
```
liulk, line 73: I see what the problem is. The Ondatra API only gives us the client stub, and it is not designed to close the underlying connection if it's new. For now, let's use the Default() stub, but consider filing an issue with Ondatra to rectify the RawAPIs design.

Accessing unexported fields using reflection is problematic.

feature/experimental/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+	mu.Lock()
+	defer mu.Unlock()
+	g.lastElectionID = results[len(results)-1].CurrentServerElectionID
+}
+
+func (g *GRIBIHandler) GetLastElectionID(t *testing.T) (low, high uint64) {
+	return g.lastElectionID.Low, g.lastElectionID.High
+}
+
+func (g *GRIBIHandler) BeMaster(t *testing.T) {
```
liulk, line 127: Suggested new name: `BecomeLeader`

feature/experimental/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	results := g.fluentC.Results(t)
+	mu.Lock()
+	defer mu.Unlock()
+	g.lastElectionID = results[len(results)-1].CurrentServerElectionID
+	return g.lastElectionID.Low, g.lastElectionID.High
+}
+
+// set the election id
+func (g *GRIBIHandler) UpdateElectionID(t *testing.T, lowElecId, highElecId uint64) {
```
liulk, line 104: `t *testing.T` should be `t testing.TB` consistently throughout.

liulk (changes requested):

liulk:
Overall I think this helper is going to be really useful. We have a few tests that could be migrated to use this helper also. Here is one that we open sourced recently: https://github.com/openconfig/featureprofiles/tree/main/feature/gribi/ate_tests/base_leader_election_test

mojiiba:
@liulk Thank you so much for the comments. Sure, I will address the comment and migrate the published test case  to use the new helpers.

liulk:
I just remembered that some of the tests prefer to batch each route's NH, NHG, and IPv4Entry into one request with the modify-await-chk, and these tests might not gain very much from this particular helper.

* Route 1: NH1, NHG1, IPv4E1
* Route 2: NH2, NHG2, IPv4E2
* etc.

In other words, the batching in those tests is done in row order.

But this particular helper batches NH1, NH2, etc. into a batch for modify-await-chk, then NHG1, NHG2, etc. as a batch, then IPv4E1, IPv4E2, etc. as a batch. The batching is done in columnar order.

Use judgment whether a test will benefit from this helper before trying to migrate them.

mojiiba:
> 

How about having a separate function (AddRoute) to cover the batch cases as well?

liulk:
That could work!

feature/experimental/gribi/gribi.go
mojiiba: done

feature/experimental/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+	lowElecId, highElecId := g.LearnElectionID(t)
+	if lowElecId == maxUint64 {
+		highElecId = highElecId + 1
+	} else {
+		lowElecId = lowElecId + 1
+	}
+	g.UpdateElectionID(t, lowElecId, highElecId)
+}
+
+func (g *GRIBIHandler) AddNHG(t *testing.T, nhgIndex uint64, nhs map[uint64]uint64, instance string) {
```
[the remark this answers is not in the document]
mojiiba: done

feature/experimental/gribi/gribi.go
mojiiba: done

feature/experimental/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+}
+
+func (g *GRIBIHandler) awaitTimeout(ctx context.Context, t testing.TB, timeout time.Duration) error {
+	subctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	return g.fluentC.Await(subctx, t)
+}
+
+// get the election id (last known election id) from a dut
```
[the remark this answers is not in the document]
mojiiba: done

feature/experimental/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+	v := reflect.Indirect(reflect.ValueOf(g.gribiC)).FieldByName("cc")
+	clientConn, ok := (reflect.NewAt(v.Type(), unsafe.Pointer(v.UnsafeAddr())).Elem().Interface()).(*grpc.ClientConn)
+	if ok {
+		clientConn.Close()
+	} else {
+		t.Fatalf("can not close gribi connection")
+	}
+}
+
+func (g *GRIBIHandler) awaitTimeout(ctx context.Context, t testing.TB, timeout time.Duration) error {
```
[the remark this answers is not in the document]
mojiiba: done

feature/experimental/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
+package gribi
```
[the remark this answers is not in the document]
mojiiba: done

feature/experimental/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+	fluentC        *fluent.GRIBIClient
+	fibAck         bool
+	persistence    bool
+	redundancyMode fluent.RedundancyMode
+	lastElectionID *grbp.Uint128
+}
+
+var mu sync.Mutex
+
+func NewGRIBIFluent(t testing.TB, dut *ondatra.DUTDevice, persistance, fibAck bool) *GRIBIHandler {
```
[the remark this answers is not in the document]
mojiiba: done

feature/experimental/gribi/gribi.go
mojiiba: @liulk  using  Default() is not going to help here since Ondatra does not provide close for default gribi client as well.
I have discussed this with @greg-dennis (here)[https://github.com/openconfig/ondatra/pull/3] and he suggested using reflection. Anyway, If you agree we have this code here until we finalize the solution for this.

mojiiba:
@liulk We have addressed your comments and migrated the public gribi test cases to use the helpers. Please see the attached log for cisco XR. 

I will add other apis for adding route, flush, delete, etc in the next PR.

[test-cisco-log.txt](https://github.com/openconfig/featureprofiles/files/8367167/test-cisco-log.txt)

feature/experimental/gribi/gribi.go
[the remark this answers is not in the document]
greg-dennis: How is this all different from
https://github.com/openconfig/gribigo/blob/main/fluent/fluent.go
?

I'm not sure we should have two fluent gribi APIs in openconfig.

feature/experimental/gribi/gribi.go
greg-dennis: Yeah, I suggested getting the connection field by reflection because the need to close the connection seemed like a rather rare requirement, making it hard to justify the additional weight and complexity on the binding that would be needed to support it. We also have no resources to design the changes to accommodate the use case at this time, and so far this is the one and only request for that feature. I don't love it, but I also don't see the generated proto code changing in a way to break it any time soon.

Two suggestions:
(1) I suspect "reflect.Indirect(reflect.ValueOf(g.gribiC))" would be better as "reflect.ValueOf(g.gribiC).Elem()"
(2) Does every test really need to close the connection? I thought it was a fairly narrow test case that actually needed the connection closed. Might be better to confine this to just that test case rather than making it part of a common library. The above-mentioned gribigo/fluent library has found no need to ever close the connection.

liulk: You can definitely reuse the gRIBI client stub over many fluent clients. The abstraction that defines a gRIBI session (election ID, persistence) is contained entirely within the Modify streaming RPC.

Hence, `RawAPIs().GRIBI().Default()` is the better way. There is no need to close this default transport since it's just in the background and indefinitely reusable.

feature/experimental/gribi/gribi.go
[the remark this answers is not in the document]
liulk: I think the value proposition of this package is that it streamlines modify-await-chk that you would otherwise have to do separately using the fluent client. The modify and chk have to agree on the entries being programmed, and this package provides a way to batch program these entries. This is a more friendly way to write gRIBI tests.

feature/experimental/gribi/gribi.go
mojiiba: Thanks for the clarification!  I changed the code as you suggested. There is no need to use reflection.

mojiiba:
@liulk Please let me know if you need anything else to be addressed? Please note that I do not have access to disco-idea-817 project, so I can not run the build and pass the related check.

liulk:
I think I will need more time to look at the TE-4.1 changes which look non-trivial. If you can split it to a separate PR, that would unblock this PR.

I'll also be doing a more thorough Go style review of the helper later today.

internal/gribi/gribi.go
```diff
@@ -0,0 +1,203 @@
...
+			WithNextHopOperation(nhIndex).
+			WithOperationType(constants.Add).
+			WithProgrammingResult(expectedResult).
+			AsResult(),
+		chk.IgnoreOperationID(),
+	)
+}
+
+// AddNH adds an IPV4 entry
+func (g *GRIBIHandler) AddIPV4Entry(t testing.TB, nhgIndex uint64, nhgInstance, prefix, instance string, expectedResult fluent.ProgrammingResult) {
```
liulk, line 183: Rename function to AddIPv4 to be consistent with the naming pattern above? They all omitted Entry.

liulk, line 183: Reorder the arguments so the mapping relationship is more clear?

```
func (g *GRIBIHandler) AddIPv4(t testing.TB, prefix string, nhgIndex uint64, instance, nhgInstance string, expectedResult fluent.ProgrammingResult)
```

liulk (changes requested):

internal/gribi/gribi.go
```diff
@@ -0,0 +1,203 @@
...
+func (g *GRIBIHandler) LearnElectionID(t testing.TB) (low, high uint64) {
+	t.Helper()
+	t.Logf("learn GRIBI Election ID from dut: %s", g.dut.Name())
+	g.fluentC.Modify().UpdateElectionID(t, g.lastElectionID.Low, g.lastElectionID.High)
+	err := g.AwaitTimeout(context.Background(), t, timeout)
+	if err != nil {
+		t.Fatalf("learnElectionID Error: %v", err)
+	}
+	results := g.fluentC.Results(t)
+	mu.Lock()
```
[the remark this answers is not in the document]
mojiiba: Thanks for pointing this out. I was trying to use lock because of election id.  However, I noticed the gribigo client already saves the last known id, so I removed election id from the struct and the locking as well.

internal/gribi/gribi.go
mojiiba: done

mojiiba: done

internal/gribi/gribi.go
```diff
@@ -0,0 +1,203 @@
...
+		fluent.OperationResult().
+			WithNextHopOperation(nhIndex).
+			WithOperationType(constants.Add).
+			WithProgrammingResult(expectedResult).
+			AsResult(),
+		chk.IgnoreOperationID(),
+	)
+}
+
+// AddNH adds an IPV4 entry
```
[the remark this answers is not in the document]
mojiiba: done

internal/gribi/gribi.go
```diff
@@ -0,0 +1,203 @@
...
+	lowElecId, highElecId := g.LearnElectionID(t)
+	if lowElecId == maxUint64 {
+		highElecId = highElecId + 1
+	} else {
+		lowElecId = lowElecId + 1
+	}
+	g.UpdateElectionID(t, lowElecId, highElecId)
+}
+
+// AddNHG adds a NextHOP group entry
```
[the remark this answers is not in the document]
mojiiba: done

internal/gribi/gribi.go
```diff
@@ -0,0 +1,203 @@
...
+func (g *GRIBIHandler) AddNHG(t testing.TB, nhgIndex uint64, nhWeights map[uint64]uint64, instance string, expectedResult fluent.ProgrammingResult) {
+	NGH := fluent.NextHopGroupEntry().WithNetworkInstance(instance).WithID(nhgIndex)
```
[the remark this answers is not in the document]
mojiiba: done

internal/gribi/gribi.go
```diff
@@ -0,0 +1,203 @@
...
+		fluentClient.Connection().WithInitialElectionID(g.lastElectionID.Low, g.lastElectionID.High).WithRedundancyMode(fluent.ElectedPrimaryClient)
+	}
+	if fibACK {
+		fluentClient.Connection().WithFIBACK()
+	}
+	fluentClient.Start(context.Background(), t)
+	fluentClient.StartSending(context.Background(), t)
+	err := g.AwaitTimeout(context.Background(), t, timeout)
+	if err != nil {
+		t.Fatalf("can not establish the session: %v", err)
```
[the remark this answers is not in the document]
mojiiba: changed as you suggested. Thanks!

internal/gribi/gribi.go
```diff
@@ -0,0 +1,203 @@
+package gribi
```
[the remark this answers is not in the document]
mojiiba: added the comment

internal/gribi/gribi.go
```diff
@@ -0,0 +1,191 @@
...
+
+// Fluent resturns the fluent client that can be used to directly call the gribi fluent APIs
+func (g *GRIBIHandler) Fluent(t testing.TB) *fluent.GRIBIClient {
+	return g.fluentC
+}
+
+// Start function start establish a client connection with the gribi server.
+// By default the client is not the leader and for that function BecomeLeader
+// needs to be called.
+
```
liulk, line 41: Remove empty line?

internal/gribi/gribi.go
```diff
@@ -0,0 +1,191 @@
...
+		highElecId = highElecId + 1
+	} else {
+		lowElecId = lowElecId + 1
+	}
+	g.UpdateElectionID(t, lowElecId, highElecId)
+}
+
+// AddNHG adds a NextHopGroupEntry with a given index, and a map of next hop entry indices to the weights,
+// in a given network instance.
+func (g *GRIBIHandler) AddNHG(t testing.TB, nhgIndex uint64, nhWeights map[uint64]uint64, instance string,
```
liulk, line 126: Avoid line wrap?

https://github.com/golang/go/wiki/CodeReviewComments#line-length

internal/gribi/gribi.go
```diff
@@ -0,0 +1,191 @@
...
+			WithNextHopGroupOperation(nhgIndex).
+			WithOperationType(constants.Add).
+			WithProgrammingResult(expectedResult).
+			AsResult(),
+		chk.IgnoreOperationID(),
+	)
+}
+
+// AddNH adds a NextHopEntry with a given index to an address within a given network instance.
+func (g *GRIBIHandler) AddNH(t testing.TB, nhIndex uint64, address, instance string,
```
liulk, line 148: Avoid line wrap?

https://github.com/golang/go/wiki/CodeReviewComments#line-length

internal/gribi/gribi.go
```diff
@@ -0,0 +1,191 @@
...
+// in a given network instance.
+func (g *GRIBIHandler) AddNHG(t testing.TB, nhgIndex uint64, nhWeights map[uint64]uint64, instance string,
+	expectedResult fluent.ProgrammingResult) {
+	nhg := fluent.NextHopGroupEntry().WithNetworkInstance(instance).WithID(nhgIndex)
+	for nhIndex, weight := range nhWeights {
+		nhg.AddNextHop(nhIndex, weight)
+	}
+	g.fluentC.Modify().AddEntry(t, nhg)
+	if err := g.AwaitTimeout(context.Background(), t, timeout); err != nil {
+		t.Fatalf("got unexpected error from server adding a NHG, got: %v, want: nil", err)
```
liulk, line 134: Reporting error doesn't need "want: nil"

```
t.Fatalf("Error waiting to add NHG: %v", err)
```

internal/gribi/gribi.go
```diff
@@ -0,0 +1,191 @@
...
+func (g *GRIBIHandler) AddNH(t testing.TB, nhIndex uint64, address, instance string,
+	expectedResult fluent.ProgrammingResult) {
+	g.fluentC.Modify().AddEntry(t,
+		fluent.NextHopEntry().
+			WithNetworkInstance(instance).
+			WithIndex(nhIndex).
+			WithIPAddress(address))
+
+	if err := g.AwaitTimeout(context.Background(), t, timeout); err != nil {
+		t.Fatalf("got unexpected error from server adding NH, got: %v, want: nil", err)
```
liulk, line 157:
```
t.Fatalf("Error waiting to add NH: %v", err)
```

internal/gribi/gribi.go
```diff
@@ -0,0 +1,191 @@
...
+	expectedResult fluent.ProgrammingResult) {
+	ipv4Entry := fluent.IPv4Entry().WithPrefix(prefix).
+		WithNetworkInstance(instance).
+		WithNextHopGroup(nhgIndex)
+	if nhgInstance != "" && nhgInstance != instance {
+		ipv4Entry.WithNextHopGroupNetworkInstance(nhgInstance)
+	}
+	g.fluentC.Modify().AddEntry(t, ipv4Entry)
+	if err := g.AwaitTimeout(context.Background(), t, timeout); err != nil {
+		t.Fatalf("got unexpected error from server adding NH, got: %v, want: nil", err)
```
liulk, line 180:
```
t.Fatalf("Error waiting to add IPv4: %v", err)
```

liulk (changes requested):

internal/gribi/gribi.go
mojiiba: done

internal/gribi/gribi.go
mojiiba: done

internal/gribi/gribi.go
mojiiba: done

internal/gribi/gribi.go
mojiiba: done

internal/gribi/gribi.go
mojiiba: done

internal/gribi/gribi.go
```diff
@@ -0,0 +1,188 @@
...
+		fluent.OperationResult().
+			WithCurrentServerElectionID(lowElecId, highElecId).
+			AsResult(),
+	)
+}
+
+// BecomeLeader learns the latest election id and the make the client leader by increasing the election id by one.
+func (g *GRIBIHandler) BecomeLeader(t testing.TB) {
+	t.Logf("trying to be a master with increasing the election id by one on dut: %s", g.DUT.Name())
+	lowElecId, highElecId := g.learnElectionID(t)
```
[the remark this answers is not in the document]
mojiiba: done

internal/gribi/gribi.go
```diff
@@ -0,0 +1,188 @@
...
+	if lowElecId == maxUint64 {
```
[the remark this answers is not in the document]
mojiiba: Thanks! Done

internal/gribi/gribi.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/openconfig/gribigo/fluent"
+	"github.com/openconfig/ondatra"
+)
+
+const (
+	timeout   = time.Minute
+	maxUint64 = 1<<64 - 1
+)
+
+// GRIBIHandler provides access to GRIBI APIs of the DUT.
```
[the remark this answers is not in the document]
mojiiba: done

internal/gribi/gribi.go
```diff
@@ -0,0 +1,188 @@
...
+type GRIBIHandler struct {
+	DUT         *ondatra.DUTDevice
+	FibACK      bool
+	Persistence bool
+	fluentC     *fluent.GRIBIClient
```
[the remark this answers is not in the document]
mojiiba: done

internal/gribi/gribi.go
```diff
@@ -0,0 +1,188 @@
+// Package gribi provides helper APIs to simplify writing gribi test cases.
```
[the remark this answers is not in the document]
mojiiba: done

internal/gribi/gribi.go
```diff
@@ -0,0 +1,188 @@
...
+// It uses fluent APIs and provides wrapper functions to manage sessions and change clients? role
+// roles easily without a need to keep track of the server election id.
+// It also packs modify operations with the corresponding verifications to
+// prevent code duplications and increase the test code readability.
+
```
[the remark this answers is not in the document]
mojiiba: done

internal/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+		g.fluentC.Connection().WithInitialElectionID(1, 0).
+			WithRedundancyMode(fluent.ElectedPrimaryClient).WithPersistence()
+	} else {
+		g.fluentC.Connection().WithInitialElectionID(1, 0).
+			WithRedundancyMode(fluent.ElectedPrimaryClient)
+	}
+	if g.FibACK {
+		g.fluentC.Connection().WithFIBACK()
+	}
+	g.fluentC.Start(context.Background(), t)
```
liulk, line 81:
```
ctx := context.Background()
```

And then reuse ctx below.

internal/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+	g.fluentC.StartSending(context.Background(), t)
+	err := g.AwaitTimeout(context.Background(), t, timeout)
+	return err
+}
+
+// Close function closes the gribi session with the dut by stopping the fluent client.
+func (g *GRIBIHandler) Close(t testing.TB) {
+	t.Helper()
+	t.Logf("closing GRIBI connection for dut: %s", g.DUT.Name())
```
liulk, line 90: Log strings are generally capitalized, unlike error strings.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

internal/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+	subctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	return g.fluentC.Await(subctx, t)
+}
+
+// learnElectionID learns the current server election id by sending
+// a dummy modify request with election id 1.
+func (g *GRIBIHandler) learnElectionID(t testing.TB) (low, high uint64) {
+	t.Helper()
+	t.Logf("learn GRIBI Election ID from dut: %s", g.DUT.Name())
```
liulk, line 108: Capitalize log string?

internal/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+	results := g.fluentC.Results(t)
+	electionID := results[len(results)-1].CurrentServerElectionID
+	return electionID.Low, electionID.High
+}
+
+// UpdateElectionID updates the election id of the dut.
+// The function fails if the requsted election id is less than the server election id.
+func (g *GRIBIHandler) UpdateElectionID(t testing.TB, lowElecId, highElecId uint64) {
+	t.Helper()
+	t.Logf("setting GRIBI Election ID for dut: %s to low=%d,high=%d", g.DUT.Name(), lowElecId, highElecId)
```
liulk, line 123: Capitalize?

internal/gribi/gribi.go
```diff
@@ -0,0 +1,209 @@
...
+	chk.HasResult(t, g.fluentC.Results(t),
+		fluent.OperationResult().
+			WithCurrentServerElectionID(lowElecId, highElecId).
+			AsResult(),
+	)
+}
+
+// BecomeLeader learns the latest election id and the make the client leader by increasing the election id by one.
+func (g *GRIBIHandler) BecomeLeader(t testing.TB) {
+	t.Logf("trying to be a master with increasing the election id by one on dut: %s", g.DUT.Name())
```
liulk, line 138: Capitalize?

liulk (changes requested):
Sorry, just noticed some more nits. Should be the last ones. Thanks for your patience!

mojiiba:
> Sorry, just noticed some more nits. Should be the last ones. Thanks for your patience!

Thanks a lot for the careful and thorough review!

liulk (approved):

PR #109, opened 2022-03-25
CI test

wenovus:


wenovus:
This is a mystery to me, how come the `go vet` and `staticcheck` times are so long? I thought it was for sure the generated code but even after excluding them it's still taking a very long time. Unfortunately I have trouble running `staticcheck` locally so I can't check whether it's reproducible locally right now.

https://github.com/openconfig/featureprofiles/runs/5696985209?check_suite_focus=true
https://github.com/openconfig/featureprofiles/runs/5697267629?check_suite_focus=true

wenovus:
This has been resolved.

PR #114, opened 2022-03-29
Contribution guide directory and folder naming

dplore:
Updated test folder and test file naming

PR #120, opened 2022-04-01
Rename oc to fpoc so it is a bit unique for integration

sthesayi:
This is for integration in Google where the previous naming is too generic. Also added a GH workflow for building/testing the config library.

There are a bunch of files touched but the changes are

1) rename yang/oc to yang/fpoc. This affects the imports on multiple files. Also the generated ygot API goes from oc -> fpoc directory and hence a bunch of files there that you can simply ignore.
2) Add workflows/go.yml for building and testing Config Lib with coverage report. The ondatra tests are skipped as they are not runnable without testbed.
3) The staticroute config lib tests were failing so fixed them so the go.yml is happy. Also added some more test cases, primarily negative.


dplore:
We received some comments that it could be easier to differentiate between configuration generation, config generation testing and feature profile/DUT  tests?    Perhaps you have a suggestion?

sthesayi:
> We received some comments that it could be easier to differentiate between configuration generation, config generation testing and feature profile/DUT tests? Perhaps you have a suggestion?

You mean how we name the files?

sthesayi:
@dplore I see your comment might be related to https://github.com/openconfig/featureprofiles/pull/114. I think the naming suggested there makes sense for the feature profile tests. These tests should be in ate_tests/ otg_tests/ or tests/ directly. The config lib code will reside directly in the feature profile directory and the corresponding unit test will just be named <fp>_test.go in the same directory. This is the approach I coded up in go.yml for config lib coverage.

feature/staticroute/staticroute_test.go
```diff
@@ -60,21 +61,25 @@ func TestAugment(t *testing.T) {
 			},
 		},
 	}, {
-		desc:   "Static route with one next-hop",
-		static: New().WithRoute("1.1.1.1", []string{"1.2.3.44"}),
-		inNI:   &oc.NetworkInstance{},
-		wantNI: &oc.NetworkInstance{
-			Protocol: map[oc.NetworkInstance_Protocol_Key]*oc.NetworkInstance_Protocol{
+		desc:   "Static route Multiple Next Hops",
```
bstoll, line 64: nit:
"Static route with multiple next-hops"

bstoll (approved):

feature/staticroute/staticroute_test.go
sthesayi: I will get this in my next commit if thats ok. Don't want to lose the approval on a cosmetic change.

PR #122, opened 2022-04-06
Add session_id in binding.proto to use FetchSession to connect to ATE

mingyangcisco:
The current static binding creates a new ATE session for each run. The test/ondatra doesn't clear the port ownership. When running the testcase again, it may hit the port not available. Based on that, re-using the existing session id could be helpful since ports will be reclaim in the same session. Also session_id/FetchSession could help enable connection to IXIA windows REST API client. 
Opened issue https://github.com/openconfig/featureprofiles/issues/123#issue-1195206882
Sample Usage:

```
  ixnetwork {
    # Change this to the Web UI hostname, if it differs from the Ixia
    # chassis name.
    target: "ixia-hostname"
    skip_verify: true
    username: "admin"
    password: "admin"
    session_id: 1
  }
```


topologies/binding/binding.go
```diff
@@ -162,9 +162,15 @@ func (b *staticBind) DialIxNetwork(ctx context.Context, ate *binding.ATE) (*bind
...
 	}
-	ixs, err := ixw.IxNetwork().NewSession(ctx, ate.Name)
-	if err != nil {
-		return nil, err
+	var ixs *ixweb.Session
+	var ixs_err error
+	if dialer.SessionId > 0 {
+		ixs, ixs_err = ixw.IxNetwork().FetchSession(ctx, int(dialer.SessionId))
+	} else {
+		ixs, ixs_err = ixw.IxNetwork().NewSession(ctx, ate.Name)
```
liulk, line 170: Per @greg-dennis's comment, let's also delete the new session in staticBind's Release() if we created it.

https://github.com/openconfig/featureprofiles/issues/123#issuecomment-1091045073

liulk: At the moment, Ondatra isn't calling DialIxNetwork multiple times, but for the sake of correctness, we should remember all the sessions that were created by this bind and delete them on Release().

For this to work, we may need to split the code that creates the ixw client (perhaps move that to dialer.newIxWebClient) and the code that creates new session or fetches existing ones, so that Release() has a ixw client it can use to delete a session.

greg-dennis: Alternatively, in the absence of a SessionId, `Reserve` can create a new session and store the session ID. Then DialIxNetwork unconditionally fetches the session ID (be it one that was set from the outset, or created by Reserve). Then Release deletes the session, if one was created.

liulk: Let's go for @greg-dennis's suggestion. I think this means that:

 1. staticBind should maintain its own singleton instance of ixweb client to be obtained from dialer.newIxWebClient. This will be used by Reserve() and Release().
 2. In Reserve(), if ATE session ID is missing, call ixweb client NewSession. Or use FetchSession. Remember this as a singleton instance of *binding.IxSession.
 3. DialIxNetwork always returns the singleton instance of the *binding.IxSession.
 4. In Release(), it only calls DeleteSession if the session was created by Reserve().

mingyangcisco: Thanks @liulk and @greg-dennis for the suggestion. I'm working on those changes now and will try to push commit later today.

mingyangcisco: Hi @liulk and @greg-dennis, I added the logic to use singleton instance to IxWeb and IxSession. Those would be initiated in Reserve() function and would be used in Release() function to delete new session if created. Could you please help check if it matches you expectation?

topologies/binding/options.go
```diff
@@ -140,6 +141,22 @@ func (d *dialer) newHTTPClient() *http.Client {
 	return &http.Client{Transport: tr}
 }
 
+// newIxWebClient makes an IxWeb session using the binding options.
+func (d *dialer) newIxWebClient(ctx context.Context) (*ixweb.IxWeb, error) {
+	hc := d.newHTTPClient()
+	password := d.GetPassword()
+	username := d.GetUsername()
+	if len(password) == 0 || len(username) == 0 {
```
liulk, line 149: Let's make some changes here if you don't mind.

```
if username == "" && password == "" {
  ...
}
```

topologies/binding/binding.go
```diff
@@ -46,6 +47,13 @@ var _ = binding.Binding(&staticBind{})
 
 const resvID = "STATIC"
 
+var (
+	muIxWeb      sync.Mutex
+	muIxSession  sync.Mutex
+	ateIxWeb     = make(map[string]*ixweb.IxWeb)
```
liulk, line 53: Let's make it not as a globally-scoped singleton, but a singleton that is tied to the staticBind instance. And I think you're right that each ATE will need its own session, so we will need a session map nonetheless.

(I guess there isn't a good name for per-instance singleton? Sorry for the confusion.)

liulk: @greg-dennis if this is scoped to the staticBind instance, does it need to be protected by a mutex (i.e. thread-safe as opposed to thread-compatible)?

liulk (changes requested):

topologies/binding/options.go
mingyangcisco: sure. I will make those changes now

topologies/binding/binding.go
mingyangcisco: > Let's make it not as a globally-scoped singleton, but a singleton that is tied to the staticBind instance. And I think you're right that each ATE will need its own session, so we will need a session map nonetheless.
> 
> (I guess there isn't a good name for per-instance singleton? Sorry for the confusion.)

do I need to change ateIxSession to be tied to staticBind instance as well?

liulk: Yes, let's make them tied to staticBind as well.

mingyangcisco: sure. thanks for quick confirmation. I will also make reserveATE/releaseATE/ixWebForATE/ixSessionForATE tied to staticBind.

topologies/binding/options.go
mingyangcisco: changes are done

topologies/binding/binding.go
mingyangcisco: Hi @liulk, based on your suggestion, I made ateIxWeb, ateIxSession and related function tied to staticBind. I keep those mutex for now and will make changes based on discussion above. 
For singleton ateIxWeb/ateIxSession map, the current key is ate.Name which is same as ATE chassis IP. I think it would unique per ate for most of the most of cases in the binding file unless someone use different ate.Id with the same ATE chassis. What do you think about it? Do we need to change the key to be ate.Id?

topologies/binding/binding.go
```diff
@@ -299,3 +300,81 @@ func ports(tports []*opb.Port, bports []*bindpb.Port) (map[string]*binding.Port,
...
+			if err := ixw.IxNetwork().DeleteSession(ctx, ixs.ID()); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (b *staticBind) ixWebForATE(ctx context.Context, ateName string, d dialer) (*ixweb.IxWeb, error) {
+	b.muIxWeb.Lock()
```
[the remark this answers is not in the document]
greg-dennis: Is the question here: can a binding's implementation of Reserve and Release be called concurrently? No they cannot. If we find a case in which they are, that would be a bug on Ondatra to prevent the binding from being called in that way. Our philosophy is that the framework ought to relieve the burden on the binding implementation as much as possible.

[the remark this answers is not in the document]
liulk: Okay, @mingyangcisco please remove the mutex.

I also decided to take a moment and write a blog post about it with example code, in case it will be helpful in the future: https://lifecs.likai.org/2022/04/the-correct-way-to-lock-get-or-create.html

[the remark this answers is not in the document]
mingyangcisco: Cool. Thanks @liulk. I just removed the locks and will go through your blog.

liulk (approved):

Issue #128, opened 2022-04-08
Gribi tests - Additional configuration required for Arista ceos


octpetre:
Hello,

I'm working on the gribi/otg_test/base_leader_election_test.go and I ran into an issue while trying to configure the Arista ceos. 
The ate test case was written for hardware eos which works fine, but apparently the ceos requires the following extra configuration:
- Enabling Arfa mode. This requires injecting a ceos-config file with TFA_VERSION=2 in /mnt/flash on the container for boot up.
- Placing all of the interface configuration on port-channels. A port-channel of 1 member is sufficient.
- Adding this configuration piece in the container startup config 
              **_management api models   
                  ipv4-unicast
                  ipv6-unicast_**

You can reproduce the error I'm seeing by cloning my forked repo at [git@github.com:octpetre/featureprofiles.git](url) and running the following steps:
0. You will require the proper repo images pulled into your cluster (I'm using kind for this)
1. kne_cli create ~/featureprofiles/topologies/kne/ixia-arista-3ports.textproto
2. cd featureprofiles/feature/gribi/otg_tests/base_leader_election_test/
3. go test -v . -kne-config ~/featureprofiles/topologies/kne/testbed.kne-3ports.yml -testbed ~/featureprofiles/topologies/atedut_3.testbed

The script fails at steps 365/366
**ipv4Path := args.dut.Telemetry().NetworkInstance(instance).Afts().Ipv4Entry(ateDstNetCIDR)**

Thanks,

liulk:
Link from Octavian: https://github.com/octpetre/featureprofiles/blob/main/feature/gribi/otg_tests/base_leader_election_test/base_leader_election_test.go#L366

liulk:
@sulrich could you help Octavian? I have not used cEOS at all. I can confirm that base_leader_election_test is passing on 7808 hardware using the 20220411 image.

liulk:
I just noticed that the `management api models` you have in your init_arista.config differs from our arista_ceos.config.

Yours: https://github.com/octpetre/featureprofiles/blob/main/topologies/kne/init_arista.config#L33

```
management api models
   ipv4-unicast
   ipv6-unicast
```

Ours: https://github.com/openconfig/featureprofiles/blob/main/topologies/kne/arista_ceos.config#L41

```
management api models
   provider aft
      ipv4-unicast
      ipv6-unicast
```

(This is also the same as what we have on hardware.)

[...]

PR #139, opened 2022-04-12
Update Arista cEOS config for gNMI/gRIBI best practices

liulk:


liulk:
#128

liulk:
@sulrich I don't know if cEOS supports `hardware fec programmed all` emulation. If not, I'll remove it.

sulrich:
> @sulrich I don't know if cEOS supports `hardware fec programmed all` emulation. If not, I'll remove it.

aiui, there's no equivalent for this with ceos-lab

bstoll:
The "shutdown;no shutdown" etc style commands should not be necessary.  This is loaded as the startup-config on a fresh boot - not changing the configuration after it is already running.  This configuration file is taken from the command "show running-config" (or "write mem") which will never include "shutdown;no shutdown" etc style commands.

I verified AFT streaming works without these changes, so as far as I can tell this doesn't change anything. However, the ordering of the config wasn't exactly right - see [diff](https://github.com/bstoll/featureprofiles/commit/c87bb4b92fe785a92375149e040bc6d652878b0d) that has this fixed up a little bit.  But I don't see any difference in behavior with these changes - it works both before and after.

I left `hardware fec programmed all` out because I don't think it does anything in cEOS, but it does take the command - it should appear after the line `ip routing` if we add it.

liulk:
I removed "shutdown / no shutdown" per your comment, but I moved the provider aft section and hardware fec before the gnmi/gribi sections, so that the statements that would normally require restart of gnmi and gribi would be in place before these agents start.

bstoll (approved):

PR #181, opened 2022-05-05
Adjust staticcheck CI settings to lower memory use

bstoll:
Github runners have [7GB](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) of RAM allocated to them.  Static check analysis appears to exceed this limit and crashes sometimes.  This is mostly due to the large generated code in Ondatra/ygot that this repo uses heavily.

Static checks on my dev machine report the following stats:
```
/usr/bin/time -v staticcheck ./...
...
Elapsed (wall clock) time (h:mm:ss or m:ss): 10:07.36
Maximum resident set size (kbytes): 5965512
...
```

Adjusting the GOGC setting, we can lower the memory usage at the cost of some additional CPU time:
```
GOGC=30 /usr/bin/time -v staticcheck ./...
...
Elapsed (wall clock) time (h:mm:ss or m:ss): 11:14.21
Maximum resident set size (kbytes): 4088248
...
```

We can also get a big improvement by adding ~/.cache/staticcheck to the cache.  A cached run example:
```
GOGC=30 /usr/bin/time -v staticcheck ./...
...
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:02.77
Maximum resident set size (kbytes): 73468
...

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -8,7 +8,7 @@ jobs:
     - name: Install go
       uses: actions/setup-go@v2
       with:
-        go-version: '1.17'
+        go-version: '1.18'
     - name: Install protobuf
       uses: arduino/setup-protoc@v1
       with:
@@ -46,7 +46,7 @@ jobs:
     - name: Install go
       uses: actions/setup-go@v2
       with:
-        go-version: '1.17'
+        go-version: '1.18'
     - name: Checkout code
       uses: actions/checkout@v2
     - name: Cache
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by liulk — APPROVED** [review]

**Comment by dominikh** [comment]
I'm currently looking into some of that poor performance, which is indeed caused largely by ygot.

The first change to Staticcheck, which should land soonish, seems promising, at least on the speed front:

```
before: 861.65s user 9.52s system 266% cpu 5:26.48 total
after:  193.85s user 8.77s system 852% cpu 23.765 total
```

liulk:
Oh wow, looks like the new change both reduced total CPU usage and increased parallelism. Very cool!

PR #183, opened 2022-05-06
Use a const string for the default network instance name

xw-g:


xw-g:
@dplore should we change the "default" to all upper case per https://github.com/openconfig/public/blob/e3fa98e0d27cf7fecf75c71d31316f45c83ddfd6/release/models/network-instance/openconfig-network-instance.yang#L1149-L1161 ?

xw-g:
A friendly ping to @dplore 

Also, another related question: What's gNMI's streaming expectation on the default instance? Should all vendors return "DEFAULT"?

> @dplore should we change the "default" to all upper case per https://github.com/openconfig/public/blob/e3fa98e0d27cf7fecf75c71d31316f45c83ddfd6/release/models/network-instance/openconfig-network-instance.yang#L1149-L1161 ?

xw-g:
cc: @nflath

liulk:
Some vendor currently uses "default" lowercase and does not allow "DEFAULT" uppercase. How about making it a shared flag in the deviations package for now, and we can file a vendor bug to make them OC compliant?

https://github.com/openconfig/featureprofiles/blob/main/internal/deviations/deviations.go

```
var (
  DefaultNI = flag.String("deviation_default_ni", "DEFAULT", "The name to use for the default network instance.")
)
```

Elsewhere, use *deviations.DefaultNI.

xw-g:
> I think `defaultNIName` is a little verbose, so how about just `defaultNI`?
> 
> We also have some preexisting tests that use a constant but using a slightly different name, just:
> 
> ```
> const instance = "default"
> ```
> 
> See:
> 
> * https://github.com/openconfig/featureprofiles/blob/3c2c85ca55ece07984dbddb3c795beb9b15da4dd/feature/gribi/ate_tests/weighted_balancing_test/setup_test.go#L95
> * https://github.com/openconfig/featureprofiles/blob/67d79b451a034d82d31319e1435a3c770cfe6069/feature/gribi/ate_tests/base_leader_election_test/base_leader_election_test.go#L50
> 
> I'm okay to use either `instance` or `defaultNI` but let's standardize it across all tests?

Done. Refactored the naming to defaultNI.

xw-g:
Regarding if we want to use a flag in the deviations package for now, I want to hear @dplore 's comment on https://github.com/openconfig/featureprofiles/pull/183#issuecomment-1128071087. 

Maybe @dplore or the OC model community already have some plans to align the behaviors across vendors.

liulk:
Deprecated by #288

PR #184, opened 2022-05-10
Import RT-5.2: Aggregate Interfaces test

bstoll:


internal/fptest/ports.go
```diff
@@ -15,11 +15,33 @@
...
+func PortChannelName(t testing.TB, dut *ondatra.DUTDevice, i int) string {
+	t.Helper()
+	if i <= 0 {
+		t.Fatalf("port-channel index must be >= 1: %d", i)
+	}
+	switch dut.Vendor() {
+	case ondatra.ARISTA:
+		return fmt.Sprintf("Port-Channel%d", i)
+	case ondatra.CISCO:
+		return fmt.Sprintf("Port-channel%d", i)
```
mingyangcisco, line 36: For Cisco IOS XR, the interface would be "Bundle-Ether%d".

feature/interface/aggregate/ate_tests/aggregate_test/aggregate_test.go
```diff
@@ -0,0 +1,441 @@
...
+	i.Description = ygot.String(p.String())
+	i.Type = ethernetCsmacd
+
+	if *deviations.InterfaceEnabled {
+		i.Enabled = ygot.Bool(true)
+	}
+
+	e := i.GetOrCreateEthernet()
+	e.AggregateId = ygot.String(tc.aggID)
+	// Speed of 99 Gbps is a temporary workaround indicating that the
```
liulk, line 160: This might be a little tricky. Ondatra has a pending change that adds PMD() to ondatra.Port. It's staged internally but not yet open sourced.

https://pkg.go.dev/github.com/openconfig/ondatra#Port

When it's published, the code should look like this:

```
if p.PMD() == "PMD_100G_FR" {
  ...
}
```

What do you think, leave the speed 99 workaround for now and follow up later?

bstoll: I think we should commit it as-is and fix it once it is open sourced.  If we commit this as-is, I will assign an issue to myself to do this.

internal/fptest/ports.go
bstoll: Thanks! I updated it.

internal/fptest/ports.go
```diff
@@ -15,11 +15,33 @@
...
+	"fmt"
 	"sort"
+	"testing"
 
 	"github.com/openconfig/ondatra"
 )
 
+// PortChannelName returns the port channel interface name which is
+// vendor specific.
+func PortChannelName(t testing.TB, dut *ondatra.DUTDevice, i int) string {
```
liulk, line 27: I know that PortChannelName is coming from the internal test that we've already written, but on second thought I'd like to make some changes.

1. I think RT-5.2 is the only test that uses this function, so let's move this to aggregate_test.go and make it unexported.
2. Rename the function to lagName so it's vendor neutral.

bstoll: This is also used by the gnmi_1_5 test.  I considered moving it into the test, but this seems like a helper that is necessary for any test involving aggregate interfaces.  Given that it is used in multiple tests, should we duplicate it in each test?

I agree on point 2.

liulk: I see. In that case, let's add some unit tests for LAGName(), for both positive and negative cases, and maybe consider rewriting the function to remove the dependency on `t testing.TB` by returning an error instead?

liulk (approved):
Brilliant!

PR #187, opened 2022-05-18
Update AFT telemetry counter path

bstoll:
OC model change in openconfig/public#558 changed the path for these counters.


dplore (approved):
LGTM.  This was due to https://github.com/openconfig/public/pull/558

PR #193, opened 2022-05-23
Chassis Reboot Test: Gather DUT datetime after sleep

### LINKED ISSUE #192: Reboot Test Failure
Reboot test fails with following logs sometimes:

```
featureprofiles/feature/gnoi/tests/complete_chassis_reboot/complete_chassis_reboot_test.go:129: Time elapsed 0.29 seconds since reboot was requested.
featureprofiles/feature/gnoi/tests/complete_chassis_reboot/complete_chassis_reboot_test.go:136: Get latest system time: got 2022-05-23T10:07:06-07:00, want newer time than 2022-05-23T10:07:06-07:00
featureprofiles/feature/gnoi/tests/complete_chassis_reboot/complete_chassis_reboot_test.go:129: Time elapsed 10.44 seconds since reboot was requested.
```

Time accuracy is only to 1 second and the check is failing because less than 1 second has elapsed on the loop.

bstoll:
The first `prevTime` and `latestTime` time could be collected in the same second causing this test to fail occasionally.  This improves the time handling a bit and guarantees some delay has passed.  Resolves #192.


sezhang2 (approved):
LGTM

Issue #221, opened 2022-06-07
ondatra.go:56] PushConfig unimplemented


cprabha:
Hi,

I did git pull today morning and unable to run scripts now due to error "ondatra.go:56] PushConfig unimplemented".
It was working before the git pull.

I have debugged the issue further and it shows getting error at below point. It looks like we are trying to execute reset on dut instead of ate.

Is this due to changes pushed in  topologies/binding/binding.go ??? (https://github.com/openconfig/featureprofiles/pull/206)

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
func (b *staticBind) reset(ctx context.Context) error {
	for _, bdut := range b.r.GetDuts() {
		if err := applyCli(ctx, bdut, b); err != nil {   ============> gives error here " pushconfig unimplemented"
			return err
		}
		if err := applyGnmi(ctx, bdut, b); err != nil {
			return err
		}
	}
	return nil
}
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

Error:
=========
F0607 11:51:04.651470  544968 ondatra.go:56] PushConfig unimplemented

Thanks,
Prabha

Ankur19:
Hi @cprabha Thanks for the heads up! Didn't realize PushConfig() is unimplemented. You can still run the scripts by setting the **pushConfig** variable to _false_ in **staticBind**: https://github.com/openconfig/featureprofiles/blob/main/topologies/binding/binding.go#L44

cprabha:
Hi @Ankur19 
Thank you. I have topology where dut and ate both are present. If I disable pushconfig will it work for ate ?
Please suggest.

Ankur19:
Yes! @cprabha the staticBind.pushConfig variable is only for DUTs and not for ATEs.

cprabha:
Thank you @Ankur19 .
I will try with pushConfig flag set to false.

liulk:
This should be fixed by #225 . Let @Ankur19 know if it's still not fixed.

PR #225, opened 2022-06-07
Update binding to implement device reset

Ankur19:


topologies/binding/binding.go
```diff
@@ -299,43 +303,71 @@ func readGnmi(path string) (*gpb.SetRequest, error) {
...
 	if err != nil {
 		return err
 	}
-	conn, err := dialer.dialGRPC(ctx)
+	sc, err := dialer.dialSSH()
 	if err != nil {
 		return err
 	}
-	defer conn.Close()
+	defer sc.Close()
```
liulk, line 332: Rather than sc.Close(), do cli.Close() instead.

Currently cli.Close() will just call sc.Close() but the main idea is that the SSH transport is owned by the CLI object, so we should not operate on it anymore once it's given to newCLI().

Ankur19: Done!

topologies/binding/new.go
```diff
@@ -32,7 +32,7 @@ var (
 	pluginArgs  = flag.String("plugin-args", "", "arguments for the vendor binding")
 	bindingFile = flag.String("binding", "", "static binding configuration file")
 	kneConfig   = flag.String("kne-config", "", "YAML configuration file")
-	pushConfig  = flag.Bool("push-config", true, "push device reset config supplied to static binding")
+	pushConfig  = flag.Bool("push-config", false, "push device reset config supplied to static binding")
```
liulk, line 35: We can make this default true again, since if neither cli nor gnmi set are given, reset is now no-op.

Ankur19: Done!

liulk (approved):

liulk:
This should fix #221

PR #233, opened 2022-06-10
Add list_tests.sh to output device tests in CSV.

liulk:


tools/list_tests.sh
```diff
@@ -0,0 +1,47 @@
...
+        echo "$(red WARNING:)" "not a valid test path: $line" >&2
+        feature="${path}"
+        ;;
+    esac
+    feature="${feature#testing/}"
+    feature="${feature#feature/}"
+
+    echo "\"${feature}\",\"${id}\",\"${desc}\",\"${path}\""
+  done < <(
+    find -L feature testing/feature -name \*.md -exec egrep '^# [A-Za-z]+-[0-9]+\.[0-9]+:' \{} \+
```
bstoll, line 37: What is `testing/feature`?  Doesn't look like a directory that exists.

bstoll (approved):

tools/list_tests.sh
liulk: Explained in chat.

PR #236, opened 2022-06-13
Remove 'show' command parsing via gNMI Get.

robshakir:
```
 * (M) feature/system/gnmi/cliorigin/ate_test/cli_origin_test.go
  - Remove use of gNMI.Get as means to have arbitrary CLI commands run. gNMI
    Get should not be used in this manner, but rather gNOI commands that have
    the equivalent functionality should be used.
```

This behaviour isn't specified anywhere in the gNMI specification, and I don't believe it is intended to be. If we think that this kind of functionality should be available, we need to primarily update the specifications before adding tests.

### DIFF
--- a/feature/system/gnmi/cliorigin/ate_tests/cli_origin_test/cli_origin_test.go
+++ b/feature/system/gnmi/cliorigin/ate_tests/cli_origin_test/cli_origin_test.go
@@ -16,7 +16,6 @@ package cli_origin_test
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


bstoll (approved):
I agree this is correct ([referece](https://github.com/openconfig/reference/blob/master/rpc/gnmi/mixed-schema.md#special-values-of-origin)) but I also think this behavior should exist in some form.

PR #251, opened 2022-06-17
deviations' config

mojiiba:
Hello, Currently deviations are enabled by default that causes issue for vendors whose implementation does not deviate from the OC. Ideally, the deviations should be disabled by default. This pull adds a simple function `ConfigVendorDeviations` that can be called at the beginning of the each test to enable/disable the deviations based on the device vendor.

### DIFF
--- a/feature/gribi/ate_tests/base_leader_election_test/base_leader_election_test.go
+++ b/feature/gribi/ate_tests/base_leader_election_test/base_leader_election_test.go
@@ -245,7 +245,6 @@ func testIPv4LeaderActiveChange(ctx context.Context, t *testing.T, args *testArg
 
 func TestElectionIDChange(t *testing.T) {
 	dut := ondatra.DUT(t, "dut")
-	deviations.ConfigVendorDeviations(dut.Vendor())
 	ctx := context.Background()
 
 	// Configure the DUT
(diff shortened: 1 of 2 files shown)


xw-g:
Thanks @mojiiba for the contribution. 

I tend to **disagree** with the proposal in this PR, because: 
* Since it's a flag to the feature profile test, we might want to keep the deviation flag management outside feature profile test.
* Following the idea of this PR, we might accidentally expose a vendor's capability or limitation in the test code, which is not something we want for Feature Profile :). 

But I'll let @liulk as the formal reviewer.

mojiiba:
Thanks for the comment. I understand. In that case, can we set the flags’ default to false?

From: Xiao Wang ***@***.***>
Date: Friday, June 17, 2022 at 4:18 PM
To: openconfig/featureprofiles ***@***.***>
Cc: Moji Ba (mbagherz) ***@***.***>, Mention ***@***.***>
Subject: Re: [openconfig/featureprofiles] deviations' config (PR #251)

Thanks @mojiiba<https://github.com/mojiiba> for the contribution.

I tend to disagree with the proposal in this PR, because:

  *   Since it's a flag to the feature profile test, we might want to keep the deviation flag management outside feature profile test.
  *   Following the idea of this PR, we might accidentally expose a vendor's capability or limitation in the test code, which is not something we want for Feature Profile :).

But I'll let @liulk<https://github.com/liulk> as the formal reviewer.

liulk:
I think deviation management has to be done as part of the test result collection workflow outside of the test, so we can track which deviations are required in order for tests to pass, and the intent is to gradually phase out these deviations. It is equally problematic if a device requires a deviation to be set to false. These requirements need to be eradicated over time.

For now, I am okay with setting the --deviation_interface_enabled to false, so let's remove ConfigureVendorDeviations. I've also filed an issue to specifically test the non-deviated setting of subinterface enabled in #253 so we have a way to track this non-compliant behavior.

liulk:
@bstoll FYI, we'll have to set deviations explicitly for our continuous test run workflow.

liulk (approved):

PR #261, opened 2022-06-22
Open source ordering ack test

bstoll:
Co-authored-by: Likai Liu <liulk@google.com>

### DIFF
--- a/feature/gribi/ate_tests/ordering_ack_test/ordering_ack_test.go
+++ b/feature/gribi/ate_tests/ordering_ack_test/ordering_ack_test.go
@@ -1,0 +1,503 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


feature/gribi/ate_tests/ordering_ack_test/ordering_ack_test.go
```diff
@@ -0,0 +1,503 @@
...
+		t.Fatalf("Await got error for ModifyRequest: %v", err)
+	}
+
+	res := args.c.Results(t)
+	chk.HasResult(t, res,
+		fluent.OperationResult().
+			WithOperationID(2).
+			WithOperationType(constants.Add).
+			WithIPv4Operation(ateDstNetCIDR).
+			WithProgrammingResult(fluent.ProgrammingFailed).
```
mingyangcisco, line 281: We are seeing different expectation regarding out of order when those are sent in one modify request vs multiple modify requests. gribigo [compliance testcase](https://github.com/openconfig/gribigo/blob/a471e9b7c03e2d121a3b5d23e44c17fb37514c47/compliance/compliance.go#L542) is expecting forward reference to work when multiple modify requests are sent. should the behavior be consistent? @xw-g

mingyangcisco: sure. thanks

liulk (approved):

PR #264, opened 2022-06-22
Open source supervisor switchover test

bstoll:
Co-authored-by: Sean Zhang <sezhang@google.com>


liulk (approved):

liulk (changes requested):
Move the test under feature/gnoi/tests?

https://github.com/openconfig/featureprofiles/tree/main/feature/gnoi/tests

liulk:
I think the discussion about whether gnoi should be a root feature, a subfeature under system, or just be part of the system feature would take much longer to settle. At the moment, I just want to avoid having many test locations when open sourcing.

dplore:
There is precedent to place tests which are concerned with using gNOI protocol in the feature/gnoi folder and gRIBI in the feature/gribi folder.   So IMO more gNOI tests should IMO go into feature/gnoi.   

The other folders are named after their openconfig root paths.  But maybe we should have used gnmi/<root_oc_path> instead to be consistent with the naming we used for gNOI and gRIBI.  I think this is historical because we didn't consider gRIBI and gNOI when we very first started.  I agree that this latter topic isn't super important to resolve and is beyond the scope of this PR.

bstoll:
@liulk I updated the PR to use the paths we discussed in the meeting.  If this sounds good to you, I can make another PR to get all the GNOI related features/tests into sync with this style.  #281 is another very similar PR to this one, so I will copy any decisions from here into that PR.

PR #286, opened 2022-06-29
CONTRIBUTING guide refactored to make code style more clear

liulk:
- Code style references inclusive language and testing blog.
- Fixed singleton test example.
- Added "code should follow the test plan".


CONTRIBUTING.md
```diff
@@ -31,123 +31,112 @@ Note that we use the Google CLA because the OpenConfig project is
...
-OpenConfig working group has strived for since its inception in 2014. Thank you
-for your contributions!
+All code and documentation should follow
+[Google developer documentation style guide](https://developers.google.com/style/word-list)
+for the use of inclusive language.
 
-## Test development style guide
+It is recommended that tests follow
+[Testing on the Toilet](https://testing.googleblog.com/) for best practices on
+design patterns.
```
dplore, line 47: Since this is blog format, IMO it is difficult to derive any particular requirements or actions an author should perform.   Are there any specific design patterns we want to request be used?  

Perhaps another way would be to accumulate a list of recommendations.  We can start with just the obvious ones and add to them over time:

Recommended styles and design patterns used where appropriate in featureprofiles include:
- [Prefer table driven tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests)
- [Test clarity](https://testing.googleblog.com/2014/03/testing-on-toilet-what-makes-good-test.html)
- [Flaky test challenges](https://testing.googleblog.com/2021/03/test-flakiness-one-of-main-challenges.html)

CONTRIBUTING.md
```diff
@@ -31,123 +31,112 @@ Note that we use the Google CLA because the OpenConfig project is
...
+## Directory Organization
 
 The directory tree is organized as follows:
 
 *   `cloudbuild/` contains google cloud build scripts for running virtual
     routers in containers on [KNE](https://github.com/google/kne)
 *   `feature/` contains definition and tests of feature profiles.
 *   `feature/experimental` contains new features and tests which are not yet
-    proven to be stable
+    categorized or not confirmed to pass.
```
dplore, line 57: Should state how a test is moved from experimental and into feature/

CONTRIBUTING.md
```diff
@@ -31,123 +31,112 @@ Note that we use the Google CLA because the OpenConfig project is
...
-*   `feature/interface/tests/` contains the interfaces test suite without
-    traffic generation.
-*   `feature/bgp/` container for BGP feature profiles and test suites.
-*   `feature/bgp/addpath` contains feature profiles related to BGP.
-*   `feature/bgp/addpath/README.md` documents the BGP feature profile.
-*   `feature/bgp/addpath/ate_tests` contains test code related to BGP Add Path
-    features.
-*   `internal/deviations` contains code which overrides standard tests where
+`feature/<featurename>/[<sub-feature>/]<tests|ate_tests|otg_tests|kne_tests>/<test_name>/<test_name>.go`.
+For example:
```
dplore, line 68: should we add example of where configuration generation code goes?  This has been a subject of confusion in the past.

CONTRIBUTING.md
```diff
@@ -210,9 +199,63 @@ Number Reservation for Documentation Use ([RFC 5398]). In particular:
...
+```
+    * (M) yang/keychain/openconfig-keychain.yang
+      - Add a typedef for referencing a keychain from other modules.
+    * (M) yang/isis/*
+      - Fix support for hello-authentication to allow for references to a
+        specific keychain as defined in the keychain model.
+      - Fix support for authentication of *SNP packets, referencing a
+        keychain that can be used to auth these packets.
+      - move IS-IS model to openconfig-inet-types rather than ietf-inet-types.
+```
```
dplore, line 261: Ideas:
1.  should we request the author include a link to test run output?  (perhaps via a github gist?)   
2. If we can see believable results of a test passing, would that allow skipping going through the step of  experimental folder?

dplore (dismissed):
I think it's good as is, but I have a few suggestions to consider

CONTRIBUTING.md
liulk: Listing some concrete design pattern is a great suggestion. I'll add the table driven test from the golang wiki, plus two others that I've actually referenced during PR review.

CONTRIBUTING.md
liulk: Generally, after they are categorized and confirmed to pass, but I think we may want to reserve our discretion on a case by case scenario.

CONTRIBUTING.md
liulk: Since we did agree to split the config gen library from the device tests, I'll update this section as part of that work.

CONTRIBUTING.md
liulk: 1. I'd certainly like to see some evidence from the test run, but I'm not sure if gist is a good way to manage test output. For what it's worth, test log is typically very long, and I don't know if it could be truncated silently.
2. One time I suggested exactly that on a PR whose test code looked very clean and believable, and it also had a clear canonical location, but on second thought I think having the experimental gating might still be a good idea. We may need to first align internally on where the canonical feature location is, e.g. for gNOI.

dplore (approved):
Thanks for the updates!

PR #287, opened 2022-06-29
update feature profiles to latest ondatra version

greg-dennis:


bstoll (approved):
LGTM looks like there is some overlap with #285.

Issue #253, opened 2022-06-17
gNMI-1.11 add check that subinterface ipv4/ipv6 enabled can be explicitly set to true


liulk:
Per #251, some devices do not support setting these paths:

  * /interfaces/interface/subinterfaces/subinterface/ipv4/config/enabled
  * /interfaces/interface/subinterfaces/subinterface/ipv6/config/enabled

which are part of the [OpenConfig standard](http://ops.openconfig.net/branches/models/master/docs/openconfig-interfaces.html).

It was discovered when running tests with `--deviation_interface_enabled` which currently defaults to true. The decision is to set the deviation default to false, but add a specific non-deviated check in the test plan so we can track the non-compliant behavior for those devices. "gNMI-1.11: Telemetry: Interface Packet Counters" seems like a logical place where such check could be added without much overhead.

https://github.com/openconfig/featureprofiles/tree/main/feature/gnmi/ate_tests/telemetry_interface_packet_counters_test

PR #294, opened 2022-06-30
guard ipv4/enabled with deviations.InterfaceEnabled

mojiiba:
This pull request guards the ipv4/enable for [telemetry_interface_packet_counters_test.go](https://github.com/openconfig/featureprofiles/compare/main...mojiiba:featureprofiles:moji_interface_deviation?expand=1#diff-df1d54880bfbdae0493ef1e2d51a80c91bb932962e512b437de73bbd75834ef4) with deviations.InterfaceEnabled simialr to other tests.

### DIFF
--- a/feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
+++ b/feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
@@ -18,6 +18,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/openconfig/featureprofiles/internal/deviations"
 	"github.com/openconfig/featureprofiles/internal/fptest"
 	"github.com/openconfig/ondatra"
 	"github.com/openconfig/ondatra/telemetry"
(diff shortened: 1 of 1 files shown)


bstoll:
Can you explain the purpose of this change a bit?  My understanding is the InterfaceEnabled deviation exists because some implementations were not respecting the default value (enabled) for subinterface configuration, so we had to explicitly specify for the value to be enabled on subinterfaces with a deviation flag.

We could guard s.Enabled with the deviation, but this PR seems to be guarding the reads of the value on the deviation (and only on IPv4, not on IPv6?); shouldn't we always be able to read the value?  The [test plan](https://github.com/openconfig/featureprofiles/tree/main/feature/gnmi/ate_tests/telemetry_interface_packet_counters_test) is asking to be able to read this value, so I am not sure it is correct to guard it on the deviation flag.

mojiiba:
> 

@bstoll  Thanks for the comment. I missed the comment made by @liulk  [here](https://github.com/openconfig/featureprofiles/issues/253).  It appears that gNMI-1.11  exercises the ipv4/enabled as part of the test logic.  So, I am closing the pull request for now. 

Concerning the ipv6 versus ipv4, based on my understanding, when ipv6 is enabled, the router turns on ipv6 neighbor discovery and allocates a link-local address, but for IPV4, I am not sure what is the purpose? So, we need tp discuss this offline and revisit this pull request after that.

PR #298, opened 2022-07-01
Expire Cloud Build SSH keys after 1 day

bstoll:
Cloud Build tasks were failing with the following error:

`ERROR: (gcloud.compute.ssh) INVALID_ARGUMENT: Login profile size exceeds 32 KiB. Delete profile values to make additional space.`

This should prevent the profile from growing.


bstoll:
The GCP account switched from using instance-metadata to requiring [OS Login](https://cloud.google.com/compute/docs/oslogin) and that resulted in this change.  It was intended for keys to be generated for each CI run and then destroyed, but OS Login changed the way this works a bit and it looks like we now needs to specify the key timeout.

PR #308, opened 2022-07-12
Fix Aggregate Load Balancing weight ratio test

bstoll:
The "Ratio" subtest fails consistently in the aggregate link load balancing test.  The test would display the reason for the failure already in log messages:

```
outPkts normalized got: [0 0 0 0.33202573136537744 0.33593268058457215 0.3320415880500504]
want: [0 0 0 0]
...
Packet distribution ratios -want,+got:
          []float64{
          	0,
          	0,
          	0,
        - 	0,
        + 	0.33202573136537744,
        + 	0.33593268058457215,
        + 	0.3320415880500504,
          }
```

These lists get compared and are expected to be similar to eachother.  With this PR, the test returns the following output:

```
outPkts normalized got: [0.3359486895501926 0.3320247218493458 0.33202658860046164]
want: [0.3333333333333333 0.3333333333333333 0.3333333333333333]
```

Against a DUT, the test now passes.  This test was run in a testbed with a total of 4 ports (although it should support more).

### DIFF
--- a/feature/interface/aggregate/ate_tests/balancing_test/balancing_test.go
+++ b/feature/interface/aggregate/ate_tests/balancing_test/balancing_test.go
@@ -327,9 +327,10 @@ var approxOpt = cmpopts.EquateApprox(0 /* frac */, 0.01 /* absolute */)
 // portWants converts the nextHop wanted weights to per-port wanted
 // weights listed in the same order as atePorts.
 func (tc *testCase) portWants() []float64 {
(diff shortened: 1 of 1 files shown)


bstoll:
#309 needs to get pulled in to fix the OC path validation CI check.

dplore:
Thanks @bstoll ,  #309 is merged now so you can rebase/merge it in

PR #309, opened 2022-07-12
Sflow feature profile update

bstoll:
The OpenConfig model for sflow was updated in [openconfig/public#522](https://github.com/openconfig/public/pull/522).  This syncs the feature profile up with the latest model.


dplore (approved):
LGTM, thanks!

PR #325, opened 2022-07-20
Binding reset to use prototext and propagate error correctly (fixes #322)

### LINKED ISSUE #322: Unmarshaling Setrequest and error handling for reset function
Hi, the code for unmarshalling SetRequest from a file uses proto.Unmarshal rather than prototext.Unmarshal that causes failure for loading config in prototext format. Is this intentional? If not can I open a pull request to fix it?

https://github.com/openconfig/featureprofiles/blob/736ba3a9593960cf1763b02098c4b1ec56db87ee/topologies/binding/binding.go#L300

Also,  the code  ignores  errors for applying config during reset that can cause confusion. We need to either terminate the test execution if the reset is unsuccessful or log a warning message.  I can also open a pull request for this if you agree. 

https://github.com/openconfig/featureprofiles/blob/736ba3a9593960cf1763b02098c4b1ec56db87ee/topologies/binding/binding.go#L272

liulk:


liulk:
#322

PR #341, opened 2022-07-25
Add ATE telemetry test to the OTG test.

robshakir:
```
 * (M) {go.mod, go.sum}
   - Revert to gosnappi version that works against -2801 image.
 * (A) topologies/kne/arista_ceos_otg.textproto
   - Add a topology for an Arista DUT to an Ixia-c container.
 * (M) topologies/otg_tests/topology_test/topology_test.go
   - Add a check that the interface telemetry from the ATE is
     operational as a gate to accepting an image.
   - Change naming of ATE to "ate" to allow standard testbeds to be
     used.
```


dplore:
Note, this has some conflicts with #340.  Which gosnappi version should we use?

robshakir:
That is a good debate that I was planning to raise here.

`gosnappi` is related to the version of `ixia-c` that is being run, so if one is not using HEAD of `ixia-c` then it's not necessarily possible to move to latest. I propose that we have some co-ordinated version (gated with passing `otg_tests/topology_test` at the very least) that means that we move to a new `gosnappi` version. @octpetre - WDYT?

liulk:
Sorry I merged #340 before seeing this.

octpetre:
go.mod was updated in #340 to the latest gosnappi version 0.8.5

robshakir:
Once I'm updated to `-3002` then I can pick up this new version.

@bstoll - today, I think we should be able to run OTG tests as part of GCB CI. Is the current Cloud Build workflow set up to pull `ixia-c` containers?

robshakir:
Now we're all on the same `ixia-c` version, this is ready to be reviewed again, thanks.

bstoll:
> Once I'm updated to `-3002` then I can pick up this new version.
> 
> @bstoll - today, I think we should be able to run OTG tests as part of GCB CI. Is the current Cloud Build workflow set up to pull `ixia-c` containers?

I don't think we have the correct permissions to directly access the gep-kne project containers.  We could copy them, but it would be a lot easier if they were shared.  I will try to figure this out.

robshakir:
Thanks Brandon -- I'd discuss this with Alex Masi, he has integration tests for KNE that I expect that do this.

robshakir:
friendly ping.

bstoll (approved):

PR #343, opened 2022-07-26
Update ping_test.go

sezhang2:
- Replace the hard-coded Loopback0 interface name with Ondatra helper function call
 - Skip the check on some optional fields on the first echo reply summary message

### DIFF
--- a/feature/gnoi/system/tests/ping_test/ping_test.go
+++ b/feature/gnoi/system/tests/ping_test/ping_test.go
@@ -399,6 +399,27 @@ func TestGNOIPing(t *testing.T) {
 	}}
 
 	gnoiClient := dut.RawAPIs().GNOI().Default(t)
+	for _, tc := range cases {
+		t.Run(tc.desc, func(t *testing.T) {
+			t.Logf("Sent ping request: %v\n\n", tc.pingRequest)
+
+			pingClient, err := gnoiClient.System().Ping(context.Background(), tc.pingRequest)
+			if err != nil {
+				t.Fatalf("Failed to query gnoi endpoint: %v", err)
+			}
+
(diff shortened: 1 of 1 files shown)


bstoll:
I believe the Cloud Build task is failing because the container isn't generating the TLS certificates correctly.

I created https://github.com/openconfig/kne/pull/175 , which seems to fix the issue on my test VM.  The check failing is not related to anything in this PR specifically.

feature/gnoi/system/tests/ping_test/ping_test.go
```diff
@@ -435,17 +431,21 @@ func TestGNOIPing(t *testing.T) {
...
-				if responses[i].Bytes < tc.expectedReply.Bytes {
-					t.Errorf("Ping Bytes: got %v, want >= %v", responses[i].Bytes, tc.expectedReply.Bytes)
-				}
-				if responses[i].Sequence < tc.expectedReply.Sequence {
-					t.Errorf("Ping time: got %v, want >= %v", responses[i].Sequence, tc.expectedReply.Sequence)
-				}
-				if responses[i].Ttl < tc.expectedReply.Ttl {
-					t.Errorf("Ping TTL: got %v, want >= %v", responses[i].Ttl, tc.expectedReply.Ttl)
+
+				if i < len(responses)-1 {
```
Ankur19, line 435: This doesn't look great. I feel we should make the top level if statement to `len(responses)-1` and then add the logic for last element outside the loop.
e.g.
```
for i := 0; i < len(responses)-1; i++ {
// log echo response data
}
summary := responses[len(responses)-1]
// log summary response data

Ankur19 (approved):
LGTM!

PR #346, opened 2022-07-27
Telemetry: Port Speed Test

ggtsp:
Open source  Telemetry: Port Speed Test
Co-authored-by: Likai Liu <liulk@google.com>


internal/quirks/manufacturer/manufacturer.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/openconfig/ondatra"
+)
+
+// PortChannelName returns the port channel interface name which is
+// vendor specific.
+func PortChannelName(t testing.TB, dut *ondatra.DUTDevice, i int) string {
+	t.Helper()
+	if i <= 0 {
+		t.Fatalf("port-channel index must be >= 1: %d", i)
```
robshakir, line 29: Is this true across all implementations? I think we can have `ae0` in JUNOS. I'd recommend not having validations in this function.

robshakir (changes requested):

feature/gnmi/ate_tests/telemetry_port_speed_test/README.md
```diff
@@ -0,0 +1,39 @@
...
+        Validate /interfaces/interface/aggregation/state/lag-speed is reported
+        as N*port speed.
+    *   Disable each port at ATE and determine that the effective speed is
+        reduced by the expected amount.
+    *   Turn ports sequentially up at the ATE, and determine that the effective
+        speed is increased as expected.
+
+## Config Parameter Coverage
+
+No configuration coverage.
```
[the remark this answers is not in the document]
liulk: This PR is supposed to be a straightforward move from internal code depot and the internal test plan doc to open source, so perhaps we could just label the coverage as "TBD" for now and unblock the move?

feature/gnmi/ate_tests/telemetry_port_speed_test/telemetry_port_speed_test.go
```diff
@@ -0,0 +1,371 @@
...
+	if *deviations.InterfaceEnabled {
+		i.Enabled = ygot.Bool(true)
+	}
+
+	s := i.GetOrCreateSubinterface(0)
+	s4 := s.GetOrCreateIpv4()
+	if *deviations.InterfaceEnabled {
+		s4.Enabled = ygot.Bool(true)
+	}
+	a4 := s4.GetOrCreateAddress(a.IPv4)
```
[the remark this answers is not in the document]
liulk: Usually a move PR only changes the code to fix import, so maybe defer these changes to a future PR?

feature/gnmi/ate_tests/telemetry_port_speed_test/telemetry_port_speed_test.go
```diff
@@ -0,0 +1,371 @@
...
+
+func (tc *testCase) configMemberDUT(i *telemetry.Interface, p *ondatra.Port) {
+	i.Description = ygot.String(p.String())
+	if *deviations.InterfaceEnabled {
+		i.Enabled = ygot.Bool(true)
+	}
+
+	e := i.GetOrCreateEthernet()
+	e.AggregateId = ygot.String(tc.aggID)
+	// Speed of 99 Gbps is a temporary workaround indicating that the
```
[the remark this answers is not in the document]
liulk: It had not been viable to detect 100G-FR optics from Ondatra for a very long time until very recently. I'm responsible for the cleanup.

Let's leave it in for now to keep the change during open sourcing move at minimum.

feature/gnmi/ate_tests/telemetry_port_speed_test/telemetry_port_speed_test.go
```diff
@@ -0,0 +1,371 @@
...
+		WithDefaultGateway(dutIPs.IPv4)
+	agg.IPv6().
+		WithAddress(ateIPs.IPv6CIDR()).
+		WithDefaultGateway(dutIPs.IPv6)
+
+	// Fail early if the topology is bad.
+	tc.top.Push(t)
+
+	ok := false
+	for n := 3; !ok; n-- {
```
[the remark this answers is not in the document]
liulk: I think the original author of this test copied the topology_test as the basis for this test, and topology_test implemented retrying because it was used for troubleshooting testbeds. Normally, tests should not need to do this.

Let's replace lines 247-261 to be just:

```
tc.top.Push(t).StartProtocols(t)
```

internal/quirks/manufacturer/manufacturer.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/openconfig/ondatra"
+)
+
+// PortChannelName returns the port channel interface name which is
+// vendor specific.
+func PortChannelName(t testing.TB, dut *ondatra.DUTDevice, i int) string {
```
liulk, line 26: We don't need this anymore. Ondatra now provides netutil.NextBundleInterface.

https://pkg.go.dev/github.com/openconfig/ondatra/netutil#NextBundleInterface

Example: https://github.com/openconfig/featureprofiles/blob/b8054dcbfa3c998035fd2dc8940991dd314553fc/feature/interface/aggregate/ate_tests/aggregate_test/aggregate_test.go#L432

internal/quirks/manufacturer/manufacturer.go
liulk: The original internal version had a comment addressing this very question, but somehow the comment got lost. Nevertheless, we're not using this function anymore.

liulk (changes requested):

feature/gnmi/ate_tests/telemetry_port_speed_test/telemetry_port_speed_test.go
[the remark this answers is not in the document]
ggtsp: made the change at this and other locations, can we keep them?

feature/gnmi/ate_tests/telemetry_port_speed_test/README.md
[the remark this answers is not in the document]
ggtsp: marked as TBD

feature/gnmi/ate_tests/telemetry_port_speed_test/telemetry_port_speed_test.go
```diff
@@ -0,0 +1,352 @@
...
+
+	agg.IPv4().
+		WithAddress(ateIPs.IPv4CIDR()).
+		WithDefaultGateway(dutIPs.IPv4)
+	agg.IPv6().
+		WithAddress(ateIPs.IPv6CIDR()).
+		WithDefaultGateway(dutIPs.IPv6)
+
+	// Fail early if the topology is bad.
+	tc.top.Push(t)
```
liulk, line 245: Line 245 is redundant, remove?

liulk (changes requested):

liulk (approved):

feature/gnmi/ate_tests/telemetry_port_speed_test/telemetry_port_speed_test.go
```diff
@@ -0,0 +1,350 @@
...
+	top *ondatra.ATETopology
+
+	dutPorts []*ondatra.Port
+	atePorts []*ondatra.Port
+	aggID    string
+}
+
+func (*testCase) configDUT(i *telemetry.Interface, a *attrs.Attributes) {
+	i.Description = ygot.String(a.Desc)
+	if *deviations.InterfaceEnabled {
```
robshakir, line 94: I don't think `deviations.InterfaceEnabled` should cover this - having `/interfaces/interface/config/enabled=true` seems like something that should be supported on all devices. I believe the deviation and discussion was about whether we do this specifically for IPv4. @dplore - PTAL.

feature/gnmi/ate_tests/telemetry_port_speed_test/telemetry_port_speed_test.go
```diff
@@ -0,0 +1,350 @@
...
+		t.Fatalf("Testbed requires at least 2 ports, got: %v", tc.atePorts)
+	}
+
+	// Don't use WithLACPEnabled which is for emulated Ixia LACP.
+	agg := tc.top.AddInterface(ateIPs.Name)
+	lag := tc.top.AddLAG("lag").WithPorts(tc.atePorts...)
+	lag.LACP().WithEnabled(tc.lagType == lagTypeLACP)
+	agg.WithLAG(lag)
+
+	// Disable FEC for 100G-FR ports because Novus does not support it.
```
robshakir, line 226: Should this be conditional on a deviation too, it seems to be referencing something specific to a particular model of test interface?

robshakir (approved):
Two top-level questions, but LGTM.

feature/gnmi/ate_tests/telemetry_port_speed_test/telemetry_port_speed_test.go
dplore: That is correct, `deviations.InterfaceEnabled` is only intended to cover the ipv4 and ipv6 subinterface leaves.  This doesn't break this test because it is also required by openconfig that configured interfaces are enabled , but makes it slightly more permissive than necessary.

Issue #329, opened 2022-07-21
Default Network-Instance Case


aredmon8551:
https://github.com/openconfig/featureprofiles/blob/1b105fe0d3bdf7f866dc52aa15e4ea5327be01c1/feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go#L141

Hey @liulk - should this be 'DEFAULT' rather than 'default'?  

https://github.com/openconfig/public/blob/master/release/models/network-instance/openconfig-network-instance.yang#L1184

Concerned that vendors may have diverged in the naming convention used for this - EOS looks to use 'default', Junos uses 'DEFAULT' per the NI model's guidance, not sure about SR-Linux/IOS-XR.

aredmon8551:
Looks like this is under discussion, as well, here:
https://github.com/openconfig/featureprofiles/pull/183

earies:
@dplore for comment here as I believe this was brought up in a few other recent issues/PRs as well

dplore:
`"DEFAULT"` is the correct string.  Requesting @bstoll take this one.

dplore:
Related to #334  #347  #337

dplore:
We will use this issue to track scanning all the tests for references to the default network instance and update strings to use 'DEFAULT'.  @amrindrr will take this up

mojiiba:
Hi @dplore , Can we use deviations.DefaultNetworkInstance rather than the "DEFAULT" for the test cases? Otherwise, we will not be able to use the deviation if is needed.

PR #352, opened 2022-07-27
Fix hard-coded "default" to *deviations.DefaultNetworkInstance

liulk:
#329

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -378,7 +378,7 @@ been clarified that the name for the default network instance should be
 uppercase `"DEFAULT"`. Some legacy devices are still using lowercase
 `"default"`, so device tests should use the deviation
 `*deviations.DefaultNetworkInstance` which allows them to work on those legacy
-devices whlie they are being updated. Non-device unit tests may hard-code
+devices while they are being updated. Non-device unit tests may hard-code
 `"DEFAULT"`.
 
 ## Pull Requests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -371,6 +371,16 @@ Number Reservation for Documentation Use ([RFC 5398]). In particular:
 Both ranges have 16 total numbers each. The hexadecimal notation makes it more
 obvious where the range starts and stops.
 
+## Default Network Instance
+
+In OpenConfig [PR #599](https://github.com/openconfig/public/pull/599), it has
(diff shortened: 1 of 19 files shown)


CONTRIBUTING.md
```diff
@@ -371,6 +371,16 @@ Number Reservation for Documentation Use ([RFC 5398]). In particular:
...
+## Default Network Instance
+
+In OpenConfig [PR #599](https://github.com/openconfig/public/pull/599), it has
+been clarified that the name for the default network instance should be
+uppercase `"DEFAULT"`. Some legacy devices are still using lowercase
+`"default"`, so device tests should use the deviation
+`*deviations.DefaultNetworkInstance` which allows them to work on those legacy
+devices whlie they are being updated. Non-device unit tests may hard-code
+`"DEFAULT"`.
+
```
dplore, line 383: Can this be moved to a new file such as DEVIATIONS.md?   The contributing guide can have a link added.  The  deviations document can include documentation of deviations as desired, but also importantly explain the intended use of deviations as temporary workarounds in order to demonstrate other portions of test and that deviations are treated like bugs, to be resolved in order to gain compliance.

Another option is to add a link to [deviations.go](https://github.com/openconfig/featureprofiles/blob/1bfcd6e1ef943d56c1d340e1a2405e02decf552c/internal/deviations/deviations.go) and expand on the documentation there.  But a markdown may look better, standout as documentation and enable easier to navigate links to specific deviations and their usage.

dplore (commented):
Does this make deviations.DefaultNetworkInstance a standard versus an exception which is to be removed?    ie: the content in deviations.DefaultNetworkInstance is actually the expected value (which of course can be overridden by an argument to the test invocation).

CONTRIBUTING.md
liulk: As discussed, the reason this is in CONTRIBUTIONS.md is because this is being referenced by a pull-request check to explain why the string `"default"` is disallowed. None of the other deviations are checked in this manner.

Rather than adding a DEVIATIONS.md, I just added more detail to the deviations.go flag's help string.

To make sure tests are using the standardized `"DEFAULT"`, I will send a CL to Ondatra to add the constant to netutil, and follow up with another PR to redirect the deviation flag's default value to that constant.

Let me know if there is anything else you'd like me to add.

bstoll (approved):

PR #353, opened 2022-07-28
Fixing failed counter test for unsupported paths

amrindrr:
- Commented out the lookup for unsupported paths as its confirmed that the subinterface discard counters are not supposed to be supported

- Added the await check before the interface config (ConfigureDUTIntf) to make sure interface is in operational state before fetching the counters.

### DIFF
--- a/feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
+++ b/feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
@@ -139,8 +139,8 @@ func TestInterfaceCounters(t *testing.T) {
 		// counter: ipv6Counters.InDiscardedPkts().Lookup(t),
 	}, {
 		// desc: "IPv6OutDiscardedPkts",
-		// TODO: Uncomment counter out-discarded-pkts after the issue fixed.
 		path:    ipv6CounterPath + "out-discarded-pkts",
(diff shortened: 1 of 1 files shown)


feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
```diff
@@ -131,11 +135,13 @@ func TestInterfaceCounters(t *testing.T) {
 	}, {
 		// desc: "IPv6InDiscardedPkts",
 		path:    ipv6CounterPath + "in-discarded-pkts",
-		counter: ipv6Counters.InDiscardedPkts().Lookup(t),
+		// TODO: Uncomment counter in-discarded-pkts after the issue fixed.
+		// counter: ipv6Counters.InDiscardedPkts().Lookup(t),
```
Ankur19, line 139: Instead of skipping based on `desc` and `counter`, we can skip based on a deviation flag. 

```
tc := {
		desc: "IPv6OutDiscardedPkts",
		path:    ipv6CounterPath + "out-discarded-pkts",
		counter: ipv6Counters.OutDiscardedPkts().Lookup,
}

if (*deviations.SubInterfacePacketCountersSupported) {
 if !tc.counter(t).IsPresent() {
    t.Errorf("Get IsPresent status for path %q: got false, want true", tc.path)
 }
}
```

amrindrr:
Creating new PR with CLA authentication

Ankur19 (changes requested):

PR #355, opened 2022-07-28
OTG Test - gNMI-1.10: Telemetry: Basic Check

octpetre:


feature/gnmi/otg_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -173,6 +175,8 @@ func TestInterfaceStatusChange(t *testing.T) {
 }
 
 func TestHardwarePort(t *testing.T) {
+	t.Skip("Hardware port not supported on KNE.")
```
liulk, line 178: @greg-dennis is there a way we can detect from Ondatra API that the test is running against KNE? I was thinking about something like this:

```
if ondatra.IsKNE() {
  t.Skip("Hardware port not supported on KNE.")
}
```

feature/gnmi/otg_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -669,61 +673,76 @@ func TestIntfCounterUpdate(t *testing.T) {
...
+	otgutils.LogFlowMetrics(t, otg, config)
+	ateInPkts := otg.Telemetry().Flow(flowName).Counters().InPkts().Get(t)
+	ateOutPkts := otg.Telemetry().Flow(flowName).Counters().OutPkts().Get(t)
 
 	if ateOutPkts == 0 {
-		t.Errorf("Get(out packets for flow %q: got %v, want nonzero", flow.Name(), ateOutPkts)
+		t.Errorf("Get(out packets for flow %q: got %v, want nonzero", flowName, ateOutPkts)
 	}
-	lossPct := ate.Telemetry().Flow(flow.Name()).LossPct().Get(t)
+	lossPct := float32((ateOutPkts - ateInPkts) * 100 / ateOutPkts)
```
liulk, line 743: float32 conversion needs to happen before the division, otherwise it will already be too late. Alternatively, maybe we can keep it as an integer and below compare it as `if lossPct >= 1` just like elsewhere.

liulk (changes requested):

feature/gnmi/otg_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
octpetre: Kept it as integer.

feature/gnmi/otg_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
greg-dennis: There are a couple options. One is that the WBB bind still has a flag for that, right? So you could perhaps look at the value of that flag. The other is to use the device attribute looks like Model() or Version(), one of which I would imaging has KNE-specific values. I would hesitate to add an IsKNE() lookup, because KNE is one of potentially may different bindings.

liulk: I think in KNE, the dut.Model() would look something like `ARISTA_CEOS`, but there is a TODO suggesting that it will change.

https://github.com/openconfig/ondatra/blob/9c82dc14390cc1018cbc014bd7823b4655ad235c/knebind/solver/solver.go#L219

That also means we need a long list of model exclusions, which is not really practical.

I suppose the featureprofiles binding knows that it is loading the kne binding rather than the static binding. But if someone is using a vendor-specific binding over the plugin interface that happens to use knebind, then we won't know.

https://github.com/openconfig/featureprofiles/blob/25da3d2411ec9754716ddaa62f4d9408c8c1ba6c/topologies/binding/new.go#L60

My counter-proposal is that binding.Dims could add a KNE bool, only set by KNE, so this code:

https://github.com/openconfig/ondatra/blob/9c82dc14390cc1018cbc014bd7823b4655ad235c/knebind/solver/solver.go#L220-L227

Would look like:

```
	dims := &binding.Dims{
		Name:   node.GetName(),
		Vendor: vendor,
		// TODO: Determine appropriate hardware model and software version
		HardwareModel:   typeName,
		SoftwareVersion: typeName,
		Ports:           make(map[string]*binding.Port),
		KNE:             true,  // <-- new.
	}
```

Tests would then use dut.KNE() to query this bool. What do you think?

liulk: To unblock this PR, unconditional skipping is fine for now.

liulk (approved):

PR #357, opened 2022-07-28
Fixing Go fmt errors

amrindrr:
Fixing Go fmt errors


amrindrr:
Will submit new PR with CLA authentication

PR #362, opened 2022-07-30
Clarify the intent of deviations

dplore:


internal/deviations/deviations.go
```diff
@@ -11,12 +11,16 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
-// Package deviations documents and controls vendor deviation
-// workarounds that exist in the featureprofiles test suite.
 //
-// This is a temporary workaround and will be deprecated in the
-// future.
```
robshakir, line 19: please add a blank line here, this will allow the static analysis to identify license header and the subsequent package documentation comment.

internal/deviations/deviations.go
```diff
@@ -11,12 +11,16 @@
...
+// Package deviations documents defines arguments to enable
+// workarounds for a featureprofiles test suite.
+//
+// Deviations are useful to temporarily enable passing for other dependent
+// subcases to give visibility to incremental progress to achieving
+// compliance to test.
+//
+// Passing with a deviation enabled is considered non-compliant to the
+// openconfig featureprofiles test.
```
robshakir, line 23:
```suggestion
// OpenConfig featureprofiles test.
```

internal/deviations/deviations.go
```diff
@@ -11,12 +11,16 @@
...
-// workarounds that exist in the featureprofiles test suite.
 //
-// This is a temporary workaround and will be deprecated in the
-// future.
+// Package deviations documents defines arguments to enable
+// workarounds for a featureprofiles test suite.
+//
+// Deviations are useful to temporarily enable passing for other dependent
+// subcases to give visibility to incremental progress to achieving
+// compliance to test.
```
robshakir, line 20: can we include here:

1. what is the process to add a deviation -- is each one that is added created as a field in this file, with an associated github issue that notes why it exists?
2. what is the process to remove a deviation? i.e., we do not want these to be long-lived things, so there should be some process by which they are removed.
3. if i want to run the /actual/ tests, can we document how this is done -- this should be (imho) the default thing that is being run -- with special flags being needed to involve deviations (ideally, each deviation individually) such that different target types are supported.

robshakir (changes requested):

internal/deviations/deviations.go
dplore: Good idea to clarify.  I proposed some text for you to review

internal/deviations/deviations.go
```diff
@@ -12,11 +12,33 @@
...
+//
+// To remove a deviation
+//   - Submit a pull request which proposes to resolve the relevent
+//     github issue by removing the deviation and it's usage within tests
+//   - Typically the author or an affiliate of the author's organization
+//     is expected to remove a deviation they introduced.
+//
+// To enable the deviations for a given test run
+//   - By default, deviations are not enabled and instead require the
+//     test invocation to set an arguement to enable the deviation.
```
robshakir, line 39:
```suggestion
//     test invocation to set an argument to enable the deviation.
```

robshakir (commented):
this seems good to me, i think it clearly describes the constraints here.

should we have a label for the issues that describe the flags, and/or require that the issue is linked in a description or comment pertaining to the flag that is defined?

liulk:
Fixed the newly introduced default VRF names in #364

internal/deviations/deviations.go
```diff
@@ -12,11 +12,33 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package deviations documents and controls vendor deviation
-// workarounds that exist in the featureprofiles test suite.
+// Package deviations documents defines arguments to enable
```
liulk, line 15: "documents defines arguments" -> "defines the arguments"?

internal/deviations/deviations.go
```diff
@@ -12,11 +12,33 @@
...
+// workarounds for a featureprofiles test suite.
```
liulk, line 16: "for a" -> "for the"?

internal/deviations/deviations.go
```diff
@@ -12,11 +12,33 @@
...
 //
-// This is a temporary workaround and will be deprecated in the
-// future.
+// Deviations are useful to temporarily enable passing for other dependent
```
liulk, line 18: I'm not sure what "enable passing for other dependent subcases" means.

Maybe "enable tests to pass when a device is not OpenConfig compliant"?

internal/deviations/deviations.go
```diff
@@ -12,11 +12,33 @@
...
+// subcases to give visibility to incremental progress to achieving
+// compliance to test.  Deviations should be granular, typically affecting
+// one sub-test, one OpenConfig path or small Openconfig sub-tree.
```
liulk, line 21: Openconfig -> OpenConfig.

internal/deviations/deviations.go
```diff
@@ -12,11 +12,33 @@
...
+//
+// Passing with a deviation enabled is considered non-compliant to the
+// OpenConfig featureprofiles test.
+//
+// To add a deviation
```
liulk, line 26: Add `:`?

internal/deviations/deviations.go
```diff
@@ -12,11 +12,33 @@
...
+//   - Submit a github issue explaining the need for the deviation
+//   - Submit a pull request to add a flag to this file and updates to
+//     the tests where it is intended to be used.
+//
+// To remove a deviation
```
liulk, line 31: Add `:`?

internal/deviations/deviations.go
```diff
@@ -12,11 +12,33 @@
...
+//   - Submit a pull request which proposes to resolve the relevent
+//     github issue by removing the deviation and it's usage within tests
+//   - Typically the author or an affiliate of the author's organization
+//     is expected to remove a deviation they introduced.
+//
+// To enable the deviations for a given test run
```
liulk, line 37: Add `:`?

liulk (changes requested):

internal/deviations/deviations.go
dplore: Would it be useful to include text like this:
`If a test case contains five 'subtests', subtests 2-5 may be dependent on subtest 1.  If a device does not yet implement subcase 1, subcases 2-5 may not be able to pass.  With a deviation for subcase 1, a device can demonstrate passing for subcases 2-5 and non-compliance for subcase 1.   This enhances visibility.`



I would like to avoid suggesting that introducing a deviation is useful to make a test pass as that can be interpreted  very broadly.

liulk: How about this? "Deviations are useful to temporarily bypass non-compliant failure conditions using workarounds so further testing could be done."

This wording avoids any mentioning of enabling or passing.

dplore: Done

internal/deviations/deviations.go
```diff
@@ -12,11 +12,34 @@
...
 
-// Package deviations documents and controls vendor deviation
-// workarounds that exist in the featureprofiles test suite.
+// Package deviations defines the arguments to enable temporary workarounds
+// for the featureprofiles test suite.
 //
-// This is a temporary workaround and will be deprecated in the
-// future.
+// Deviations may be introduced to temporarily work around non-compliant issues
+// so further sub-tests can be implmeneted.  Deviations should be
```
liulk, line 19: implmeneted -> implemented

internal/deviations/deviations.go
```diff
@@ -12,11 +12,34 @@
...
+// small in scope, typically affecting one sub-test, one OpenConfig
+// path or small OpenConfig sub-tree.  Deviations are enabled using
+// commmand line parameters.
```
liulk, line 22: commmand -> command

internal/deviations/deviations.go
```diff
@@ -12,11 +12,34 @@
...
+//
+// Passing with a deviation enabled is considered non-compliant to the
+// OpenConfig featureprofiles test.
+//
+// To add a deviation:
+//   - Submit a github issue explaining the need for the deviation
```
liulk, line 28: Let's punctuate the sentence with a period?

internal/deviations/deviations.go
```diff
@@ -12,11 +12,34 @@
...
+//   - Submit a pull request referencing the above issue to add a flag to
+//     this file and updates to the tests where it is intended to be used.
+//
+// To remove a deviation:
+//   - Submit a pull request which proposes to resolve the relevent
```
liulk, line 33: Period?

liulk (changes requested):

internal/deviations/deviations.go
```diff
@@ -16,21 +16,21 @@
...
 //
 // To add a deviation:
-//   - Submit a github issue explaining the need for the deviation
+//   - Submit a github issue explaining the need for the deviation.
 //   - Submit a pull request referencing the above issue to add a flag to
 //     this file and updates to the tests where it is intended to be used.
 //
 // To remove a deviation:
-//   - Submit a pull request which proposes to resolve the relevent
+//   - Submit a pull request which proposes to resolve the relevant.
```
robshakir, line 33: This looks like a full stop in the middle of a sentence now?

internal/deviations/deviations.go
liulk: Sorry, my bad here.

robshakir (approved):

liulk (approved):

PR #369, opened 2022-08-02
Interface packet counter deviation 

amrindrr:
This change add deviation functionality for ipv4/ipv6 discard packet counters. By default it checks for discard counters and the test will fail if paths are not supported.

To pass the test, add deviation (deviation_subinterface_counter_supported) - that will skip the discard counter lookup


amrindrr:
Duplicate : https://github.com/openconfig/featureprofiles/pull/372

PR #373, opened 2022-08-02
Sub Interface packet counter deviation

amrindrr:
This change add deviation functionality for ipv4/ipv6 discard packet counters. By default it checks for discard counters and the test will fail if paths are not supported.

To pass the test, add deviation (deviation_subinterface_counter_supported) - that will skip the discard counter lookup

### DIFF
--- a/feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
+++ b/feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
@@ -18,6 +18,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/openconfig/featureprofiles/internal/deviations"
 	"github.com/openconfig/featureprofiles/internal/fptest"
 	"github.com/openconfig/ondatra"
 	"github.com/openconfig/ondatra/telemetry"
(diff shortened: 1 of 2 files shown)


feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
```diff
@@ -96,49 +97,58 @@ func TestInterfaceCounters(t *testing.T) {
...
 		path    string
-		counter *telemetry.QualifiedUint64
-	}{{
+		counter func(testing.TB) *telemetry.QualifiedUint64
+	}
+	type counterCases []counterCase
+	cases := make(counterCases, 0)
+	cases = append(cases, counterCase{
 		desc:    "InUnicastPkts",
 		path:    intfCounterPath + "in-unicast-pkts",
```
sezhang2, line 109: Can we use only variable cases instead of 2?

feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
```diff
@@ -96,49 +97,58 @@ func TestInterfaceCounters(t *testing.T) {
...
-		counter: intfCounters.InUnicastPkts().Lookup(t),
-	}, {
+		counter: intfCounters.InUnicastPkts().Lookup,
```
sezhang2, line 110: what is the reason to replace Lookup(t) with Lookup?

feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
amrindrr: "cases" is variable of type counterCases. This is required since only typed slice variables can be used with "append" function.

feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
amrindrr: Lookup() is a fatal function which is causing the test to fail if the path is not valid. Definition of "counter" is changed to lookup which is in turn calling lookup() function only for the supported test cases (causing the test to pass)
So the lookup() function will be called in the testing loop via counter(t).

internal/deviations/deviations.go
```diff
@@ -30,4 +30,7 @@ var (
 		"Device requires that aggregate Port-Channel and its members be defined in a single gNMI Update transaction at /interfaces; otherwise lag-type will be dropped, and no member can be added to the aggregate.")
 
 	DefaultNetworkInstance = flag.String("deviation_default_network_instance", "DEFAULT", "The name used for the default network instance for VRF.  This has been standardized in OpenConfig as \"DEFAULT\" but some legacy devices are using \"default\"; tests should use this deviation as a temporary workaround.")
+
+	SubInterfacePacketCountersSupported = flag.Bool("deviation_subinterface_counter_supported", true,
```
Ankur19, line 34: nit:  `deviation_subinterface_counter_supported` should this be `deviation_subinterface_packet_counters_supported`?

sezhang2 (approved):

internal/deviations/deviations.go
amrindrr: Done

Ankur19 (approved):
LGTM!

PR #379, opened 2022-08-04
Get component list by type instead of matching the component name

sezhang2:
- Added func findComponentsByType()

Test passed:
  - http://sponge2/1e6c1475-0172-4885-81f4-79c1af9f4a72

### DIFF
--- a/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
+++ b/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
@@ -32,10 +32,10 @@ import (
 )
 
 const (
-	controlcardType = telemetry.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_CONTROLLER_CARD
-	linecardType    = telemetry.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_LINECARD
-	activeController    = telemetry.PlatformTypes_ComponentRedundantRole_PRIMARY
-	standbyController    = telemetry.PlatformTypes_ComponentRedundantRole_SECONDARY
+	controlcardType   = telemetry.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_CONTROLLER_CARD
+	linecardType      = telemetry.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_LINECARD
+	activeController  = telemetry.PlatformTypes_ComponentRedundantRole_PRIMARY
+	standbyController = telemetry.PlatformTypes_ComponentRedundantRole_SECONDARY
 )
 
 func TestMain(m *testing.M) {
--- a/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
+++ b/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
@@ -215,7 +215,7 @@ func findComponentsByType(t *testing.T, dut *ondatra.DUTDevice, cType telemetry.
 
 			switch v := componentType.(type) {
 			case telemetry.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT:
-				if v ==cType {
+				if v == cType {
 					s = append(s, c)
 				}
 			default:
--- a/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
+++ b/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
@@ -75,7 +75,7 @@ func TestMain(m *testing.M) {
 //    https://github.com/fullstorydev/grpcurl
 //
 
-func TestStandbySupervisorReboot(t *testing.T) {
+func TestStandbyControllerCardReboot(t *testing.T) {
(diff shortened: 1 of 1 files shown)


feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
```diff
@@ -200,12 +202,28 @@ func TestLinecardReboot(t *testing.T) {
...
+			t.Logf("Component %s has type: %v", c, componentType)
+
+			switch v := componentType.(type) {
+			case telemetry.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT:
+				switch v {
+				case cType:
+					s = append(s, c)
+				default:
+					continue
+				}
```
wenovus, line 223:
```suggestion
				if v == cType {
					s = append(s, c)
				}
```

feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
```diff
@@ -200,12 +202,28 @@ func TestLinecardReboot(t *testing.T) {
...
+			default:
+				t.Fatalf("Error extracting component state, could not type assert union. union: %v", componentType)
```
wenovus, line 225:
```suggestion
				t.Fatalf("Expected component type to be a hardware component, got (%T, %v)", componentType, componentType)
```

wenovus (approved):

dplore:
nit on naming:  Can we replace the references to `supervisor` with `controllercard`?  This way the names we are using in the test code are using the same names as  openconfig.  For example:

TestStandbySupervisorReboot becomes: TestStandbyControllerCardReboot

sezhang2:
> nit on naming: Can we replace the references to `supervisor` with `controllercard`? This way the names we are using in the test code are using the same names as openconfig. For example:
> 
> TestStandbySupervisorReboot becomes: TestStandbyControllerCardReboot

Done

feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
sezhang2: Done

feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
sezhang2: Done

sezhang2:
Test passed after the change:
  - http://sponge2/d673e6c5-9574-4856-ac8d-48512a196b1a

aaronmillisor (commented):
While this commit is improving the vendor neutrality in the detection of controller cards, the detection of  which one is standby is being done based on the description of the component, which is not vendor neutral. Should this change also handle this issue, or be a separate PR?

sezhang2:
> 

I will make changes in the next PR based on redundant-role. Does it work for Cisco?

aaronmillisor:
That works! Thanks!

sezhang2:
> That works! Thanks!

I just made the changes based on redundant-role.

feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
```diff
@@ -214,12 +230,14 @@ func findMatchedComponents(t *testing.T, dut *ondatra.DUTDevice, r *regexp.Regex
...
+		role := dut.Telemetry().Component(supervisor).RedundantRole().Get(t)
+		t.Logf("Component(supervisor).RedundantRole().Get(t): %v, Role: %v", supervisor, role)
+		if role == standbyController {
 			standbyRP = supervisor
-		} else {
+		} else if role == activeController {
 			activeRP = supervisor
+		} else {
+			t.Fatalf("Expected controller %s to be active or standby, got %v", supervisor, role)
 		}
```
wenovus, line 241: (optional nit): a switch statement here could be easier to read since you're checking the value of a single variable.

feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
```diff
@@ -214,12 +230,14 @@ func findMatchedComponents(t *testing.T, dut *ondatra.DUTDevice, r *regexp.Regex
 func findStandbyRP(t *testing.T, dut *ondatra.DUTDevice, supervisors []string) (string, string) {
 	var activeRP, standbyRP string
```
wenovus, line 231: (no action required on this PR) If these remain empty strings it means (unless a supervisor is the empty string) that it was not found. Should this be documented as a doc comment for the function? If this is not expected should an error be returned? What happens if there are multiple supervisors with the same role?

Considering this issue was here before I think this is best addressed in a separate PR.

wenovus (approved):

feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
sezhang2: Ack. We fail the test if the number of controller is < 2.

Will add the check to expect one active and one standby controller in a separate PR.

PR #382, opened 2022-08-04
OTG Test - RT-1.5: BGP Prefix Limit

octpetre:


octpetre:
@dplore or @liulk could you please assign a reviewer?
Thanks,

liulk (approved):

PR #386, opened 2022-08-05
tcheck library for table-driven telemetry tests

dan-lepage:
This package provides a way to do table-driven testing of Ondatra telemetry paths with varying node types and minimal boilerplate. See the module documentation for more details.


bstoll (approved):
This LGTM but it might be good to have @liulk look it over (if he hasn't already).

liulk (dismissed):
The `/* ... */` styled comments does seem a little out of place because most of the code elsewhere already use `//` styled comment. But if this practice is confined to the package, it should be okay.

dan-lepage:
@liulk I added an `AwaitUntil` method, which solves the problem of having dozens of checks each wait - now you can choose a deadline, e.g. `time.Now().Add(time.Second)`, and make repeated `AwaitUntil` calls, and if the device is failing the whole set will only wait as long as the deadline.

liulk (approved):

PR #387, opened 2022-08-05

sezhang2:
Look up the controller card by type instead of matching the component name.

### DIFF
--- a/feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go
+++ b/feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go
@@ -15,23 +15,26 @@ package supervisor_switchover_test
 
 import (
 	"context"
-	"regexp"
 	"sort"
-	"strings"
 	"testing"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/openconfig/featureprofiles/internal/fptest"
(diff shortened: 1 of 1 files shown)


feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go
```diff
@@ -162,29 +163,49 @@ func TestSupervisorSwitchover(t *testing.T) {
...
+			default:
+				t.Fatalf("Expected component type to be a hardware component, got (%T, %v)", componentType, componentType)
+			}
 		}
 	}
 	return s
 }
 
-func findActiveRP(t *testing.T, dut *ondatra.DUTDevice, supervisors []string) (string, string) {
+func findStandbyRP(t *testing.T, dut *ondatra.DUTDevice, supervisors []string) (string, string) {
```
wenovus, line 190: Since this is named `findStandbyRP`, should this return just the standbyRP, or should it have a different name to be more accurate?

wenovus (approved):

feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go
sezhang2: Agreed. Since the similar function is also used in other test code, I plan to move them to a library as helper function and update the function name as well.

PR #390, opened 2022-08-08
Cleanup speed == 99 to use PMD == 100G-FR

liulk:


robshakir (approved):
Thanks for this -- glad to see the magic number going away :-)

PR #396, opened 2022-08-09
Move leader_failover_test out of experimental

Ankur19:
8ba20945-8c41-410f-91a5-97f8a8134e51


mojiiba:
Hi @Ankur19 this tests uses DELETE mode for gribi session that was not a mandatory requirement to implement. Can we still  make the DELETE tests optional if we move out this tests to gribi folder?

liulk:
Let's complete the move for now, and I'll make a follow up PR. I have a task to make gRIBI DELETE client optional in tests.

liulk (approved):

PR #400, opened 2022-08-10
aggregate_test verify DUT oper-status before configure ATE

liulk:


liulk:
Invocation ID: 659613f9-9e31-4d4d-be08-7a0107e65780

feature/interface/aggregate/ate_tests/aggregate_test/aggregate_test.go
```diff
@@ -294,40 +292,47 @@ const (
...
 	}
 }
+
 func (tc *testCase) verifyInterfaceDUT(t *testing.T, dp *ondatra.Port) {
 	dip := tc.dut.Telemetry().Interface(dp.Name())
 	di := dip.Get(t)
 	fptest.LogYgot(t, dp.String(), dip, di)
+
+	// LAG members may fall behind, so wait for them to be up.
+	dip.OperStatus().Await(t, time.Minute, opUp)
```
bstoll, line 309: di needs to be collected after the Await (or just use the values returned by Await) doesn't it?  If the Await is doing anything, the values saved in di are the old states, and they will just fail in the next few lines even if the interface does come up after a delay.

liulk: But if I move the dip.Get(t) after the await, then if the interface is down, then await will fatal out and di is never logged.

liulk: Removed the di.GetOperStatus() check which used potentially stale value.

bstoll (approved):

PR #401, opened 2022-08-11
Update traceroute_test.go

sezhang2:
Enabled the traceroute test cases and replaced the fake response

Test passed on different platforms.


feature/gnoi/system/tests/traceroute_test/traceroute_test.go
```diff
@@ -82,7 +83,8 @@ func TestMain(m *testing.M) {
 func TestGNOITraceroute(t *testing.T) {
 	dut := ondatra.DUT(t, "dut")
 
-	lo0 := dut.Telemetry().Interface("Loopback0").Subinterface(0)
+	lbIntf := netutil.LoopbackInterface(t, dut, 0)
```
mojiiba, line 86: netutil.LoopbackInterface only works if loopback 0 is configured on DUT with ipv4/6. We either need to configure the looback in test or change netutil.LoopbackInterface to configure the loopback if it is not configured.

sezhang2: Loopback 0 should be configured with IPv4/IPv6 addresses as the base config of the device. In order to run the test, the device needs to be reachable with the base config. Since it is a functionality test, we also try not to touch other features such as interface config using gnmi.

mojiiba: Thanks for the reply. This assumption is not consistent with other featureprofile tests. We configure interfaces in all tests. In some tests we even configure default network instance which is already exist on the device without the base config. That would be great if we do not make an assumption for the base config, or at least mention the assumption in  readme.

sezhang2: I will update readme files for tests using loopback0 in the next pull.

I will also follow-up with base config requirement for feature profile test.

dplore:
FYI, created #405 for documenting base config expectations

mojiiba:
> FYI, created #405 for documenting base config expectations

thanks a lot!

bstoll (approved):

PR #403, opened 2022-08-11
TE-11.2: Backup NHG: Multiple NH

### LINKED ISSUE #70: TE-11.2: Backup NHG: Multiple NH

manan-patel:
Added TC TE-11.2: Backup NHG: Multiple NH 
Uplifted GRIBI NHG call for supporting additional parameters 
Added Flush API

Resolves #70

### DIFF
--- a/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
+++ b/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
@@ -32,9 +32,9 @@ import (
 const (
 	ipv4PrefixLen = 30
 	ipv6PrefixLen = 126
-	dstPfx        = "1.0.0.0/8"
-	dstPfxMin     = "1.0.0.0"
-	dstPfxMask    = "8"
+	dstPfx        = "203.0.113.0/24"
+	dstPfxMin     = "203.0.113.0"
+	dstPfxMask    = "24"
 )
 
 // testArgs holds the objects needed by a test case.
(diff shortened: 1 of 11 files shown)


manan-patel:
As per the WBB vendor plan ipv4 address, 1.0.0.0/8 needs to be used. Can you please confirm which address needs to be used

dplore:
> As per the WBB vendor plan ipv4 address, 1.0.0.0/8 needs to be used. Can you please confirm which address needs to be used

Thank you for the contribution.  In this case, lets use the [contribution guide - ip address assignment](https://github.com/openconfig/featureprofiles/blob/main/CONTRIBUTING.md#ip-addresses-assignment) guidelines.

So for this test, the destination prefix should be changed to be within the range `TEST-NET-3: (203.0.113.0/24): data plane destination network addresses used for traffic testing; split as needed.`.      Since this test doesn't call for multiple subnets,  think you could use `203.0.113.0/24`

manan-patel:
Hi Darren, can you please assign a reviewer?

dplore:
Hi thanks for your contribution!  I will assign a reviewer now and queue as candidate for our next sprint.  It may be 1-2 weeks before there is action.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -0,0 +1,381 @@
...
+	return top
+}
+
+// configureDUT configures port1, port2, port3 and port4 on the DUT.
+func configureDUT(t *testing.T, dut *ondatra.DUTDevice) {
+	d := dut.Config()
+
+	p1 := dut.Port(t, "port1")
+	i1 := &telemetry.Interface{Name: ygot.String(p1.Name())}
+	d.Interface(p1.Name()).Replace(t, configInterfaceDUT(i1, &dutPort1))
```
bstoll, line 168: There is a helper in the attrs library that will simplify this a bit and get rid of configInterfaceDUT:

`d.Interface(p1.Name()).Replace(t, dutPort1.NewInterface(p1.Name()))`

The helper looks like it does the same thing as configInterfaceDUT(), but it also covers some deviations for some devices with bugs.

bstoll (approved):
Thanks, this LGTM.

bstoll (approved):

internal/gribi/gribi.go
```diff
@@ -63,6 +63,25 @@ func (c *Client) Fluent(t testing.TB) *fluent.GRIBIClient {
...
 }
 
+// OptionalParam is used to pass additional parameters to gribi APIs
+type OptionalParam interface {
+	// IsTestOpt marks the TestOpt as implementing this interface.
+	IsOptionalParam()
+}
+
+// BKNHG is used to pass the backup nexthop group for adding next hop group
+type BKNHG struct {
```
liulk, line 73: Maybe we don't need an interface, just make this an NHGOptions struct, rename id to BackupNHG, and only apply this if id > 0?

liulk (changes requested):

internal/gribi/gribi.go
manan-patel: The reason I am using an interface is that we may need to extend current function parameters, add other functions in gribi file in the future and I want to avoid changing calls on the caller's side. I can still make the change if you prefer struct.

bstoll (dismissed):

bstoll:
> /gcbrun

This PR is older and the Cloud Build task fails.  If this was rebased to include #526 it should pass, but I don't believe the changes here are relevant to the cloud build check either.

internal/gribi/gribi.go
liulk: This should not require the caller to change anything when you add new fields:

```
client := &gribi.Client{...}
client.AddNHG(t, 42, ..., &gribi.NHGOptions{BackupNHG: 10})
```

Besides, gribi is an internal package, so the usage is limited to this repo which we have full control over. Someone who introduces a breaking change could see this in pull request checks and will be able to fix it.

bstoll:
@liulk I included the suggested change in https://github.com/openconfig/featureprofiles/pull/403#discussion_r981556680 .  I had to rebase the PR to make this merge successfully with the AddNHG changes.  Does this look good?

liulk (approved):

PR #408, opened 2022-08-12
Branch TE-6.2 out of TE-6.1

xw-g:
Branch non-default VRF scenario out of TE-6.1, and make its own TE-6.2. 

The logic of TE-6.2 is already LGTMed internally by @nflath.


feature/gribi/ate_tests/route_removal_non_default_vrf/README.md
```diff
@@ -0,0 +1,56 @@
+# TE-6.2: Route Removal In Non Default VRF
```
liulk, line 1: The folder name need to end in _test so we know it's a test package.

`feature/gribi/ate_tests/route_removal_non_default_vrf_test`

liulk (changes requested):
Mostly LGTM, just have 1 comment.

feature/gribi/ate_tests/route_removal_non_default_vrf/README.md
xw-g: Done. Thanks!

liulk (approved):

Issue #383, opened 2022-08-04
Issues with RT-5.1 test (Singleton Interface)


mojiiba:
Hi, 
There are two issue with [RT-5.1](https://github.com/openconfig/featureprofiles/tree/main/feature/interface/singleton/ate_tests/singleton_test). Can u please help to resolve the issues.
- The test sets the MTU at the subinterface/ipv4 and ipv6 level (https://github.com/openconfig/featureprofiles/blob/d29b222f35ae6bf795ad667e08f67e63218ac9f6/feature/interface/singleton/ate_tests/singleton_test/singleton_test.go#L121) without setting the MTU for the physical interface. The MTU for the interface should be set to tc.MTU+14 (`i.Mtu = ygot.Uint16(tc.mtu + 14)`), otherwise the IPV4/6 should be dropped that will result in test failure. 
- The test assumes that the number of output unicast packets from ATE that is captured at the flow level should be equal or larger than the number of inbound unicast packets  by the DUT (https://github.com/openconfig/featureprofiles/blob/d29b222f35ae6bf795ad667e08f67e63218ac9f6/feature/interface/singleton/ate_tests/singleton_test/singleton_test.go#L337). This assumption is not correct due to 
1- The flow statistics does not account for the control packets, but DUT's interface counter does account for it.
2- Also, ATE sends ICMPv6 since the ipv6 is configured on the ATE interface. Below figure shows that three control packet is received by the DUT that  results in inbound packet to be larger than (+3) outbound packets.  

<img width="1386" alt="Screen Shot 2022-08-04 at 11 58 03 AM" src="https://user-images.githubusercontent.com/91647529/182947277-e27dd28e-76e7-4544-b223-28f0ebb65ed6.png">

To fix this issue, either we should have a threshold (+3) to account for the control packets or change the test logic to read the number of packets in a way to account for the control packets. I think ATE statistics at the port level  account for control packets.

earies:
@dplore @bstoll - I would suggest if any IP (or upper layer) MTU is being altered that a prerequisite would be to alter the interface MTU otherwise IP MTUs cannot be set.  We should not introduce any internal implicit behaviors of treatment to the interface MTU if only an IP MTU is set but rather this should be explicit intent.

It is likely the YANG node descriptions for the `mtu` leafs should be updated to reflect this behavioral expectation as well

liulk:
[...]

PR #412, opened 2022-08-15
Rectified deviation so "true" always indicates deviated behavior.

liulk:


sezhang2 (approved):

internal/deviations/deviations.go
```diff
@@ -12,22 +12,25 @@
...
+// sub-tests can be implemented.  Deviations should be small in scope, typically affecting
+// one sub-test, one OpenConfig path or small OpenConfig sub-tree.  Deviations are enabled
+// using command line parameters.
 //
-// Passing with a deviation enabled is considered non-compliant to the
-// OpenConfig featureprofiles test.
+// If a device only passes a test by setting the deviation to true, that is considered
+// non-compliant behavior.  Conversely, if a device only passes a test by setting the
+// deviation to false, that is also non-compliant.  Compliant devices must pass with the
+// deviation flags set to either true or false; failing in either case is non-compliant.
```
dplore, line 26: I think a device is compliant if it passes without deviation.  Enabling a deviation I think should mean the test results as non-compliant with the featureprofiles test requirements.  In the future, we can introduce properties to the test log output to programmatically indicate that a deviation is enabled.

I can see scenarios where a compliant device cannot pass with a deviation enabled. Any deviation defined which requires a non-compliant behavior would fall into this set of scenarios.  For example, usage of the default network instance name.  A device can only have one default name for an instance and it must be 'DEFAULT'.  If there is a deviation enabled to define the default network instance name to 'default' and the test is run with this deviation enabled, a compliant device would fail.

dplore (commented):
I like the requirement for deviation defaults to false, but I have a question about a compliant device passing with and without deviations.

internal/deviations/deviations.go
liulk: I didn't mention default VRF because it's not a boolean flag, but I can see why someone might confuse the wording to construe not being able to use a different name for default VRF as non-compliant. I have adjusted the wording in the help string to clarify this particular exception, but I'm also open to renaming `--deviation_default_network_instance` because it is impossible to create a compliant workaround with a different value.

The main idea is that workarounds should only use compliant behavior, and I would like to see tests adhere to this principle unless there is a good reason.

liulk: Upon closer reading of the standard:

https://github.com/openconfig/public/blob/master/release/models/network-instance/openconfig-network-instance.yang#L1178-L1185

It looks like not being able to use an operator assigned value "default" or otherwise is non-compliant. So we don't have to make this an exception after all.

internal/deviations/deviations.go
```diff
@@ -12,22 +12,27 @@
...
+// A test may introduce deviations in order to temporarily work around non-compliant
+// issues so further sub-tests can make progress.  Deviations should be small in scope,
+// typically affecting one sub-test, one OpenConfig path or small OpenConfig sub-tree.
+// Deviations are enabled using command line parameters.
 //
-// Passing with a deviation enabled is considered non-compliant to the
-// OpenConfig featureprofiles test.
+// The intended behavior to be tested should always be without the deviation, so if a
+// device could not pass without deviation, that is considered non-compliant behavior.
+// Ideally, device should pass both with and without a deviation which means the deviation
```
dplore, line 25: I think we should be very clear that passing with deviation is always non-compliant.  The additional information is good description to elaborate further on behavior when deviations are enabled, but we should keep some simple statement that passing with deviation is non-compliant.

internal/deviations/deviations.go
```diff
@@ -47,13 +52,14 @@ import "flag"
 // Vendor deviation flags.
 var (
 	InterfaceEnabled = flag.Bool("deviation_interface_enabled", false,
-		"Device requires interface enabled leaf booleans to be explicitly set to true.")
+		"Device requires interface enabled leaf booleans to be explicitly set to true.  Fully-compliant devices should pass both with and without this deviation.")
```
dplore, line 55: I think adding this additional description can cause confusion.  Passing without deviations is required for compliance.  Passing with a deviation is not compliant.  If the deviation is enabled, one cannot tell if the device is compliant.  The deviation must be disabled to know if the device is compliant.

dplore (commented):
This is close but I'd still like more clarity that passing with a deviation is not compliant.

internal/deviations/deviations.go
liulk: What you are saying is that a device that passes without deviation is still non-compliant if it happens to also pass with deviation, and this is not correct.

Most of the existing deviations are implemented using only fully compliant behavior. The only exception so far is the one we will have to reintroduce for #383 to omit setting L2 MTU as a deviation. That's because if we have these scenarios:

  - Device A: enforces that L2 MTU >= L3 MTU and L2 MTU default is 1514. (compliant)
  - Device B: allows L3 MTU to override L2 MTU. (also compliant)
  - Device C: can only set L3 MTU but not L2 MTU (not compliant)

Here is the matrix for a deviation that omits the L2 MTU, for a test that wants to set L3 MTU to 9000.

| Device  | Without Deviation | With Deviation |
| ---- | ---- | ---- |
| Device A  | Pass | Fail (L3 MTU 9000 > L2 MTU 1514) |
| Device B | Pass | Pass |
| Device C | Fail (cannot set L2 MTU) | Pass |

As you can see, if we say "passing with deviation is always non-compliant" then you would disqualify Device B.

internal/deviations/deviations.go
liulk: As discussed yesterday, the deviation itself should only be implemented using fully-compliant mechanisms.

* Device A: cannot set the subinterface enabled leaf. (not compliant)
* Device B: requires setting the interface and subinterface enabled leaf to true. (not compliant)
* Device C: interface and subinterface are both enabled by default (compliant)

| Device | Without Deviation | With Deviation |
| ---- | ---- | ---- |
| Device A | Pass | Fail (can't set leaf) |
| Device B | Fail (interface and subinterface are disabled) | Pass |
| Device C | Pass | Pass |

Again, you would favor Device A as compliant even though it is not, and disqualify Device C even though it is actually fully-compliant.

mojiiba: @liulk  concerning `deviation_interface_enabled`, I think we need to treat ipv4/enabled and ipv6/enabled differently here.  I do not fully understand why a device that does not support setting ipv4/enabled (Device A) is not compliant, if it supports ipv4/enabled by default. 

I understand this is a deviation with OC, but that would be great if we focus on use cases for FP deviations. Do we have a use case that this leafs need to be used when device (i.e., Device C) enable ipv4 by default?

liulk: I originally considered introducing a --deviation_subinterface_enabled in a follow up PR, but it sounds like you want separate --deviation_subinterface_ipv4_enabled and --deviation_subinterface_ipv6_enabled? What is the rationale for different IPv4 and IPv6 enabled behaviors?

Here is the OC verbiage, and this is the criteria I used for determining OC compliance.

* `/interfaces/interface/config/enabled` (default true) - This leaf contains the configured, desired state of the interface. Systems that implement the IF-MIB use the value of this leaf in the 'running' datastore to set IF-MIB.ifAdminStatus to 'up' or 'down' after an ifEntry has been initialized, as described in RFC 2863. Changes in this leaf in the 'running' datastore are reflected in ifAdminStatus, but if ifAdminStatus is changed over SNMP, this leaf is not affected.
* `/interfaces/interface/subinterfaces/subinterface/config/enabled` (default true) - This leaf contains the configured, desired state of the interface. Systems that implement the IF-MIB use the value of this leaf in the 'running' datastore to set IF-MIB.ifAdminStatus to 'up' or 'down' after an ifEntry has been initialized, as described in RFC 2863. Changes in this leaf in the 'running' datastore are reflected in ifAdminStatus, but if ifAdminStatus is changed over SNMP, this leaf is not affected.
* `/interfaces/interface/subinterfaces/subinterface/ipv4/config/enabled` (default true) - Controls whether IPv4 is enabled or disabled on this interface. When IPv4 is enabled, this interface is connected to an IPv4 stack, and the interface can send and receive IPv4 packets.
* `/interfaces/interface/subinterfaces/subinterface/ipv6/config/enabled` (default true) - Controls whether IPv6 is enabled or disabled on this interface. When IPv6 is enabled, this interface is connected to an IPv6 stack, and the interface can send and receive IPv6 packets.

There is also the "unnumbered" leaves which allows the L3 stack to be turned on without an L3 address. Currently, all our tests assign an IP address.

* `/interfaces/interface/subinterfaces/subinterface/ipv4/unnumbered/config/enabled` (default false) - Indicates that the subinterface is unnumbered. By default the subinterface is numbered, i.e., expected to have an IP address configuration.
* `/interfaces/interface/subinterfaces/subinterface/ipv6/unnumbered/config/enabled` (default false) - Indicates that the subinterface is unnumbered. By default the subinterface is numbered, i.e., expected to have an IP address configuration.

This means:

* Device A: cannot set the subinterface enabled leaf.
    * Not compliant because it doesn't support a path that is clearly defined in OC.
* Device B: requires setting the interface and subinterface enabled leaf to true.
    * Not compliant because OC default is enabled true, so the test should not have to set them.
* Device C: interface and subinterface are both enabled by default.
    * Compliant according to OC.

Now, I understand that Darren has a plan to adjust the OC verbiage to remove the ipv4/enabled and ipv6/enabled leaves. That is fine, and that will make both Device A and Device C compliant, and Device B will fail both with and without deviations because the test will not be able to set subinterface enabled, which becomes non-compliant under new OC verbiage.

What I want to establish here is the deviations guideline so that it adheres to the OC verbiage. This is not a good forum to discussing the merits of the OC verbiage itself.

Does that make sense?

mojiiba: Thanks for the clarification!

dplore: We may have different definitions for compliance.  IMO, the definition of compliance for featureprofiles test is the DUT is compliant to the requirements of the test.  This is a useful definition I think for the following reasons:

* An operator wants to be assured that a specific configuration will yield the exact same behavior on all devices which pass the test.
* OpenConfig and network devices in general may provide more than one way to achieve a desired behavior.

If a test passes with a deviation enabled, the operator needs to know the behavior requirement in the test is not being met.   The configuration the operating is generating is either not fully supported or will have some different behavior (deviation) than what the test is requiring.  This is why I think it is useful to defining passing with a deviation as not compliant with the test.  

I do agree that we can require that deviations only implement openconfig.

dplore (changes requested):
I see one last change regarding the assertion to "use any operator-assigned values" for the DEFAULT_INSTANCE.

internal/deviations/deviations.go
liulk: Just for posterity to summarize my part of our discussion: my definition of compliance targets the OpenConfg public model rather than the test, and that's because:

 1. Tests should not use behavior outside of OpenConfig even when implementing a deviation.
 2. Tests are not written in a rigorous language that could be used to standardize OpenConfig behavior.

If we make tests be the compliance target, it can contradict the language in the OpenConfig public model and become a "de facto" standard even though they do not go through the same rigorous standardization process. But our definitions of compliance will converge when we agree that all tests, with or without deviations, only depend on OpenConfig behavior.

Tests typically only test a subset of OpenConfig, so they are already more lenient and represent the "low bar" of OpenConfig compliance. Deviation can make them even more lenient, and that's why the bar is even lower. Device should meet the higher bar without the deviation.

internal/deviations/deviations.go
```diff
@@ -47,13 +61,14 @@ import "flag"
...
+	AggregateAtomicUpdate = flag.Bool("deviation_aggregate_atomic_update", false,
+		"Device requires that aggregate Port-Channel and its members be defined in a single gNMI Update transaction at /interfaces; otherwise lag-type will be dropped, and no member can be added to the aggregate.  Fully-compliant devices should pass both with and without this deviation.")
 
-	DefaultNetworkInstance = flag.String("deviation_default_network_instance", "DEFAULT", "The name used for the default network instance for VRF.  This has been standardized in OpenConfig as \"DEFAULT\" but some legacy devices are using \"default\"; tests should use this deviation as a temporary workaround.")
+	DefaultNetworkInstance = flag.String("deviation_default_network_instance", "DEFAULT",
+		"The name used for the default network instance for VRF.  The default name in OpenConfig is \"DEFAULT\" but some legacy devices still use \"default\".  Fully-compliant devices should be able to use any operator-assigned values.")
```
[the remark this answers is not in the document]
liulk: I added some language in the comment to differentiate between "OpenConfig compliance" and "test plan compliance" tiers. I think this way we have the correct vocabulary to state the deviation behavior requirements without confusing each other about what compliance means.

internal/deviations/deviations.go
aaronmillisor: To comment quickly on this part:

> it sounds like you want separate --deviation_subinterface_ipv4_enabled and --deviation_subinterface_ipv6_enabled? What is the rationale for different IPv4 and IPv6 enabled behaviors?

I believe my opinion here is a bit outside of the scope of the feature profile itself as it pertains to the merits of the ipv4 enabled leaf, but I would say that most implementations have a mechanism for having 'ipv6 enabled' which turns on ipv6 neighbor discovery and link-local addressing. In IPv6 it is possible for two hosts to communicate by simply 'turning on' ipv6, which is not the case in ipv4. This is a fundamental difference between ipv4 and ipv6, and can result in a difference in vendor behavior when doing mapping to openconfig as the meaning of ipv4 enabled is considerably more ambiguous than the the meaning of ipv6 enabled.

liulk: Thanks for the clarification. Our tests using this deviation always assigns a global address, so this deviation flag is not meant to test unnumbered stack turnup.

We don't have a test plan written for the unnumbered stack behavior. In the future if someone adds a test for that, I think the test will additionally have to set `ipv6/unnumbered/config/enabled` to true per OC spec. The test will probably be very specific about which "enabled" combination to use, so it will not use a deviation.

internal/deviations/deviations.go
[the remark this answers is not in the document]
dplore: Thanks, it helps.  I like the definitions and clarity that featureprofiles tests assert 'tier 1' compliance.

dplore (approved):
Thanks for all the discussion and updates!

Issue #273, opened 2022-06-27
cli_origin_test.go : "shutdown" interface operation to be vendor-agnostic


avaneesh90:
Hi,

Request cli_origin.go script  "shutdown" interface operation to be vendor-agnostic for other vendor that use different operation command to bring down interface.

**Currently script is usring rpc call:** 

<rpc xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="2"><load-configuration action = "merge" origin="juniper" format="text" path =""><configuration-text>
interface et-0/0/5
  shutdown   **>>>>>> This should be vendor-agnostic**
</configuration-text></load-configuration></rpc>
]]>]]>

var intfConfig string = `
interface %s
  %s
`
	cases := []struct {
		desc                string
		intfOper            string
		expectedAdminStatus telemetry.E_Interface_AdminStatus
		expectedOperStatus  telemetry.E_Interface_OperStatus
	}{
		{
			desc:                "shutdown interface",
			**intfOper:            "shutdown",       >>>>>>>>>  This should be vendor-agnostic**
			expectedAdminStatus: telemetry.Interface_AdminStatus_DOWN,
			expectedOperStatus:  telemetry.Interface_OperStatus_DOWN,
		},

Thanks,
Avaneesh

dplore:
cli commands are expected to be vendor specific.    The goal is to test gNMI setRequest for origin:CLI.   Because we aren't using origin: openconfig, this is not expected to be a vendor neutral openconfig behavior test.

That said,  I expect code should be added to use a vendor appropriate cli command for testing.  
@avaneesh90 would you be interested in submitting a PR to add this?

avaneesh90:
Thank you @dplore, yes we are interested to do this

PR #423, opened 2022-08-16
Update CLI Origin Test to be vendor agnostic

avaneesh90:
Hi ,

I have added the changes for below github issues in this pull request:
https://github.com/openconfig/featureprofiles/issues/273

Thanks,
Avaneesh

### DIFF
--- a/feature/system/gnmi/cliorigin/ate_tests/cli_origin_test/cli_origin_test.go
+++ b/feature/system/gnmi/cliorigin/ate_tests/cli_origin_test/cli_origin_test.go
@@ -67,11 +67,34 @@ func TestMain(m *testing.M) {
 func TestOriginCliConfig(t *testing.T) {
 	dut := ondatra.DUT(t, "dut")
 	dp1 := dut.Port(t, "port1")
+	var intfAdminDown, intfAdminUp, intfConfig string
 
-	var intfConfig string = `
-interface %s
-  %s
-`
+	switch dut.Vendor() {
+	case ondatra.JUNIPER:
+		intfAdminDown = "disable;"
+		intfAdminUp = "enable;"
+		intfConfig = `
+        	interfaces {
+            	    %s {
+                        %s
+                    }
+                }
+                `
+	case ondatra.CISCO:
+		intfAdminDown = "shutdown"
+        	intfAdminUp = "no shutdown"
+		intfConfig = `
+        	interface %s
+          	%s
+        	`
+	case ondatra.ARISTA:
+        	intfAdminDown = "shutdown"
+        	intfAdminUp = "no shutdown"
+		intfConfig = `
+        	interface %s
+          	    %s
+        	`
+	}
 	cases := []struct {
 		desc                string
 		intfOper            string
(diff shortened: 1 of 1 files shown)


bstoll:
Hi @avaneesh90 , could you correct the CI issues with the PR?  It looks like you need to run gofmt on the file (perhaps from go1.19).

feature/system/gnmi/cliorigin/ate_tests/cli_origin_test/cli_origin_test.go
```diff
@@ -67,26 +67,49 @@ func TestMain(m *testing.M) {
 func TestOriginCliConfig(t *testing.T) {
 	dut := ondatra.DUT(t, "dut")
 	dp1 := dut.Port(t, "port1")
+	var intfAdminDown, intfAdminUp, intfConfig string
 
-	var intfConfig string = `
-interface %s
-  %s
-`
+	switch dut.Vendor() {
```
liulk, line 72: Let's factor the vendor CLIs out to their own functions, in particular so fmt.Sprintf() does not use a variable as a format string which is error prone because the variable can change, whereas the rest of the arguments to fmt.Sprintf() can't change.

```
func juniperCLI(intf string, enabled bool) string {
  op := "disable"
  if enabled {
    op = "enable"
  }
  return fmt.Sprintf(`
interfaces {
  %s {
    %s;
  }
}
`, intf, op)
}

func ciscoCLI(intf string, enabled bool) string { ... } // Similar to above.
func aristaCLI(intf string, enabled bool) string { ... } // Similar to above.
```

Here instead of three string variables, use a single function variable.

```
var cliFn func(intf string, enabled bool) string
switch dut.Vendor() {
case ondatra.JUNIPER:
  cliFn = juniperCLI
case ondatra.CISCO:
  cliFn = ciscoCLI
case ondatra.ARISTA:
  cliFn = aristaCLI
default:
  t.Fatalf("Vendor %v is not supported.", dut.Vendor())
}
```

feature/system/gnmi/cliorigin/ate_tests/cli_origin_test/cli_origin_test.go
```diff
@@ -67,26 +67,49 @@ func TestMain(m *testing.M) {
...
+          	%s
+        	`
+	case ondatra.ARISTA:
+		intfAdminDown = "shutdown"
+		intfAdminUp = "no shutdown"
+		intfConfig = `
+        	interface %s
+          	    %s
+        	`
+	}
```
liulk, line 97: Let's make sure there is a default case.

liulk (changes requested):

feature/system/gnmi/cliorigin/ate_tests/cli_origin_test/cli_origin_test.go
```diff
@@ -122,6 +128,47 @@ interface %s
...
+		op = "enable"
+	}
+	return fmt.Sprintf(`
+  interfaces {
+	%s {
+	  %s;
+	}
+  }
+  `, intf, op)
+}
```
liulk, line 171:
```suggestion
}

```

feature/system/gnmi/cliorigin/ate_tests/cli_origin_test/cli_origin_test.go
```diff
@@ -122,6 +128,47 @@ interface %s
...
+
+func ciscoCLI(intf string, enabled bool) string {
```
liulk, line 145: Cisco should be shutdown/no shutdown?

feature/system/gnmi/cliorigin/ate_tests/cli_origin_test/cli_origin_test.go
```diff
@@ -122,6 +128,47 @@ interface %s
...
+  interfaces {
+	%s {
+	  %s;
+	}
+  }
+  `, intf, op)
+}
+
+func aristaCLI(intf string, enabled bool) string {
+	op := "disable"
```
liulk, line 160: Arista should be shutdown/no shutdown?

liulk (changes requested):

liulk (changes requested):
Just noticed one last thing before approval. Thanks for making all the requested changes!

avaneesh90:
Hi @liulk, What else is pending for changes, please update

liulk (approved):

liulk:
> Hi @liulk, What else is pending for changes, please update

There was a suggested edit (probably easier to see in diff view so I understand why you might have missed it). I applied it for you.

PR #425, opened 2022-08-16
Update TE-3.5 README

xw-g:
* In the case of FIB ACK, RIB ACK is optional. gRIBI does allow a device to only return FIB ACK (skip RIB_ACK) after a success FIB programming. This is also aligned with current implementation that checks either FIB ACK or RIB ACK (with flag), but not both.  
* Remove case#2:
	* If case#2 means IPv4Entry -> NHG -> NH, then it's already covered by case#4.
	* If case#2 means hierarchical resolution, then it's covered by TE-3.1
	* case#2 is not implemented anyway.

### DIFF
--- a/feature/gribi/ate_tests/ordering_ack_test/README.md
+++ b/feature/gribi/ate_tests/ordering_ack_test/README.md
@@ -11,11 +11,8 @@ Ensure that acknowledgements are sent as is expected by gRIBI controller.
     the requested `ack_type`.
 *   Install the following entries and determine whether the expected result is
     observed:
-    *   A `NextHopGroup` referencing a `NextHop` is responded to with RIB+FIB
+    *   A `NextHopGroup` referencing a `NextHop` is responded to with FIB
         ACK, and is reported through the AFT telemetry.
-    *   A `NextHopGroup` referencing a `NextHop` with a referencing `IPv4Entry`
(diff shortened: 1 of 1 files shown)


liulk (approved):

feature/gribi/ate_tests/ordering_ack_test/README.md
```diff
@@ -11,11 +11,8 @@ Ensure that acknowledgements are sent as is expected by gRIBI controller.
     the requested `ack_type`.
 *   Install the following entries and determine whether the expected result is
     observed:
-    *   A `NextHopGroup` referencing a `NextHop` is responded to with RIB+FIB
+    *   A `NextHopGroup` referencing a `NextHop` is responded to with FIB
         ACK, and is reported through the AFT telemetry.
-    *   A `NextHopGroup` referencing a `NextHop` with a referencing `IPv4Entry`
```
liulk, line 16: Wait why is this deleted?

liulk (changes requested):

feature/gribi/ate_tests/ordering_ack_test/README.md
xw-g: See the PR description?

liulk (approved):

PR #427, opened 2022-08-16
Add missing state paths

amrindrr:
Fix the list of config paths which are missing corresponding state path


feature/qos/feature.textproto
```diff
@@ -63,33 +84,60 @@ telemetry_path {
...
   path: "/qos/scheduler-policies/scheduler-policy/schedulers/scheduler/inputs/input/config/queue"
 }
+telemetry_path {
+  path: "/qos/scheduler-policies/scheduler-policy/schedulers/scheduler/inputs/input/state/queue"
+}
 config_path {
   path: "/qos/scheduler-policies/scheduler-policy/schedulers/scheduler/inputs/input/config/weight"
 }
 telemetry_path {
-  path: "/qos/scheduler-policies/scheduler-policy/schedulers/scheduler/inputs/input/state/queue"
```
dplore, line 88: ah nice catch, looks like this was a duplicate

dplore (approved):
LGTM, good catch on the duplicate.  It seems we shouldn't have duplicates within an individual file.  But between different features, perhaps duplicates could be intentional (two  features may have some overlapping path dependencies)

Did you use some script to generate these?  Is that script in opensource?  If yes, you can reference the usage in the PR comments for others to observe.

bstoll (approved):

feature/qos/feature.textproto
amrindrr: Yes I wrote script in /featureprofiles/tools/validate_paths.go to check for missing paths. I will open source the code so that it runs automatically in the backend.

PR #431, opened 2022-08-17
Restore setting of L2 MTU and --deviation_omit_l2_mtu

liulk:
#383


dplore (dismissed):
LGTM. Only thing I can think to suggest is we could clarify the mtus are l3mtus or ipMtu on line 403 and in the test case mtu could become l3mtu or similar.  This may help disambiguate which mtu we are talking about.  But I don't think it's critical.

liulk:
Added clarification for RT-5.1

dplore (approved):

Issue #275, opened 2022-06-27
local_bgp_test.go: Supported keepalive and hold timer values


aredmon8551:
In one instance of local_bgp_test.go, the keepalive timer is set to 50 and the hold-timer is set to 100 seconds.  
https://github.com/openconfig/featureprofiles/blob/main/feature/bgp/tests/local_bgp_test/local_bgp_test.go#L239

BGP's RFC suggests a keepalive timer of hold-down/3:
   `A reasonable maximum time between KEEPALIVE messages would be one third
   of the Hold Time interval.  KEEPALIVE messages MUST NOT be sent more
   frequently than one per second.  An implementation MAY adjust the
   rate at which it sends KEEPALIVE messages as a function of the Hold
   Time interval.`

Note, too, that only the hold-timer is conveyed in BGP OPEN messages - the keepalive must be derived (or explicitly configured) by both peers.  Keepalive intervals are largely locally significant, as well - as long as one is received (or an UPDATE message) within the hold-down period then the session will remain ESTABLISHED.

Juniper only supports keepalive values of hold-down/3 - this is a fundamental behavior of our implementation that cannot be easily changed.  

Can we update this test case to use values that align with the RFC's guidance and can be supported by all major vendor implementations by using the formula of hold-down = keepalive * 3?

Edit: grammar

dplore:
@bstoll what do you think?

jhaas-pfrc:
I had been asked to look into this issue as part of internal reviews.  This sort of issue comes up every few years.

BGP's core specification, RFC 4271 in section 10, provides this guidance:

   An implementation of BGP MUST allow the HoldTimer to be configurable
   on a per-peer basis, and MAY allow the other timers to be
   configurable.

It's quite fine for an implementation to support a configurable keepalive timer.  Requiring it to be configurable means the test case is out of variance with the RFC.

I will also advise that tests against the timers of conformant implementations have a tendency to be fragile due to excess pedantry.  Implementations are generally expected to apply jitter.  From the same section of the RFC:

To minimize the likelihood that the distribution of BGP messages by a
   given BGP speaker will contain peaks, jitter SHOULD be applied to the
   timers associated with MinASOriginationIntervalTimer, KeepaliveTimer,
   MinRouteAdvertisementIntervalTimer, and ConnectRetryTimer.  A given
   BGP speaker MAY apply the same jitter to each of these quantities,
   regardless of the destinations to which the updates are being sent;
[...]

PR #439, opened 2022-08-18
BGP keepalive timer should be 1/3 of hold timer

bstoll:
This is to address #275.  This test is not evaluating the details of hold time behavior, but rather just setting BGP session parameters and ensuring sessions come up.

### DIFF
--- a/feature/bgp/tests/local_bgp_test/local_bgp_test.go
+++ b/feature/bgp/tests/local_bgp_test/local_bgp_test.go
@@ -47,6 +47,9 @@ var (
 const (
 	dutAS = 64500
 	ateAS = 64501
+
+	keepAlive = 50
+	holdTime  = keepAlive * 3 // Should be 3x keepAlive, see RFC 4271 - A Border Gateway Protocol 4, Sec. 10
 )
 
 func bgpWithNbr(as uint32, routerID string, nbr *telemetry.NetworkInstance_Protocol_Bgp_Neighbor) *telemetry.NetworkInstance_Protocol_Bgp {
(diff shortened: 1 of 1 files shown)


dplore (dismissed):

dplore:
LGTM, adding codeowners to review

sezhang2 (dismissed):

feature/bgp/tests/local_bgp_test/local_bgp_test.go
```diff
@@ -236,7 +236,7 @@ func TestParameters(t *testing.T) {
 				PeerAs:          ygot.Uint32(ateAS),
 				NeighborAddress: ygot.String(ateIP),
 				Timers: &telemetry.NetworkInstance_Protocol_Bgp_Neighbor_Timers{
-					HoldTime:          ygot.Uint16(100),
+					HoldTime:          ygot.Uint16(150),
```
liulk, line 239: If there is a relationship, could we make them a constant like this?

```
const (
  keepAlive = 50
  holdTime = keepAlive * 3 // Must be at least 3 times, see (reference).
)
```

If there is a source that explains why this has to be case, let's add it?

liulk (changes requested):

feature/bgp/tests/local_bgp_test/local_bgp_test.go
bstoll: I added a const for the values.

sezhang2 (approved):

liulk (approved):

Issue #137, opened 2022-04-11
gNMI-1.1: cli Origin Support


dplore:

dplore:
Resolved by #474

PR #474, opened 2022-08-31
OpenSource gNMI mixed_oc_cli_origin_support_test

hariramnat:
Co-authored-by: Hariram Natarajan hariramnat@google.com, Sean Zhang sezhang@google.com

### DIFF
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
@@ -55,18 +55,3 @@ required in the emergency case that is covered by this requirement.
         ```
 
 *   TODO: Validate that DUT port-1 description still has `"foo2"`.
-
-*   TODO: Push overlapping mixed SetRequest specifying OpenConfig before CLI for
-    DUT port-1.
-
-    *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
-        value at `/interfaces/interface/config/description` to `"foo1"`.
-
-    *   `origin: "cli"` containing vendor configuration.
-
-        ```
-        interface <DUT port-1>
-          description foo2
-        ```
-
-*   TODO: Validate that DUT port-1 description still has `"foo2"`.
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
@@ -33,7 +33,7 @@ required in the emergency case that is covered by this requirement.
 
         ```
         interface <DUT port-1>
-          description foo1
+          description from cli
         ```
 
     *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
@@ -41,7 +41,7 @@ required in the emergency case that is covered by this requirement.
 
 *   Validate that DUT port-1 description has `"foo2"`.
 
-*   TODO: Push overlapping mixed SetRequest specifying OpenConfig before CLI for
+*   Push overlapping mixed SetRequest specifying OpenConfig before CLI for
     DUT port-1.
 
     *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
@@ -54,4 +54,4 @@ required in the emergency case that is covered by this requirement.
           description foo2
         ```
 
-*   TODO: Validate that DUT port-1 description still has `"foo2"`.
+*   Validate that DUT port-1 description still has `"foo2"`.
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
@@ -39,7 +39,7 @@ required in the emergency case that is covered by this requirement.
     *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
         value at `/interfaces/interface/config/description` to `"from oc"`.
 
-*   Validate that DUT port-1 description has `"foo2"`.
+*   Validate that DUT port-1 description is `"from oc"`.
 
 *   Push overlapping mixed SetRequest specifying OpenConfig before CLI for
     DUT port-1.
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
@@ -37,7 +37,7 @@ required in the emergency case that is covered by this requirement.
         ```
 
     *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
-        value at `/interfaces/interface/config/description` to `"foo2"`.
+        value at `/interfaces/interface/config/description` to `"from oc"`.
 
 *   Validate that DUT port-1 description has `"foo2"`.
 
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
@@ -1,0 +1,55 @@
+# gNMI-1.1: cli Origin
+
+## Summary
+
+Ensure that both CLI and OC configuration can be pushed to the device at the
+same time
+
+## Procedure
+
+Note: this test is intended to cover only the case of pushing some configuration
+along with OC paths - since it is unknown what CLI configuration would be
+required in the emergency case that is covered by this requirement.
+
+*   TODO: Push base configuration to DUT specifying an interface configuration
+    for DUT port-1 and DUT port-2.
+
+*   Validate mixed OC/CLI schema by sending both CLI and OC config concurrently
+    for two different interfaces.
+
+*   Push configuration using SetRequest specifying:
+
+    *   `origin: "cli"` - containing modelled configuration.
+
+        ~~~
+        interface <DUT port-1>
+        description foo1
+        ```
+        ~~~
+
+    *   `origin: ""` (openconfig, default origin) - containing modelled
+        configuration for DUT port-2.
+
+*   Validate that DUT port-1 and DUT port-2 description through telemetry.
+
+*   Validate order dependence by trying to modify the same config using CLI and
+    OC.
+
+*   Push configuration using SetRequest specifying:
+
+    *   `origin: "cli"` - containing modelled configuration.
+
+        ~~~
+        interface <DUT port-1>
+        description foo1
+        ```
+        ~~~
+
+    *   `origin: ""` (openconfig, default origin) - containing modelled
+        configuration for DUT port-1.:
+
+*   Validate that DUT port-1 description through telemetry.
+
+## Config Parameter coverage
+
+## Telemetry Parameter coverage
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
@@ -51,7 +51,7 @@ required in the emergency case that is covered by this requirement.
 
         ```
         interface <DUT port-1>
-          description foo2
+          description from cli
         ```
 
 *   Validate that DUT port-1 description is `"from cli"`.
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
@@ -1,55 +1,59 @@
-# gNMI-1.1: cli Origin
+# gNMI-1.12: Mixed OpenConfig/CLI Origin
 
 ## Summary
 
 Ensure that both CLI and OC configuration can be pushed to the device at the
-same time
+same time.
 
 ## Procedure
 
 Note: this test is intended to cover only the case of pushing some configuration
 along with OC paths - since it is unknown what CLI configuration would be
-required in the emergency case that is covered by this requirement.
+required in the emergency case that is covered by this requirement. * Push
+non-overlapping mixed SetRequest specifying CLI for DUT port-1 and OpenConfig
+for DUT port-2.
 
-*   TODO: Push base configuration to DUT specifying an interface configuration
-    for DUT port-1 and DUT port-2.
+````
+*   `origin: "cli"` containing vendor configuration.
 
-*   Validate mixed OC/CLI schema by sending both CLI and OC config concurrently
-    for two different interfaces.
+    ~~~
+    interface <DUT port-1>
+      description foo1
+    ~~~
 
-*   Push configuration using SetRequest specifying:
+*   `origin: ""` (openconfig, default origin) setting the DUT port-2
+    string value at `/interfaces/interface/config/description` to `"foo2"`.
 
-    *   `origin: "cli"` - containing modelled configuration.
 
-        ~~~
-        interface <DUT port-1>
-        description foo1
+*   Validate the DUT port-1 and DUT port-2 descriptions through telemetry.
+
+*   Push overlapping mixed SetRequest specifying CLI before OpenConfig for DUT
+    port-1.
+
+    *   `origin: "cli"` containing vendor configuration.
+
         ```
-        ~~~
-
-    *   `origin: ""` (openconfig, default origin) - containing modelled
-        configuration for DUT port-2.
-
-*   Validate that DUT port-1 and DUT port-2 description through telemetry.
-
-*   Validate order dependence by trying to modify the same config using CLI and
-    OC.
-
-*   Push configuration using SetRequest specifying:
-
-    *   `origin: "cli"` - containing modelled configuration.
-
-        ~~~
         interface <DUT port-1>
-        description foo1
+          description foo1
         ```
-        ~~~
 
-    *   `origin: ""` (openconfig, default origin) - containing modelled
-        configuration for DUT port-1.:
+    *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
+        value at `/interfaces/interface/config/description` to `"foo2"`.
 
-*   Validate that DUT port-1 description through telemetry.
+*   Validate that DUT port-1 description has `"foo2"`.
 
-## Config Parameter coverage
+*   TODO: Push overlapping mixed SetRequest specifying OpenConfig before CLI for
+    DUT port-1.
 
-## Telemetry Parameter coverage
+    *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
+        value at `/interfaces/interface/config/description` to `"foo1"`.
+
+    *   `origin: "cli"` containing vendor configuration.
+
+        ```
+        interface <DUT port-1>
+          description foo2
+        ```
+
+*   TODO: Validate that DUT port-1 description still has `"foo2"`.
+````
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
@@ -9,25 +9,20 @@ same time.
 
 Note: this test is intended to cover only the case of pushing some configuration
 along with OC paths - since it is unknown what CLI configuration would be
-required in the emergency case that is covered by this requirement. * Push
-non-overlapping mixed SetRequest specifying CLI for DUT port-1 and OpenConfig
-
-# for DUT port-2.
-
-along with OC paths, since it is unknown what CLI configuration would be
 required in the emergency case that is covered by this requirement.
 
-````
-*   `origin: "cli"` containing vendor configuration.
+*   Push non-overlapping mixed SetRequest specifying CLI for DUT port-1 and
+    OpenConfig for DUT port-2.
 
-    ~~~
-    interface <DUT port-1>
-      description foo1
-    ~~~
+    *   `origin: "cli"` containing vendor configuration.
 
-*   `origin: ""` (openconfig, default origin) setting the DUT port-2
-    string value at `/interfaces/interface/config/description` to `"foo2"`.
+        ```
+        interface <DUT port-1>
+          description foo1
+        ```
 
+    *   `origin: ""` (openconfig, default origin) setting the DUT port-2 string
+        value at `/interfaces/interface/config/description` to `"foo2"`.
 
 *   Validate the DUT port-1 and DUT port-2 descriptions through telemetry.
 
@@ -45,48 +40,19 @@ required in the emergency case that is covered by this requirement.
         value at `/interfaces/interface/config/description` to `"foo2"`.
 
 *   Validate that DUT port-1 description has `"foo2"`.
-*   Push non-overlapping mixed SetRequest specifying CLI for DUT port-1 and
-    OpenConfig for DUT port-2.
-
-    *   `origin: "cli"` containing vendor configuration.
-
-        ~~~
-        interface <DUT port-1>
-          description foo1
-        ~~~
-
-    *   `origin: ""` (openconfig, default origin) setting the DUT port-2
-        string value at `/interfaces/interface/config/description` to `"foo2"`.
-
-*   Validate the DUT port-1 and DUT port-2 descriptions through telemetry.
-
-*   Push overlapping mixed SetRequest specifying CLI before OpenConfig for
-    DUT port-1.
-
-    *   `origin: "cli"` containing vendor configuration.
-
-        ~~~
-        interface <DUT port-1>
-          description foo1
-        ~~~
-
-    *   `origin: ""` (openconfig, default origin) setting the DUT port-1
-        string value at `/interfaces/interface/config/description` to `"foo2"`.
-
-*   Validate that DUT port-1 description has `"foo2"`.
 
 *   TODO: Push overlapping mixed SetRequest specifying OpenConfig before CLI for
     DUT port-1.
 
-    *   `origin: ""` (openconfig, default origin) setting the DUT port-1
-        string value at `/interfaces/interface/config/description` to `"foo1"`.
+    *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
+        value at `/interfaces/interface/config/description` to `"foo1"`.
 
     *   `origin: "cli"` containing vendor configuration.
 
-        ~~~
+        ```
         interface <DUT port-1>
           description foo2
-        ~~~
+        ```
 
 *   TODO: Validate that DUT port-1 description still has `"foo2"`.
 
@@ -104,4 +70,3 @@ required in the emergency case that is covered by this requirement.
         ```
 
 *   TODO: Validate that DUT port-1 description still has `"foo2"`.
-````
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
@@ -54,4 +54,4 @@ required in the emergency case that is covered by this requirement.
           description foo2
         ```
 
-*   Validate that DUT port-1 description still has `"foo2"`.
+*   Validate that DUT port-1 description is `"from cli"`.
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
@@ -45,7 +45,7 @@ required in the emergency case that is covered by this requirement.
     DUT port-1.
 
     *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
-        value at `/interfaces/interface/config/description` to `"foo1"`.
+        value at `/interfaces/interface/config/description` to `"from oc"`.
 
     *   `origin: "cli"` containing vendor configuration.
 
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
@@ -164,12 +164,7 @@ func TestMixedOriginOCCLIConfig(t *testing.T) {
 	dp1 := dut.Port(t, "port1")
 	dp2 := dut.Port(t, "port2")
 	gnmiClient := dut.RawAPIs().GNMI().Default(t)
-	// `origin: "cli"` - containing vendor  configuration.
-	// ~~~
-	// interface <DUT port-1>
-	// description foo1
-	//  ```
-	// ~~~
+	// `origin: "cli"` - containing vendor configuration.
 	intf1Config := interfaceDescriptionCLI(dp1, "foo1")
 	if intf1Config == "" {
 		t.Fatalf("Please add vendor support for %v", dut.Vendor())
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
@@ -84,6 +84,9 @@ func TestCLIBeforeOpenConfig(t *testing.T) {
 	//  string value at `/interfaces/interface/config/description` to `"from oc"`.
 	resolvedPath := dut.Config().Interface(dp.Name()).Description()
 	path, _, errs := ygot.ResolvePath(resolvedPath)
+	if errs != nil {
+		t.Fatalf("Could not resolve path: %v", errs)
+	}
 
 	gpbSetRequest := &gpb.SetRequest{
 		Update: []*gpb.Update{
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
@@ -123,6 +123,13 @@ func TestOpenConfigBeforeCLI(t *testing.T) {
         resolvedPath := dut.Config().Interface(dp.Name()).Description()
         path, _, errs := ygot.ResolvePath(resolvedPath)
 
+        // `origin: "cli"` - containing vendor configuration.
+        intfConfig := interfaceDescriptionCLI(dp, "from cli")
+        if intfConfig == "" {
+                t.Fatalf("Please add vendor support for %v", dut.Vendor())
+        }
+        t.Logf("Building the CLI config:\n%s", intfConfig)
+
         gpbSetRequest := &gpb.SetRequest{
                 Update: []*gpb.Update{
 			buildCLIUpdate(intfConfig),
--- a/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
+++ b/feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
@@ -91,9 +91,6 @@ func TestCLIBeforeOpenConfig(t *testing.T) {
 			buildOCUpdate(path, "from oc"),
 		},
 	}
-	if errs != nil {
-		t.Fatalf("Could not resolve path: %v", errs)
-	}
 	t.Log("gnmiClient Set both CLI and OpenConfig modelled config")
 	t.Log(gpbSetRequest)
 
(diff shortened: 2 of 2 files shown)


feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,146 @@
+/*
```
liulk, line 1: I think you added this license manually because it uses the block styled comment. We should standardize on the line styled comments.

If you remove the license block, and subsequently use the `addlicense -l apache .` command, you should get the correct style.

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,55 @@
...
+Ensure that both CLI and OC configuration can be pushed to the device at the
+same time
+
+## Procedure
+
+Note: this test is intended to cover only the case of pushing some configuration
+along with OC paths - since it is unknown what CLI configuration would be
+required in the emergency case that is covered by this requirement.
+
+*   TODO: Push base configuration to DUT specifying an interface configuration
```
liulk, line 14: Why is this TODO?

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,146 @@
...
+		t.Fatalf("Could not resolve path: %v", errs)
+	}
+	t.Log("gnmiClient Set both CLI and OpenConfig modelled config")
+	t.Log(gpbSetRequest)
+
+	response, err := gnmiClient.Set(context.Background(), gpbSetRequest)
+	if err != nil {
+		t.Fatalf("gnmiClient.Set() with unexpected error: %v", err)
+	}
+	t.Log("gnmiClient Set Response for CLI and OpenConfig modelled  config")
```
liulk, line 101:
```suggestion
	t.Log("gnmiClient Set Response for CLI and OpenConfig modelled config")
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,146 @@
...
```
liulk, line 136:
```suggestion
	t.Log("gnmiClient Set Response for CLI and OpenConfig modelled config")
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+}
+
+func TestMixedOriginOCCLIConfig(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	dp1 := dut.Port(t, "port1")
+	dp2 := dut.Port(t, "port2")
+	gnmiClient := dut.RawAPIs().GNMI().Default(t)
+	intf1Config := interfaceDescriptionCLI(dp1, "foo1")
+	t.Logf("Building the CLI config:\n%s", intf1Config)
```
liulk, line 116: Let's add:

```
if intf1Config == "" {
  t.Fatal("Please add vendor support for %v", dut.Vendor())
}
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/openconfig/ondatra"
+)
+
+func TestMain(m *testing.M) {
+	fptest.RunTests(m)
+}
+
+func interfaceDescriptionCLI(dp *ondatra.Port, desc string) string {
+	switch dp.Device().Vendor() {
+	case ondatra.ARISTA:
```
liulk, line 37: I believe Cisco uses the same syntax as well.
```suggestion
	case ondatra.ARISTA, ondatra.CISCO:
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,55 @@
+# gNMI-1.1: cli Origin
+
+## Summary
+
+Ensure that both CLI and OC configuration can be pushed to the device at the
+same time
+
+## Procedure
```
liulk, line 8: I'm going to push an edit with slight rewording. If you're okay with the new wording, can you copy each procedure step to a code comment at the appropriate places?

liulk (changes requested):

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
hariramnat: No actually I added it using addlicense only. Have changed it now.

liulk: If the license was missing, addlicense should prefer line styled comment. But if you run addlicense on a file that already has a license, it will not do anything.

Like I said, you have to remove the license first, then rerun addlicense.

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,68 @@
...
+        interface <DUT port-1>
+          description foo1
+        ~~~
+
+    *   `origin: ""` (openconfig, default origin) setting the DUT port-1
+        string value at `/interfaces/interface/config/description` to `"foo2"`.
+
+*   Validate that DUT port-1 description has `"foo2"`.
+
+*   TODO: Push overlapping mixed SetRequest specifying OpenConfig before CLI for
```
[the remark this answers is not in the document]
earies: You are introducing a behavior that is not agreed upon by the community with any overlap or order of operations here.  This is not what each implementation does so I suggest this be brought up in the wider OC community first before any codification - cc: @robshakir @dplore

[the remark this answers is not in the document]
hariramnat: Ack

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
hariramnat: Done

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
hariramnat: Done!

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
hariramnat: We actually do not push base config in cliOrigin and mixed_cli_origin, which is why in both these tests this step is marked as TODO. We just push specific config relevant to that interface. This step can probably be removed for this test as its not required.

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
hariramnat: Done. The previous comment came out before the git push which is why it was not visible

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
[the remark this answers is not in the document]
liulk: @earies I will refer you to gNMI specification section 3.4:

https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#34-modifying-state

> The order of the replace and update fields MUST be treated as significant within a single SetRequest message. If a single path is specified multiple times for a single operation (i.e., within update or replace), then the state of the target MUST reflect the application of all of the operations in order, even if they overwrite each other.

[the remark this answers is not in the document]
hariramnat: Hey @earies this is to verify that what's the expected behavior in cases of such an overlap. The above specification does detail the requirements for such an operation. Do let us know if you have any more questions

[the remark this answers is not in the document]
earies: @liulk @hariramnat - Unfortunately there are still ambiguities that have not been thoroughly discussed and ratified when dealing with mixed schemas/origins.  The section you reference does reference the ordering of `Update` message application within a single `SetRequest` however there is an assumption that systems treat _all_ schemas equally with direct access to underlying objects when infact some implementations have had a preference model based off most explicit vendor native value wins.  OpenConfig (and any 3rd party modeling) is an abstraction layer built on top and can be done in a variety of ways.

For instance, you could send a `SetRequest` payload with the first `Update` containing `origin=openconfig` and the second `origin=cli` (or vice versa) but if there is a node that overlaps (lets pick on BGP ASN) then the most specific vendor native (cli in this case) always wins.

I still suggest further discussion as this is not how all implementations are built today.
Since this has always been per implementation behaviors from day 1 - this has significant architectural considerations

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,107 @@
...
+Note: this test is intended to cover only the case of pushing some configuration
+along with OC paths - since it is unknown what CLI configuration would be
+required in the emergency case that is covered by this requirement. * Push
+non-overlapping mixed SetRequest specifying CLI for DUT port-1 and OpenConfig
+
+# for DUT port-2.
+
+along with OC paths, since it is unknown what CLI configuration would be
+required in the emergency case that is covered by this requirement.
+
```
dplore, line 19: Looks like there is some cut/paste error here?  Can you correct?

liulk: @hariramnat this commit is bad: https://github.com/openconfig/featureprofiles/pull/474/commits/dfa5c011b05a3287fff4e745bdd09392782da1ba

hariramnat: Fixed it now.

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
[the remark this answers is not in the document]
dplore: @rgwilton , @rolandphung could you comment on this thread?  

I think it's most  important to understand the use case for why we want the behavior described in this PR.

The use case is to allow composition of an setRequest containing both OC and CLI origins.  We propose that the content be processed in the order it appears in the setRequest.  This allows flexible composition of configuration updates with OC and CLI.  While this example is contrived it is meant to generate questions about how a NOS will behave when given a setRequest with this content.  Our goal is to get deterministic configuration behavior from the NOS in the presence of multiple origins.

In [func TestOrderDependence](https://github.com/openconfig/featureprofiles/pull/474/files#diff-36bbeeef4c3dacadf0b003e85e03370579cd36775fe2d292d37d25bdd95638d4R72) there is an expectation that when pushing configuration using two different origins which overlap that the last to appear takes precedence.   

Note  @hariramnat , the way the test is written at the moment doesn't yet assert this, as it could also pass if the CLI always takes precedence.   Can you add another interface description update where origin CLI tries to set the value to some other unwanted value before the current two updates?

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,57 @@
...
+        interface <DUT port-1>
+          description foo1
+        ```
+
+    *   `origin: ""` (openconfig, default origin) setting the DUT port-2 string
+        value at `/interfaces/interface/config/description` to `"foo2"`.
+
+*   Validate the DUT port-1 and DUT port-2 descriptions through telemetry.
+
+*   Push overlapping mixed SetRequest specifying CLI before OpenConfig for DUT
```
liulk, line 29: @dplore this is the case with CLI before OC, as currently implemented in the test. The OC before CLI case is what I added to Hariram's test plan, and because it hasn't been implemented, I marked it as TODO.

@hariramnat should we implement the TODO case in this PR?

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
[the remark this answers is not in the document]
hariramnat: @dplore, @liulk Done. Have added a testcase now to test whether the latest origin takes precedence or not.

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
hariramnat: Done.This is implemented now.

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+		Val: &gpb.TypedValue{
+			Value: &gpb.TypedValue_AsciiVal{
+				AsciiVal: value,
+			},
+		},
+	}
+	return update
+}
+
+// Push overlapping mixed SetRequest specifying CLI before OpenConfig for  DUT port-1.
```
liulk, line 71: The comment says CLI before OpenConfig but the request below puts OpenConfig before CLI. In my suggested edit I'm going with this comment.

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+	// ~~~
+	intfConfig := interfaceDescriptionCLI(dp, "want")
+	if intfConfig == "" {
+		t.Skip("Vendor is not supported.")
+	}
+	gnmiClient := dut.RawAPIs().GNMI().Default(t)
+	t.Logf("Building the CLI config:\n%s", intfConfig)
+
+	// `origin: ""` (openconfig, default origin) setting the DUT port-1
+	//  string value at `/interfaces/interface/config/description` to `"foo2"`.
```
liulk, line 89:
```suggestion
	// `origin: ""` (openconfig, default origin) setting the DUT port-1
	//  string value at `/interfaces/interface/config/description` to `"from oc"`.
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+	resolvedPath := dut.Config().Interface(dp.Name()).Description()
+	path, _, errs := ygot.ResolvePath(resolvedPath)
+
+	gpbSetRequest := &gpb.SetRequest{
+		Update: []*gpb.Update{
+			buildOCUpdate(path, "not want"),
+			buildCLIUpdate(intfConfig),
+		},
+	}
```
liulk, line 98:
```suggestion
	gpbSetRequest := &gpb.SetRequest{
		Update: []*gpb.Update{
			buildCLIUpdate(intfConfig),
			buildOCUpdate(path, "from oc"),
		},
	}
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+	if errs != nil {
+		t.Fatalf("Could not resolve path: %v", errs)
+	}
```
liulk, line 101:
```suggestion
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+	if intfConfig == "" {
+		t.Skip("Vendor is not supported.")
+	}
+	gnmiClient := dut.RawAPIs().GNMI().Default(t)
+	t.Logf("Building the CLI config:\n%s", intfConfig)
+
+	// `origin: ""` (openconfig, default origin) setting the DUT port-1
+	//  string value at `/interfaces/interface/config/description` to `"foo2"`.
+	resolvedPath := dut.Config().Interface(dp.Name()).Description()
+	path, _, errs := ygot.ResolvePath(resolvedPath)
```
liulk, line 91:
```suggestion
	path, _, errs := ygot.ResolvePath(resolvedPath)
	if errs != nil {
		t.Fatalf("Could not resolve path: %v", errs)
	}
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+                t.Fatalf("gnmiClient.Set() with unexpected error: %v", err)
+        }
+        t.Log("gnmiClient Set Response for CLI and OpenConfig modelled config")
+        t.Log(response)
+
+        // Validate that DUT port-1 description has `"foo2"`
+        got := dut.Telemetry().Interface(dp.Name()).Description().Get(t)
+        if got != "want" {
+                t.Errorf("Get(DUT port description): got %v, want %v", got, "want")
+        }
```
liulk, line 164:
```suggestion
        // Validate that DUT port-1 description is `"from cli"`
        got := dut.Telemetry().Interface(dp.Name()).Description().Get(t)
        want := "from cli"
        if got != want {
                t.Errorf("Get(DUT port description): got %v, want %v", got, want)
        }
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+	dut := ondatra.DUT(t, "dut")
+	dp1 := dut.Port(t, "port1")
+	dp2 := dut.Port(t, "port2")
+	gnmiClient := dut.RawAPIs().GNMI().Default(t)
+	// `origin: "cli"` - containing vendor  configuration.
+	// ~~~
+	// interface <DUT port-1>
+	// description foo1
+	//  ```
+	// ~~~
```
liulk, line 178:
```suggestion
	// `origin: "cli"` - containing vendor configuration.
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+	dp := dut.Port(t, "port1")
+	// `origin: "cli"` - containing vendor  configuration.
+	// ~~~
+	// interface <DUT port-1>
+	// description foo1
+	//  ```
+	// ~~~
+	intfConfig := interfaceDescriptionCLI(dp, "want")
+	if intfConfig == "" {
+		t.Skip("Vendor is not supported.")
```
liulk, line 83:
```suggestion
		t.Fatalf("Please add vendor support for %v", dut.Vendor())
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+func TestOrderDependenceForCli(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	dp := dut.Port(t, "port1")
+	// `origin: "cli"` - containing vendor  configuration.
+	// ~~~
+	// interface <DUT port-1>
+	// description foo1
+	//  ```
+	// ~~~
+	intfConfig := interfaceDescriptionCLI(dp, "want")
```
liulk, line 81:
```suggestion

	// `origin: "cli"` - containing vendor configuration.
	intfConfig := interfaceDescriptionCLI(dp, "from cli")
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+	if intfConfig == "" {
+		t.Skip("Vendor is not supported.")
+	}
+	gnmiClient := dut.RawAPIs().GNMI().Default(t)
```
liulk, line 85:
```suggestion
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+			buildCLIUpdate(intfConfig),
+		},
+	}
+	if errs != nil {
+		t.Fatalf("Could not resolve path: %v", errs)
+	}
+	t.Log("gnmiClient Set both CLI and OpenConfig modelled config")
+	t.Log(gpbSetRequest)
+
+	response, err := gnmiClient.Set(context.Background(), gpbSetRequest)
```
liulk, line 105:
```suggestion
	gnmiClient := dut.RawAPIs().GNMI().Default(t)
	response, err := gnmiClient.Set(context.Background(), gpbSetRequest)
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+        // interface <DUT port-1>
+        // description foo1
+        //  ```
+        // ~~~
+        intfConfig := interfaceDescriptionCLI(dp, "not want")
+        if intfConfig == "" {
+                t.Skip("Vendor is not supported.")
+        }
+        gnmiClient := dut.RawAPIs().GNMI().Default(t)
+        t.Logf("Building the CLI config:\n%s", intfConfig)
```
liulk, line 134:
```suggestion
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+        // `origin: ""` (openconfig, default origin) setting the DUT port-1
+        //  string value at `/interfaces/interface/config/description` to `"foo2"`.
+        resolvedPath := dut.Config().Interface(dp.Name()).Description()
+        path, _, errs := ygot.ResolvePath(resolvedPath)
+
```
liulk, line 140:
```suggestion

        // `origin: "cli"` - containing vendor configuration.
        intfConfig := interfaceDescriptionCLI(dp, "from cli")
        if intfConfig == "" {
                t.Fatalf("Please add vendor support for %v", dut.Vendor())
        }
        t.Logf("Building the CLI config:\n%s", intfConfig)

```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+	t.Log(response)
+
+	// Validate that DUT port-1 description has `"foo2"`
+	got := dut.Telemetry().Interface(dp.Name()).Description().Get(t)
+	if got != "want" {
+		t.Errorf("Get(DUT port description): got %v, want %v", got, "want")
+	}
+}
+
+// Push overlapping mixed SetRequest specifying OpenConfig before Cli for  DUT port-1.
```
liulk, line 119: This entire function seems to be using spaces instead of tabs. Please use gofmt to reformat.

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+        // ~~~
+        intfConfig := interfaceDescriptionCLI(dp, "not want")
+        if intfConfig == "" {
+                t.Skip("Vendor is not supported.")
+        }
+        gnmiClient := dut.RawAPIs().GNMI().Default(t)
+        t.Logf("Building the CLI config:\n%s", intfConfig)
+
+        // `origin: ""` (openconfig, default origin) setting the DUT port-1
+        //  string value at `/interfaces/interface/config/description` to `"foo2"`.
```
liulk, line 137:
```suggestion
        // `origin: ""` (openconfig, default origin) setting the DUT port-1
        //  string value at `/interfaces/interface/config/description` to `"from oc"`.
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
liulk: Move it below OC.

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+        gpbSetRequest := &gpb.SetRequest{
+                Update: []*gpb.Update{
+			buildCLIUpdate(intfConfig),
+                        buildOCUpdate(path, "want"),
+                },
+        }
+        if errs != nil {
+                t.Fatalf("Could not resolve path: %v", errs)
+        }
```
liulk, line 149:
```suggestion
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+        if intfConfig == "" {
+                t.Skip("Vendor is not supported.")
+        }
+        gnmiClient := dut.RawAPIs().GNMI().Default(t)
+        t.Logf("Building the CLI config:\n%s", intfConfig)
+
+        // `origin: ""` (openconfig, default origin) setting the DUT port-1
+        //  string value at `/interfaces/interface/config/description` to `"foo2"`.
+        resolvedPath := dut.Config().Interface(dp.Name()).Description()
+        path, _, errs := ygot.ResolvePath(resolvedPath)
```
liulk, line 139:
```suggestion
        path, _, errs := ygot.ResolvePath(resolvedPath)
        if errs != nil {
                t.Fatalf("Could not resolve path: %v", errs)
        }
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+                        buildOCUpdate(path, "want"),
+                },
+        }
+        if errs != nil {
+                t.Fatalf("Could not resolve path: %v", errs)
+        }
+        t.Log("gnmiClient Set both CLI and OpenConfig modelled config")
+        t.Log(gpbSetRequest)
+
+        response, err := gnmiClient.Set(context.Background(), gpbSetRequest)
```
liulk, line 153:
```suggestion
        gnmiClient := dut.RawAPIs().GNMI().Default(t)
        response, err := gnmiClient.Set(context.Background(), gpbSetRequest)
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+        //  string value at `/interfaces/interface/config/description` to `"foo2"`.
+        resolvedPath := dut.Config().Interface(dp.Name()).Description()
+        path, _, errs := ygot.ResolvePath(resolvedPath)
+
+        gpbSetRequest := &gpb.SetRequest{
+                Update: []*gpb.Update{
+			buildCLIUpdate(intfConfig),
+                        buildOCUpdate(path, "want"),
+                },
+        }
```
liulk, line 146:
```suggestion
        gpbSetRequest := &gpb.SetRequest{
                Update: []*gpb.Update{
                        buildOCUpdate(path, "from oc"),
                        buildCLIUpdate(intfConfig),
                },
        }
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+	// `origin: ""` (openconfig, default origin) setting the DUT port-2
+	//  string value at `/interfaces/interface/config/description` to `"foo2"`.
+	resolvedPath := dut.Config().Interface(dp2.Name()).Description()
+	path, _, errs := ygot.ResolvePath(resolvedPath)
+
+	gpbSetRequest := &gpb.SetRequest{
+		Update: []*gpb.Update{
+			buildOCUpdate(path, "foo2"),
+			buildCLIUpdate(intf1Config),
```
liulk, line 193:
```suggestion
			buildCLIUpdate(intf1Config),
			buildOCUpdate(path, "foo2"),
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,57 @@
...
+    port-1.
+
+    *   `origin: "cli"` containing vendor configuration.
+
+        ```
+        interface <DUT port-1>
+          description foo1
```
liulk, line 36:
```suggestion
          description from cli
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,57 @@
...
+        ```
+
+    *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
+        value at `/interfaces/interface/config/description` to `"foo2"`.
```
liulk, line 40:
```suggestion
        value at `/interfaces/interface/config/description` to `"from oc"`.
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+*   Validate that DUT port-1 description has `"foo2"`.
```
liulk, line 42:
```suggestion
*   Validate that DUT port-1 description is `"from oc"`.
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+*   Push overlapping mixed SetRequest specifying OpenConfig before CLI for
+    DUT port-1.
+
+    *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
+        value at `/interfaces/interface/config/description` to `"foo1"`.
```
liulk, line 48:
```suggestion
        value at `/interfaces/interface/config/description` to `"from oc"`.
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+    *   `origin: "cli"` containing vendor configuration.
+
+        ```
+        interface <DUT port-1>
+          description foo2
```
liulk, line 54:
```suggestion
          description from cli
```

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,57 @@
...
+        ```
+
+*   Validate that DUT port-1 description still has `"foo2"`.
```
liulk, line 57:
```suggestion
*   Validate that DUT port-1 description is `"from cli"`.
```

liulk (changes requested):
I made some suggested edits to change the sentinel strings from "foo1" and "foo2" to "from cli" and "from oc" for clarity.

Also moved some code to shorten the distance between declaring something and when it's used.

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
hariramnat: Ack. Done, committed all changes

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
liulk: Please apply?

liulk (approved):

feature/system/gnmi/cliorigin/mixed_oc_cli_origin_support_test/README.md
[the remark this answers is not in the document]
dplore: Thanks for the thoughtful comments!  Let's discuss in the openconfig community call.  The topic can be something like:  "Merge behavior between different gNMI origins in a gNMI setRequest"

dplore:
Resolves #137

rgwilton:
Expanding on my comments on the call:

For me, I think key point at an external data model level whether the same or different data nodes are being written for "CL"I vs "OC YANG".

If there are multiple edits to the *same external* data node, then the last edit to *that data node* takes effect. This is common sense intuition and matches all other management protocols.

The case where it is necessary to perform a merge operation is where a *single internal configuration property* is set (directly or indirectly) by *two or more data nodes in the external models*.  At the top level, there are two cases to consider:

1) In the case that the two (or more) external data nodes that are being mapped are all in the same model family (e.g., both are in OC YANG) then if we cannot consistently map those external values into the equivalent internal configuration property then we must reject the configuration change, because otherwise we would not be enacting some of the OC configuration, and the client would have no way of knowing value in the OC model has precedence.  This behavior applies whether the OC configuration is configured explicitly or an in-scope default value (e.g., a parent list entry has been created).  An example of such a configuration mapping issue is where there is only a coarse config knob in the internal model and more fine-grained equivalent config knobs in the external model).  It is possible to map the set of fine-grained config knobs in the external model to the coarse config knob in the internal model as long they consistently collectively map to the same value for the coarse config knob.  Note, the ideal solution here is to extend the internal model with fine-grained config knobs as well.

2. The second case is where the two external data nodes being mapped are in different model families (e.g., one is configured via a device native CLI, and the other is configured via OC YANG).  This is the scenario where it is far more likely that very similar of equivalent configuration exists in both OC and Native model families.  This scenario is also more complex, since it must consider how: a) two explicitly configured values interact, b) how explicitly configured values in one model interact with in-scope default values in the other model, and c) how in-scope default values between the two model families interact.  

For the second option, I believe that the most consistent semantics (for OC) are:

1) If conflicting explicitly configured values are configured in two model families, then the configuration should be rejected.  E.g., if the client explicitly configures MTU 4000 in OC, and explicitly configures MTU 3000 in the CLI/Native then the right thing to do here is to reject the configuration change.
2) Any explicitly configured value overrides a default/implicit value from another model family.  E.g., an explicitly configured interface MTU in OC should override the device default MTU for the interface in the CLI/Native schema, even if that interface has been configured in the CLI/Native configuration, e.g., to enable some other per-interface config knob not expressed in the OC schema.
3) In the case where there are different in-scope default values between the two external model families that are getting mapped, then the default value from the OC model family should take precedence over the default value in the native model family.  E.g., if OC defined the default MTU of an interface to be 2000 bytes, and the CLI/Native schema defined the default MTU of an interface to be 1500 bytes, and the interface existed in both CLI/Native & OC configuration, then since the assumption is that the device is primarily being configured via OC then the default value in OC should override the default value in the CLI/Native schema.

Finally, there are a couple of variants to the options above:

1) it should probably be possible via device level configuration to allow the client to decide which configuration family takes precedence (e.g., CLI/Native or OC), depending on whether the device is primarily being configured via OC or via CLI/Native.

2) rather than rejecting conflicting explicitly configured values, one could instead allow them to be resolved the same way as for defaults.  In yesterday's call, I think that there was a suggestion that the value configured on the CLI should take precedence over an OC configured, but I suspect that this would be confusing in many deployments unless conflicting defaults values are handled similarly.  Ultimately, I think that explicitly rejecting conflicting values is like to cause less surprise for clients and is a probably better way to go.

Rob

liulk:
Hi Rob, thanks for the thoughtful feedback. Could we move the discussion to #740 so it is tracked appropriately? If a verdict is reached where we must change the test, it would not be viable to track that work in a PR that's already merged.

PR #477, opened 2022-09-01
Add tool to generate wiki docs

bstoll:
It is currently difficult to see all test plans in a single location.  This PR creates a Github action triggered on every commit that will update the Github Wiki with all the test plans.

An example of the output can be seen on my fork at https://github.com/bstoll/featureprofiles/wiki

### DIFF
--- a/.github/workflows/wiki.yml
+++ b/.github/workflows/wiki.yml
@@ -3,8 +3,6 @@ name: Update Wiki
 on:
   push:
     branches: [ main ]
-  pull_request:
-    branches: [ main ]
 
 jobs:
   update-wiki:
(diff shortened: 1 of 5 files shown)


bstoll:
I think there is a permissions issue with the Github actions token needing permission to read/edit the wiki.  Will look more into this...

liulk (dismissed):
The sidebar looks a little crammed. Maybe in the future we can make a full index page?

Not critical right now. Thanks for putting in the work!

bstoll:
It looks like the initial issue was because we didn't have a Wiki created yet.  I was trying to write the wiki in the pull request, but the Github Actions token does not have permission to write to the wiki when executed from the PR (anyone can create a PR, so that would be a security problem).

The task shouldn't be run from a PR anyway. It should only run on merges to the main branch.  It looks like the actions that are run in the main branch do have correct permissions in the logs, so I think it will work once merged.

liulk (approved):

PR #480, opened 2022-09-01
RT-5.1 singleton unicast packet fix less-than check

liulk:
#383 

Id: 5bf479c3-0dcf-4594-9629-0ce69dcf410a

### DIFF
--- a/feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
+++ b/feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
@@ -339,8 +339,8 @@ func (tc *testCase) testFlow(t *testing.T, packetSize uint16, ipHeader ondatra.H
 	} else if avg := octets / outPkts; avg > uint64(tc.mtu) {
 		t.Errorf("Flow source packet size average got %d, want <= %d (MTU)", avg, tc.mtu)
 	}
-	if outPkts < p1InDiff.unicast {
-		t.Errorf("Incorrect number of packets inbound received got %d, want > %d", outPkts, p1InDiff.unicast)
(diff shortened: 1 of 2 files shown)


feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -339,8 +339,8 @@ func (tc *testCase) testFlow(t *testing.T, packetSize uint16, ipHeader ondatra.H
 	} else if avg := octets / outPkts; avg > uint64(tc.mtu) {
 		t.Errorf("Flow source packet size average got %d, want <= %d (MTU)", avg, tc.mtu)
 	}
-	if outPkts < p1InDiff.unicast {
-		t.Errorf("Incorrect number of packets inbound received got %d, want > %d", outPkts, p1InDiff.unicast)
+	if p1InDiff.unicast < outPkts {
+		t.Errorf("DUT received too few source packets: got %d, want >= %d", p1InDiff.unicast, outPkts)
```
xw-g, line 343: nit: fewer?

xw-g (approved):

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
liulk: "too fewer" is not the correct grammar.

https://ell.stackexchange.com/questions/55053/why-are-too-less-and-too-fewer-wrong

"too few" is the correct qualifier for packet because "packet" is a countable noun.

https://ell.stackexchange.com/questions/182732/is-too-few-equivalent-to-not-enough

PR #481, opened 2022-09-02
changed call from replace to update to avoid overwriting LAG config

manan-patel:
The automation configures lag mode and then overwrites it with ip address configuration with existing "Replace" call, changing it to "Update" call appends to the configuration.

### DIFF
--- a/feature/interface/aggregate/ate_tests/aggregate_test/aggregate_test.go
+++ b/feature/interface/aggregate/ate_tests/aggregate_test/aggregate_test.go
@@ -214,7 +214,7 @@ func (tc *testCase) configureDUT(t *testing.T) {
 	tc.configDstAggregateDUT(agg, &dutDst)
 	aggPath := d.Interface(tc.aggID)
 	fptest.LogYgot(t, tc.aggID, aggPath, agg)
-	aggPath.Replace(t, agg)
+	aggPath.Update(t, agg)
 
 	srcp := tc.dutPorts[0]
 	srci := &telemetry.Interface{Name: ygot.String(srcp.Name())}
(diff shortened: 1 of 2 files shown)


feature/interface/aggregate/ate_tests/aggregate_test/aggregate_test.go
```diff
@@ -214,7 +214,7 @@ func (tc *testCase) configureDUT(t *testing.T) {
 	tc.configDstAggregateDUT(agg, &dutDst)
 	aggPath := d.Interface(tc.aggID)
 	fptest.LogYgot(t, tc.aggID, aggPath, agg)
-	aggPath.Replace(t, agg)
+	aggPath.Update(t, agg)
```
liulk, line 217: We definitely want to keep Replace.

The tc.setupAggregateAtomically() was guarded by a deviation because some vendors make it impossible to create an empty lag before adding members to it. It only contained enough update to allow the lag to exist.

The replace here should contain the actual config that the test plan intended.

liulk (changes requested):

feature/interface/aggregate/ate_tests/aggregate_test/aggregate_test.go
manan-patel: Thank you for the quick response, we'll take a look.

liulk: Any update?

manan-patel: We are working with the engineering team to resolve the issue.

dplore: Hi @manan-patel , any new updates?

mojiiba: WE have fixed this issue on our side. Please close the PR.

PR #484, opened 2022-09-06
feature.textproto for qos add dscp

sezhang2:
Added missing qos config and telemetry paths
  - Reference:  https://ops.openconfig.net/branches/models/master/openconfig-qos.html#


bstoll:
I am thinking ECN should be in its own feature profile, with only the "enable-ecn" leaf going into it.   ECN would require the base QOS feature still.

sezhang2:
> I am thinking ECN should be in its own feature profile, with only the "enable-ecn" leaf going into it. ECN would require the base QOS feature still.

Ack. Deleted enable-ecn from feature profile. We will discuss if we need to create a separate ecn feature profile.

liulk (approved):

PR #490, opened 2022-09-07
Rnhaddad/p4rt wbb

rnhaddad:
P4RT WBB P4info wrapper package and example code.
This experimental code is meant to show how the Ondatra framework can be integrated over a wrapper code that can simplify the P4RT integration.

### DIFF
--- a/feature/experimental/p4rt/wbb.p4info.pb.txt
+++ b/feature/experimental/p4rt/wbb.p4info.pb.txt
@@ -315,6 +315,68 @@ tables {
 }
 tables {
   preamble {
+    id: 33554691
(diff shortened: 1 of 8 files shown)


rnhaddad:
@hariramnat Can you please take a look as discussed?

dplore:
Hi, do you intend to merge this code as a test case? Or should we mark it as a draft and not assign it for review?  It seems like it is for reference purposes only and not to be merged as a test case, right?

dplore:
I will close this now that we have a merged p4rt test?   Ref: #495 and #498    Please add a comment if this was not your intention.

PR #497, opened 2022-09-09
Establish docs/code-style.md and docs/testplan-style.md

liulk:
* (A) docs/code-style.md
  - Contains code style guide formerly found in CONTRIBUTING.md
* (A) docs/testplan-style.md
  - Newly introduced test plan style guide.
* (M) CONTRIBUTING.md
  - Moved code style out to docs/code-style.md, replaced it with rationale and references to code and test plan style guides.
* (M) .github/workflows/pull_request.yml
  - Updated reference to docs/code-style.md
* (M) tools/allowed_file_types.sh
  - Updated reference to docs/code-style.md 
* (M) tools/check_ip_addresses.pl 
  - Updated reference to docs/code-style.md


docs/testplan-style.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Vendor Neutrality
+
+Test plans should use a vendor-neutral mechanism to configure the DUT, e.g. in
+the OpenConfig public model or other OpenConfig protocols such as gNMI, gRIBI,
+and gNOI. It should depend only on the behavior of the open specification.
+
+### Device/Topology Neutrality
+
+Tests are to be built against a minimal topology that consists of a single
```
robshakir, line 66: We should say "where possible" here -- since there might be cases where this is undesirable, or not possible. In these cases, these could still be functional tests. Equally, the reference implementation for some management plane protocols could be something other than ATE.

docs/testplan-style.md
```diff
@@ -0,0 +1,243 @@
...
+
+Each test plan must choose from one of the following testbeds and adhere to it
+for the entirety of the test plan:
+
+*   atedut\_2: two interconnected ports between ATE and DUT.
+*   atedut\_4: four interconnected ports between ATE and DUT.
+*   atedut\_12: 4 singleton ports and 8 breakout ports interconnected between
+    ATE and DUT. Ondatra testbed models the logical ports, but the test is
+    responsible for activating the breakout.
+*   dutdut: four interconnected ports between two DUTs.
```
robshakir, line 78: Having `dutdut` here is perhaps not consistent with the need to have single DUT, single ATE tests as defined above.

I wonder also whether being prescriptive about these topologies is assuming that one knows what the test cases that are being written are. Is there any reason that someone couldn't write a test that has 32, 64, 96, ... ports available to it - for e.g., scaling testing.

docs/testplan-style.md
```diff
@@ -0,0 +1,243 @@
...
+
+Test plans must not hardcode assumptions about the presence of specific hardware
+components (e.g. switch chip, controller card, line card). If needed, it may
+accept a "device profile" which parameterizes the expectations about the device.
```
robshakir, line 82: I haven't seen this concept anywhere in the current code. Is it a new concept? Should we perhaps have demonstrated that this works as we expect it to through some examples before declaring it a requirement?

What does a device profile look like?

docs/testplan-style.md
```diff
@@ -0,0 +1,243 @@
...
+If the test plan is about a feature that does not make sense for a given device
+profile, the test should be skipped. For example, a test about controller card
+switchover should be skipped if the device profile indicates that the device is
+expected to have only one controller card, rather than skipping based on the
+number of controller cards detected from the device.
+
+### Functional Testing vs. Fuzzing
+
+Initially, the test plan will consist of manually identified functional features
```
robshakir, line 91: Does this need to be the case? Is there any reason that one cannot define a test plan that is fuzzing? Is this solely because of where the current tests that are defined are coming from rather than what could be contributed by anyone to this repository?

robshakir: To be more concrete, this feels like a priority call that could be made by each author of a new test. For some folks, fuzzing might be the most important thing :-)

docs/testplan-style.md
```diff
@@ -0,0 +1,243 @@
...
+used as pass/fail criteria of the test, such that telemetry is validated as a
+core component of each test.
+
+## Readable Language
+
+All code and documentation should follow
+[Google developer documentation style guide](https://developers.google.com/style/word-list)
+for the use of inclusive language.
+
+Test plan should be written in a natural language. Avoid introducing your own
```
robshakir, line 114: Can you explain what this means? In some tests it is hard to define things solely in prose, and having some form of parseable pseudo-code seems quite useful. It'd be good to understand the intention behind this statement.

docs/testplan-style.md
```diff
@@ -0,0 +1,243 @@
...
+notation because people who read the test plan are not going to be familiar with
+the notation. We need the test plan to be self-explanatory.
+
+### Example: Good
```
robshakir, line 118: @xw-g When I've reviewed some of the changes that you've suggested, I've found something that looks like "bad" here quite instructive. Do you have thoughts here?

liulk (commented):
@robshakir @xw-g @dplore 

Before I respond to the comments individually, I want to emphasize that the style guides are not "Holy Scripture" that is set in stone once written. They are intended to be a living document describing what the best practices are. When the use case arises, we can relax some of the mandates within reason.

However, it is motivated by the needs that have arisen from the current use cases. All of the content are motivated by things I would like to see done better and should try to do consistently in the future.

robshakir:
I agree that they can be iterated on, but once they're committed, I think it becomes something that we then use as guidance for other contributors. I think that this means that they need to be well-founded ideas that are generally thought of as things that will help, with enough detail that they can be implemented by project contributors.

My primary concern here is that it's not clear what the justification for some of these is from the perspective of enabling wider functional testing of devices.

liulk:
I sense that you don't trust me on my judgment why this test plan style guide is needed. You have to trust me on this.

We need to take into account that the test plan writer and the test code implementer are often different people, and a chronic pain point we've had, myself speaking as someone who implemented several test plans written by someone else, is that the test plan did not provide enough detail about what behavior needs to be tested.

For example, I'm now revisiting a TODO in RT-5.1 Singleton Interfaces which mentions in passing the need to verify negotiated-port-speed. It did not say that it is only populated when auto-negotiation is enabled, but most optics don't support auto-negotiation, so the test has to test both with and without auto-negotiation. This is just an unwritten consequence of the test plan. Furthermore, IEEE 802.3-2012 says that devices should allow auto-negotiation to be enabled even though PHY does not support it, so that's another behavior that needs to be tested.

My point is, the devil is in the detail that's omitted by the test plan author, and that's why it's hard to predict the productivity of the FNT team. Just because a test plan is written does not mean it is implementation ready. I feel that the test code author has to do too much research, and this time spent is not properly acknowledged.

Although there will always be details that one cannot foresee, what makes my life easier as a test code implementer is reading a well-written test plan with an attention to detail. I have seen some helpful styles from the existing test plans, and I believe you wrote the vast majority of them. But what I'm also beginning to see is that different authors start introducing their idiosyncratic styles. It's the same concern from a readability point of view. When reading the test plans written by other authors, one has to mentally switch gear to cater to the style of that author, and that is hurting productivity.

ps. Why does a test plan need to specify the IP addresses? Not all software engineers understand that you can't just self-assign random address. They are assigned by IANA, with special ranges reserved for specific purposes by the relevant RFCs. Same goes for AS numbers. When a testbed uses inband management like our team internally does for some legacy reason, a poorly chosen IP address for testing can wreck havoc to other parts of the network.

Trust me.

robshakir:
Let's discuss this tomorrow in person. I don't think this is a trust issue -- but rather a question of the right bar to set, and the right place to focus efforts for the goals that we have. 

Personally, I believe that the specification of the test plan is better done in code, with the usual approach of comments for understanding that test code. The existence of the test plan itself is a little more to do with happenstance than trying to provide a specification for a development that is being done. I think this is where we're not aligned.

PR #499, opened 2022-09-11
Add KillProcess RPC into GNOI feature profile

hariramnat:
Add KillProcess RPC into GNOI feature profile


dplore (approved):
LGTM, ref: https://github.com/openconfig/gnoi/blob/main/system/system.proto#L87

PR #516, opened 2022-09-15
Admin - copy of ate-tests RT-5.2, RT-5.3 and TE-3.2

octpetre:


octpetre:
I've also added the atedut_lag.testbed file required in ondatra reservation of LAG groups

liulk (approved):

PR #527, opened 2022-09-16
Fix failed BGP tests

hariramnat:


sezhang2 (approved):

robshakir:
@sezhang2 @hariramnat @dplore I note in this PR we changed loss tolerance from 1-2%. Was there a reason to do this please? We should be encoding expectation rather than observed behaviour in the tests.

PR #530, opened 2022-09-16
Modify RT-5.1: singleton to verify negotiated-port-speed

liulk:
Test Result Id: 6d925554-a3f5-44bc-9804-39eb9cae5ce5

### DIFF
--- a/feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
+++ b/feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
@@ -90,8 +90,20 @@ var (
 	}
 )
 
+// autoMode specifies the type of auto-negotiation behavior testing: forced, auto, and
+// auto-negotiation while also specifying duplex and speed.  The last case is permitted by
+// IEEE Std 802.3-2012 and OpenConfig.
+type autoMode int
+
+const (
+	forcedNegotiation autoMode = iota
+	autoNegotiation
+	autoNegotiationWithDuplexSpeed
+)
+
 type testCase struct {
-	mtu uint16 // This is the L3 MTU, i.e. the payload portion of an Ethernet frame.
+	mtu  uint16 // This is the L3 MTU, i.e. the payload portion of an Ethernet frame.
+	auto autoMode
 
 	dut *ondatra.DUTDevice
 	ate *ondatra.ATEDevice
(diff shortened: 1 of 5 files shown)


feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -111,11 +123,17 @@ var portSpeed = map[ondatra.Speed]telemetry.E_IfEthernet_ETHERNET_SPEED{
...
-		e.DuplexMode = telemetry.Ethernet_DuplexMode_FULL
+	e := i.GetOrCreateEthernet()
+	if tc.auto == autoNegotiation || tc.auto == autoNegotiationWithDuplexSpeed {
+		e.AutoNegotiate = ygot.Bool(true)
+	} else {
 		e.AutoNegotiate = ygot.Bool(false)
-		e.PortSpeed = speed
+	}
+	if tc.auto == forcedNegotiation || tc.auto == autoNegotiationWithDuplexSpeed {
+		if speed, ok := portSpeed[dp.Speed()]; ok {
```
xw-g, line 133: do we want to handle scenarios where `ok == false`?

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -198,6 +220,22 @@ func (tc *testCase) verifyInterfaceDUT(
 		t.Errorf("%s oper-status got %v, want %v", dp, got, opUp)
 	}
 
+	if speed, ok := portSpeed[dp.Speed()]; ok {
```
xw-g, line 223: Similar comment as above about `ok != true`.

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -198,6 +220,22 @@ func (tc *testCase) verifyInterfaceDUT(
...
+		if tc.auto == forcedNegotiation || tc.auto == autoNegotiationWithDuplexSpeed {
+			if got := dip.Ethernet().PortSpeed().Get(t); got != speed {
+				t.Errorf("%s port-speed got %v, want %v", dp, got, speed)
+			}
+		}
+		if tc.auto == autoNegotiation || tc.auto == autoNegotiationWithDuplexSpeed {
+			if dip.Ethernet().AutoNegotiate().Get(t) {
```
xw-g, line 230: if `dip.Ethernet().AutoNegotiate().Get(t) == false`, that means the port does not support auto negotiation and auto negotiation is ignored, right? I guess in this case we still want to check port speed?

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -400,19 +434,62 @@ func (tc *testCase) run(t *testing.T) {
...
+// while overriding port speed and duplex.
+func TestNegotiate(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	ate := ondatra.ATE(t, "ate")
+
+	for auto, name := range autoModeName {
+		t.Run(name, func(t *testing.T) {
+			top := ate.Topology().New()
+			tc := &testCase{
+				mtu:  1500,
```
xw-g, line 475: nit: some comment on this magic 1500 number? (e.g. // fix the L3 MTU to 1500 because this test focus on auto negotiation).

xw-g (approved):
A few comments to consider before merging.

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
liulk: The port-speed telemetry reflects only what the config sets, and it's checked above on line 225. On the other hand, the negotiated-port-speed is only set if the port supports auto negotiation.

In the autoNegotiationWithDuplexSpeed case, it will trigger both port-speed and negotiated-port-speed checks, so it's working as intended.

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
liulk: Seems like you're able to figure this out without the explanation, so it goes without saying. 😀

PR #532, opened 2022-09-19
Update complete_chassis_reboot_test.go

sezhang2:
Removed duplicated  SW versions before comparison 
  - Test passed on hardware platform

### DIFF
--- a/feature/gnoi/system/tests/complete_chassis_reboot/complete_chassis_reboot_test.go
+++ b/feature/gnoi/system/tests/complete_chassis_reboot/complete_chassis_reboot_test.go
@@ -98,7 +98,8 @@ func TestChassisReboot(t *testing.T) {
 			}},
 	}
 
-	expectedVersion := dut.Telemetry().ComponentAny().SoftwareVersion().Get(t)
+	versions := dut.Telemetry().ComponentAny().SoftwareVersion().Get(t)
+	expectedVersion := FetchUniqueItems(t, versions)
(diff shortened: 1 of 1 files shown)


amrindrr (approved):

feature/gnoi/system/tests/complete_chassis_reboot/complete_chassis_reboot_test.go
```diff
@@ -176,3 +178,17 @@ func TestChassisReboot(t *testing.T) {
...
+
+func FetchUniqueItems(t *testing.T, s []string) []string {
+	itemExisted := make(map[string]bool)
+	var uniqueList []string
+	for _, item := range s {
+		if _, ok := itemExisted[item]; !ok {
+			itemExisted[item] = true
+			uniqueList = append(uniqueList, item)
+		} else {
+			t.Logf("Detected duplicated item: %v", item)
```
wenovus, line 190: If this is helpful for debugging then would be nice to add some context to what is being duplicated.

wenovus (approved):

feature/gnoi/system/tests/complete_chassis_reboot/complete_chassis_reboot_test.go
sezhang2: I am adding the log to collect if we found the duplicated items on different platforms. 

I will update this code when I have more info.

PR #533, opened 2022-09-19
Component helper function 

amrindrr:
Use helper function defined here
https://github.com/openconfig/featureprofiles/blob/main/internal/components/components.go


sezhang2 (approved):

sezhang2 (changes requested):
Need to fix the static analysis error

sezhang2 (approved):

PR #501, opened 2022-09-12
complete_chassis_reboot_test: fix gnoi connection issue

mojiiba:
Hi, the gnoi reboot tests use the dut.RawAPIs().GNOI().Default(t). This call use the cached connection. However, since rebooting device closes the GNOI, so the cached connection will not be usable by the next testcase.

PR #539, opened 2022-09-21

sezhang2:
Fixed the similar connection issue like #501 using dut.RawAPIs().GNOI().New(t) instead of dut.RawAPIs().GNOI().Default()


liulk (approved):

bstoll:
Clarifying from an offline conversation:

This change does not do anything in the current state, but when another test case is added to this (depending on execution order) it will be necessary for each case to use a unique connection.

sezhang2:
> Clarifying from an offline conversation:
> 
> This change does not do anything in the current state, but when another test case is added to this (depending on execution order) it will be necessary for each case to use a unique connection.

Ack.  I plan to add another switchover, so there will be 2 switchovers. I makes similar change so that I will not miss it in future.

PR #546, opened 2022-09-22
Update README.md

sezhang2:
Added /interfaces/interface/state/physical-channel

### DIFF
--- a/feature/gnmi/ate_tests/telemetry_basic_check_test/README.md
+++ b/feature/gnmi/ate_tests/telemetry_basic_check_test/README.md
@@ -24,7 +24,7 @@ following features:
 
 *   Interface physical channel
 
-    *   Check interface physical-channel exists and correct.
+    *   Check interface physical-channel exists.
         *   /interfaces/interface/state/physical-channel
 
 *   Interface status change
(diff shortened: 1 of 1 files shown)


bstoll:
I think this should either define what "correct" is (marked as a TODO), or just test for existence. Judging from the code, all that is implemented is to test for existence of some values:

https://github.com/openconfig/featureprofiles/blob/5108f1151bdeb95e39685d1b298a506d3c941268/feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go#L134-L143

If we wanted to get more detailed on what correct is, I think we could validate that the channels can be looked up under /components/transceiver/physical-channels/channel (ie, something like `dut.Telemetry().Component(dp.Name()).Transceiver().Channel(phyChannel[n]).Get(t)`).  Otherwise, this leaf could just be returning random numbers and we would have no idea.

Also, this is going to be a bit confusing because this test plan exists at feature/gnmi/otg_tests/telemetry_basic_check_test/README.md.  Could you copy whatever changes you make in this file to the file in the otg_tests directory too?

sezhang2:
> 

Updated the doc to reflect the test checks the existence of  interface physical-channel telemetry path.

Looks like we have duplicated README and test code for ate and otg test. We need to investigate how to use same test code to support both ate and otg tests to avoid the duplicated efforts.

bstoll (approved):

Issue #277, opened 2022-06-27
gribigo_compliance_test.go


cprabha:
Hi,

Is there any pre-configuration needed to run gribigo_compliance_test.go ??

In compliance.go file , found using prefixes like below. should we configure them prior to run test?

WithIPAddress("192.0.2.1"))
WithIPAddress("2.2.2.2"))
WithIPAddress("192.0.2.2"))

Thanks,
Prabha

cprabha:
@bstoll , please could you share your inputs on this issue. Thank you.

bstoll:
This test does assume GRIBI is running/configured on the DUT, but it does not forward traffic - it is only doing control plane validation.  I don't think you should need prefixes configured.  Were you running into issues because of this?

cprabha:
@bstoll , For  gRiBI programmed route nexthop will be in inactive state if we dont configure interfaces?

dplore:
Hi, redirecting to @xw-g to help answer

xw-g:
I think this is a good question. In the case of `FIB ACK`, the compliance test does need the DUT to be able to resolve the NH L3 IP in someway. 

1. @cprabha, maybe we can use the `skipFIBACK` flag to skip the FIB ACK case first. 
2. @bstoll, do you think we can specify, or configure, the testbed as part of the compliant test? I think 2 L3 interface (UP), or 2 L3 sub-interfaces(UP) should be good enough.

cprabha:
@xw-g Thanks for sharing , I will set skipfalg for FIBACK.
Also please could you share number of links needed if we are configuring interfaces on DUT

xw-g:
Compliance tests does not involve traffic. With a quick glance, I see at most 2 NH with next-hop IP mentioned in a NHG. So I would guess we probably only need 2 L3 interfaces/sub-interfaces.

cprabha:
@xw-g sure, let me configure two interfaces on DUT and re run test.
Thanks,
Prabha

bstoll:
We discussed we will add steps to configure interfaces to allow nexthops to resolve to actual interfaces and update gribigo tests to use IPs allocated out of test ranges (instead of ie 2.2.2.2).  This should help run the tests in a fiback scenario where successful resolution is required.

cprabha:
Thank you @bstoll

PR #551, opened 2022-09-22
Add ATE to GRIBI Compliance Test

bstoll:
Add 3 interfaces as documented in the [GRIBI compliance tests](https://github.com/openconfig/gribigo/blob/98b0f3bbf1f750542fc505f9a3e24d6d9ce67b3d/compliance/compliance.go#L1129) to the test setup.  FIBACK tests need to be able to resolve the next-hops to some valid next-hop.

I will still need to go over the gribigo compliance tests to ensure all the next-hops are using one of these 3 IPs.

See #277.


feature/gribi/ate_tests/gribigo_compliance_test/gribigo_compliance_test.go
```diff
@@ -103,3 +132,40 @@ func TestCompliance(t *testing.T) {
...
+	p3 := dut.Port(t, "port3")
+
+	dut.Config().Interface(p1.Name()).Replace(t, dutPort1.NewInterface(p1.Name()))
+	dut.Config().Interface(p2.Name()).Replace(t, dutPort2.NewInterface(p2.Name()))
+	dut.Config().Interface(p3.Name()).Replace(t, dutPort3.NewInterface(p3.Name()))
+
+	d := &telemetry.Device{}
+	ni := d.GetOrCreateNetworkInstance(*nonDefaultNI)
+	ni.Type = telemetry.NetworkInstanceTypes_NETWORK_INSTANCE_TYPE_L3VRF
+	ni.GetOrCreateProtocol(telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, "static")
```
sezhang2, line 149: Is "static" vendor independent?

feature/gribi/ate_tests/gribigo_compliance_test/gribigo_compliance_test.go
```diff
@@ -103,3 +132,40 @@ func TestCompliance(t *testing.T) {
...
+	fptest.LogYgot(t, "nonDefaultNI", nip, nip.Get(t))
+}
+
+// configreATE configures port1-3 on the ATE.
+func configureATE(t *testing.T, ate *ondatra.ATEDevice) *ondatra.ATETopology {
+	top := ate.Topology().New()
+
+	p1 := ate.Port(t, "port1")
+	p2 := ate.Port(t, "port2")
+	p3 := ate.Port(t, "port3")
```
sezhang2, line 162: Can you add the topology with the specific port names like port1, port2, port3 for both dut and ate in README or test code?

sezhang2 (dismissed):
Add license header by running addlicense -l apache

feature/gribi/ate_tests/gribigo_compliance_test/gribigo_compliance_test.go
bstoll: I believe different implementations have different requirements here, but that is a more general issue we need to sort out still.  I think strictly speaking any value can be used here, but that seems like it has caused some problems.

This will likely need to change to a deviation in the future once we sort out what it _should_ be.

feature/gribi/ate_tests/gribigo_compliance_test/gribigo_compliance_test.go
bstoll: Done

bstoll:
The comments should be addressed by latest commit.

sezhang2 (approved):

PR #592, opened 2022-09-29
added networkinstance type

cprabha:
Hi,

I have added network instance type, its missing from previous commits.(#317)

Thanks,
Prabha


bstoll:
Is this PR still necessary?  I just merged #567 which removed configuring the router-id at the network instance level entirely, so there are no modifications to the default network instance.

I also was not following why this change was required to begin with either: shouldn't the default network instance already have the type field set?  Why does it need to be replaced?

cprabha:
Hi Brandon,

Instance type config is required when DUT is assumed to have no network instance configured prior to running the test.
If instance type is not configured prior to script run, config failure will be seen as instance type is mandatory.

If DUT already have instance type configured, it also will be of type DEFAULT_INSTANCE.
Script also does replace with same type, hence replacing instance type is not a problem such DUT devices.

Thanks,
Prabha




Juniper Business Use Only

Issue #351, opened 2022-07-27
protocl name for static route policy 


mojiiba:
Hi, a few of the tests (e.g., https://github.com/openconfig/featureprofiles/blob/3da579ef74aa08d90476105630fe213376e67cf1/feature/gribi/tests/gribigo_compliance_test/gribigo_compliance_test.go#L56) configure a static route and use "static" as the protocol name as shown below:

```
/network-instances/network-instance[name=non-default-vrf]
{
  "openconfig-network-instance:config": {
    "name": "non-default-vrf"
  },
  "openconfig-network-instance:name": "non-default-vrf",
  "openconfig-network-instance:protocols": {
    "protocol": [
      {
        "config": {
          "identifier": "openconfig-policy-types:STATIC",
          "name": "static"
        },
        "identifier": "openconfig-policy-types:STATIC",
        "name": "static"
      }
    ]
  }
}
```

This value is vendor specific and we need to have a flag to pass the name. I can have a pull request to add the flag, however I am not sure where do you prefer to add the flag since this will be used by few tests?

xw-g:
@bstoll

xw-g:
I'm ok with starting a flag. @bstoll to comment further.

xw-g:
#600

PR #600, opened 2022-09-30
TE-5.1: add a flag for static router protocl name gRIBI Get RPC)

mojiiba:
This pull fixes issue https://github.com/openconfig/featureprofiles/issues/351. I will change the other tests if you agree with this change.


feature/gribi/ate_tests/get_rpc_test/get_rpc_test.go
```diff
@@ -318,7 +322,7 @@ func testIPv4LeaderActive(ctx context.Context, t *testing.T, args *testArgs) {
 	// are returned, with no entry returned for 198.51.100.192/64.
 	dc := args.dut.Config()
 	ni := dc.NetworkInstance(*deviations.DefaultNetworkInstance).
-		Protocol(telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, "STATIC")
+		Protocol(telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, *protocolName)
```
Ankur19, line 325: Suggest making this `DEFAULT`, since that is the OC compliant default static protocol name.

If `DEFAULT` doesn't work, then add a `deviation`

mojiiba: Hi @Ankur19, to the best of my knowledge  OC does not specify the name for static protocol name.  This value is vendor specific and can be anything. So, I do not think using values other than "DEFAULT" is an OC deviation.

xw-g: @mojiiba you PR makes the name configurable for different testbed (which I think it does make sense). [The OC](https://github.com/openconfig/public/blob/95948a506f12c904e7414e7d9cc27c467bee9070/release/models/network-instance/openconfig-network-instance.yang#L1271) does specify default value as `DEFAULT`, I"m not sure why the flag value can not be set to `DEFAULT`, since it's configurable anyway?

mojiiba: @xw-g sorry for the confusion. Flag can be set to anything including DEFAULT. I meant value other than "DEFAULT" should not be considered as a deviation.

bstoll:
I think this is a good case to use a global deviation.  This is a recurring problem across tests.

I would propose adding something like this to `internal/deviations/deviations.go`:

```
DefaultProtocolStaticName = flag.String("deviation_default_protocol_static_name", "DEFAULT",
		"The name used for the static routing protocol.  The default name in OpenConfig is \"DEFAULT\" but some devices use other names.")
```

The OpenConfig model does set DEFAULT as the preferred value.  This was a somewhat recent change:

https://github.com/openconfig/public/blob/95948a506f12c904e7414e7d9cc27c467bee9070/release/models/network-instance/openconfig-network-instance.yang#L1269-L1281

mojiiba:
> I think this is a good case to use a global deviation. This is a recurring problem across tests.
> 
> I would propose adding something like this to `internal/deviations/deviations.go`:
> 
> ```
> DefaultProtocolStaticName = flag.String("deviation_default_protocol_static_name", "DEFAULT",
> 		"The name used for the static routing protocol.  The default name in OpenConfig is \"DEFAULT\" but some devices use other names.")
> ```
> 
> The OpenConfig model does set DEFAULT as the preferred value. This was a somewhat recent change:
> 
> https://github.com/openconfig/public/blob/95948a506f12c904e7414e7d9cc27c467bee9070/release/models/network-instance/openconfig-network-instance.yang#L1269-L1281

Sure @bstoll   I will do what you suggested.

xw-g:
+1 to @bstoll's comment to make it a global deviation flag.

xw-g:
@mojiiba please ping me and @bstoll when next commit comes in. thanks!

mojiiba:
> @mojiiba please ping me and @bstoll when next commit comes in. thanks!

@bstoll I have added the flag as you suggested. Please let me know if anything else is needed.

internal/deviations/deviations.go
```diff
@@ -87,4 +87,6 @@ var (
 	GRIBIRIBAckOnly = flag.Bool("deviation_gribi_riback_only", false, "Device only supports RIB ack, so tests that normally expect FIB_ACK will allow just RIB_ACK.  Full gRIBI compliant devices should pass both with and without this deviation.")
 
 	NextHopAFTNotSupported = flag.Bool("deviation_nexthop_aft_not_supported", false, "Device currently doesnot support AFT Next Hop Telemetry. A fully compliant device should support all types of AFT telemetry without this deviation.")
+
+	DefaultProtocolStaticName = flag.String("deviation_default_protocol_static_name", "DEFAULT", "The name used for the static routing protocol.  The default name in OpenConfig is \"DEFAULT\" but some devices use other names.")
```
xw-g, line 91: should it be just `StaticProtocolName` with the default value as `DEFAULT`.

xw-g (commented):
I defer my LGTM to @bstoll

internal/deviations/deviations.go
mojiiba: I took the suggestion from @bstoll, but I can change it if you prefer StaticProtocolName.

mojiiba:
@xw-g @bstoll  can you please merge this pull?

bstoll (approved):

PR #601, opened 2022-09-30
TE-2.1, TE-4.2:   Add GRIBIPreserveOnly flag and fix refrencing issue

mojiiba:
Hi This pull adds GRIBIPreserveOnly guard to TE-2.1 and TE-4.2. It also  fixes referencing issue for deleting entries in TE-2.1. We can not delete a NHG before deleting all dependent ipv4 entries and the same holds for NH. The reason that current tests does not catch this issue is because it uses delete mode and does not verify RIB ack.

### DIFF
--- a/feature/gribi/ate_tests/leader_failover_test/leader_failover_test.go
+++ b/feature/gribi/ate_tests/leader_failover_test/leader_failover_test.go
@@ -250,7 +250,9 @@ func TestLeaderFailover(t *testing.T) {
 	}
 
 	t.Run("SINGLE_PRIMARY/PERSISTENCE=DELETE", func(t *testing.T) {
-
+		if *deviations.GRIBIPreserveOnly {
+			t.Skip()
+		}
 		// Set parameters for gRIBI client clientA.
 		// Set Persistence to false.
 		clientA := &gribi.Client{
@@ -289,6 +291,9 @@ func TestLeaderFailover(t *testing.T) {
 	})
 
 	t.Run("ShouldDelete", func(t *testing.T) {
+		if *deviations.GRIBIPreserveOnly {
+			t.Skip()
+		}
 
 		t.Logf("Verify through Telemetry and Traffic that the route to %s has been deleted after gRIBI client disconnected", ateDstNetCIDR)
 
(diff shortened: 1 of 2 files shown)


feature/gribi/ate_tests/leader_failover_test/leader_failover_test.go
```diff
@@ -250,7 +250,9 @@ func TestLeaderFailover(t *testing.T) {
 	}
 
 	t.Run("SINGLE_PRIMARY/PERSISTENCE=DELETE", func(t *testing.T) {
-
+		if *deviations.GRIBIPreserveOnly {
+			t.Skip()
```
bstoll, line 254: Can we add a skip reason?  I see other tests we have used t.Skip("Skipping due to --deviations_gribi_preserve_only").  This makes logs easier to read.

bstoll (commented):
Thanks, this looks good.

feature/gribi/ate_tests/leader_failover_test/leader_failover_test.go
mojiiba: thanks!, addressed

feature/gribi/ate_tests/leader_failover_test/leader_failover_test.go
```diff
@@ -250,7 +250,9 @@ func TestLeaderFailover(t *testing.T) {
 	}
 
 	t.Run("SINGLE_PRIMARY/PERSISTENCE=DELETE", func(t *testing.T) {
-
+		if *deviations.GRIBIPreserveOnly {
```
liulk, line 253: TE-4.2 is an indicator for preserve DELETE so we don't skip.

feature/gribi/ate_tests/leader_failover_test/leader_failover_test.go
```diff
@@ -289,6 +291,9 @@ func TestLeaderFailover(t *testing.T) {
 	})
 
 	t.Run("ShouldDelete", func(t *testing.T) {
+		if *deviations.GRIBIPreserveOnly {
```
liulk, line 294: Don't skip.

liulk (commented):
I noticed a few other places where I used the wrong style (should have used t.Skip, not t.Skipf) and used the wrong flag name (should be --deviation_gribi_preserve_only, not --deviations). I'll follow up with a PR with these changes and incorporate your change here.

mojiiba:
@liulk thank you so much! closing this since it is no longer needed.

PR #602, opened 2022-10-03

sezhang2:
- Added switchoverReady check before switchover
- Deleted the redundant check on interface status

Test passed on hardware platform.

### DIFF
--- a/feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go
+++ b/feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go
@@ -19,7 +19,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/openconfig/featureprofiles/internal/components"
 	"github.com/openconfig/featureprofiles/internal/fptest"
 	"github.com/openconfig/ondatra"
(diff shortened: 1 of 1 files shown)


feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go
```diff
@@ -81,6 +80,13 @@ func TestSupervisorSwitchover(t *testing.T) {
 	rpStandbyBeforeSwitch, rpActiveBeforeSwitch := findStandbyRP(t, dut, supervisors)
 	t.Logf("Detected rpStandby: %v, rpActive: %v", rpStandbyBeforeSwitch, rpActiveBeforeSwitch)
 
+	switchoverReady := dut.Telemetry().Component(rpActiveBeforeSwitch).SwitchoverReady()
+	switchoverReady.Await(t, 30*time.Minute, true)
```
amrindrr, line 84: I am wondering if 30 mins is too much. As it may timeout the test or Guitar workflow when run in production.

amrindrr (approved):
LGTM

feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go
sezhang2: 30 minutes is the current default reservation timeout. The test run timeout is longer.

feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go
```diff
@@ -81,6 +80,13 @@ func TestSupervisorSwitchover(t *testing.T) {
...
+	t.Logf("SwitchoverReady().Get(t): %v", switchoverReady.Get(t))
+	if !switchoverReady.Get(t) {
+		t.Errorf("switchoverReady.Get(t): got %v, want %v", switchoverReady.Get(t), true)
```
wenovus, line 87: Ideally I'd like to see the following to avoid duplicating constant values and typos in the future:
```go
if got, want := switchoverReady.Get(t), true; got != want {
...., got, want)
}
```
But the rest of the test doesn't seem to be following this style. So I'm ok with keeping with the prevailing style here.

@liulk What do you think about this style? I think the got/want makes it clearer and less typo-prone.

wenovus (approved):

feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go
sezhang2: Agreed. I will make some global changes to some of my tests in the next PR.

PR #603, opened 2022-10-03
Static routing

nikiitp18:
added ATE config but on running it is showing error

static_route_test.go:109: 
        *** Pushing config to ate...
        
        
    static_route_test.go:109: PushConfig(t) on ATE{Dims:Dims{Name:ate Vendor:IXIA HardwareModel:IXIA_TG SoftwareVersion:IXIA_TG Ports:map[port1:Port{Name:eth4 Speed:SPEED_UNSPECIFIED CardModel: PMD:PMD_UNSPECIFIED} port2:Port{Name:eth1 Speed:SPEED_UNSPECIFIED CardModel: PMD:PMD_UNSPECIFIED} port3:Port{Name:eth2 Speed:SPEED_UNSPECIFIED CardModel: PMD:PMD_UNSPECIFIED} port4:Port{Name:eth3 Speed:SPEED_UNSPECIFIED CardModel: PMD:PMD_UNSPECIFIED}]}}: errors:
        - Error occurred while setting Protocol config for user common:could not found remapped
          location name for eth4 from map[eth1:service-ate-port-eth1.ixia-c.svc.cluster.local:5555+service-ate-port-eth1.ixia-c.svc.cluster.local:50071
          eth2:service-ate-port-eth2.ixia-c.svc.cluster.local:5555+service-ate-port-eth2.ixia-c.svc.cluster.local:50071]
--- FAIL: TestGNMI (6.35s)
FAIL


feature/staticroute/otg_tests/static_route_test.go
```diff
@@ -0,0 +1,112 @@
...
+	// dp4 := dut.Port(t, "port4")
+
+	ifCfg1 := &telemetry.Interface{
+		Name:        ygot.String(dp1.Name()),
+		Description: ygot.String("From Ixia"),
+	}
+
+	ifCfg1.GetOrCreateSubinterface(0).
+		GetOrCreateIpv4().
+		GetOrCreateAddress("192.0.2.10").
```
Ankur19, line 33: I don't think you are allowed to have `last bit == 0` in the interface IP addresses. So maybe not use `192.0.2.10`?

feature/staticroute/otg_tests/static_route_test.go
```diff
@@ -0,0 +1,112 @@
...
+		PrefixLength = ygot.Uint8(31)
+
+	ifCfg2 := &telemetry.Interface{
```
Ankur19, line 36: Additionally, check out `deviations.InterfaceEnabled`. I think `Enabled` value may be required in `telemetry.Interface` and its subinterfaces.

robshakir:
The test failure here looks like it might be related to not having a 4 port ATE specified in the KNE topology. Which of the KNE topologies with OTG are you using?

feature/staticroute/otg_tests/static_route_test.go
robshakir: This is fine - the last bit being set to zero is no problem :-) The usable addresses are specified based on the subnet mask that is configured - which is sometimes why specific addresses (network and broadcast addresses) are not usable.

nikiitp18:
> The test failure here looks like it might be related to not having a 4 port ATE specified in the KNE topology. Which of the KNE topologies with OTG are you using?

using the same arista_ceos_otg.textproto
actually when i tried the same thing with 2 ports and generate the flow and then it is working fin. i also printed in and out packets as well

bstoll:
Is this related to a copy of an ATE test ported to OTG?  I could not find a README or explanation for the purpose of this.  Is this intended to be a new test addition?

I created #623 because I found this directory confusing too.

robshakir:
It's a new test that is being worked on outside of the current test plan on static routing. Apologies, this PR should be draft.

feature/staticroute/TwoPortConfig/static_route_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/openconfig/ondatra"
+	"github.com/openconfig/ondatra/telemetry"
+	"github.com/openconfig/ygot/ygot"
+)
+
+func TestMain(m *testing.M) {
+	fptest.RunTests(m)
+}
+
+func TestGNMI(t *testing.T) {
```
robshakir, line 17: We might want to this call this something like `TestStaticRoute`.

feature/staticroute/TwoPortConfig/static_route_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Configure a DUT
+
+	dut := ondatra.DUT(t, "dut")
+	dp1 := dut.Port(t, "port1")
+	dp2 := dut.Port(t, "port2")
+
+	ifCfg1 := &telemetry.Interface{
```
robshakir, line 25: We might want to say something like:

```golang

config := map[string]string{
  "port1": "192.0.2.10",
  "port2": "192.0.2.20",
}

for port, ipaddr := range config {
   // Do the port configuration here.
   ifCfg := ...GetOrCreateInterface(port)....GetOrCreateAddress(ipaddr)
   dut.Config().Interface(...).Update(t, ifcfg)
}
```

robshakir: Or actually... should we look at [`Attributes`](https://github.com/openconfig/featureprofiles/blob/71497a3b4c5d5806e8425f5b7ee748231d5a53ec/internal/attrs/attrs.go#L35)

```golang

config := []*Attributes{{
  IPv4: "192.0.2.20",
  Name: "port1",
}, {
  IPv4: "192.0.2.30",
  Name: "port2",
}}

for _, value := range config {
   ... same configuration with value.Name and value.IPv4 ...
   ... can include the IPv4 prefix length ...
}
```

feature/staticroute/TwoPortConfig/static_route_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	//  Configure an ATE
+
+	ate := ondatra.ATE(t, "ate")
+	ap1 := ate.Port(t, "port1")
+	ap2 := ate.Port(t, "port2")
+
+	top := ate.OTG().NewConfig(t)
+
+	top.Ports().Add().SetName(ap1.ID())
```
robshakir, line 64: Again, we might want to refactor this in loop that does both ports. And like above, we could look at [`Attributes`](https://github.com/openconfig/featureprofiles/blob/71497a3b4c5d5806e8425f5b7ee748231d5a53ec/internal/attrs/attrs.go#L35).

feature/staticroute/TwoPortConfig/static_route_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	flow := top.Flows().Add().SetName("Flow")
+	flow.TxRx().Device().SetTxNames([]string{"src.IPv4"}).SetRxNames([]string{"dst1.IPv4"})
+	
+	// enabling flow metrics
+	flow.Metrics().Msg().Enable = ygot.Bool(true)
+	
+	endpoint := flow.Packet().Add().Ipv4()
+	endpoint.Src().SetValue("192.0.2.11")
+	endpoint.Dst().SetValue("192.0.2.21")
```
robshakir, line 85: We need to change the destination address to an IP address in our static route (`10.0.0.0/24`).

feature/staticroute/TwoPortConfig/static_route_test.go
```diff
@@ -0,0 +1,101 @@
...
+		PrefixLength = ygot.Uint8(31)
+
+	dut.Config().Interface(dp1.Name()).Update(t, ifCfg1)
+	dut.Config().Interface(dp2.Name()).Update(t, ifCfg2)
+
+	Ni1 := &telemetry.NetworkInstance{}
+
+	Ni1.GetOrCreateProtocol(telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, "static").
+		GetOrCreateStatic("10.0.0.0/24").
+		GetOrCreateNextHop("h1").NextHop = telemetry.UnionString("192.0.2.10")
```
robshakir, line 52: Our flow goes to ATE port 1 on line 78 - but we might want to check whether this can be forwarded to ATE port 2.

One way that we might make this clearer is we said:

```golang

atePort2 := &Attributes{
   IPv4: "192.0.2.xx",
}

...NextHop = telemetry.UnionString(atePort2.IPv4)
```

robshakir: But now we lost our looping...

```golang
atePort1 := &Attributes{ ... }
atePort2 := &Attributes{ ... }

atePorts := []*Attributes{atePort1, atePort2}
for _, port := range atePorts { ... }
```

feature/staticroute/TwoPortConfig/static_route_test.go
```diff
@@ -0,0 +1,101 @@
...
+	top.Ports().Add().SetName(ap2.ID())
+	i2 := top.Devices().Add().SetName(ap2.ID())
+	eth2 := i2.Ethernets().Add().SetName("dst1.Eth").SetPortName(i2.Name()).SetMac("02:1a:c0:00:02:02")
+	eth2.Ipv4Addresses().Add().SetName("dst1.IPv4").SetAddress("192.0.2.21").SetGateway("192.0.2.20").SetPrefix(int32(31))
+
+	ate.OTG().PushConfig(t, top)
+	ate.OTG().StartProtocols(t)
+
+	flow := top.Flows().Add().SetName("Flow")
+	flow.TxRx().Device().SetTxNames([]string{"src.IPv4"}).SetRxNames([]string{"dst1.IPv4"})
```
robshakir, line 78: Here we need to make sure that our destination address will be routed to the destination port we specify -- i.e., the `NextHop` we specified above.

dplore:
FYI: I formally assigned @robshakir to help ensure we know who should to review.  @Ankur19 and @bstoll you can let Rob handle this one, thanks!

robshakir:
Thanks Darren. This is where Nikhil and I are iterating on his work on test development, ACK on the reviews/reviewing approach.

feature/staticroute/otg_tests/static_route_test.go
```diff
@@ -66,41 +90,12 @@ func TestGNMI(t *testing.T) {
 	}
 
 	Ni1 := &telemetry.NetworkInstance{}
-
 	Ni1.GetOrCreateProtocol(telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, "static").
 		GetOrCreateStatic("10.0.0.0/24").
 		GetOrCreateNextHop("h1").NextHop = telemetry.UnionString("192.0.2.23")
```
robshakir, line 95: If you adopt the approach to use `map[string]attrs.Attributes` from above then this could be made a bit clearer to the test reader, since you could have `...NextHop = telemetry.UnionString(atePorts["dst3"].IPv4)`.

Making this kind of change helps the person that is examining the test later not need to keep track of the IPs that are assigned themselves.

feature/staticroute/otg_tests/static_route_test.go
```diff
@@ -116,17 +111,16 @@ func TestGNMI(t *testing.T) {
 	ate.OTG().PushConfig(t,top)
 	ate.OTG().StartProtocols(t)	
 
-
 	flow := top.Flows().Add().SetName("Flow")
 	flow.TxRx().Device().SetTxNames([]string{"src.IPv4"}).SetRxNames([]string{"dst1.IPv4"})
 	flow.Metrics().Msg().Enable = ygot.Bool(true)
 	endpoint := flow.Packet().Add().Ipv4()
-	endpoint.Src().SetValue("192.0.2.13")
+	endpoint.Src().SetValue("192.0.2.23")
```
robshakir, line 118: Same point as above here, if you use the map, then this can become more intuitive to the user -- e.g., `atePorts["src"].IPv4`.

robshakir (commented):
Looks good -- some more comments to iterate and get this code to the most 'readable' state.

Note, here we're iterating to make sure that we write code that is easy to maintain. We're going to write this once, but probably a lot more people will read it (if the test fails, or they want to understand what the test does), so we want to optimise for those readers.

feature/staticroute/otg_tests/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Validate static routing for a DUT.
+
+## Procedure
+
+*   Configure a DUT that has 4 interfaces to an ATE device configured with IPv4 and IPv6 addresses at both ends.
+*   Configure ATE port-1 connected to DUT port-1, and ATE port 2-4 connected to DUT port 2-4 with the relevant IPv4 addresses.
+*   Using the OpenConfig static routing configuration validate simple static routes for IPv4:
+    *   Add a static route for an IPv4 prefix 10.0.0.0/24 that routes packets from ATE port 1 to the next-hop address on each ATE port 2-4:
+        *   Validate that packets are received at each ATE port 2-4 when sourced from ATE port 1.
```
robshakir, line 13:
```suggestion
    *   Validate that packets are received at each ATE port 2-4 when sourced from ATE port 1.
```

feature/staticroute/otg_tests/README.md
```diff
@@ -0,0 +1,40 @@
...
+*   /interfaces/interface/config/enabled
+*   /interfaces/interface/subinterfaces/subinterface/ipv4/addresses/address/config/prefix-length
+*   /network-instances/network-instance/protocols/protocol/config/name
+*   /network-instances/network-instance/protocols/protocol/static-routes/static/prefix
+*   /network-instances/network-instance/protocols/protocol/static-routes/static/next-hops/next-hop/config/next-hop
+
+
+## Telemetry Parameter Coverage
+
+*   /interfaces/interface/state/counters/out-octets
```
robshakir, line 28: Since we don't read these from the DUT (just the ATE) then I don't think that we cover these.

robshakir (changes requested):
This looks great -- some minor changes and then I think we're ready to make sure that one of the fntest team reviews the code!

robshakir:
@nikiitp18 -- we should also look at this error: https://github.com/openconfig/featureprofiles/actions/runs/3263324865/jobs/5361974716

The fntest team would like us to use some documentation range prefixes rather than `10.0.0.0/8`. :-)

feature/staticroute/otg_tests/ipv4_static_routing_test/ipv4_static_routing_test.go
```diff
@@ -0,0 +1,192 @@
...
+		ifCfg := &telemetry.Interface{
+			Name:        ygot.String(pn),
+			Description: ygot.String(attributes.Desc),
+		}
+		ifCfg.GetOrCreateSubinterface(0).GetOrCreateIpv4().Enabled = ygot.Bool(true)
+		ifCfg.GetOrCreateSubinterface(0).
+			GetOrCreateIpv4().
+			GetOrCreateAddress(attributes.IPv4).PrefixLength = ygot.Uint8(attributes.IPv4Len)
+
+		dut.Config().Interface(pn).Update(t, ifCfg)
```
wenovus, line 113: Please use the new ygnmi library instead of the old `dut.Config()`:
```go
gnmi.Replace(t, dut, gnmi.OC().Interface(pn).Config(), attributes.NewOCInterface(pn))
```

feature/staticroute/otg_tests/ipv4_static_routing_test/ipv4_static_routing_test.go
```diff
@@ -0,0 +1,192 @@
...
+	for dstport := range atePorts {
+		if dstport == "port1" {
+			continue
+		}
+
+		ni := &telemetry.NetworkInstance{}
+		ni.GetOrCreateProtocol(telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, "static").
+			GetOrCreateStatic("192.0.2.0/24").
+			GetOrCreateNextHop("h").NextHop = telemetry.UnionString(atePorts[dstport].IPv4)
+		dut.Config().NetworkInstance("default").Update(t, ni)
```
wenovus, line 150: Please use the new ygnmi library (see feature/gribi/otg_tests/ipv4_entry_test/ipv4_entry_test.go for examples):
```
gnmi.Update(...)
```

PR #605, opened 2022-10-03
AddToOTG function to attr package

octpetre:
Adding AddToOTG function to internal/attrs/attrs.go
This is very useful for quick OTG configurations

### DIFF
--- a/internal/attrs/attrs.go
+++ b/internal/attrs/attrs.go
@@ -22,6 +22,7 @@ package attrs
 import (
 	"fmt"
 
+	"github.com/open-traffic-generator/snappi/gosnappi"
 	"github.com/openconfig/featureprofiles/internal/deviations"
 	"github.com/openconfig/ondatra"
 	oc "github.com/openconfig/ondatra/telemetry"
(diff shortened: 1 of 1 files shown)


internal/attrs/attrs.go
```diff
@@ -126,3 +127,23 @@ func (a *Attributes) AddToATE(top *ondatra.ATETopology, ap *ondatra.Port, peer *
 	}
 	return i
 }
+
+// AddToOTG adds basic elements to a gosnappi configuration
+func (a *Attributes) AddToOTG(top gosnappi.Config, ap *ondatra.Port, peer *Attributes) {
+	top.Ports().Add().SetName(ap.ID())
+	dev := top.Devices().Add().SetName(a.Name)
```
liulk, line 134: A single attributes struct only covers a single interface, so each interface will individually call AddToOTG, and this looks like each interface will add its own device?

octpetre: Yes, the test will need to call the AddToOTG for each port/interface to add basic IPv4/IPv6 config provided inthe *Attributes parameters.

octpetre: The "device" is an element of the gosnappi configuration and is not the ondatra "OTG device". Here is a sample code on how this function would be used 
```// configureATE configures port1-3 on the ATE.
func configureATE(t *testing.T, ate *ondatra.ATEDevice) gosnappi.Config {
    top := ate.OTG().NewConfig(t)

    p1 := ate.Port(t, "port1")
    p2 := ate.Port(t, "port2")
    p3 := ate.Port(t, "port3")

    atePort1.AddToOTG(top, p1, &dutPort1)
    atePort2.AddToOTG(top, p2, &dutPort2)
    atePort3.AddToOTG(top, p3, &dutPort3)

    return top
}
```

liulk (approved):

Issue #424, opened 2022-08-16
osinstall_test.go script need to update tc.dualSup checks "For Junos Evolved image transfer and activate standby are not needed""


avaneesh90:
Hi ,

Need to update the script for vender specific:

Below code changes needed at https://github.com/openconfig/featureprofiles/blob/main/feature/gnoi/os/tests/osinstall/osinstall_test.go#L91:

if tc.dualSup {
		switch dut.Vendor() {
		case ondatra.JUNIPER:
			fmt.Println("For Junos Evolved image transfer and activate standby are not needed")
		case ondatra.CISCO:
			tc.transferOS(ctx, t, true)
			tc.activateOS(ctx, t, true)
		case ondatra.ARISTA:
			tc.transferOS(ctx, t, true)
			tc.activateOS(ctx, t, true)
		}
	}

bstoll:
We generally do not want to introduce changes like this as the goal of the tests is to not have any mentions of any specific vendors.  The process for software upgrades is documented in the [GNOI protobuf](https://github.com/openconfig/gnoi/blob/master/os/os.proto) and requires supporting these steps in order to upgrade a standby routing engine.  I'm not familiar with the background on why transfer/activate are not needed. Is this a temporary bug workaround?

aaronmillisor:
I can't speak for the Juniper case, but in IOS XR there is no concept of manually transferring images between RP's. You can transfer _files_ manually, but the OS management has its own special handling. The state is automatically synchronized between RP's in the system as part of the install process for the software. Some other platforms require you to manually transfer/sync images between controller cards as files but this is not universal, even though the gNOI protobuf assumes that it is.

dplore:
Thanks for reporting the issue.    To be compliant to the [gNOI specification for OS installs](https://github.com/openconfig/gnoi/blob/6bddacd72ef8aa06304b887df414a43c8e63ff27/os/os.proto#L67-L109)  and [osinstall_test](https://github.com/openconfig/featureprofiles/tree/main/feature/gnoi/os/tests/osinstall#readme) the device must be able to have an OS install be completed as specified without vendor specific deviations.

xw-g:
@aaronmillisor and @avaneesh90, I think there might be some confusion here. (maybe because the name of the wrapper function?). 

My opinions are: 
* It's good that some vendors/OS take care of the image sync. 
[...]

PR #622, opened 2022-10-06
Adding changes for https://github.com/openconfig/featureprofiles/issu…

avaneesh90:
…es/424

### DIFF
--- a/feature/gnoi/os/tests/osinstall/osinstall_test.go
+++ b/feature/gnoi/os/tests/osinstall/osinstall_test.go
@@ -73,6 +73,7 @@ func TestOSInstall(t *testing.T) {
 
 	ctx := context.Background()
 	dut := ondatra.DUT(t, "dut")
+	dp := dut.Port(t, "port1")
 
 	reader, err := packageReader(ctx)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -88,9 +89,14 @@ func TestOSInstall(t *testing.T) {
 	tc.fetchStandbySupervisorStatus(ctx, t)
 	tc.transferOS(ctx, t, false)
 	tc.activateOS(ctx, t, false)
-	if tc.dualSup {
-		tc.transferOS(ctx, t, true)
-		tc.activateOS(ctx, t, true)
+	switch dp.Device().Vendor() {
+	case ondatra.ARISTA, ondatra.CISCO:
```
mojiiba, line 93: For Cisco we also do not need transfer and activate for standby rp, please move the cisco to the next case.

bstoll:
We don't want to add specific vendor flags in tests like this, so this is not something we can do.  It also wouldn't line up with the GNOI spec for how dual supervisor upgrades are handled.  If this was understood as a bug or unintended temporary state we could have a deviation.

It sounds like either the device behavior needs to change to support the GNOI spec, or you would like to make a change to the GNOI OS upgrade spec.

dplore:
I've added @xw-g for this review who will help find a solution.

xw-g:
> I've added @xw-g for this review who will help find a solution.

https://github.com/openconfig/featureprofiles/issues/424#issuecomment-1273924798

xw-g:
> …es/424

Please help keep the title in a clean format thanks :) I just changed your PR title.

xw-g:
Let's continue the discussion first in https://github.com/openconfig/featureprofiles/issues/424#issuecomment-1273924798

xw-g:
duplicate of https://github.com/openconfig/featureprofiles/pull/675?

avaneesh90:
Refering to https://github.com/openconfig/featureprofiles/pull/675

PR #624, opened 2022-10-07
added network instance type to default

cprabha:
Hi,

Instance type config is required when DUT is assumed to have no network instance configured prior to running the test.
If instance type is not configured prior to script run, config failure will be seen as instance type is mandatory.

If DUT already have instance type configured, it also will be of type DEFAULT_INSTANCE.
Script also does replace with same type, hence replacing instance type is not a problem such DUT devices.

Ref PR #: https://github.com/openconfig/featureprofiles/pull/592

Thanks,
Prabha

### DIFF
--- a/feature/bgp/policybase/ate_tests/route_installation_test/route_installation_test.go
+++ b/feature/bgp/policybase/ate_tests/route_installation_test/route_installation_test.go
@@ -467,7 +467,6 @@ func TestEstablish(t *testing.T) {
 	dutConfNIPath := dut.Config().NetworkInstance(*deviations.DefaultNetworkInstance)
 	dutConfNIPath.Type().Replace(t, telemetry.NetworkInstanceTypes_NETWORK_INSTANCE_TYPE_DEFAULT_INSTANCE)
 
-
 	// Configure BGP+Neighbors on the DUT
 	t.Logf("Start DUT BGP Config")
 	dutConfPath := dut.Config().NetworkInstance(*deviations.DefaultNetworkInstance).Protocol(telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP").Bgp()
@@ -522,7 +521,6 @@ func TestBGPPolicy(t *testing.T) {
 	t.Log("Configure Network Instance type ")
 	dutConfNIPath := dut.Config().NetworkInstance(*deviations.DefaultNetworkInstance)
 	dutConfNIPath.Type().Replace(t, telemetry.NetworkInstanceTypes_NETWORK_INSTANCE_TYPE_DEFAULT_INSTANCE)
-
 
 	cases := []struct {
 		desc                      string
--- a/feature/bgp/policybase/ate_tests/route_installation_test/route_installation_test.go
+++ b/feature/bgp/policybase/ate_tests/route_installation_test/route_installation_test.go
@@ -462,6 +462,12 @@ func TestEstablish(t *testing.T) {
 	t.Logf("Start DUT interface Config")
 	configureDUT(t, dut)
 
+	// Configure Network instance type on DUT
+	t.Log("Configure Network Instance type")
+	dutConfNIPath := dut.Config().NetworkInstance(*deviations.DefaultNetworkInstance)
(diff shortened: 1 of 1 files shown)


bstoll:
I believe we are expecting the default network instance to have the type set correctly so this shouldn't be necessary.  If this is required, this sounds like an implementation bug.  In that case, I think we could add it wrapped with a deviation flag.

I don't think it is reasonable to require every test to set this value.

dplore:
OC currently does not define a default for a network instance type.  Therefore I believe we should explicitly set the network-instance type.  Otherwise (as pointed out earlier) it is ambiguous what type will be used as the NOS could use any type it wants and still be compliant.  I do believe the model is weak on this point and have filed an issue at https://github.com/openconfig/public/issues/726.

Reference:
https://github.com/openconfig/public/blob/f3f05d5f6e38ca7c46671ea5a4c11b62d264c10b/release/models/network-instance/openconfig-network-instance.yang#L1193-L1208

cprabha:
@bstoll , Please could you review and provide approval for this pull request.

Thanks,
Prabha

bstoll:
I generally disagree with this approach.  We may be able to define this as a deviation and then mark the use cases of it with a deviation flag, but I don't think it is good that every time an operator wants to work on the "default" network instance they have to update(or verify) a leaf value.

I can see that the "type" could be marked as required in the OC yang, but regardless of that outcome the device doesn't seem like it is behaving correctly here.  Either the "default" network-instance exists and the required "type" value is set on it, or the "default" network-instance does not exist and we need to create it (with a type value).  How did the device end up in the state where the "default" network-instance exists, but is does not have a type value set?  I don't think this is a case of defining a default value of "type" for all network-instances.  Didn't some configuration (likely not even OC) result in the creation of a "default" network instance?

dplore:
I can see two possible resolutions:

1.  Explicitly create the default network instance and set it's type in this test.  Update OpenConfig model to make the type a required value.  

2. Require that the default network instance be used by default when no network instance is specified.  Update OpenConfig model to define a default type.

I think option 1 is the most clear and should not require a deviation.  Maybe there is another better option?

bstoll:
After thinking about this some more I am coming around to being OK with the original ask.  I was approaching it from the perspective that the default network instance must exist as a result of interface membership, and therefor the network-instance configuration must become populated by something valid.  I still think this is this confusing/bad behavior.

Being explicit in network-instance configuration does sound reasonable too.

Does this mean we should also be explicitly replacing the "enabled" and the "enabled-address-families" leafs to IPV4/IPV6 as well?

earies:
> I generally disagree with this approach. We may be able to define this as a deviation and then mark the use cases of it with a deviation flag, but I don't think it is good that every time an operator wants to work on the "default" network instance they have to update(or verify) a leaf value.

They wouldn't - what is seen is just a result of individual test executions assuming a prior environment exists.

> Does this mean we should also be explicitly replacing the "enabled" and the "enabled-address-families" leafs to IPV4/IPV6 as well?

My opinion is that `enabled-address-families` be removed from the model-set.  See https://github.com/openconfig/public/pull/738

At the same time, I will propose removal of the `enabled`  node under `/network-instances/network-instance/config`.  I would like to see a pointer to > 0 implementations that can support this not to mention not being able to do this on the DEFAULT_INSTANCE whatsoever and what that would mean.

I agree w/ @dplore option 1 above and provided comment on https://github.com/openconfig/public/issues/726 as well.

We don't need a default value here as that is subjective.  There can only be a single `DEFAULT_INSTANCE` so it would need to be another type.  An operator should be explicit in setting this value - just that the value must be set since a dangling instance without a type is not viable.

bstoll:
My opinion of `enabled-address-families` was similar to yours: It's not clear in the documentation what impact the knob is supposed to have, and it is obviously a big deal.

I think there are some practical use case for "disable", but it's certainly not common or something we are doing in any tests here.

We are discussing bringing some of this base interface/network instance configuration into a helper, and I hope we can pull this change out (and all the other instances of it) so every test does not need to repeat this configuration.  In the meantime, I'll stop being a thorn on this issue.

Thanks for the comments/thoughts.

/gcbrun

bstoll (approved):

PR #638, opened 2022-10-12
Update IS-IS tests to use ygnmi and check.

dan-lepage:
This removes the hacky assertion helper that the IS-IS tests used previously and updates them to use the new `check` library, as well as the new OC package generated with `ygnmi`. Both of these are tools we'd like to standardize on going forwards.

This PR also fixes several issues with the IS-IS `base_adjacencies_test`:
* The `spf-runs` counter is no longer hard-coded to have a specific value (this was unreliable, as the device may choose to run SPF additional times), and instead we simply test that after an adjacency forms the counter has gone up.
* We no longer expect the number of hellos sent to be 0 before adjacencies form (the device will send these once IS-IS is configured, even before the adjacency has actually be established).
* When testing MD5 authentication the ATE is now properly configured to use MD5 encryption.
* All the tests now allow a greater delay between when a configuration value is set and when the device's state reflects this change (a device now can take up to 5s for IS-IS to show up as enabled after configuration is pushed, instead of checking it microseconds later).
* The new `--deviation_allow_nil_for_defaults` is only used by the IS-IS tests at present, but when set certain checks that check that specific nodes are equal to their default values will also pass if the device returns no value at all (this is NOT compliant OpenConfig behavior - if an unset node has a default value then fetching the operational state should return that default value, not nil - but some devices do not yet follow this guideline).


feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
```diff
@@ -15,228 +15,313 @@
...
 	t.Run("read_config", func(t *testing.T) {
-		assert.Value(t, isisRoot.Global().Net(), []string{"49.0001.1920.0000.2001.00"})
-		assert.Value(t, isisRoot.Global().LevelCapability(), telemetry.IsisTypes_LevelType_LEVEL_1_2)
-		assert.Value(t, isisRoot.Global().Af(telemetry.IsisTypes_AFI_TYPE_IPV4, telemetry.IsisTypes_SAFI_TYPE_UNICAST).Enabled(), true)
-		assert.Value(t, isisRoot.Global().Af(telemetry.IsisTypes_AFI_TYPE_IPV6, telemetry.IsisTypes_SAFI_TYPE_UNICAST).Enabled(), true)
-		assert.Value(t, isisRoot.Level(2).Enabled(), true)
-		assert.Value(t, isisRoot.Interface(ts.DUT.Port(t, "port1").Name()).Enabled(), true)
-		assert.Value(t, isisRoot.Interface(ts.DUT.Port(t, "port1").Name()).CircuitType(), telemetry.IsisTypes_CircuitType_POINT_TO_POINT)
+		for _, vd := range []check.Validator{
+			check.Equal(isisRoot.Global().Net().State(), []string{"49.0001.1920.0000.2001.00"}),
```
sezhang2, line 71: Define hardcoded value as constant.

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
```diff
@@ -15,228 +15,313 @@
...
+			EqualToDefault(isisRoot.Global().LevelCapability().State(), oc.Isis_LevelType_LEVEL_1_2),
+			check.Equal(isisRoot.Global().Af(oc.IsisTypes_AFI_TYPE_IPV4, oc.IsisTypes_SAFI_TYPE_UNICAST).Enabled().State(), true),
+			check.Equal(isisRoot.Global().Af(oc.IsisTypes_AFI_TYPE_IPV6, oc.IsisTypes_SAFI_TYPE_UNICAST).Enabled().State(), true),
+			check.Equal(isisRoot.Level(2).Enabled().State(), true),
+			check.Equal(port1ISIS.Enabled().State(), true),
+			check.Equal(port1ISIS.CircuitType().State(), oc.Isis_CircuitType_POINT_TO_POINT),
+		} {
+			t.Run(vd.RelPath(isisRoot), func(t *testing.T) {
+				if err := vd.AwaitUntil(deadline, ts.DUTClient); err != nil {
+					t.Error(err)
```
sezhang2, line 81: Added vd.AwaitUntil(deadline, ts.DUTClient) to the error msg.

sezhang2 (dismissed):

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
dan-lepage: I don't think it belongs in the error message - these checks are all designed to produce readable error messages like

```
/network-instances/network-instance[name=default]/protocols/protocol[identifier=ISIS][name=DEFAULT]/isis/level/state/enabled: got no value, want true
```

From the point of view of anyone reading the test output, this is the information they want, not the fact that it was part of an `AwaitUntil` call specifically. (I know Go testing style guidelines recommend putting the function being called in the error message, but that only makes sense if the function being called is the thing you're testing - we aren't testing `AwaitUntil` here, we're testing that this particular path has the correct value).

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
dan-lepage: Fixed.

sezhang2 (approved):

internal/deviations/deviations.go
```diff
@@ -67,6 +67,8 @@ import "flag"
 
 // Vendor deviation flags.
 var (
+	MixedOcCliPrefersCli = flag.Bool("deviation_mixed_occli_prefers_cli", false,
```
liulk, line 70: Why is this deviation added in this PR? The name also stutters. "MixedOriginPrefersCLI" would have been more accurate.

internal/deviations/deviations.go
```diff
@@ -95,5 +97,8 @@ var (
 
 	NextHopAFTNotSupported = flag.Bool("deviation_nexthop_aft_not_supported", false, "Device currently doesnot support AFT Next Hop Telemetry. A fully compliant device should support all types of AFT telemetry without this deviation.")
 
+	AllowNilForDefaults = flag.Bool("deviation_allow_nil_for_defaults", false,
```
liulk, line 100: I think "allow" is not really accurate in the name because it's not the device allowing nil, but rather the test is allowing it.

MissingValueForDefaults is probably a better name. We want the deviation to describe the device behavior, not the test behavior.

internal/deviations/deviations.go
dan-lepage: Whoops, I botched that during a rebase. I'll make a PR to remove it.

PR #640, opened 2022-10-12
RT-6.1: core_lldp_tlv_population_test.go

yashikakn:
https://github.com/openconfig/featureprofiles/blob/main/feature/lldp/tests/core_lldp_tlv_population_test/core_lldp_tlv_population_test.go#L117

In above code, gnmi get-config data is compared with telemetry state data. In case where lldp is disabled, the config will still list interface configured , but the corresponding state data will not be present. Hence, added an if condition to skip the check for lldp disabled case. Please review.

Thanks.


dplore (approved):
LGTM.  By [definition of the test](https://github.com/openconfig/featureprofiles/pull/640/files#diff-23f527b315d59fe05d447a44cad7c8c6e9e7921bb495e34615d420d2545461a2R95-R96), the state being compared here is LLDP state.  If the test has disabled LLDP, the LLDP state will not be populated, but the test still attempts to compare.  

The change only attempts the LLDP state to config if LLDP is enabled.  This makes sense.

dplore:
@alexmasi for go style review

feature/lldp/tests/core_lldp_tlv_population_test/core_lldp_tlv_population_test.go
```diff
@@ -114,10 +114,13 @@ func verifyNodeConfig(t *testing.T, nodeTelemetry *device.DevicePath, port *onda
 	} else {
 		t.Errorf("LLDP SystemName is not proper, got %s", state.GetSystemName())
 	}
-	if conf.GetInterface(port.Name()).GetName() != state.GetInterface(port.Name()).GetName() {
-		t.Errorf("LLDP interfaces/interface/state/name got: %s, want: %s.", state.GetInterface(port.Name()).GetName(),
-			conf.GetInterface(port.Name()).GetName())
+	if lldpEnabled {
```
alexmasi, line 117: this nested conditional can be combined to reduce indentation: https://google.github.io/styleguide/go/decisions#conditionals-and-loops

it appears short circuiting is not required here so you can do something like the following:

```
stateMatch := conf.GetInterface(port.Name()).GetName() == state.GetInterface(port.Name()).GetName()
if lldpEnabled && !stateMatch {
  t.Errorf()
}
```

an alternative that may reduce line length even further is:
```
got := state.GetInterface(port.Name()).GetName()
want := conf.GetInterface(port.Name()).GetName()
if lldpEnabled && got != want {
  t.Errorf("LLDP interfaces/interface/state/name = %s, want %s", got, want)
}
```

alexmasi (changes requested):
oops I forgot to "request changes" for my comment

feature/lldp/tests/core_lldp_tlv_population_test/core_lldp_tlv_population_test.go
yashikakn: Hi @alexmasi , Thanks for reviewing. Made changes accordingly. Please take a look.

yashikakn:
> LGTM. By [definition of the test](https://github.com/openconfig/featureprofiles/pull/640/files#diff-23f527b315d59fe05d447a44cad7c8c6e9e7921bb495e34615d420d2545461a2R95-R96), the state being compared here is LLDP state. If the test has disabled LLDP, the LLDP state will not be populated, but the test still attempts to compare.
> 
> The change only attempts the LLDP state to config if LLDP is enabled. This makes sense.

@dplore : Thanks for reviewing.

alexmasi (approved):

alexmasi:
LGTM for golang

dplore (approved):

PR #641, opened 2022-10-12
Add ipv4/enabled deviation

mojiiba:
This pull adds a deviation for ipv4/enabled leaf. It is expected that ipv4 to be enabled by default for the devices with this deviation.

### DIFF
--- a/feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
+++ b/feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
@@ -341,7 +341,7 @@ func ConfigureDUTIntf(t *testing.T, dut *ondatra.DUTDevice) {
 		// per: https://github.com/openconfig/featureprofiles/issues/253
 		i.Enabled = ygot.Bool(true)
 		s.Enabled = ygot.Bool(true)
-		if !*deviations.IPV4Enabled {
+		if !*deviations.IPv4MissingEnabled {
 			v4.Enabled = ygot.Bool(true)
 		}
 		v6.Enabled = ygot.Bool(true)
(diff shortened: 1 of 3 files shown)


internal/deviations/deviations.go
```diff
@@ -70,6 +70,8 @@ var (
 	InterfaceEnabled = flag.Bool("deviation_interface_enabled", false,
 		"Device requires interface enabled leaf booleans to be explicitly set to true.  Full OpenConfig compliant devices should pass both with and without this deviation.")
 
+	IPV4Enabled = flag.Bool("deviation_ipv4_enabled", false, "Device does not support interface/ipv4/enabled leaf. The ipv4 is enabled when it is configured.")
```
liulk, line 73: Deviation should use false value to indicate compliant behavior, and true value when device is not compliant. See the package doc above.

Therefore, we should do something like this:

```suggestion
	IPv4MissingEnabled = flag.Bool("deviation_ipv4_missing_enabled", false, "Device does not support interface/ipv4/enabled, so suppress configuring this leaf.")
```

feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
```diff
@@ -341,14 +341,17 @@ func ConfigureDUTIntf(t *testing.T, dut *ondatra.DUTDevice) {
...
 
 		dut.Config().Interface(intf.intfName).Replace(t, i)
 
 		t.Logf("Validate that IPv4 and IPv6 addresses are enabled: %s", intf.intfName)
 		subint := dut.Telemetry().Interface(intf.intfName).Subinterface(0)
-		if !subint.Ipv4().Enabled().Get(t) {
+
+		if !*deviations.IPV4Enabled && !subint.Ipv4().Enabled().Get(t) {
 			t.Errorf("Ipv4().Enabled().Get(t) for interface %v: got false, want true", intf.intfName)
 		}
```
liulk, line 356: Probably easier to read:

```suggestion
		if !*deviations.IPv4MissingEnabled {
			if !subint.Ipv4().Enabled().Get(t) {
				t.Errorf("Ipv4().Enabled().Get(t) for interface %v: got false, want true", intf.intfName)
			}
		}
```

liulk (changes requested):

internal/deviations/deviations.go
mojiiba: thank you! addressed

feature/gnmi/ate_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
mojiiba: thanks, addressed

liulk (approved):

PR #642, opened 2022-10-13
Add deviation for interface counter 

mojiiba:
This pull adds a deviation to use gnmi get at the container level for reading interface counters. It also guards the ipv4/enabled with the related deviation.


internal/deviations/deviations.go
```diff
@@ -70,6 +70,8 @@ var (
 	InterfaceEnabled = flag.Bool("deviation_interface_enabled", false,
 		"Device requires interface enabled leaf booleans to be explicitly set to true.  Full OpenConfig compliant devices should pass both with and without this deviation.")
 
+	CounterLeavesNotUpdated = flag.Bool("deviation_counters_leaves_notupdated", false, "Device updates counters when they are queried at the container level.")
```
liulk, line 73:
```suggestion
	InterfaceCountersFromContainer = flag.Bool("deviation_interface_counters_from_container", false, "Device only supports querying interface counters from the state container, not from individual counter leaves.")
```

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -897,7 +906,9 @@ func ConfigureDUTIntf(t *testing.T, dut *ondatra.DUTDevice) {
 		}
 		i.GetOrCreateEthernet()
 		s := i.GetOrCreateSubinterface(0).GetOrCreateIpv4()
-		s.Enabled = ygot.Bool(true)
+		if *deviations.InterfaceEnabled {
+			s.Enabled = ygot.Bool(true)
+		}
```
liulk, line 911: Already done in #641 ?

liulk (changes requested):

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
mojiiba: no, I missed this one

internal/deviations/deviations.go
mojiiba: thank you, addressed

liulk (approved):

PR #647, opened 2022-10-16
RT-1.7 : replaced Replace(t, nil) to Delete(t)

singh-prem:
RT-1.7 Local BGP Test is currently using Replace(t, nil) to delete configuration which is against GNMI specification.
This PR replaces Replace(t, nil) to Delete(t).


bstoll (approved):
reference:

https://github.com/openconfig/reference/blob/c243b35b36e366852f9476c87fb2efe6e9050dfe/rpc/gnmi/gnmi-specification.md?plain=1#L1070-L1073

PR #653, opened 2022-10-18
RT-5.1 - Deviations for DuplexMode ,AutoNegotiate, PortSpeed and Origin

agvarghe-cisco:
Modification done to singleton_test ->
- Deviation for DuplexMode and AutoNegotiate and PortSpeed
- Deviation for subinterfaces/subinterface/neighbors/neighbors/state/origin


internal/deviations/deviations.go
```diff
@@ -67,6 +67,12 @@ import "flag"
 
 // Vendor deviation flags.
 var (
+	PortSpeedDuplexAutoNegotiateEnabled = flag.Bool("port_speed_enabled", false,
```
liulk, line 70: Please explain what is the device behavior that needs to be deviated, and the expected OC behavior so we know when the device becomes OC compliant, then this deviation can be removed.

Look at other deviations to see how they are worded. We cannot accept deviations without justification.

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -235,20 +238,20 @@ func (tc *testCase) verifyInterfaceDUT(
...
 		}
 	}
-
-	disp := dip.Subinterface(0)
-
-	// IPv4 neighbor discovered by ARP.
-	dis4np := disp.Ipv4().Neighbor(atea.IPv4)
-	if got := dis4np.Origin().Get(t); got != dynamic {
-		t.Errorf("%s IPv4 neighbor %s origin got %v, want %v", dp, atea.IPv4, got, dynamic)
+	if *deviations.MissingSubinterfaceOriginLeaf {
```
liulk, line 241: Deviation must be designed so that "false" is OC compliant, and "true" is deviated behavior. This is the opposite and not acceptable.

liulk (changes requested):

internal/deviations/deviations.go
agvarghe-cisco: @liulk liulk , Addressed review commenst

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
agvarghe-cisco: @liulk , Changed it

internal/deviations/deviations.go
```diff
@@ -68,10 +68,10 @@ import "flag"
 // Vendor deviation flags.
 var (
 	PortSpeedDuplexAutoNegotiateEnabled = flag.Bool("port_speed_enabled", false,
```
liulk, line 70:
```suggestion
	MissingEthernetPhy = flag.Bool("missing_ethernet_phy", false,
```

internal/deviations/deviations.go
```diff
@@ -68,10 +68,10 @@ import "flag"
...
-		"Deviation for DuplexMode and AutoNegotiate and PortSpeed")
+		"Device does not support DuplexMode, AutoNegotiate and PortSpeed leafs, so suppress configuring these leafs")
```
liulk, line 71:
```suggestion
		"Device does not support configuring Ethernet PHY settings (duplex-mode, auto-negotiate and port-speed). Fully compliant devices should allow these to be configured.")
```

internal/deviations/deviations.go
```diff
@@ -68,10 +68,10 @@ import "flag"
...
 
-	MissingSubinterfaceOriginLeaf = flag.Bool("subinterface_origin_check", false,
-		"Deviation for subinterfaces/subinterface/neighbors/neighbors/state/origin")
+	SubinterfaceOriginLeaf = flag.Bool("subinterface_origin_check", false,
+		"Device does not support querying subinterfaces/subinterface/neighbors/neighbors/state/origin leaf,  so the test will skip checking them.")
```
liulk, line 74:
```suggestion
	MissingNeighborOrigin = flag.Bool("missing_neighbor_origin", false,
		"Device does not support querying the MAC origin of a subinterface neighbor. Fully compliant device should report whether the MAC entry is static or dynamic.")
```

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -238,7 +238,7 @@ func (tc *testCase) verifyInterfaceDUT(
 			}
 		}
 	}
-	if *deviations.MissingSubinterfaceOriginLeaf {
+	if !*deviations.SubinterfaceOriginLeaf {
```
liulk, line 241: If we are only skipping the origin, we should still get something from the neighbor to ensure that the neighbor exists.

liulk (changes requested):

liulk:
@vosipchu @xw-g for Quattro review.

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -235,20 +238,20 @@ func (tc *testCase) verifyInterfaceDUT(
...
 		}
 	}
-
-	disp := dip.Subinterface(0)
-
-	// IPv4 neighbor discovered by ARP.
-	dis4np := disp.Ipv4().Neighbor(atea.IPv4)
-	if got := dis4np.Origin().Get(t); got != dynamic {
-		t.Errorf("%s IPv4 neighbor %s origin got %v, want %v", dp, atea.IPv4, got, dynamic)
+	if !*deviations.MissingNeighborOrigin {
```
liulk, line 241: If we are just missing the origin, can we still check the neighbor IP?

* [/interfaces/interface/subinterfaces/subinterface/ipv4/neighbors/neighbor/state/ip](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-interfaces.html#interfaces-interface-subinterfaces-subinterface-ipv4-neighbors-neighbor-state-ip)

* [/interfaces/interface/subinterfaces/subinterface/ipv6/neighbors/neighbor/state/ip](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-interfaces.html#interfaces-interface-subinterfaces-subinterface-ipv6-neighbors-neighbor-state-ip)

liulk (changes requested):

internal/deviations/deviations.go
```diff
@@ -67,6 +67,12 @@ import "flag"
 
 // Vendor deviation flags.
 var (
+	MissingEthernetPhy = flag.Bool("missing_ethernet_phy", false,
+		"Device does not support configuring Ethernet PHY settings (duplex-mode, auto-negotiate and port-speed). Fully compliant devices should allow these to be configured.")
+
+	MissingNeighborOrigin = flag.Bool("missing_neighbor_origin", false,
```
guoshiuan, line 73: The flag should be "deviation_missing_neighbor_origin".

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -263,19 +266,20 @@ func (tc *testCase) verifyInterfaceDUT(
 			}
 		}
 	}
+if !*deviations.MissingNeighborOrigin {
```
liulk, line 269: Still not addressed: https://github.com/openconfig/featureprofiles/pull/653#discussion_r1043675090

liulk (changes requested):

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
agvarghe-cisco: Will check this and get back

liulk: Is the entire neighbor subtree missing? Maybe the deviation should be renamed MissingNeighbor to better reflect the missing part?

Also during the investigation, I came to realize that BGP also has a neighbor subtree, so MissingSubinterfaceNeighbor is probably a better name in order to disambiguate.

liulk (changes requested):

agvarghe-cisco:
Closing since duplex and auto negotiate is removed via https://github.com/openconfig/featureprofiles/pull/946

PR #675, opened 2022-10-19
gNOI-4.1: Add OSActivateNoReboot and OSInstallForStandbyRP deviations 

mojiiba:
This pull adds two deviations `OSActivateNoReboot` and  `OSInstallForStandbyRP` to accommodate devices that install images automatically on  standby rp and does not require separate eboot to activate the new os.


feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -87,20 +88,33 @@ func TestOSInstall(t *testing.T) {
 	}
 	tc.fetchStandbySupervisorStatus(ctx, t)
 	tc.transferOS(ctx, t, false)
-	tc.activateOS(ctx, t, false)
-	if tc.dualSup {
+
+	if *deviations.OSActiavteRequiresReboot {
```
xw-g, line 92: This deviation only applies to duaSup case, right? Which means you probably need the if `if tc.dualSup` condition?

xw-g:
+ @vosipchu, @vcheny

feature/gnoi/os/tests/osinstall/osinstall_test.go
mojiiba: no, it applies for both cases since some vendors do not support activating os without doing reboot.

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -87,20 +88,33 @@ func TestOSInstall(t *testing.T) {
...
+		tc.activateOS(ctx, t, false /*standby*/, false /*noreboot*/)
+	} else {
+		tc.activateOS(ctx, t, false /*standby*/, true /*noreboot*/)
+	}
+
+	if !*deviations.NoOSInstallForStandbyRP && tc.dualSup {
 		tc.transferOS(ctx, t, true)
```
xw-g, line 99: Do you still need transferOS to backup in the case of `NoOSInstallForStandbyRP`?

mojiiba: no, there is no need to trasnfer

xw-g: That's my point.  You still have `tc.transferOS(ctx, t, true)` in your PR, or I read it wrong?

mojiiba: yes, I have it, but it is guarded by  !*deviations.NoOSInstallForStandbyRP.  This code block will be skipped when NoOSInstallForStandbyRP is set to true.

xw-g: ah, sorry I missed the double negative. Still it's a bit hard to read. Say if a dual SUP chassis requires reboot in activation, then you line#93 will reboot the whole chassis immediately, right? which means your line#98 logic will reach to a offline chassis?  Anyway you can help refactor it to make it more readable?

mojiiba: sure, I will do that

mojiiba: It seems I also missed the double negative regarding the noreboot flag and I may need a few more days to address this.

mojiiba: @xw-g  thanks for the patience, I have addressed your comments.

mojiiba:
> Hi @mojiiba, could you please fix the typo: OSActiavteRequiresReboot -> OSActivateRequiresReboot
> 
> Thanks!



> Hi @mojiiba, could you please fix the typo: OSActiavteRequiresReboot -> OSActivateRequiresReboot
> 
> Thanks!



> Hi @mojiiba, could you please fix the typo: OSActiavteRequiresReboot -> OSActivateRequiresReboot
> 
> Thanks!

thanks, have addressed the comment

rszarecki (approved):

PR #676, opened 2022-10-19
RT-5.1 - Add Tolerance for ipv6 control packets and traffic fix

agvarghe-cisco:
Following Changes done to singleton_test
-     Tolerance for ipv6 control packets
-     For packet larger than MTU configured case , compare with dropped packet counters


vosipchu:
Xiao, Yan, can you please have a look? 
The changes, in short:
1) adding 6 packets as a margin for ipv6 traffic (multicast based neighbor discovery. Might be observed up to two times due to test length, with three packets needed in each run)
2) Not checking drops for MTU+ sized packets (they are dropped). 

LGTM from my side

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -50,8 +50,9 @@ func TestMain(m *testing.M) {
 // where WW:XX:YY:ZZ are the four octets of the IPv4 in hex.  The 0x02
 // means the MAC address is locally administered.
 const (
-	plen4 = 30
-	plen6 = 126
+	plen4                      = 30
+	plen6                      = 126
+	Ipv6ControlPacketTolerance = 6
```
vosipchu, line 55: https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps
maybe we can make it shorter? something like "v6PktTolerance" 
Yan, Xiao? Thoughts?

sachendras: Not checking drops for MTU+ sized packets (they are dropped) 
 -- IIUC recommendation is to not check drops for MTU+ sized packets. If so, Isnt that the test requirement to ensure MTU+ packets are not forwarded? So shouldnt the test check for drops?

Following per the test requirement. 

For IPv4 and IPv6:
With traffic flow from ATE port-1 to ATE port-2, ensure:
For MTUs of 1500, 5000, 9212:
  - Packets with size greater than the configured MTU with DF-bit set are not transmitted. <<<<<---------
  - Packets with size of configured MTU are received.
  - Packets with size less than the configured MTU are received.

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -372,13 +373,19 @@ func (tc *testCase) testFlow(t *testing.T, packetSize uint16, ipHeader ondatra.H
...
 	if outPkts == 0 {
 		t.Error("Flow did not send any packet")
 	} else if avg := octets / outPkts; avg > uint64(tc.mtu) {
 		t.Errorf("Flow source packet size average got %d, want <= %d (MTU)", avg, tc.mtu)
 	}
-	if p1InDiff.unicast < outPkts {
-		t.Errorf("DUT received too few source packets: got %d, want >= %d", p1InDiff.unicast, outPkts)
+
+	if pktlmtu {
+		t.Logf("Skipping packet check for packets larger than MTU")
```
vcheny, line 384: Why do we want to skip checking packets larger than MTU? If they are actually dropped, can we fix the counter?

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -388,7 +395,9 @@ func (tc *testCase) testFlow(t *testing.T, packetSize uint16, ipHeader ondatra.H
 	} else if avg := octets / inPkts; avg > uint64(tc.mtu) {
 		t.Errorf("Flow destination packet size average got %d, want <= %d (MTU)", avg, tc.mtu)
 	}
-	if inPkts < p2OutDiff.unicast {
+
+	// Tolerance added for ipv6 control packets
+	if inPkts+Ipv6ControlPacketTolerance < p2OutDiff.unicast {
```
vcheny, line 400: Instead of adding a tolerance, can we check the unicast counters instead if they are available?

mojiiba: the ipv6 advertisement message are unicast. the issue is that ixia does not account for control packets  for flow counters.

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
mojiiba: when packets are dropped they are not counted as unicast. if we want to do packet counter comparison for packet larger than  MTU,  I think we  should compare based on the dropped packet counters rather unicast packet.  we can explorer this as another pr if you want.

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
mojiiba: yes, i agree the test should check for drop counter rather than unicast counters. can we add that check in another pr?

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
vosipchu: NS message is a solicit multicast, a NA message is a unicast. So, the exchange is just 50% unicast

xw-g (changes requested):
* It would be nice to add background/motivation in the PR description. It will help the reviewers a lot. 

* Also, it would be nice to keep one PR one thing, this will help speed up review process. Here we are mixing `pktlmtu` and `Ipv6ControlPacketTolerance`, which are different topics.

jjing1122:
cc: @jjing1122

PR #681, opened 2022-10-21
Adding AssociationType as Server to NTP Server config

agvarghe-cisco:
Adding AssociationType as Server to NTP Server config

### DIFF
--- a/internal/deviations/deviations.go
+++ b/internal/deviations/deviations.go
@@ -67,6 +67,9 @@ import "flag"
 
 // Vendor deviation flags.
 var (
+	NtpAssociationType = flag.Bool("ntp_association_type_enabled", false,
+		"Device requires AssociationType as Server to be explicitly set to true.")
+
 	InterfaceEnabled = flag.Bool("deviation_interface_enabled", false,
 		"Device requires interface enabled leaf booleans to be explicitly set to true.  Full OpenConfig compliant devices should pass both with and without this deviation.")
 
--- a/feature/system/ntp/tests/server_test.go
+++ b/feature/system/ntp/tests/server_test.go
@@ -17,6 +17,7 @@ package system_ntp_test
 import (
 	"testing"
 
+	"github.com/openconfig/featureprofiles/internal/deviations"
 	"github.com/openconfig/ondatra"
 	"github.com/openconfig/ondatra/telemetry"
 )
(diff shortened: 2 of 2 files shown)


bstoll:
Could you elaborate on why this change is needed?  The AssociationType leaf has a default value of [SERVER](https://github.com/openconfig/public/blob/00a27a5f0f3c472205ef293ab238e0530b610811/release/models/system/openconfig-system.yang#L679) so it seems like this change would be redundant.

agvarghe-cisco:
> Could you elaborate on why this change is needed? The AssociationType leaf has a default value of [SERVER](https://github.com/openconfig/public/blob/00a27a5f0f3c472205ef293ab238e0530b610811/release/models/system/openconfig-system.yang#L679) so it seems like this change would be redundant.
@bstoll , AssociationType leaf if mandatory for ntp server config ,  can we have a deviation until it's fixed

bstoll:
Yes, this could be done with a deviation.  I think I would prefer if this was documented such that the NOS does not populate default values in gNMI requests and not something specific to configuring NTP server.

vosipchu:
Cisco, can you append code for deviation itself? One there we will do a quick lgtm round and i will commit. Thanks! 
Also, when will this default behavior be implemented?
thanks

agvarghe-cisco:
> [vosipchu](/vosipchu)

@ vosipchu, Added AssociationType as deviation. Will let know when default behavior is implemented

vosipchu:
Sash, Yan, need your view on this deviation. If anything is not clear, ping me with questions. Thanks
@sachendras @vcheny

sachendras:
LGTM!
Lets also track the final release for delivery and ETA.

vcheny (approved):

vosipchu (approved):

internal/deviations/deviations.go
```diff
@@ -67,6 +67,9 @@ import "flag"
 
 // Vendor deviation flags.
 var (
+	NtpAssociationType = flag.Bool("ntp_association_type_enabled", false,
```
liulk, line 70: Please be advised that NTP must be capitalized according to Go initialism.

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

Deviation flags must begin with `--deviations_`.

Also, the help string should explain what the OpenConfig behavior is for a fully compliant device.

Fixing these issues in #762

PR #682, opened 2022-10-21
Fix IS-IS tests

dan-lepage:
The `init-fails`, `part-changes`, and `manual-address-drop-from-area` are not generally used and shouldn't be tested or required by the IS-IS feature spec (init-fails and manual-...-area were already removed from the spec, but not from the corresponding tests or test plans). This removes them entirely.

This also adds a nil deviation check in the LSP update check, since some vendors return nil instead of false when the value isn't set.

### DIFF
--- a/feature/experimental/isis/ate_tests/lsp_updates_test/lsp_updates_test.go
+++ b/feature/experimental/isis/ate_tests/lsp_updates_test/lsp_updates_test.go
@@ -41,7 +41,7 @@ func TestOverloadBit(t *testing.T) {
 	setBit := isisPath.Global().LspBit().OverloadBit().SetBit()
 	deadline := time.Now().Add(time.Second)
 	checkSetBit := check.Equal(setBit.State(), false)
-	if *deviations.AllowNilForDefaults {
+	if *deviations.MissingValueForDefaults {
 		checkSetBit = check.EqualOrNil(setBit.State(), false)
 	}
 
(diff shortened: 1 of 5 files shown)


feature/experimental/isis/ate_tests/lsp_updates_test/lsp_updates_test.go
```diff
@@ -40,8 +40,13 @@ func TestOverloadBit(t *testing.T) {
 	overloads := isisPath.Level(2).SystemLevelCounters().DatabaseOverloads()
 	setBit := isisPath.Global().LspBit().OverloadBit().SetBit()
 	deadline := time.Now().Add(time.Second)
+	checkSetBit := check.Equal(setBit.State(), false)
+	if *deviations.AllowNilForDefaults {
```
liulk, line 44: Can we change the deviation flag to MissingValueForDefaults? I think "allow" is not really accurate in the name because it's not the device allowing nil, but rather the test is allowing it. We generally want the deviation to describe the device behavior, not the test behavior. I also commented on #638.

liulk (changes requested):

feature/experimental/isis/ate_tests/lsp_updates_test/lsp_updates_test.go
dan-lepage: Fixed. I also removed the `MixedOcCliPrefersCli` deviation, which was added accidentally due to a rebasing mistake.

liulk (approved):

PR #683, opened 2022-10-21
Create packet_link_qualification_test with fake response

sezhang2:
- The fake response can be removed if link qualification is supported by the DUT


hariramnat (approved):

sezhang2:
The test passed in my local run, so I submitted it as an experimental code

PR #685, opened 2022-10-21
pull_request.yml upgrade actions/cache from v2 to v3

liulk:


bstoll:
I think there are a bunch of actions that need to be updated in all the workflows if we want to get the deprecated warnings resolved , but this is a start.  I think #684 will fix the actual issue with the check failing though.

bstoll (approved):

PR #686, opened 2022-10-21
rundata collects runtime data about the test environment

liulk:


bstoll:
Still going through the details, but where is the entry point for `runtime.Properties()` (fptest.RunTests?)  Can this PR also include the hook to call the ondatra report AddTestProperty?

liulk:
If I add it now it's going to go live. I want to get vendor comment before doing that.

But the idea is to add it either here: https://github.com/openconfig/featureprofiles/blob/main/topologies/binding/binding.go#L85

Or in NotifyExitCode() which is not yet overridden by the static binding, but straightforward to add.

bstoll (approved):

Issue #302, opened 2022-07-06
gNOI Ping Test Question


Hobbydos:
I was wondering if this is expected behavior:
in `feature/gnoi/tests/ping_test/ping_test.go` the following testcase sends ping request with size=0 but seems to expect a reply of 64
Sorry having multiple code blocks, just using them to get to the sourced reference of `minimumPingReplySize          = 64` 

However, the proto file for gNOI references that if the request has a size of 0, then the reply size should be default (usually 56)
[/openconfig/gnoi/blob/master/system/system.proto#L176](https://github.com/openconfig/gnoi/blob/master/system/system.proto#L176)
Are we expecting size of 56b then encap of 8b to = 64?
```
{
		desc: "Check ping with IPv4 destination",
		pingRequest: &spb.PingRequest{
			Destination: ipv4Addrs[0].GetIp(),
		},
		expectedReply: commonExpectedIPv4Reply,
		expectedStats: commonExpectedReplyStats,
	},
```

Ref to commonExpectedIPv4Reply
```
	commonExpectedIPv4Reply := &spb.PingResponse{
		Source:   ipv4Addrs[0].GetIp(),
		Time:     minimumPingTime,
		Bytes:    minimumPingReplySize,
		Sequence: minimumPingReplySequence,
		Ttl:      minimumPingReplyTTL,
	}
```

Ref to minimumPingReplySize
```
const (
	minimumPingRequestSize        = 64
	maximumDefaultPingRequestSize = 1512
	maximumPingRequestSize        = 9202
	minimumPingReplySize          = 64
	icmpHeaderSize                = 20
	// Set Minimum value to 1 to verify the field is not empty.
	minimumPingReplyTTL      = 1
	minimumPingReplySequence = 1
	minimumPingTime          = 1
	minimumSent              = 1
	minimumReceived          = 1
	minimumMinTime           = 1
	minimumAvgTime           = 1
	minimumMaxTime           = 1
	//StdDeviation would be 0 if we only send 1 ping.
	minimumStdDev = 1
)
```

This causes test case failures
```
        
=== RUN   TestGNOIPing/Check_ping_with_IPv4_destination
    ping_test.go:403: Sent ping request: destination:"192.168.99.99"
        
    ping_test.go:419: Got ping responses: Items: 6
[...]

PR #687, opened 2022-10-21
Changed minimumPingReplySize

Hobbydos:
changed minimumPingReplySize to 56 to be in line with https://github.com/openconfig/gnoi/blob/main/system/system.proto#L176


bstoll:
I think there are some shortcomings with this test and comparing the correct number of bytes sent and received, but I am not sure this is a correct change.  The number of bytes received should be 8 more than the number sent (this is the size of the ICMP header).  The tests are sending 64 for a minimum, so I would expect to receive 72 bytes in responses.  I don't believe this is clearly documented in the GNOI ping behavior, so if you have a different interpretation I would like to know.  The test transmits 64 bytes, so I expect the response to be at least 64 bytes, so lowering this to 56 doesn't make sense to me.

I am open to reworking this aspect of the test, as I think this could be improved (i.e., `icmpHeaderSize` seems strange).

Hobbydos:
@bstoll, Thanks for the feedback. To preface this I had originally raised [issue 302](https://github.com/openconfig/featureprofiles/issues/302) where I outlined that the testcases which have `minimumPingRequestSize` set are correctly handled. However, when it isn't set, it seems like the default value expected by the testcase is `64bytes`:
```
=== RUN   TestGNOIPing/Check_ping_with_IPv4_destination
    ping_test.go:403: Sent ping request: destination:"192.168.99.99"
        
    ping_test.go:419: Got ping responses: Items: 6
        , Content: [source:"192.168.99.99"  time:4602238  bytes:56  sequence:1  ttl:64 source:"192.168.99.99"  time:3387708  bytes:56  sequence:2  ttl:64 source:"192.168.99.99"  time:4430538  bytes:56  sequence:3  ttl:64 source:"192.168.99.99"  time:3408598  bytes:56  sequence:4  ttl:64 source:"192.168.99.99"  time:4406217  bytes:56  sequence:5  ttl:64 source:"192.168.99.99"  time:20568469  sent:5  received:5  min_time:3387708  avg_time:4047059  max_time:4602238  std_dev:597212
        
    ping_test.go:428: Check each ping reply 1 out of 6.
          source:"192.168.99.99"  time:4602238  bytes:56  sequence:1  ttl:64
    ping_test.go:443: Ping Bytes: got 56, want >= 64
```
When `minimumPingRequestSize` is set, the DUT replies with the appropriate expected value.

Hobbydos:
I don't want to overcomplicate this either, but it seems like different vendors might have different implementations. I was testing with cEOS 4.28.3M and their default is set to 64b. Maybe the best way to handle this is to look for more than 1 singular value (56, 64, might be more)? 
```
 gnoic -a 172.20.20.2:6030 -u admin -p admin --insecure system ping --format json --destination 192.168.99.99 --print-proto --count 2
"172.20.20.2:6030":
gnoi.system.PingRequest
destination:  "192.168.99.99"
count:  2
interval:  1000000000

"172.20.20.2:6030":
gnoi.system.PingResponse
source:  "192.168.99.99"
time:  25000
bytes:  64
sequence:  1
ttl:  64

{
  "target": "172.20.20.2:6030",
  "response": {
    "source": "192.168.99.99",
    "time": 25000,
    "bytes": 64,
    "sequence": 1,
    "ttl": 64
  }
```

Hobbydos:
> I see your point. If we change `minimumPingRequestSize` to 56 as well (as this is explicitly documented), I think that would fix the situation you are describing and keep the other test cases as correct as they can be for now.

That's correct, since the test case looks at `>= minimumPingRequestSize`, this should cover for other vendors who might default to 64 bytes.

xw-g:
@Hobbydos have you made changes based on @bstoll 's comment?

Hobbydos:
> 

I'll add the change to `minimumPingRequestSize` as well. I was just testing the changes to ensure it doesn't fail on different vendor implementations.

xw-g (approved):

xw-g:
@bstoll to review and merge.

bstoll:
/gcbrun

Thanks, this LGTM.

PR #691, opened 2022-10-25
Changes for Issue-466. NextHop index as integer string.

trathod1:
Changes for below error hit during get_rpc_test.go script execution -

[InvalidArgument] failed to set .network-instance{.name=="DEFAULT"}.next-hop-groups.group{.name=="oc-static-198.51.100.192/26"}.nexthop{.index==0}.index - Invalid value "AUTO_0": Must be an integer in range 0..65535


trathod1:
Closing as of now due to CLA.

PR #692, opened 2022-10-25
change GetOrCreateNextHop from 'AUTO_0' to '0'

Hobbydos:
Changes for below error hit during get_rpc_test.go script execution -

[InvalidArgument] failed to set .network-instance{.name=="DEFAULT"}.next-hop-groups.group{.name=="oc-static-198.51.100.192/26"}.nexthop{.index==0}.index - Invalid value "AUTO_0": Must be an integer in range 0..65535

### DIFF
--- a/feature/gribi/ate_tests/get_rpc_test/get_rpc_test.go
+++ b/feature/gribi/ate_tests/get_rpc_test/get_rpc_test.go
@@ -322,7 +322,7 @@ func testIPv4LeaderActive(ctx context.Context, t *testing.T, args *testArgs) {
 	static := &telemetry.NetworkInstance_Protocol_Static{
 		Prefix: ygot.String(staticCIDR),
 	}
-	static.GetOrCreateNextHop("AUTO_0").NextHop = telemetry.UnionString(atePort2.IPv4)
+	static.GetOrCreateNextHop("0").NextHop = telemetry.UnionString(atePort2.IPv4)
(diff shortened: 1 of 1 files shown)


brwallis:
@xw-g I think it's common in implementations to key the next hop on an index - and index _implies_ a number. Maybe the more relevant question is should the OC modeling itself change to model this as an actual integer. Most of the other tests in this repo provide a integer encoded as a string as the value (which implementors can simply convert to an int), but this specific test uses 'AUTO_0'.

aaronmillisor:
I agree with Bruce. This seems like a bug in the model, no? In the AFT model (to which I would expect static-route to map), the index is a uint64. I'm not seeing any other model which contains a leaf called 'index' that is a string.

xw-g:
@dplore can you help share your opinion here about the OC? cc: @wenovus too.

dplore:
I am ok with changing [static route index](https://github.com/openconfig/public/blob/00a27a5f0f3c472205ef293ab238e0530b610811/release/models/local-routing/openconfig-local-routing.yang#L206) to an int.    I'd prefer the model to support a larger range than 0..65535 however.    A quick scan shows the majority type used for `leaf index` is uint64.   Shall we make the static index uint64?

dplore (approved):
A future/separate PR can be filed in openconfig/public for the model change.  I am ok with this change.

PR #700, opened 2022-10-26
RT-3.1: Policy based VRF selection

bstoll:
Add test RT-3.1.

This is a continuation of #521 with 2d83ab4 changes & rebased to main.


bstoll:
Most of these changes are in #712.

PR #713, opened 2022-10-28
TE 6.2: remove unnecessary call of configureDUT 

mojiiba:
This pull removes an unnecessary call tp configureDUT for the second time. I believe this code is left their unintentionally. If I am wrong, can you please clarify why do we need to configure DUT twice?


vosipchu:
I'm fine overall, but want Yan to review and confirm. 
Sash as well @sachendras

vcheny:
lgtm. Do we see different test result w/ and w/o the 2nd configureDUT() call?

mojiiba:
> lgtm. Do we see different test result w/ and w/o the 2nd configureDUT() call?

yes, the second configure dut may fail on devices that do not allow replacing some of interfaces' configuations when they are added in to a vrf and activated.

vcheny:
Any reason why the existing interfaces can't be replaced? thx.

mojiiba:
> Any reason why the existing interfaces can't be replaced? thx.

I am not sure about the detail reason, but I can guess this simplifies the implementation and improves the code reliability since changing an active interface in a vrf also have impacts on routing tables and etc. This is not even supported in cli.  If you want I can investigate this internally and then we can have an offline discussion about this.

Anyway, in the context of this test that tries to test gribi flush on non-default vrf, the second config  is unnecessary  and can be removed. We can also have a separate discussion if you want this behavior to be changed.

vcheny:
Thank for the explanations. Yeah, for this specific test the 2nd config seems redundant.
This behavior doesn't seem desirable though. Is it only specific to interfaces in VRF? Do we have to delete it first and then add it back w/ new config in order to make changes?

vosipchu (approved):

vcheny (approved):
lgtm

xw-g (approved):
Approve given that logic has been approved by vcheny@ and vosipchu@

sachendras:
lgtm

PR #715, opened 2022-10-28
Add deviation to support deprecated vlan-id config

bstoll:
Vendor support for vlan-id configuration can vary - some support vlan/config/vlan-id and others support the newer vlan/match/single-tagged/config/vlan-id leaf.  This deviation allows us to support both methods.


bstoll:
I forgot to add a link to the relevant OC for this change:

https://github.com/openconfig/public/blob/00a27a5f0f3c472205ef293ab238e0530b610811/release/models/vlan/openconfig-vlan.yang#L271-L290

Ankur19 (approved):
LGTM!

PR #513, opened 2022-09-14
Fix RT-1.3: BGP route propagation needs NI type, router-id and peer-group

yashikakn:
Hi,
Resolves #312. Please review and provide approval. 

Thanks.

Issue #391, opened 2022-08-09
RT-1.3: BGP Route Propagation needs uplifting for router-id, global address family, route-policy for ebgp


singh-prem:
RT-1.3: BGP Route Propagation currently does not have 
1. router-id
2. global address family initialization such that initialized address family can be used under neighbors
3. route-policy used for filtering for an ebgp session

I can take up uplifting this test if no-one has any issues

trathod1:
Observing same issue. Script failing due to - 
       1. Missing PeerGroup configuration with route-policy (required for ebgp) and address-family enabling.
       2. Missing Address-family configuration at global and neighbor (ipv4)

xw-g:
is this related to this PR? https://github.com/openconfig/featureprofiles/pull/513

bstoll:
Closing as this does appear to be resolved by that PR.

PR #716, opened 2022-10-28
Porting changes openconfig#312 to OTG location to same script.

stthummuri:


feature/bgp/addpath/otg_tests/route_propagation_test/route_propagation_test.go
```diff
@@ -460,6 +466,11 @@ func TestBGP(t *testing.T) {
 			}
 
 			dut := ondatra.DUT(t, "dut")
+			t.Log("Configure Network Instance")
+			dutConfNIPath := dut.Config().NetworkInstance(*deviations.DefaultNetworkInstance)
+			dutConfNIPath.Type().Replace(t, telemetry.NetworkInstanceTypes_NETWORK_INSTANCE_TYPE_DEFAULT_INSTANCE)
+			dutConfNIPath.RouterId().Replace(t, dutPort2.IPv4)
```
jasdeep-hundal, line 472: While router ID is a constant for these tests, keep it as part of the each test case like in the corresponding ATE test (https://github.com/openconfig/featureprofiles/blob/1a438857b49d6b423ff3a694912cd41a0b26afdf/feature/bgp/addpath/ate_tests/route_propagation_test/route_propagation_test.go#L194). The relevant DUT configuration for a particular test case is made more obvious that way.

jasdeep-hundal (changes requested):

feature/bgp/addpath/otg_tests/route_propagation_test/route_propagation_test.go
stthummuri: Made the changes

jasdeep-hundal:
Needs a gofmt.

sachendras:
I got pinged by IXIA that the introduction of peer-group in https://github.com/openconfig/featureprofiles/pull/513/files broke the OTG version of RT-1.3. The KNE instance ran on Arista ceos-lab now seeing the gNMI error "found bad element: peer-group".
IMO, as per https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-bgp-peer-groups-peer-group, peer-group is a standard that should be supported by all implementations. I see similar requirement by Cisco as well in https://github.com/openconfig/featureprofiles/issues/391. Hence, support for the same should also exist in Arista ceos-lab.

PR #721, opened 2022-10-31
Adding Type as GetRequest_CONFIG in GetRequest

agvarghe-cisco:
Adding Type as GetRequest_CONFIG in GetRequest

Get will return entire snapshot in the GetResponse message if type is not specified and it not supported  in some platforms due to efficiency reason.

### DIFF
--- a/feature/system/gnmi/get/tests/system_gnmi_get_test/system_gnmi_get_test.go
+++ b/feature/system/gnmi/get/tests/system_gnmi_get_test/system_gnmi_get_test.go
@@ -66,6 +66,7 @@ func TestGNMIGet(t *testing.T) {
 			Path: []*gpb.Path{{
 				// empty path indicates the root.
 			}},
+			Type:     gpb.GetRequest_CONFIG,
 			Encoding: gpb.Encoding_JSON_IETF,
 		},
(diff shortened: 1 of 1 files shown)


guoshiuan:
Tried to run it with 8808, and got the following error:

=== RUN   TestGNMIGet/response_for_/_honours_encoding
    [third_party/openconfig/featureprofiles/feature/system/gnmi/get/tests/get_test.go:112](https://cs.corp.google.com/piper///depot/google3/third_party/openconfig/featureprofiles/feature/system/gnmi/get/tests/get_test.go?l=112): cannot call Unmarshal for path , err: NONE is not a valid value for enum field Facility, type oc.E_SystemLogging_SYSLOG_FACILITY

--- FAIL: TestGNMIGet (10.30s)
    --- PASS: TestGNMIGet/single_response_for_/ (4.53s)
    --- FAIL: TestGNMIGet/response_for_/_honours_encoding (4.76s)
FAIL

guoshiuan:
The "cannot call Unmarshal for path , err: NONE is not a valid value for enum field Facility, type oc.E_SystemLogging_SYSLOG_FACILITY" is caused by the base config (the logging) we use; it is not fatal so this PR should be safe to be merged.

PR #727, opened 2022-11-01
OTG Test - RT-2.2: IS-IS LSP Updates

### LINKED ISSUE #620: OTG: RT-2.2: IS-IS LSP Updates

octpetre:
OTG conversion of **_RT-2.2: IS-IS LSP Updates_**
Output log found [here](https://drive.google.com/file/d/1tMLsXlqa2u00eUOlIq_QC4nl9C6iCNR7/view?usp=share_link)
Fixes #620


feature/experimental/isis/otg_tests/internal/session/session.go
```diff
@@ -118,30 +123,43 @@ func addISISOC(dev *oc.Root, areaAddress, sysID, ifaceName string) {
...
+
+	devIsisInt := devIsis.Interfaces().
+		Add().
+		SetEthName(dev.Ethernets().Items()[0].Name()).
+		SetName("devIsisInt").
+		SetNetworkType(gosnappi.IsisInterfaceNetworkType.POINT_TO_POINT).
+		SetLevelType(gosnappi.IsisInterfaceLevelType.LEVEL_2)
+
+	devIsisInt.Advanced().
+		SetAutoAdjustMtu(true).SetAutoAdjustArea(true).SetAutoAdjustSupportedProtocols(true)
```
jasdeep-hundal, line 146: Not too familiar with IS-iS, are these settings needed for the test? (We do not have the equivalent in the IxNetwork ATE implementation).

feature/experimental/isis/otg_tests/internal/session/session.go
```diff
@@ -201,9 +220,6 @@ func (s *TestSession) WithISIS() *TestSession {
 // IS-IS configuration of ts.ATETop
 func (s *TestSession) ConfigISIS(ocFn func(*oc.NetworkInstance_Protocol_Isis), ateFn func(*ixnet.ISIS)) {
```
jasdeep-hundal, line 221: I still see the ateFn argument here used in OTG tests, is the intent to remove it later? (If so, why is it no longer needed here?)

feature/experimental/isis/otg_tests/internal/session/attrs.go
```diff
@@ -122,3 +123,25 @@ func (a *Attributes) AddToATE(top *ondatra.ATETopology, ap *ondatra.Port, peer *
 	}
 	return i
 }
+
+// AddToOTG adds basic elements to a gosnappi configuration
+func (a *Attributes) AddToOTG(top gosnappi.Config, ap *ondatra.Port, peer *Attributes) gosnappi.Device {
+	top.Ports().Add().SetName(ap.ID())
```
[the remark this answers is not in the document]
octpetre: This is possible in gosnappi, but in this helper we don't need to do that. I've only encounter a single test where multiple devices were required on a port so if we need to change this we should rather edit internal/attrs/attrs.go instead

feature/experimental/isis/otg_tests/internal/session/session.go
octpetre: The first 2 are true by default and the 3rd one may not be needed. I will remove all.

feature/experimental/isis/otg_tests/internal/session/session.go
octpetre: I removed it for now. That is actually not used in the RT-2.2 ATE test

jasdeep-hundal: That works. (We'll need to add something back in for base_adjacencies_test, but that will have a different type anyways.)

feature/experimental/isis/otg_tests/internal/session/attrs.go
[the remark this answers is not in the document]
jasdeep-hundal: Ah, I wasn't totally clear: is top.Ports().Add().SetName(ap.ID()) idempotent if adding the same port to the topology multiple times?

While it wouldn't necessarily happen in this test, we could make sure that this is not accidentally misused by adding the ports outside of this function.

jasdeep-hundal (changes requested):

feature/experimental/isis/otg_tests/internal/session/attrs.go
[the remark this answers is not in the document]
octpetre: Moved the ports part outside of AddToOTG helper.

jasdeep-hundal:
@octpetre : The base adjacencies test needs an update for the modified ConfigISIS function.

octpetre:
I will update the file in the #735. Here is not needed

octpetre:
@jasdeep-hundal could you please approve after the latest ygnmi API changes

feature/experimental/isis/otg_tests/lsp_updates_test/lsp_updates_test.go
```diff
@@ -91,9 +110,20 @@ func TestMetric(t *testing.T) {
...
 		t.Error(err)
 	}
-	// TODO: Verify the link state database on the ATE once the ATE reports this properly
-	// ateTelemPth := ts.ATEISISTelemetry(t)
-	// ateDB := ateTelemPth.Level(2).LspAny()
-	// for _, nbr := range ateDB.Tlv(telemetry.IsisLsdbTypes_ISIS_TLV_TYPE_IS_NEIGHBOR_ATTRIBUTE).IsisNeighborAttribute().NeighborAny().Get(t) {
-	// }
+
+	_, ok := gnmi.WatchAll(t, otg, gnmi.OTG().IsisRouter("devIsis").LinkStateDatabase().LspsAny().Tlvs().ExtendedIpv4Reachability().PrefixAny().Metric().State(), time.Minute, func(v *ygnmi.Value[uint32]) bool {
+		time.Sleep(1 * time.Second)
```
jasdeep-hundal, line 115: Last thing :) : Can we remove this one as well?

octpetre: Done

jasdeep-hundal (approved):

PR #729, opened 2022-11-01
Implementation proposal for network instance deviation

Hobbydos:
Implementation proposal for network instance deviation. Purpose is to allow deviations where interfaces could be assigned to network instances.


Hobbydos:
Re-Writing this PR.

PR #737, opened 2022-11-04
Proposal for network instance assignment deviations

Hobbydos:
Proposal for deviation implementation to assign interface into network instance


vcheny:
Can we have the following details?
Issues:
Proposals:
Justifications:
Long Term Solution:

xw-g:
cc: @xw-g

internal/deviations/deviations.go
```diff
@@ -99,4 +108,21 @@ var (
 	StaticProtocolName = flag.String("deviation_static_protocol_name", "DEFAULT", "The name used for the static routing protocol.  The default name in OpenConfig is \"DEFAULT\" but some devices use other names.")
 
 	DeprecatedVlanID = flag.Bool("deviation_deprecated_vlan_id", false, "Device requires using the deprecated openconfig-vlan:vlan/config/vlan-id or openconfig-vlan:vlan/state/vlan-id leaves.")
+
+	AddSubIntfToNetInst = flag.Bool("deviation_adding_subintf_to_netowork_instance", true,
+	"Vendor requiring explicitly attaching sub-interface to network-instance, this deviation should be marked true. Default expectation is sub-interface is part of default network-instance implicitly.")
```
robshakir, line 113:
```suggestion
	"If a device requires explicit attachment of a sub-interface to a network-instance, this deviation should be marked true. The default expectation is that a sub-interface is part of default network-instance implicitly.")
```

robshakir: I'd suggest that we file a bug in openconfig/public to make sure that a clarfication is made in the model that clearly states that this default association exists.

internal/deviations/deviations.go
```diff
@@ -99,4 +108,21 @@ var (
...
 )
+
+func SubIntfToNetworkInstance(t *testing.T, dconf *ondatra.Config, i *telemetry.Interface, si uint32, inst string) {
```
robshakir, line 116: There are two implementations we could have of this -- we could handle it at the end of the configuration generation - i.e., go and make sure that every interface that is configured is associated with a network instance, and if not, then make that association -- or alternatively this design. Did you consider the alternate approach in the wider implementation?

internal/deviations/deviations.go
```diff
@@ -99,4 +108,21 @@ var (
...
+	netInst := &telemetry.NetworkInstance{Name: ygot.String(inst)}
+	netInstIntf, err := netInst.NewInterface(i.GetName())
+	if err != nil {
+		t.Errorf("Error fetching NewInterface for %s", i.GetName())
```
robshakir, line 120:
```suggestion
		t.Errorf("Error fetching NewInterface for %s, error: %v", i.GetName(), err)
```

Making sure that we return the error that is received to aid with debugging.

feature/interface/staticarp/ate_tests/static_arp_test/static_arp_test.go
```diff
@@ -154,10 +154,18 @@ func configureDUT(t *testing.T, peermac string) {
...
+		if *deviations.AddSubIntfToNetInst {
+			deviations.SubIntfToNetworkInstance(t, d, i1, 0, *deviations.DefaultNetworkInstance)
+		}
+
 	p2 := dut.Port(t, "port2")
 	i2 := &telemetry.Interface{Name: ygot.String(p2.Name())}
 	d.Interface(p2.Name()).Replace(t,
 		configInterfaceDUT(i2, &dutDst, &ateDst, peermac))
+
+		if *deviations.AddSubIntfToNetInst {
```
robshakir, line 166: Indentation is incorrect -- please run `gofmt` on this file.

feature/interface/staticarp/ate_tests/static_arp_test/static_arp_test.go
```diff
@@ -154,10 +154,18 @@ func configureDUT(t *testing.T, peermac string) {
 	d.Interface(p1.Name()).Replace(t,
 		configInterfaceDUT(i1, &dutSrc, &ateSrc, peermac))
 
+		if *deviations.AddSubIntfToNetInst {
```
robshakir, line 157: Indentation is incorrect -- please run `gofmt` on this file.

internal/deviations/deviations.go
robshakir, line 116: Is this the right place to include this code -- pinging @liulk to discuss where we want to include the code that implements deviations, today it's included in the code base, but I think having this as a reusable library would be helpful.

robshakir (changes requested):
Thanks for submitting this.

I thought that this was explicitly stated somewhere that all interfaces are assumed to be part of the `DEFAULT-INSTANCE` implicitly, but I cannot find this reference right now. I think we need a change in `openconfig-network-instance.yang` that states that this is the case. I think this is solely documentation since there is no way to define default entries in a list.

The 'default' association is something implemented by most existing NOSes that I'm aware of.

internal/deviations/deviations.go
liulk: The justification is still unclear to me. It says that the device requires explicitly putting a subinterface into **a** network instance, and the added helper puts the subinterface into an arbitrary (possibly non-default) network instance. But the updated static ARP test puts it into **the default** network instance, which is not required either by OpenConfig spec nor supported by the wording of this deviation.

What happens if the test only configures the subinterface but does not put in into any network instance?

robshakir: This is exactly the problem. The assumption that is made in the schema and tests is that an interface/subinterface is by default assigned to the `DEFAULT-INSTANCE` network instance, which we've now said should be called `DEFAULT`. 

The spec doesn't explicitly say this but it *is* expected, as the tests today concrete this by not assigning interfaces to any network instance. 

It is important to note that the models allow for this explicit assignment (explicitly specifying the default is fine in all cases).

The implementation that is being discussed here does not currently make this default assignment - so we need a way to be able to make it explicitly. In practice, this is a more explicit and likely more compatible config. i.e., if one knows all the interfaces one could make this assignment in all cases and it wouldn't change the behaviour of the device.

For non-default network instances, then this configuration _is_ required explicitly because the implicit default is not the case. So it is not clear that this can be assigned to an arbitrary network-instance. A change to the deviation could be that it only assigns to the default instance - this will need to be dependent upon the default instance name deviation.

Another discussion we could have here is whether we just make this assignment explicit in all tests without the deviation. It seems pragmatic to introduce the deviation to say that this is a "more compatible config, that we don't know whether it's supported in all implementations" and then come back and determine what the preferred approach is going forward after clarifying the specification.

Let's meet if we need to discuss this further.

christophersmall: Note that the flag is "deviation_adding_subintf_to_**netowork**_instance". you probably want it to be "deviation_adding_subintf_to_**network**_instance"

liulk: What is the process for moving an interface/subinterface from one network instance to another? I'm asking because device behavior can differ.

Currently, because the interface/sub does not belong to any specific NI, tests could put them into a specific NI with a single `Replace`. But if we put them into `DEFAULT` explicitly, some devices might require the interface/sub to also be removed explicitly before they can be added to another NI.

If we want tests to consistently put interface/sub into NI, we need to allocate time to update the test plans and tests, and add this requirement to the style guide so new tests will follow it.

My inclination is to just keep the test coverage minimal. If a test does not need NI, then leave it alone. We already have VRF tests (RT-3.1, RT-3.2) that put interface/sub into NI explicitly.

mojiiba: > Hi @robshakir, Thanks for the feedback and I'm glad the conversation is started on this subject.
> 
> One additional point that I would like to bring forward is a case where we might assign an interface to a network instance and move it to a different one - (I've not done a deep dive of all test cases to see if that might happen).
> 
> In certain vendors, you might have to explicitly remove the association from it's original Network instance to successfully assign it to the new desired Network Instance.

+1 to the second point

nokia1adam: The assumption that unreferenced subintefaces always belong to DEFAULT feels like a slippery slope. What other OC modeled objects that support configurable net-instance association should follow the same assumption of belonging to DEFAULT without any explicit binding? Explicit references to DEFAULT ensure that different implementer assumptions about what DEFAULT owns by default do not affect the test result.

Also worth noting.. In many implementations there are checks done to ensure that subinterface references are within scope. For example, it would be common for an implementation to check that all the IS-IS interfaces of an IS-IS instance running in the DEFAULT net-instance actually belong to the DEFAULT network-instance. I believe there is some loss of clarity, for the implementer and operator, if the set of eligible subinterfaces in this case is the superset of unbound subinterfaces and subinterfaces explicitly bound to the DEFAULT network-instance.

internal/deviations/deviations.go
```diff
@@ -99,4 +110,21 @@ var (
 	StaticProtocolName = flag.String("deviation_static_protocol_name", "DEFAULT", "The name used for the static routing protocol.  The default name in OpenConfig is \"DEFAULT\" but some devices use other names.")
 
 	DeprecatedVlanID = flag.Bool("deviation_deprecated_vlan_id", false, "Device requires using the deprecated openconfig-vlan:vlan/config/vlan-id or openconfig-vlan:vlan/state/vlan-id leaves.")
+
+	AddSubIntfToNetInst = flag.Bool("deviation_adding_subintf_to_network_instance", true,
```
liulk, line 114: I don't think this should be a deviation.

If we agree that all tests must put subinterfaces into a VRF, someone should modify the tests to do that unconditionally.

By the way, my preferred name for the deviation would have been `--deviation_explicit_subinterface_in_vrf`.

internal/deviations/deviations.go
liulk: This code probably belongs best in https://github.com/openconfig/featureprofiles/blob/main/internal/attrs/attrs.go

Add a function `AddToDUT` that will both replace the interface and put it in a VRF. Add the VRF attribute - if empty, use *deviations.DefaultNetworkInstance.

internal/deviations/deviations.go
```diff
@@ -63,10 +63,21 @@
...
+	"testing"
+
+	"github.com/openconfig/ondatra"
+	"github.com/openconfig/ondatra/telemetry"
+	"github.com/openconfig/ygot/ygot"
+)
 
 // Vendor deviation flags.
 var (
+	NtpAssociationType = flag.Bool("ntp_association_type_enabled", false,
```
liulk, line 78: Please revert unrelated change?

liulk (changes requested):

internal/deviations/deviations.go
Hobbydos: If the decision is to add each subinterface to a specific VRF explicitly, then that works, I will remove this deviation. The initial issue we had was that there was an assumption where the (sub)interfaces would be in the `DEFAULT` VRF.

internal/deviations/deviations.go
Hobbydos: Ack'd - I'll move this as AddToDut in attrs.go - and If the NI is empty, then pass it to default network instance.

internal/deviations/deviations.go
robshakir: I think because we don't have this in all tests today, I'd like to go down the route of adding this to the deviations, and then subsequently the way that we can remove it is to determine whether:

a) the implementation here changes to take into account the difference, _or_
b) we explicitly specify the network instance for all interfaces in all deployments.

What I can do is write a specific test that covers interface to network instance assignment that we can use to determine whether we need this deviation in the medium term. I'll work on that now, but I would prefer that we move forward with this deviation for the moment.

cc: @xw-g @vcheny @vosipchu @sachendras for comment too.

mojiiba: +1 to a deviation. Interface assignment to vrfs differs for vendors and making explicit assignment mandatory (specially to default vrf) will make some devices to fail.

robshakir: I think that there are two things here:

1. Is the interface implicitly assigned to a default NI?
2. Can one configure explicitly that an interface is part of the default NI?

It seems that both of these options can coexist.

robshakir: I checked in #758 to validate this -- needs to be run against implementations.

liulk: After we validate #758 against other vendors, do we still need this as a deviation though?

Hobbydos: @robshakir, few things with #758 - I just ran a quick test.
1. The interface ID is not populated `i.Id = ygot.String(dp.Name() + "." + fmt.Sprint(p.Subintf))` which results in only the interface attempted to be assigned to the NI 
2. The update method exposes the issue that I was referring to [here](https://github.com/openconfig/featureprofiles/pull/737#discussion_r1016937740) when attempting to move an interface from one NI to another, it requires a delete (maybe Replace)?
```
[FailedPrecondition] Error in path: .network-instance{.name=="BLUE"}.interface{.name=="ethernet-1/19.0"}
            [FailedPrecondition] subinterface is already bound to .network-instance{.name=="DEFAULT"}.interface{.name=="ethernet-1/19.0"}
```

mojiiba: I agree both options can coexist. However, If a device supports option 1, then the second option will be a no-op configuration. I think a device should not be tagged as no-compliant for not supporting a no-op configuration.

Also as @Hobbydos mentioned, there is another point that need to be addressed in the [test](https://github.com/openconfig/featureprofiles/pull/758) that is related to moving an active interface from one vrf to another that may not allowed by all devices.

internal/deviations/deviations.go
```diff
@@ -99,4 +107,21 @@ var (
 	StaticProtocolName = flag.String("deviation_static_protocol_name", "DEFAULT", "The name used for the static routing protocol.  The default name in OpenConfig is \"DEFAULT\" but some devices use other names.")
 
 	DeprecatedVlanID = flag.Bool("deviation_deprecated_vlan_id", false, "Device requires using the deprecated openconfig-vlan:vlan/config/vlan-id or openconfig-vlan:vlan/state/vlan-id leaves.")
+
+	AddSubIntfToNetInst = flag.Bool("deviation_adding_subintf_to_network_instance", true,
+		"If a device requires explicit attachment of a sub-interface to a network-instance, this deviation should be marked true. The default expectation is that a sub-interface is part of default network-instance implicitly.")
```
liulk, line 112:
```suggestion
	ExplicitInterfaceInVRF = flag.Bool("deviation_explicit_interface_in_vrf", true,
		"Device requires explicit attachment of a subinterface to a network instance. OpenConfig expects an unattached subinterface to be implicitly part of the default network instance. Fully-compliant devices should pass with and without this deviation.")
```

internal/deviations/deviations.go
xw-g: Basically we are deciding between:
1. Doing something that's allowed by current OC spec, but possible brings extra requirements to other 3 vendors and our current NPIs.
2. Set a deviation for only one vendor (and take time and effort to decide on the long term solution later, e.g. clarify the OC spec and ask the vendor OS to adjust the behavior), which will not introduce extra requirements to neither the NPIs nor other vendors.

In this case, if 1) is a no-op to all vendors OS, I would go with it. However, we know for sure that 1) is not a no-op action (and it's even a challenge) to at least 2 vendors.

Therefore I think we should go with 2), which is to introduce a deviation. 

cc: @vcheny @vosipchu @sachendras

robshakir: > @robshakir, few things with #758 - I just ran a quick test.
> 
> 1. The interface ID is not populated `i.Id = ygot.String(dp.Name() + "." + fmt.Sprint(p.Subintf))` which results in only the interface attempted to be assigned to the NI

On this issue -- the `interface` and `subinterface` leaves are what specify what interface + subinterface should be included, rather than the name in the key. If this specific name is required, this sounds like something we should file a bug against.

> 2. The update method exposes the issue that I was referring to [here](https://github.com/openconfig/featureprofiles/pull/737#discussion_r1016937740) when attempting to move an interface from one NI to another, it requires a delete (maybe Replace)?
> 
> ```
> [FailedPrecondition] Error in path: .network-instance{.name=="BLUE"}.interface{.name=="ethernet-1/19.0"}
>             [FailedPrecondition] subinterface is already bound to .network-instance{.name=="DEFAULT"}.interface{.name=="ethernet-1/19.0"}
> ```

Thanks - yes, I'd missed this writing the initial test. I'll update it.

vcheny: +1 on 2). My interpretation from  DEFAULT_INSTANCE network_instance_type below is that all interfaces should be in default vrf if not specified. 

A special routing instance which acts as the 'default' or 'global' routing instance for a network device.

liulk: I don't want to be a pest, so if we want the tests to pass on a device that don't support VRF, they have a way to do it.

internal/deviations/deviations.go
```diff
@@ -108,8 +108,8 @@ var (
 
 	DeprecatedVlanID = flag.Bool("deviation_deprecated_vlan_id", false, "Device requires using the deprecated openconfig-vlan:vlan/config/vlan-id or openconfig-vlan:vlan/state/vlan-id leaves.")
 
-	AddSubIntfToNetInst = flag.Bool("deviation_adding_subintf_to_network_instance", true,
-		"If a device requires explicit attachment of a sub-interface to a network-instance, this deviation should be marked true. The default expectation is that a sub-interface is part of default network-instance implicitly.")
+	ExplicitInterfaceInVRF = flag.Bool("deviation_explicit_interface_in_vrf", true,
```
mojiiba, line 111: @liulk @Hobbydos we use false as the default value for other deviations. should we do the same for this one?

Hobbydos: I think you're right, @mojiiba - thoughts @liulk ?

internal/deviations/deviations.go
```diff
@@ -99,4 +107,21 @@ var (
 	StaticProtocolName = flag.String("deviation_static_protocol_name", "DEFAULT", "The name used for the static routing protocol.  The default name in OpenConfig is \"DEFAULT\" but some devices use other names.")
 
 	DeprecatedVlanID = flag.Bool("deviation_deprecated_vlan_id", false, "Device requires using the deprecated openconfig-vlan:vlan/config/vlan-id or openconfig-vlan:vlan/state/vlan-id leaves.")
+
+	ExplicitInterfaceInVRF = flag.Bool("deviation_explicit_interface_in_vrf", true,
```
liulk, line 111: That's correct, I somehow missed that one. Thanks for pointing that out!

```suggestion
	ExplicitInterfaceInVRF = flag.Bool("deviation_explicit_interface_in_vrf", false,
```

internal/deviations/deviations.go
liulk: Please move the functoin to attrs.go. If it's not clear what AddToDUT should do, it should functionally replace these lines in static_arp_test.go:

```
	i1 := &telemetry.Interface{Name: ygot.String(p1.Name())}
	d.Interface(p1.Name()).Replace(t,
		configInterfaceDUT(i1, &dutSrc, &ateSrc, peermac))

	if *deviations.AddSubIntfToNetInst {
		deviations.SubIntfToNetworkInstance(t, d, i1, 0, *deviations.DefaultNetworkInstance)
	}
```

liulk (changes requested):

internal/deviations/deviations.go
Hobbydos: @liulk - I'll move it over, I don't know if I'll have it ready before the end of the day today. I'll also need to run a few tests locally. I'll target End of Day Monday for a more realistic timeframe.

Hobbydos: @liulk, I finally got around to implementing/testing moving the function to AddToDUT. Changes have been pushed to this PR.

Hobbydos: Resolving conversation as this was moved to AddToDUT() in `attrs.go`

Hobbydos: Resolving Conversation as this was moved to AddToDUT() in `attrs.go`

feature/interface/staticarp/ate_tests/static_arp_test/static_arp_test.go
```diff
@@ -153,11 +153,12 @@ func configureDUT(t *testing.T, peermac string) {
 	i1 := &telemetry.Interface{Name: ygot.String(p1.Name())}
 	d.Interface(p1.Name()).Replace(t,
 		configInterfaceDUT(i1, &dutSrc, &ateSrc, peermac))
-
+	dutSrc.AddToDUT(t, d, i1, 0)
```
liulk, line 156: I was thinking that attrs.AddToDUT should just do the d.Interface().Replace() as well. It will take p1 as argument instead of i1. It should not need to take the si argument since it's okay to assume it will be 0.

internal/attrs/attrs.go
```diff
@@ -147,3 +150,24 @@ func (a *Attributes) AddToOTG(top gosnappi.Config, ap *ondatra.Port, peer *Attri
 		ip.SetAddress(a.IPv6).SetGateway(peer.IPv6).SetPrefix(int32(a.IPv6Len))
 	}
 }
+
+// AddToDut attaches a subinterface to a network instance
+func (a *Attributes) AddToDUT(t *testing.T, dconf *ondatra.Config, i *telemetry.Interface, si uint32) {
+	if *deviations.ExplicitInterfaceInVRF {
+		ni := a.NetInst
```
liulk, line 157: Use early return to reduce nesting?

```suggestion
	if !*deviations.ExplicitInterfaceInVRF {
		return
	}
	ni := a.NetInst
```

feature/interface/staticarp/ate_tests/static_arp_test/static_arp_test.go
Hobbydos: Sounds good! I'll make those changes and commit once verified/tested locally.

robshakir: Couple of comments here:

1. `AddToDUT` seems a strange name -- since it's not very clear what is actually happening -- can this be called `AssignPortsToNI`? 
2. @liulk - from a featureprofiles perspective, is the plan to transition everything to use attrs per @Hobbydos' question? It seems like this might need to be done before having a common pattern here can be implemented. Is there a LSC that we're already doing to implement that?

liulk: Name `AddToDUT` is chosen as an analogy to an existing method `AddToATE`. Perhaps not the best name, but I want to highlight the similarity of their purpose. We can change both to `SetOnDUT` and `SetOnATE` but that requires a LSC. I'm also okay to rename `AddToATE` to `SetOnATE` provided that a compatibility function `AddToATE` is retained and marked as deprecated so we remember to clean it up.

robshakir: 1. Ah, so this function is intended to do more than solely this  -- OK, thanks. I had not understood this. @Hobbydos please can we update the doc comment so that it doesn't specifically say it is going to do this.
2. Thanks -- so if we're going through these helpers, that SGTM. @Hobbydos - how many places did you find that need alternate implementations?

internal/attrs/attrs.go
```diff
@@ -147,3 +150,26 @@ func (a *Attributes) AddToOTG(top gosnappi.Config, ap *ondatra.Port, peer *Attri
...
+// AddToDut attaches a subinterface to a network instance
+func (a *Attributes) AddToDUT(t *testing.T, dconf *ondatra.Config, i *telemetry.Interface, si uint32) {
+	if !*deviations.ExplicitInterfaceInVRF {
+		return
+	}
+	ni := a.NetInst
+		if ni == "" {
+			ni = *deviations.DefaultNetworkInstance
+		}
+		netInst := &telemetry.NetworkInstance{Name: ygot.String(ni)}
```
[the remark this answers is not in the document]
liulk: We have attempted creating a full-config management system for functional tests but haven't been able to get it to work reliably. Pinged you with detail.

feature/interface/staticarp/ate_tests/static_arp_test/static_arp_test.go
Hobbydos: @robshakir; at least 25 - here's a quickly generated list.

```
grep -r "configInterfaceDUT(" | awk '{print $1}' | cut -d ":" -f1 | uniq
feature/experimental/gribi/ate_tests/dut_daemon_failure/dut_daemon_failure_test.go
feature/experimental/p4rt/ate_tests/google_discovery_protocol_packetin_test/google_discovery_protocol_packetin_test.go
feature/experimental/p4rt/ate_tests/google_discovery_protocol_packetout_test/google_discovery_protocol_packetout_test.go
feature/experimental/policy/policy_vrf_selection/ate_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
feature/experimental/policy/policy_vrf_selection/ate_tests/base_vrf_selection/base_vrf_selection_test.go
feature/experimental/policy/policy_vrf_selection/otg_tests/base_vrf_selection/base_vrf_selection_test.go
feature/experimental/policy/policy_vrf_selection/otg_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
feature/gribi/ate_tests/ack_in_presence_other_routes/route_ack_test.go
feature/gribi/ate_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go
feature/gribi/ate_tests/base_leader_election_test/base_leader_election_test.go
feature/gribi/ate_tests/get_rpc_test/get_rpc_test.go
feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
feature/gribi/ate_tests/leader_failover_test/leader_failover_test.go
feature/gribi/ate_tests/ordering_ack_test/ordering_ack_test.go
feature/gribi/ate_tests/supervisor_failure_test/supervisor_failure_test.go
feature/gribi/otg_tests/ack_in_presence_other_routes/route_ack_test.go
feature/gribi/otg_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go
feature/gribi/otg_tests/base_leader_election_test/base_leader_election_test.go
feature/gribi/otg_tests/get_rpc_test/get_rpc_test.go
feature/gribi/otg_tests/leader_failover_test/leader_failover_test.go
feature/gribi/otg_tests/ordering_ack_test/ordering_ack_test.go
feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
feature/interface/singleton/otg_tests/singleton_test/singleton_test.go
feature/interface/staticarp/ate_tests/static_arp_test/static_arp_test.go
feature/interface/staticarp/otg_tests/static_arp_test/static_arp_test.go
```

internal/attrs/attrs.go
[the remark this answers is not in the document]
robshakir: Thanks for the details -- SGTM. No need to block here :-)

robshakir:
@Hobbydos -- thanks for the changes here. It looks like your work email address isn't registered to your GitHub account, hence the CLA check failure.

Hobbydos:
> @Hobbydos -- thanks for the changes here. It looks like your work email address isn't registered to your GitHub account, hence the CLA check failure.

Sorry, I pushed the commit from the wrong Environment, I'll repush it from the correct one.

internal/attrs/attrs.go
```diff
@@ -147,3 +150,25 @@ func (a *Attributes) AddToOTG(top gosnappi.Config, ap *ondatra.Port, peer *Attri
 		ip.SetAddress(a.IPv6).SetGateway(peer.IPv6).SetPrefix(int32(a.IPv6Len))
 	}
 }
+
+// AddToDut attaches a subinterface to a network instance
+func (a *Attributes) SetOnDUT(t *testing.T, dconf *ondatra.Config, i *telemetry.Interface) {
```
liulk, line 155: I still have a few concerns with this idiom.

Originally, the use case I intended was that attrs.AddToDUT would be a complete solution for configuring an interface: take the Attributes, generate *telemetry.Interface, replace the interface on dut, and optionally add the interface to the NI. I want something functionally equivalent to configInterfaceDUT as a one stop shop for configuring the interface on the DUT and add the interface to the NI when needed.

But looking a little deeper, I realized that static_arp_test (and potentially others) still need to populate more interface leaves than what attrs has, and this is why they don't use attrs.NewInterface() but instead built their own *telemetry.Interface. It will be hard to incorporate test-specific leaves without adding unnecessary complexity to attrs, unless we refactor the tests to add the test-specific leaves separately after an interface has been configured with the common leaves.

It seems that applying the explicit interface in VRF deviation transparently to tests will require some refactoring of existing tests, so is that the reason you opted to simplify the problem by only adding the subinterface to the network instance?

If that's the approach you want to take, then I have a few more suggestions.

 1. Make this a standalone function, not a method of Attributes. Name it `AssignToNetworkInstance` so it more accurately reflects the functionality provided.
 2. It should take *ondatra.DUTDevice as argument, not *ondatra.Config.
 3. It should take the network instance name as an argument.
 4. It should take the interface name as a string, not *telemetry.Interface.
 5. It should take the subinterface index as an argument rather than hard-coded to 0.

Tests that want to put an interface to a non-default NI should be able to use this function as well, so at least for non-default NI, the behavior should not be conditional on deviations.ExplicitInterfaceInVRF. It would be more accurate to rename the deviation to deviations.ExplicitInterfaceInDefaultVRF so it is clear that an interface will always be put into a non-default VRF, but may be no-op if putting it in default VRF is optional.

liulk (changes requested):

internal/attrs/attrs.go
Hobbydos: > It seems that applying the explicit interface in VRF deviation transparently to tests will require some refactoring of existing tests, so is that the reason you opted to simplify the problem by only adding the subinterface to the network instance?

The original idea for this PR/Deviation was simply to assign an (sub)interface to a Network Instance. It seems like maybe the idea got muddled in various conversations. I think having a complete solution to configure the interface might be a little out of scope for this PR, perhaps something can be raised at a later time and the community can contribute.

> * Make this a standalone function, not a method of Attributes. Name it `AssignToNetworkInstance` so it more accurately reflects the functionality provided.
> * It should take *ondatra.DUTDevice as argument, not *ondatra.Config.
> * It should take the network instance name as an argument.
> * It should take the interface name as a string, not *telemetry.Interface.
> * It should take the subinterface index as an argument rather than hard-coded to 0.

With regards to your first point, which file should contain this function?
I previously had `si` as a uint32  argument passed to the function but you had mentioned that it would be safe to always assume `0` as the subinterface. Is this no longer the case and you'd rather I revert back to what I initially coded?

Once I have those clarifications, I'll go ahead and make the changes.

internal/deviations/deviations.go
```diff
@@ -102,6 +102,6 @@ var (
 
 	DeprecatedVlanID = flag.Bool("deviation_deprecated_vlan_id", false, "Device requires using the deprecated openconfig-vlan:vlan/config/vlan-id or openconfig-vlan:vlan/state/vlan-id leaves.")
 
-	ExplicitInterfaceInVRF = flag.Bool("deviation_explicit_interface_in_vrf", false,
-		"Device requires explicit attachment of an interface or subinterface to a network instance. OpenConfig expects an unattached interface or subinterface to be implicitly part of the default network instance. Fully-compliant devices should pass with and without this deviation.")
+	ExplicitInterfaceInDefaultVRF = flag.Bool("deviation_explicit_interface_in_def_vrf", false,
```
liulk, line 105:
```suggestion
	ExplicitInterfaceInDefaultVRF = flag.Bool("deviation_explicit_interface_in_default_vrf", false,
```

internal/fptest/networkinstance.go
```diff
@@ -0,0 +1,47 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/openconfig/featureprofiles/internal/deviations"
+	"github.com/openconfig/ondatra"
+	"github.com/openconfig/ondatra/telemetry"
+	"github.com/openconfig/ygot/ygot"
+)
+
+// AddToDut attaches a subinterface to a network instance
```
liulk, line 27:
```suggestion
// AssignToNetworkInstance attaches a subinterface to a network instance.
```

internal/fptest/networkinstance.go
```diff
@@ -0,0 +1,47 @@
...
+func AssignToNetworkInstance(t *testing.T, d *ondatra.DUTDevice, i string, ni string, si uint32) {
+	if ni == *deviations.DefaultNetworkInstance && !*deviations.ExplicitInterfaceInDefaultVRF {
+		return
+	}
+	if ni == "" && *deviations.ExplicitInterfaceInDefaultVRF {
+		ni = *deviations.DefaultNetworkInstance
+	}
```
liulk, line 34:
```suggestion
	if ni == "" {
		ni = *deviations.DefaultNetworkInstance
	}
	if ni == *deviations.DefaultNetworkInstance && !*deviations.ExplicitInterfaceInDefaultVRF {
		return
	}
```

liulk (commented):
Almost good to go. Few more suggested edits.

internal/attrs/attrs.go
```diff
@@ -37,6 +37,7 @@ type Attributes struct {
 	IPv4    string
 	IPv6    string
 	MAC     string
+	NetInst string
```
liulk, line 40:
```suggestion
```

liulk (commented):
Sorry, missed this one.

xw-g:
It would take some time to formally update the OC schema (see https://github.com/openconfig/featureprofiles/pull/737#pullrequestreview-1171312666) to explicitly specify the wanted behavior, and take time for a vendor to implement the changes, therefore I LGTM this deviation proposal. 

@vcheny @sachendras @vosipchu any concerns?

sachendras:
LGTM

internal/fptest/networkinstance.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/openconfig/featureprofiles/internal/deviations"
+	"github.com/openconfig/ondatra"
+	"github.com/openconfig/ondatra/telemetry"
+	"github.com/openconfig/ygot/ygot"
+)
+
+// AssignToNetworkInstance attaches a subinterface to a network instance.
+func AssignToNetworkInstance(t *testing.T, d *ondatra.DUTDevice, i string, ni string, si uint32) {
+	if ni == "" {
```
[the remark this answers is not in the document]
Hobbydos: Just a fatal message outlining that the network-instance provided to the function is empty would suffice?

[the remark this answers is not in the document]
xw-g: @Hobbydos  I think that would suffice.

vosipchu:
Xiao, LGTM

vcheny (approved):
lgtm

xw-g (approved):

PR #741, opened 2022-11-04
CODEOWNERS add owners for deviations

liulk:


liulk:
@xw-g: @dplore told me you have had some thoughts about this. Let me know what you think.

xw-g:
Let's discuss it offline first. I think it should include: @vcheny @vosipchu @sachendras @robshakir @xw-g

robshakir (changes requested):
I don't agree with this change. Please let's discuss this further. I'm marking this explicitly as request changes so that it blocks merging.

liulk:
Obsolete by #777

Issue #702, opened 2022-10-26
RT-2.1: Base IS-IS Process and Adjacencies


trathod1:
During execution of test `feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go`, it configures ISIS using `addISISOC` call in `feature/experimental/isis/ate_tests/internal/session/session.go`.

In this function, while configuring ISIS, it does not configure the `config.enabled` path for ISIS protocol. This is causing ISIS session to remain admin disabled.

In the below code, when `prot.Enabled = ygot.Bool(true)` is added, the ISIS session comes up fine. 
It should be added like in line-5 below.
```
1. // addISISOC configures basic IS-IS on a device.
2. func addISISOC(dev *oc.Root, areaAddress, sysID, ifaceName string) {
3.     inst := dev.GetOrCreateNetworkInstance(*deviations.DefaultNetworkInstance)
4.     prot := inst.GetOrCreateProtocol(PTISIS, ISISName)
5.     prot.Enabled = ygot.Bool(true)
6.     isis := prot.GetOrCreateIsis()
```

I see in other scripts for configuring static-route it is getting enabled at `feature/gribi/ate_tests/ack_in_presence_other_routes/route_ack_test.go` - 

```
// configStaticRoute configures a static route.
func configStaticRoute(t *testing.T, dut *ondatra.DUTDevice, prefix string, nexthop string) *telemetry.NetworkInstance_Protocol {
	ni1 := dut.Config().NetworkInstance(*deviations.DefaultNetworkInstance).Get(t)
	static := ni1.GetOrCreateProtocol(telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, *deviations.StaticProtocolName)
	static.Enabled = ygot.Bool(true)
	static.Identifier = telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC
```

sonhyoung:
@xw-g

PR #745, opened 2022-11-08
added prot enabled

Hobbydos:
Proposal for [issue 702](https://github.com/openconfig/featureprofiles/issues/702)

### DIFF
--- a/feature/experimental/isis/ate_tests/internal/session/session.go
+++ b/feature/experimental/isis/ate_tests/internal/session/session.go
@@ -105,6 +105,7 @@ func ProtocolPath() *networkinstance.NetworkInstance_ProtocolPath {
 func addISISOC(dev *oc.Root, areaAddress, sysID, ifaceName string) {
 	inst := dev.GetOrCreateNetworkInstance(*deviations.DefaultNetworkInstance)
 	prot := inst.GetOrCreateProtocol(PTISIS, ISISName)
+	prot.Enabled = ygot.Bool(true)
 	isis := prot.GetOrCreateIsis()
 	glob := isis.GetOrCreateGlobal()
 	glob.Instance = ygot.String(ISISName)
--- a/feature/experimental/isis/otg_tests/internal/session/session.go
+++ b/feature/experimental/isis/otg_tests/internal/session/session.go
@@ -105,6 +105,7 @@ func ProtocolPath() *networkinstance.NetworkInstance_ProtocolPath {
 func addISISOC(dev *oc.Root, areaAddress, sysID, ifaceName string) {
 	inst := dev.GetOrCreateNetworkInstance(*deviations.DefaultNetworkInstance)
 	prot := inst.GetOrCreateProtocol(PTISIS, ISISName)
+	prot.Enabled = ygot.Bool(true)
 	isis := prot.GetOrCreateIsis()
 	glob := isis.GetOrCreateGlobal()
 	glob.Instance = ygot.String(ISISName)


Hobbydos:
cc @robshakir  @xw-g

xw-g:
Because current OC schema does not specify the default value, also because we are doing the same thing for other protocols (like [static protocl](https://github.com/openconfig/featureprofiles/blob/06a95833b93b02dbfbe7dfc918d8124ceba89620/feature/gribi/ate_tests/ack_in_presence_other_routes/route_ack_test.go#L203)), I LGTM this change. 

cc: @vcheny @vosipchu @sachendras thoughts?

sachendras:
LGTM

mojiiba:
@xw-g @Hobbydos  can we manage this via a deviation flag? I was going to open a pull request to do the reverse of this pR was doing.

mojiiba:
Also, I noticed several tests configure static routes and there are slight differences between  them. Can we have a util function that unify this? I can open a pull request if you want.

mojiiba:
please ignore my latest comment

Hobbydos:
> @Hobbydos can you also help update related part in OTG? Thanks.

I'll work on doing OTG - Will just run a few tests for now.



> @xw-g @Hobbydos can we manage this via a deviation flag? I was going to open a pull request to do the reverse of what this pR is doing.

Is there any purpose for this being a deviation?

vcheny:
Does it make sense to make enabled leafs in the OC model as true by default? Would imagine enable leaf to be True when adding a protocol most of time.

mojiiba:
> Does it make sense to make enabled leafs in the OC model as true by default? Would imagine enable leaf to be True when adding a protocol most of time.

yes. I also think some of enabled leaves can be removed from the oc model since the intentions for the enabled leaves are not clear. One example is vrf/enabled that is not clear what device should do with it, especially when they ate set to false. There is no way for most vendors to define a vrf and the disable it. We have to note that the devices still have to provide CLI. Making OC model cli friendl very helpful to adaption of the oc.

mojiiba:
> > @Hobbydos can you also help update related part in OTG? Thanks.
> 
> I'll work on doing OTG - Will just run a few tests for now.
> 
> > @xw-g @Hobbydos can we manage this via a deviation flag? I was going to open a pull request to do the reverse of what this pR is doing.
> 
> Is there any purpose for this being a deviation?

setting enabled will make our device to fail the test. managing this via deviation will enable both of us to pass the test for now. Google team can decide to the desired behavior and we can dap

> > @Hobbydos can you also help update related part in OTG? Thanks.
> 
> I'll work on doing OTG - Will just run a few tests for now.
> 
> > @xw-g @Hobbydos can we manage this via a deviation flag? I was going to open a pull request to do the reverse of what this pR is doing.
> 
> Is there any purpose for this being a deviation?

this change will make our device fail. The flag will allow both of us to pass the tests for now. @xw-g and their team can decide on the desired behavior and we can comply accordingly.

xw-g:
@mojiiba which logic you are planning to reverse? [this one](https://github.com/openconfig/featureprofiles/blob/06a95833b93b02dbfbe7dfc918d8124ceba89620/feature/gribi/ate_tests/ack_in_presence_other_routes/route_ack_test.go#L203)?

To keep things moving, shall we do the following? 
1. Let this PR move forward and merge, because current OC schema does not set the value by default.
2. @xw-g to file an OC PR to set the `enabled` as true by default (might take some time through the OC community, or get rejected).
3. @mojiiba to work on a PR to introduce a deviation flag for `enabled`? 

thoughts? @mojiiba, @vosipchu?

mojiiba:
> @mojiiba which logic you are planning to reverse? [this one](https://github.com/openconfig/featureprofiles/blob/06a95833b93b02dbfbe7dfc918d8124ceba89620/feature/gribi/ate_tests/ack_in_presence_other_routes/route_ack_test.go#L203)?
> 
> To keep things moving, shall we do the following?
> 
>     1. Let this PR move forward and merge, because current OC schema does not set the value by default.
> 
>     2. @xw-g to file an OC PR to set the `enabled` as true by default (might take some time through the OC community, or get rejected).
> 
>     3. @mojiiba to work on a PR to introduce a deviation flag for `enabled`?
> 
> 
> thoughts? @mojiiba, @vcheny, @vosipchu, @sachendras ?

sounds good, thank you!

feature/experimental/isis/ate_tests/internal/session/session.go
```diff
@@ -105,6 +105,7 @@ func ProtocolPath() *networkinstance.NetworkInstance_ProtocolPath {
 func addISISOC(dev *oc.Root, areaAddress, sysID, ifaceName string) {
 	inst := dev.GetOrCreateNetworkInstance(*deviations.DefaultNetworkInstance)
 	prot := inst.GetOrCreateProtocol(PTISIS, ISISName)
+	prot.Enabled = ygot.Bool(true)
```
aredmon8551, line 108: Curious - why add this?  I don't think enabling/disabling individual routing protocols is widely supported.

aredmon8551: Looks like I'm wrong - IOS-like NOS's seem to support a 'shutdown' command under protocols.  Still curious about the utility of this leaf for this test.

xw-g: @aredmon8551 does this [comment](https://github.com/openconfig/featureprofiles/pull/745#issuecomment-1314620480) answer your question? We are not enforcing it. We want to keep things moving while seeking clarification at the source (the OC)

aredmon8551: Thanks!

Somewhat answers the question - though I'm missing the motivation for the addition - is this change (plus the addition of a deviation) needed to allow support for all NOS's?  Or is this for consistency?  I'm not aware of this being a mandatory leaf/config node in any implementation.  Perhaps I'm missing something.

robshakir:
> @mojiiba which logic you are planning to reverse? [this one](https://github.com/openconfig/featureprofiles/blob/06a95833b93b02dbfbe7dfc918d8124ceba89620/feature/gribi/ate_tests/ack_in_presence_other_routes/route_ack_test.go#L203)?
> 
> To keep things moving, shall we do the following?
> 
> 1. Let this PR move forward and merge, because current OC schema does not set the value by default.
> 2. @xw-g to file an OC PR to set the `enabled` as true by default (might take some time through the OC community, or get rejected).
> 3. @mojiiba to work on a PR to introduce a deviation flag for `enabled`?
> 
> thoughts? @mojiiba, @vcheny, @vosipchu, @sachendras ?

SGTM too.

feature/experimental/isis/ate_tests/internal/session/session.go
xw-g: The referenced comment basically suggests to push things forward via 2 tracks in parallel: 

1. Align with OC. The OC doesn't set default value, so we need to set the value. That's what this PR is about. Meanwhile, let's create a deviation for other vendors that does not support set it explicitly. 
3. I'm going to file a PR to OC suggesting set the default value true. 
4. If 2 merged, let's converge on what test behaviors we want. In that case, probably we will remove the requirement of explicitly set it true.

aredmon8551: Ack, makes sense.  Thanks!

xw-g:
@Hobbydos ping me again when you have OTG updated. Thanks.

Hobbydos:
@xw-g added commit for OTG change

xw-g (approved):

Issue #463, opened 2022-08-25
TE-5.1: config.name, config.identifier and static-route container is not configured by script


trathod1:
In "get_rpc_test.go" script, under "testIPv4LeaderActive" function, when it configures Static protocol with a static-route, following errors are observed during the run - 

>             [FailedPrecondition] Leafref destination '.network-instances.network-instance{.name=="DEFAULT"}.protocols.protocol{.identifier=="openconfig-policy-types:STATIC"&&.name=="static"}.config.identifier' is not set
>         Error in path: .network-instances.network-instance{.name=="DEFAULT"}.protocols.protocol{.identifier=="openconfig-policy-types:STATIC"&&.name=="static"}.name
>             [FailedPrecondition] Leafref destination '.network-instances.network-instance{.name=="DEFAULT"}.protocols.protocol{.identifier=="openconfig-policy-types:STATIC"&&.name=="static"}.config.name' is not set
>         Error in path: .network-instances.network-instance{.name=="DEFAULT"}.protocols.protocol{.identifier=="openconfig-policy-types:STATIC"&&.name=="static"}.static-routes
>             [FailedPrecondition] ../config/identifier = 'STATIC'

However, if I perform configuration in the script using below additional new code, I dont see above errors and it passes -

> 325         batch := args.dut.Config().NewBatch()
> 326         ni.Name().BatchUpdate(t, batch, "static")
> 327         ni.Identifier().BatchUpdate(t, batch, telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC)
> 328         ni.Static(staticCIDR).Prefix().BatchUpdate(t, batch, staticCIDR)
> 329         batch.Set(t)

Also, if these configurations are present in the DUT before running the script, the test passes.

Is these configs are required to be added in the script, or they are expected by script to be pre-confgured on the DUT ?
Please help investigate, incase I am missing something here.

PR #755, opened 2022-11-09
fix for issue 463 - TE5.1_get_rpc_test

Hobbydos:
Fix for [issue 463](https://github.com/openconfig/featureprofiles/issues/463) @robshakir  @xw-g


robshakir (approved):
This LGTM. It'd be good to add to these descriptions why we're changing things.

This looks to be a case where an `Update` implicitly creates something that doesn't exist but does not fill out required fields (e.g., `config/identifier` and `config/name` for a network instance).

@xw-g PTAL.

feature/gribi/ate_tests/get_rpc_test/get_rpc_test.go
```diff
@@ -317,13 +317,12 @@ func testIPv4LeaderActive(ctx context.Context, t *testing.T, args *testArgs) {
...
 	// are returned, with no entry returned for 198.51.100.192/64.
 	dc := args.dut.Config()
-	ni := dc.NetworkInstance(*deviations.DefaultNetworkInstance).
-		Protocol(telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, *deviations.StaticProtocolName)
-	static := &telemetry.NetworkInstance_Protocol_Static{
-		Prefix: ygot.String(staticCIDR),
-	}
-	static.GetOrCreateNextHop("0").NextHop = telemetry.UnionString(atePort2.IPv4)
-	ni.Static(staticCIDR).Replace(t, static)
+	ni := dc.NetworkInstance(*deviations.DefaultNetworkInstance).Get(t)
```
mojiiba, line 320: @robshakir @xw-g  we configure static route in several tests, but they all slightly differ. Should not we just add a util function to unify this? @Hobbydos this can be done as part of this pr, otherwise I can open a separate PR for to do this?

xw-g (changes requested):
This PR#755 is to address issue https://github.com/openconfig/featureprofiles/issues/463. I see there were 2 PRs went in after your reported issue#463. They are: 
* https://github.com/openconfig/featureprofiles/pull/600
* https://github.com/openconfig/featureprofiles/pull/692

Are you still seeing issues now with the HEAD from MAIN? If not, I prefer not to change existing way of configuring static route (unless there is some motivations)

Hobbydos:
> This PR#755 is to address issue #463. I see there were 2 PRs went in after your reported issue#463. They are:
> 
> * #600
> * #692
> 
> Are you still seeing issues now with the HEAD from MAIN? If not, I prefer not to change existing way of configuring static route (unless there is some motivations)

I tested and we're still failing the test with HEAD from MAIN. The reason being that the function call in the below block (L322-L324) of code doesn't set the name/identifier. https://github.com/openconfig/featureprofiles/blob/304564810ecfad13d6fd1b43df515655af084696/feature/gribi/ate_tests/get_rpc_test/get_rpc_test.go#L320-L326

Are there issues with using `GetOrCreateProtocol`?

xw-g:
@Hobbydos can you help solve the merge conflict? Thanks. And then we will do some presubmit check internally and then merge. Thanks!

Hobbydos:
@xw-g - Sorry I missed this conflict earlier. Thanks for reminding me. It should be fixed, now.

xw-g:
cc: @jjing1122

xw-g:
@Hobbydos reminder on updating OTG test :) Thanks.

xw-g:
@jjing1122 is helping running presubmit tests internally, and will drive the merge effort.

Hobbydos:
@xw-g  - just now committed the OTG test.

bstoll:
Can we do this without doing a Replace on all static routes?  I believe this will cause issues if static routes were used for inband management, etc.

Hobbydos:
> Can we do this without doing a Replace on all static routes? I believe this will cause issues if static routes were used for inband management, etc.

@bstoll, I think changing the following `gnmi.Replace()` to `gnmi.Update` would satisfy your request, correct? https://github.com/openconfig/featureprofiles/blob/a353643d9768b27912161d75e45258015410b248/feature/gribi/ate_tests/get_rpc_test/get_rpc_test.go#L326

If so, I'll commit the change.

bstoll:
Yes, I think using Update is all that is needed.

jjing1122 (approved):

xw-g (approved):

PR #762, opened 2022-11-11
Deviation style fix for NTP association type

liulk:


liulk:
This fixes style issues introduced in #681. FYI: @vcheny @vosipchu

xw-g (approved):

PR #778, opened 2022-11-14
Migrate bgp and gnmi features to new telemetry

DanG100:


feature/gnmi/otg_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
```diff
@@ -421,13 +425,11 @@ func waitOTGARPEntry(t *testing.T) {
...
-	otg.Telemetry().InterfaceAny().Ipv6NeighborAny().LinkLayerAddress().Watch(
-		t, time.Minute, func(val *otgtelemetry.QualifiedString) bool {
-			return val.IsPresent()
-		}).Await(t)
+	gnmi.WatchAll(t, otg, gnmi.OTG().InterfaceAny().Ipv4NeighborAny().LinkLayerAddress().State(), time.Minute, func(val *ygnmi.Value[string]) bool {
+		return val.IsPresent()
+	}).Await(t)
+	gnmi.WatchAll(t, otg, gnmi.OTG().InterfaceAny().Ipv6NeighborAny().LinkLayerAddress().State(), time.Minute, func(val *ygnmi.Value[string]) bool {
+		return val.IsPresent()
+	}).Await(t)
```
dan-lepage, line 433: Hmm. This looks wrong, because nothing is actually checking the returned `ok` boolean, but it looks like that was true even before this CL so we can fix that in as a separate issue.

dan-lepage (approved):

PR #779, opened 2022-11-14
Migrate gnmi, gribi, and interfaces

DanG100:


feature/gribi/otg_tests/base_leader_election_test/base_leader_election_test.go
```diff
@@ -174,10 +175,9 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) gosnappi.Config {
 
 func waitOTGARPEntry(t *testing.T) {
 	ate := ondatra.ATE(t, "ate")
-	ate.OTG().Telemetry().InterfaceAny().Ipv4NeighborAny().LinkLayerAddress().Watch(
-		t, time.Minute, func(val *otgtelemetry.QualifiedString) bool {
-			return val.IsPresent()
-		}).Await(t)
+	gnmi.WatchAll(t, ate.OTG(), gnmi.OTG().InterfaceAny().Ipv4NeighborAny().LinkLayerAddress().State(), time.Minute, func(val *ygnmi.Value[string]) bool {
+		return val.IsPresent()
+	}).Await(t)
```
dan-lepage, line 180: Does this actually do anything? If no value is present it'll timeout and return false but nothing reacts to it. Was this true of the test already?

feature/gribi/otg_tests/ipv4_entry_test/ipv4_entry_test.go
```diff
@@ -478,18 +479,17 @@ func awaitTimeout(ctx context.Context, c *fluent.GRIBIClient, t testing.TB, time
 // Waits for at least one ARP entry on the tx OTG interface
 func waitOTGARPEntry(t *testing.T) {
 	ate := ondatra.ATE(t, "ate")
-	ate.OTG().Telemetry().Interface(atePort1.Name+".Eth").Ipv4NeighborAny().LinkLayerAddress().Watch(
-		t, time.Minute, func(val *otgtelemetry.QualifiedString) bool {
-			return val.IsPresent()
-		}).Await(t)
+	gnmi.WatchAll(t, ate.OTG(), gnmi.OTG().Interface(atePort1.Name+".Eth").Ipv4NeighborAny().LinkLayerAddress().State(), time.Minute, func(val *ygnmi.Value[string]) bool {
+		return val.IsPresent()
+	}).Await(t)
```
dan-lepage, line 484: Same question as before - in what situations is this anything other than a no-op, and if it really doesn't do anything, was that already the case?

feature/interface/singleton/otg_tests/singleton_test/singleton_test.go
```diff
@@ -26,10 +26,12 @@ import (
 	"github.com/openconfig/featureprofiles/internal/fptest"
 	"github.com/openconfig/featureprofiles/internal/otgutils"
 	"github.com/openconfig/ondatra"
+	"github.com/openconfig/ygnmi/ygnmi"
 	"github.com/openconfig/ygot/ygot"
```
dan-lepage, line 31: Any idea why goimports leaves empty lines in places like this?

dan-lepage (changes requested):
There's a recurring pattern where we call Await() or Watch() without doing anything with the value; that looks like a mistake to me, but it's also not clear if this PR actually changes the behavior or if it was already not checking these properly. Can you clarify whether the behavior is actually changing at all?

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -336,22 +337,23 @@ func flapinterface(t *testing.T, ate *ondatra.ATEDevice, port string, action boo
...
-	aftPfxNHG := dut.Telemetry().NetworkInstance(*deviations.DefaultNetworkInstance).Afts().Ipv4Entry(prefix).NextHopGroup()
-	aftPfxNHGVal, found := aftPfxNHG.Watch(t, 10*time.Second, func(val *telemetry.QualifiedUint64) bool {
+	aftPfxNHG := gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance).Afts().Ipv4Entry(prefix).NextHopGroup()
+	aftPfxNHGVal, found := gnmi.Watch(t, dut, aftPfxNHG.State(), 10*time.Second, func(val *ygnmi.Value[uint64]) bool {
 		return true
 	}).Await(t)
 	if !found {
 		t.Fatalf("Could not find prefix %s in telemetry AFT", dstPfx)
 	}
+	nhg, _ := aftPfxNHGVal.Val()
```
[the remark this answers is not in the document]
DanG100: i haven't been modifying the tests behavior, but yeah this looks incorrect.

feature/gribi/otg_tests/base_leader_election_test/base_leader_election_test.go
DanG100: that's correct. the semantics of watch haven't changed, so that what the test was doing previously.

DanG100:
I've fixed the issue you've called out, in this PR. But generally, i have not been checking the tests for issues like this.

feature/interface/singleton/otg_tests/singleton_test/singleton_test.go
DanG100: empty lines between imports are generally used to group imports, i don't think a particular grouping is enforced by goimports, except that imports are sorted in a group

feature/gribi/otg_tests/ipv4_entry_test/ipv4_entry_test.go
DanG100: fixed as well, yeah it waits, but doesn't verify that predicate was true

dan-lepage (approved):
We can go ahead and merge this in, then. I will create a separate issue to track that we have tests that aren't checking the results of some Await() calls.

PR #788, opened 2022-11-16
ADMIN PR: remove -v flag and rename kne_cli to kne

greg-dennis:
Two small updates to README.
1. Passing "-v" to go test is a now a noop for an Ondatra test, as Ondatra forces go test verbose mode on for every run
2. KNE renamed "kne_cli" to just "kne"

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,25 +34,25 @@ images can be obtained by contacting Arista.
 Setup
 
 ```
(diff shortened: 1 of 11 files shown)


greg-dennis:
Subsumed by https://github.com/openconfig/featureprofiles/pull/791

PR #794, opened 2022-11-17
Update operation changed to Repalce for TE-3.6

azubaidhar:
With current Update operation,  netconf threw error on incoming packet, where as Replace operation works. this is the only change done.

Thanks,
Abdul Zubaidhar .A


Ankur19:
@azubaidhar Also want this change in `otg_tests/.../route_ack_test.go`

jjing1122:
@xw-g

azubaidhar:
gnmi.update() works.  hence no change needed in the script for now. closing the request,

PR #798, opened 2022-11-17
collect rundata from DUT

liulk:


internal/rundata/dut.go
```diff
@@ -0,0 +1,251 @@
...
+
+	const chassisType = oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_CHASSIS
+
+	chassisIDs, err := y.FindByType(ctx, chassisType)
+	if err != nil {
+		glog.Errorf("Could not find chassis: %v", err)
+		return
+	}
+
+	chassisID := chassisIDs[0]
```
Ankur19, line 58: How confident are we that the first component in the list is the correct one?

internal/rundata/dut.go
```diff
@@ -0,0 +1,251 @@
...
+		glog.Errorf("Missing component software-version: %v", err)
+	} else {
+		di.osver = softVer
+		glog.V(2).Infof("Setting osver from operating-system software-version: %s", di.osver)
+	}
+}
+
+// setFromLLDP sets dutInfo vendor from LLDP system-description (Juniper).  This is less
+// reliable because LLDP config can be changed.
+func (di *dutInfo) setFromLLDP(ctx context.Context, y components.Y) {
```
Ankur19, line 125: Should we add `Juniper` in the function name as well? The function name looks like a general use function but logic seems very specific to Juniper.

internal/rundata/dut.go
```diff
@@ -0,0 +1,251 @@
...
+
+// dutsInfo populates the DUT properties for all DUTs in the reservation.
+func dutsInfo(ctx context.Context, m map[string]string, resv *binding.Reservation) {
+	for id, dut := range resv.DUTs {
+		gnmic, err := dut.DialGNMI(ctx)
+		if err != nil {
+			glog.Errorf("Could not dial GNMI to dut %s: %v", dut.Name(), err)
+			continue
+		}
+		yc, err := ygnmi.NewClient(gnmic)
```
Ankur19, line 244: Probably better to add two functions in components`New()` and `NewWithT()` where `New()` returns `(Y, error)`?

Ankur19 (approved):
Logic looks good!

internal/rundata/dut.go
liulk: There is typically just one chassis, but any chassis will work. Does not have to be the first one.

internal/rundata/dut.go
liulk: It's supposed to be a general use function. There could be another vendor that needs the LLDP fallback.

internal/rundata/dut.go
liulk: The T version doesn't need any struct wrapper since they can just use *ondatra.DUTDevice. Also, the T version is still used by existing tests, so I don't want to have to change them in this PR.

liulk:
To the team: since I'm OOO next week, if the infrastructure change breaks the static binding, just revert the changes in this file to the previous version: `topologies/binding/binding.go`

xw-g:
> To the team: since I'm OOO next week, if the infrastructure change breaks the static binding, just revert the changes in this file to the previous version: `topologies/binding/binding.go`

Because it's an infra improvement that does not change any test logic/behaviors, I don't think we need Quattro team approval. If test infra broke due to this PR,  @liulk in this thread mentioned the steps to recovered (rollback topologies/binding/binding.go)

PR #807, opened 2022-11-21
TE-6.2: Route Removal In Non Default VRF : Commented out EnabledAddre…

avaneesh90:
TE-6.2: Route Removal In Non Default VRF : Commented out EnabledAddressFamilies variable .
Reference : https://github.com/openconfig/public/pull/738


avaneesh90:
Pull request closed refer to : https://github.com/openconfig/featureprofiles/pull/856

PR #817, opened 2022-11-28
rename wbb to p4rtutils and move to internal/

Ankur19:
Find a better place for the p4rtutils


liulk (changes requested):
Should this be internal to the tests? The p4rtutils helper is just a bunch of hardcoded values for testing.

Move p4rtutils.go to feature/experimental/p4rt/internal/p4rtutils?

liulk (approved):

PR #827, opened 2022-12-02
Add L3 9236B to MTU test.

vcheny:


xw-g:
LGTM, can you also help update the OTG tests in this same PR? thanks.  https://github.com/openconfig/featureprofiles/tree/2a5a14e075d5c09dd830d29e6ecc14ab1ed558af/feature/interface/singleton/otg_tests/singleton_test

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
```diff
@@ -480,7 +480,7 @@ func TestMTUs(t *testing.T) {
 	ate := ondatra.ATE(t, "ate")
 
 	// These are the L3 MTUs, i.e. the payload portion of an Ethernet frame.
-	mtus := []uint16{1500, 5000, 9212}
+	mtus := []uint16{1500, 5000, 9212, 9236}
```
liulk, line 483: Be advised this is the L3 MTU, i.e. the payload portion.

I didn't find any other vendor mentioning 9236 except for Arista, and this link suggests that 9236 is the L2 MTU which includes the Ethernet frame.

https://www.arista.com/en/um-eos/eos-data-transfer

If anything, according to the calculation in the link, the L3 MTU should just be 9214 (and remove 9212), but other vendors have different max MTU sizes. I would advise checking with other vendors to make sure what the max MTU should be, and create a function in this test like this:

```
func maxMTU(vendor ondatra.Vendor) uint16 {
  switch vendor {
  case ondatra.ARISTA:
    return 9214
  case ondatra.CISCO:
    return 9216
  case ondatra.JUNIPER:
    return 9192
  default:
    return 9212
  }
}
```

liulk (changes requested):
Please also update the otg_tests with the same.

https://github.com/openconfig/featureprofiles/tree/main/feature/interface/singleton/otg_tests/singleton_test

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
aredmon8551: LGTM, though supported MTUs will vary not just by vendor but also by platform/silicon.  This should work in the immediate term though would suggest we look at this more as a platform variable and perhaps a runtime argument.

liulk: Agreed about max MTU depends on hardware. After some discussion, I suggested adding a `--deviation_max_mtu` with the default to 9236. I also suggested removing 9212 and just use this deviation as the max MTU test case. That's because each MTU with all the traffic subcases take about 14-17 minutes each, and this test is already taking very long.

In other words:

```
	mtus := []uint16{1500, 5000, *deviations.MaxMTU}
```

vcheny: 9236 MTU test on Arista passed. And all three other vendors can support MTU higher than 9236 according to vendor TLs. So, we don't need a deviation for it.
Just replaced 9212 w/ 9236 and updated otg tests.

vcheny: @vosipchu @sachendras for lgtm

liulk (approved):

feature/interface/singleton/ate_tests/singleton_test/singleton_test.go
sachendras: lgtm!

PR #830, opened 2022-12-02
P4RT-3.1: ExplicitInterfaceInDefaultVRF deviation for Google Discovery Protocol: PacketIn

trathod1:
Adding AssignToNetworkInstance call for deviation ExplicitInterfaceInDefaultVRF  added via PR-737.


trathod1:
@xw-g @robshakir

xw-g:
@Hobbydos @trathod1 I think this should be gated by the new introduced deviation flag ExplicitInterfaceInDefaultVRF, right? Not all OS implementation need to add interface explicitly.

liulk:
> @Hobbydos @trathod1 I think this should be gated by the new introduced deviation flag ExplicitInterfaceInDefaultVRF, right? Not all OS implementation need to add interface explicitly.

The gating is done by the AssignToNetworkInstance() function.

https://github.com/openconfig/featureprofiles/blob/851019f36ed18226842626018401d1478f006d6d/internal/fptest/networkinstance.go#L33

xw-g:
Thanks for pointing it out. @liulk . Then my comment becomes a readability discussion. Because most vendor OS does not require explicitly putting the interface in the default NI, I feel like it might be easier for readers to read if we gate the logic (explicitly putting the interfaces in the default NI) outside of the AssignToNetworkInstance() func. 

The test would be like: 
```
	gnmi.Replace(t, dut, d.Interface(p2.Name()).Config(), configInterfaceDUT(i2, &dutPort2))
        if ExplicitInterfaceInDefaultVRF is set
	    fptest.AssignToNetworkInstance(t, dut, p2.Name(), *deviations.DefaultNetworkInstance, 0)
```
This also means removing `if ni == *deviations.DefaultNetworkInstance && !*deviations.ExplicitInterfaceInDefaultVRF ` from AssignToNetworkInstance()

Thoughts? @liulk 

FYI, internal b/260928639 is tracking suggestions to OC schema. We might want to mention it in the OC schema that interfaces should be by default added to the default NI.

liulk:
@xw-g I agree that at the very least, let's move the gating logic out of AssignToNetworkInstance into the test so it's clear that it's a deviated behavior.

I was also mulling over another alternative which is more holistic but may be more involved.

The idea is to add a function `ConfigureDefaultNetworkInstance()` which at minimum ensures that the default NI exists (they may not be on KNE or some vendors). Instead of having the test add each interface one by one, tests would always call `ConfigureDefaultNetworkInstance()` unconditionally, which would also enumerate the interfaces and only add them to default VRF if `*deviations.ExplicitInterfaceInDefaultVRF` is true.

We can still keep AssignToNetworkInstance() but make it unconditional, so tests can use it to add an interface to default or non-default VRF.

xw-g:
@liulk yes, your alternative sounds more holistic. Shall keep it in our back pocket and see how b/260928639 goes? If b/260928639 ends up with updating/clarifying OC schema that every interface has to be by default in the default NI, then we can align vendors behind the clarification, and does not need the `ConfigureDefaultNetworkInstance()`. If b/260928639 ends up differently, then let's consider your `ConfigureDefaultNetworkInstance()`. WDYT?

xw-g:
I think we can merge it upon 1 LGTM (me) + 1 LGTM (@liulk ), because:
* This PR is an usage of the newly introduced deviation flag `ExplicitInterfaceInDefaultVRF`. Changes in this PR is gated by the flag, therefore it does not affect existing test cases.
* This PR does not change the agreed logic behind `ExplicitInterfaceInDefaultVRF`. 
 
I give my LGTM here. @liulk can you help merge it if you agree? Thanks.

liulk:
Since we changed the contract for fptest.AssignToNetworkInstance, we also need to update the existing usage.

https://github.com/openconfig/featureprofiles/blob/9dc16e7c4cf9bdc421d9df2161ef59768c4c3af0/feature/interface/staticarp/ate_tests/static_arp_test/static_arp_test.go#L156

By the way, I noticed that the otg_tests version of static_arp_test is missing the AssignToNetworkInstance call. We should add it soon. In the future, if a test has both ate_tests and otg_tests versions, both need to be updated.

https://github.com/openconfig/featureprofiles/blob/9dc16e7c4cf9bdc421d9df2161ef59768c4c3af0/feature/interface/staticarp/otg_tests/static_arp_test/static_arp_test.go#L158

You can either update static_arp_test in this PR or in a separate one, but if in a separate PR, I want to merge this PR and that one back to back in order to avoid breakage.

trathod1:
> 

Updating static_arp_test along with this PR sounds good to avoid interim breakage. Let me update ate_tests version and also add this deviation changes to otg_tests version of static_arp_test, which this PR only.

liulk (approved):

Issue #770, opened 2022-11-12
gNOI-4.1: Software Upgrade does not wait for Supervisor Sync post reboot


Hobbydos:
**Describe the bug**
After the [rebootDUT function](https://github.com/openconfig/featureprofiles/blob/main/feature/gnoi/os/tests/osinstall/osinstall_test.go#L95), there should be some mechanism to wait for the Supervisor synchronization to complete before attempting verifyInstall(). This results in the DUT returning an empty string as the OS.Version for the standby supervisor.

As a brute-force method, I simply added a `time.Sleep(5 * time.Minute)` after the reboot and the test passed successfully.

**To Reproduce**
1. Setup gNOI server on the DUT
2. Run OSInstall script on dual supervisor DUT
3. Observe Got/Want in output where got is `""`

**Expected behavior**
Wait for the Standby Supervisor to be up and synchronized before attempting to fetch version 

**Additional context**
Once added, the test passed:

```
    osinstall_test.go:329: Transfer progress: 1185873920 bytes received by DUT
    osinstall_test.go:248: OS.Install supervisor image transfer complete.
    osinstall_test.go:116: OS.Activate complete.
    osinstall_test.go:214: DUT standby supervisor has valid preexisting image; skipping transfer.
    osinstall_test.go:114: OS.Activate standby supervisor complete.
    osinstall_test.go:153: Send DUT Reboot Request
    osinstall_test.go:174: DUT has not rebooted.
    osinstall_test.go:180: Waiting for reboot to complete...
    osinstall_test.go:174: DUT has not rebooted.
    osinstall_test.go:180: Waiting for reboot to complete...
    osinstall_test.go:180: Waiting for reboot to complete...
    osinstall_test.go:171: Reboot completed.
    osinstall_test.go:96: Waiting for Supervisors to Synchronize - 5 Minutes <----- t.Logf added wtih the 5 minute timer after reboot
    osinstall_test.go:277: OS.Verify complete
--- PASS: TestOSInstall (1046.85s)
PASS
```

Hobbydos:
cc @robshakir  @xw-g

xw-g:
I see that  [rebootDUT](https://github.com/openconfig/featureprofiles/blob/781937890f6188520cfefb61e197283309c60141/feature/gnoi/os/tests/osinstall/osinstall_test.go#L148
) already adds 10m in the waiting. If you want brutal force, maybe just add there. 

[...]

PR #835, opened 2022-12-02
add stndby supervisor probing loop

Hobbydos:
Added a function to loop over the standby supervisor status to ensure it's in an Available state before attempting to verifyInstall. The function will error out if the standby supervisor is not available before 15 minutes (probes status every 1 minute). This function will also handle NON_EXISTENT while the standby supervisor is rebooting. This is a candidate for https://github.com/openconfig/featureprofiles/issues/770

### DIFF
--- a/feature/gnoi/os/tests/osinstall/osinstall_test.go
+++ b/feature/gnoi/os/tests/osinstall/osinstall_test.go
@@ -94,6 +94,9 @@ func TestOSInstall(t *testing.T) {
 		tc.activateOS(ctx, t, true)
 	}
 	tc.rebootDUT(ctx, t)
+	if tc.dualSup {
+		tc.isSupervisorReady(ctx, t)
+	}
 	tc.verifyInstall(ctx, t)
 }
 
@@ -252,6 +255,33 @@ func (tc *testCase) transferOS(ctx context.Context, t *testing.T, standby bool) 
 	} else {
 		t.Log("OS.Install supervisor image transfer complete.")
 	}
+}
+
+func (tc *testCase) isSupervisorReady(ctx context.Context, t *testing.T) error {
+	var counter int8
+	const maxRetries = 15
+	for i := int8(1); i < maxRetries; i++ {
+		if counter == maxRetries {
+			return fmt.Errorf("OS.Verify RPC did not report the standby supervisor state as available")
+		}
+		counter++
+		r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})
+		if err != nil {
+			t.Fatal(err)
+		}
+		switch v := r.GetVerifyStandby().GetState().(type) {
+		case *ospb.VerifyStandby_StandbyState:
+			switch v.StandbyState.GetState() {
+			case ospb.StandbyState_UNSPECIFIED, ospb.StandbyState_UNSUPORTED:
+				t.Fatalf("OS.Verify RPC reported that the standby supervisor is %v", v.StandbyState.GetState())
+			case ospb.StandbyState_NON_EXISTENT, ospb.StandbyState_UNAVAILABLE:
+				t.Logf("OS.Verify RPC reported that the standby supervisor is %v, waiting 60s and retrying: %d/15", v.StandbyState.GetState(), counter)
+				time.Sleep(1 * time.Minute)
+			}
+		}
+	}
+	t.Logf("OS.Verify RPC Reported that the standby supervisor is AVAILABLE")
+	return nil
 }
 
 // verifyInstall validates the OS.Verify RPC returns no failures and version numbers match the


Hobbydos:
cc @xw-g  @robshakir

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -94,6 +94,9 @@ func TestOSInstall(t *testing.T) {
 		tc.activateOS(ctx, t, true)
 	}
 	tc.rebootDUT(ctx, t)
+	if tc.dualSup {
```
xw-g, line 97: tc.rebootDUT already waited 10min, and we are adding possible another 15min on top of that. Any optimization we can do here?

Also, we only improved the dualSup case, what about single Sup case? Any thing we can optimize there too? 

I feel like that either we brutal force update the rebootDUT to wait 5 more min, or we do something smarter for **both** dualSup and singleSup cases.

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -254,6 +257,33 @@ func (tc *testCase) transferOS(ctx context.Context, t *testing.T, standby bool)
 	}
 }
 
+func (tc *testCase) isSupervisorReady(ctx context.Context, t *testing.T) error {
+	var counter int8
+	const maxRetries = 15
+	for i := int8(1); i < maxRetries; i++ {
+		if counter == maxRetries {
+			return fmt.Errorf("OS.Verify RPC did not report the standby supervisor state as available")
```
xw-g, line 265: t.Fatal instead of fmt.Errorf?

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -254,6 +257,33 @@ func (tc *testCase) transferOS(ctx context.Context, t *testing.T, standby bool)
 	}
 }
 
+func (tc *testCase) isSupervisorReady(ctx context.Context, t *testing.T) error {
+	var counter int8
+	const maxRetries = 15
+	for i := int8(1); i < maxRetries; i++ {
```
xw-g, line 263: do we need to specifically call out int8? your `maxRetries` variable didn't call out type anyway.

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -254,6 +257,33 @@ func (tc *testCase) transferOS(ctx context.Context, t *testing.T, standby bool)
...
+		if counter == maxRetries {
+			return fmt.Errorf("OS.Verify RPC did not report the standby supervisor state as available")
+		}
+		counter++
+		r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})
+		if err != nil {
+			t.Fatal(err)
+		}
+		switch v := r.GetVerifyStandby().GetState().(type) {
+		case *ospb.VerifyStandby_StandbyState:
```
xw-g, line 273: you only have one case here, do you still need the switch?

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -254,6 +257,33 @@ func (tc *testCase) transferOS(ctx context.Context, t *testing.T, standby bool)
...
+			switch v.StandbyState.GetState() {
+			case ospb.StandbyState_UNSPECIFIED, ospb.StandbyState_UNSUPORTED:
+				t.Fatalf("OS.Verify RPC reported that the standby supervisor is %v", v.StandbyState.GetState())
```
xw-g, line 276: I think you want to time.Sleep 1min here too, right?

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: tc.rebootDUT waits for the chassis to be rebooted, but there seemed to be issues when the standby supervisor was rebooting, there was no case to handle certain states in those

I haven't fully tested this against singleSup devices, yet. I think this method is more comprehensible for dualSup than simply adding a 'X' Minute timer to rebootDUT, no?

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: No, if its unspecified or unsupported, I don't see a reason to sleep?

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: I suppose I could make this function not return anything and simply return t.Fatal, is that the standard for these functions in this testcase? If so, I'll make the change

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: No, we don't actually need to call out int8 here, I'll change everything to int.

feature/gnoi/os/tests/osinstall/osinstall_test.go
xw-g: my bad. I was looking at the wrong code.

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: I thought it increased readability. What suggestions might you have to replace the switch? If require, I could change to if/else, but that might be a little more complex to read.

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -270,15 +270,28 @@ func (tc *testCase) verifyInstall(ctx context.Context, t *testing.T) {
...
+		const maxRetries = 15
+		for i := 1; i < maxRetries; i++ {
+			r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})
+			if err != nil {
+				t.Fatalf("OS.Verify request failed: %s", err)
+			}
+			if counter == maxRetries {
+				t.Fatalf("OS.Verify Standby could not obtain version status in 15 minutes")
+			}
+			counter++
```
xw-g, line 283: `counters` basically is `i`, right? seems like we don't need this extra `counter` variable?

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -270,15 +270,28 @@ func (tc *testCase) verifyInstall(ctx context.Context, t *testing.T) {
...
-		}
+		var counter int
+		const maxRetries = 15
+		for i := 1; i < maxRetries; i++ {
+			r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})
+			if err != nil {
+				t.Fatalf("OS.Verify request failed: %s", err)
+			}
+			if counter == maxRetries {
+				t.Fatalf("OS.Verify Standby could not obtain version status in 15 minutes")
```
xw-g, line 281: `15 minutes` is a variable tie to the `maxRetries`, instead of a const string.

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -270,15 +270,28 @@ func (tc *testCase) verifyInstall(ctx context.Context, t *testing.T) {
 	}
 
 	if tc.dualSup {
-		if got, want := r.GetVerifyStandby().GetVerifyResponse().GetActivationFailMessage(), ""; got != want {
-			t.Errorf("OS.Verify Standby ActivationFailMessage: got %q, want %q", got, want)
-		}
+		var counter int
+		const maxRetries = 15
+		for i := 1; i < maxRetries; i++ {
+			r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})
```
xw-g, line 276: `r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})` becomes a duplicated code as above. maybe consider refactoring it to a quick helper func.

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -270,15 +270,28 @@ func (tc *testCase) verifyInstall(ctx context.Context, t *testing.T) {
 	}
 
 	if tc.dualSup {
-		if got, want := r.GetVerifyStandby().GetVerifyResponse().GetActivationFailMessage(), ""; got != want {
-			t.Errorf("OS.Verify Standby ActivationFailMessage: got %q, want %q", got, want)
-		}
+		var counter int
+		const maxRetries = 15
+		for i := 1; i < maxRetries; i++ {
```
xw-g, line 275: maybe consider moving the `tc.osc.Verify` at the end of the for loop scope. Because you already ran it once above, and you already have some result in the `r` for comparison. Only if it does not match or produces nothing, then you retry it at the end, right?

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -270,15 +270,28 @@ func (tc *testCase) verifyInstall(ctx context.Context, t *testing.T) {
...
 
-		if got, want := r.GetVerifyStandby().GetVerifyResponse().GetVersion(), *osVersion; got != want {
-			t.Errorf("OS.Verify Standby Version: got %q, want %q", got, want)
+			if got, want := r.GetVerifyStandby().GetVerifyResponse().GetActivationFailMessage(), ""; got != want {
+				t.Errorf("OS.Verify Standby ActivationFailMessage: got %q, want %q", got, want)
+			}
+
+			if got, want := r.GetVerifyStandby().GetVerifyResponse().GetVersion(), *osVersion; got != want {
+				t.Logf("Attempt %d/15 - OS.Verify Standby Version: got %q, want %q", counter, got, want)
+				time.Sleep(1 * time.Minute)
```
xw-g, line 291: @liulk is it `time.Sleep() + for loop` the recommended way in FP test to achieve the timeout logic? Or, should we use something else e.g. compare the timeStamp? Thanks.

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: Sure, I'll remove `counter` and keep track of `i`, instead.
I'll update that in next batch commit.

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: that's right, added to next batch commit.

liulk, line 281:
```suggestion
				t.Fatalf("OS.Verify Standby could not obtain version status in %d retries", maxRetries)
```

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -270,15 +270,28 @@ func (tc *testCase) verifyInstall(ctx context.Context, t *testing.T) {
...
-			t.Errorf("OS.Verify Standby ActivationFailMessage: got %q, want %q", got, want)
-		}
+		var counter int
+		const maxRetries = 15
+		for i := 1; i < maxRetries; i++ {
+			r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})
+			if err != nil {
+				t.Fatalf("OS.Verify request failed: %s", err)
+			}
+			if counter == maxRetries {
```
liulk, line 280: This condition will never be true because `counter` is always one less than `i`, and `i` can only ever be one less than `maxRetries`. In general, the for loop idiom should look like this:

```
for i := 1; i <= maxRetries; i++ {
  t.Logf("Attempt %d of %d", i, maxRetries)
  r, err := t.osc.Verify(...)
  ... // check err
  ... // check r
  if got == want {
    break
  }
  if i == maxRetries {
    t.Fatal(...)
  }
  time.Sleep(...)
}
```

liulk (changes requested):

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -270,15 +270,28 @@ func (tc *testCase) verifyInstall(ctx context.Context, t *testing.T) {
 	}
 
 	if tc.dualSup {
-		if got, want := r.GetVerifyStandby().GetVerifyResponse().GetActivationFailMessage(), ""; got != want {
-			t.Errorf("OS.Verify Standby ActivationFailMessage: got %q, want %q", got, want)
-		}
+		var counter int
```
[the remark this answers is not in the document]
Hobbydos: sure, I'll remove counter and just use `i`

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: Fixed the loop with above mentioned idiom in latest commit.

liulk (approved):

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: if we move at end of function it will consider `r` declared but not used. I think it's best we keep it at the top of loop so for every run it's updated for evaluation

liulk: Okay, I just realized what this is doing:

 1. Lines 260-270 (before the loop started), the code checks only the primary supervisor.
 2. The retry loop (guarded by whether the device has dual supervisors) only checks the standby supervisor.
 
So the two code sections serve a different purpose.

Clarification question: why do we retry only for the standby and not primary? Should we just retry for both?

```
func (tc *testCase) verifyInstall(...) {
  for ... {
    r, err := tc.osc.Verify(...)
    ... // check err
    ok := true
    ... // check primary supervisor and set ok false if got != want
    if tc.dualsup {
      ... // check backup supervisor and set ok false if got != want
    }
    if ok {
      t.Log("OS.Verify complete")
      return
    }
    if i == maxAttempt {
      t.Fatalf(...)
    }
    t.Log("Attempt %d of %d failed. Sleep and then retrying...", i, maxRetries)
    time.Sleep(...)
  }
}
```

xw-g: @liulk my reading of the code is that:
* As for the RPC call, there is no diff between primary and secondary status. It's the same `r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})` that returns a struct `VerifyResponse` which contains both the primary and secondary info. 
* `if tc.dualSup {` just does an extra read about `VerifyStandby` in the `VerifyResponse` struct. 
* That's why I was saying that the retry of the RPC `r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})` should happen at the end of the for loop, to save one extra involking (because it's already called above). 

@Hobbydos, regarding
```
I think it's best we keep it at the top of loop so for every run it's updated for evaluation
```
* If you keep it at the bottom, it still does the same refresh, right? And you save one extra duplicated invoking of the RPC.

```
if we move at end of function it will consider r declared but not used. 
```
Because you are "refreshing" the variable `r`, you don't need the colon to initial them as new variables, right? `r, err = tc.osc.Verify(ctx, &ospb.VerifyRequest{})` 

Also, @Hobbydos we prefer not to close a comment (e.g. this one) that's not yet actually closed. There were still question from @liulk and I haven't got a chance to reply your comment :).Thanks.

Hobbydos: @xw-g did you mean something like this instead?

```go
	if tc.dualSup {
		const maxRetries = 15
		for i := 1; i <= maxRetries; i++ {
			t.Logf("Attempt %d of %d", i, maxRetries)
			if err != nil {
				t.Fatalf("OS.Verify request failed: %s", err)
			}
			if got, want := r.GetVerifyStandby().GetVerifyResponse().GetActivationFailMessage(), ""; got != want {
				t.Errorf("OS.Verify Standby ActivationFailMessage: got %q, want %q", got, want)
			}
			if got, want := r.GetVerifyStandby().GetVerifyResponse().GetVersion(), *osVersion; got != want {
				t.Logf("OS.Verify Standby Version: got %q, want %q", got, want)
			} else {
				break
			}
			if i == maxRetries {
				t.Fatalf("OS.Verify Standby could not obtain version status in %d retries", maxRetries)
			}
			time.Sleep(1 * time.Minute)
			r, err = tc.osc.Verify(ctx, &ospb.VerifyRequest{})
		}
	}
	t.Log("OS.Verify complete")
}
```

liulk: I just realized that if we're retrying here, potentially we don't have to wait for reboot to complete (lines 161-187). This way, the loop here will wait for both the primary and backup supervisors to report the correct version. And instead of maxRetries, it should honor *timeout.

@bstoll what do you think?

liulk (changes requested):

feature/gnoi/os/tests/osinstall/osinstall_test.go
xw-g: @Hobbydos yes, that's what I meant.

@liulk I think line 161-187 is for waiting on reboot, but this loop here is waiting on the sync between dual Supervisors (after reboot). For some OS implementations, until the sync is done the RPC call won't reply with the version info etc. Therefore, I think we still need both.

liulk: We do have to make a small modification for the verify loop so it will retry on `status.Code(err) == codes.Unavailable` rather than t.Fatal. After this, the reboot loop is redundant because the verify loop will now correctly wait for the reboot to finish in addition to checking the OS versions.

liulk: Your conditional will mistakenly retry if err is nil. This is why your test will eventually time out and fatal when the operations were successful. We also want to retry for both single and dual supervisor devices, not just the dual.

Can you try my rewrite of verifyInstall and let me know if the code makes sense to you?

```
func (tc *testCase) verifyInstall(ctx context.Context, t *testing.T) {
	deadline := time.Now().Add(*timeout)
	for time.Now().Before(deadline) {
		r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})
		switch status.Code(err) {
		case codes.OK:
		case codes.Unavailable:
			t.Log("Reboot in progress.")
			time.Sleep(time.Minute)
			continue
		default:
			t.Fatalf("OS.Verify request failed: %v", err)
		}

		if got, want := r.GetActivationFailMessage(), ""; got != want {
			t.Fatalf("OS.Verify ActivationFailMessage: got %q, want %q", got, want)
		}
		if got, want := r.GetVersion(), *osVersion; got != want {
			t.Log("Reboot not started yet.")
			time.Sleep(time.Minute)
			continue
		}

		if tc.dualSup {
			if got, want := r.GetVerifyStandby().GetVerifyResponse().GetActivationFailMessage(), ""; got != want {
				t.Fatalf("OS.Verify Standby ActivationFailMessage: got %q, want %q", got, want)
			}

			if got, want := r.GetVerifyStandby().GetVerifyResponse().GetVersion(), *osVersion; got != want {
				t.Log("Standby not ready.")
				time.Sleep(time.Minute)
				continue
			}
		}

		t.Log("OS.Verify complete")
		return
	}

	t.Fatal("OS.Verify did not return the correct version before deadline.")
}
```

Hobbydos: With the re-works of the `rebootDUT()` function, your code is more holistic than my initial patch-attempt to simply loop over if it's `dualSup`. I think having timeout` instead of `maxRetries` is better - I should've done this from the get-go but was sidelined with the initial constraint of looping through for 15 minutes. 


I tested it in my lab, the execution time remains similar and the test passed for a DualSup DUT

```
    osinstall_test.go:291: OS.Verify complete
--- PASS: TestOSInstall (1100.99s)
PASS

*** Releasing the testbed...

ok      command-line-arguments  1118.730s
```

liulk: Update the PR with my proposal?

liulk (changes requested):

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: I'll have a commit pushed shortly.

liulk (approved):
@xw-g looks good to me now. Do you want to take another look?

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -257,29 +224,45 @@ func (tc *testCase) transferOS(ctx context.Context, t *testing.T, standby bool)
...
+		}
 
-	if tc.dualSup {
-		if got, want := r.GetVerifyStandby().GetVerifyResponse().GetActivationFailMessage(), ""; got != want {
-			t.Errorf("OS.Verify Standby ActivationFailMessage: got %q, want %q", got, want)
+		if got, want := r.GetActivationFailMessage(), ""; got != want {
+			t.Fatalf("OS.Verify ActivationFailMessage: got %q, want %q", got, want)
 		}
+		if got, want := r.GetVersion(), *osVersion; got != want {
+			t.Log("Reboot not started yet.")
```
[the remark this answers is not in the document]
xw-g: @Hobbydos

xw-g:
I will drive the presubmit test.

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -257,29 +224,45 @@ func (tc *testCase) transferOS(ctx context.Context, t *testing.T, standby bool)
...
+	for time.Now().Before(deadline) {
+		r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})
+		switch status.Code(err) {
+		case codes.OK:
+		case codes.Unavailable:
+			t.Log("Reboot in progress.")
+			time.Sleep(time.Minute)
+			continue
+		default:
+			t.Fatalf("OS.Verify request failed: %v", err)
```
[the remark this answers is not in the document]
Hobbydos: Sounds good, we can revisit later if required and make adjustments as needed.

feature/gnoi/os/tests/osinstall/osinstall_test.go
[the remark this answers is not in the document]
Hobbydos: Added in latest commit.

jjing1122:
Filed internal bug b/263261243 to track presubmit progress.

jjing1122:
the current verification code only covers when noreboot is set to true. request to add the following condition in line 239: ([https://github.com/openconfig/featureprofiles/blob/d393408d0cb9ef9bc464884bcbfe273b035c0772/feature/gnoi/os/tests/osinstall/osinstall_test.go#L239](https://www.google.com/url?q=https://github.com/openconfig/featureprofiles/blob/d393408d0cb9ef9bc464884bcbfe273b035c0772/feature/gnoi/os/tests/osinstall/osinstall_test.go%23L239&sa=D&source=buganizer&usg=AOvVaw3kXHVVMlJAv2VVCDrbL-ms))

// when noreboot is set to false, the device returns "in-progress" before initiating the reboot
        if got, want := r.GetActivationFailMessage(), "in-progress"; got == want {
            time.Sleep(time.Minute)
            continue
        }

Hobbydos:
> the current verification code only covers when noreboot is set to true. request to add the following condition in line 239: ([https://github.com/openconfig/featureprofiles/blob/d393408d0cb9ef9bc464884bcbfe273b035c0772/feature/gnoi/os/tests/osinstall/osinstall_test.go#L239](https://www.google.com/url?q=https://github.com/openconfig/featureprofiles/blob/d393408d0cb9ef9bc464884bcbfe273b035c0772/feature/gnoi/os/tests/osinstall/osinstall_test.go%23L239&sa=D&source=buganizer&usg=AOvVaw3kXHVVMlJAv2VVCDrbL-ms))
> 
> // when noreboot is set to false, the device returns "in-progress" before initiating the reboot if got, want := r.GetActivationFailMessage(), "in-progress"; got == want { time.Sleep(time.Minute) continue }

Added in latest commit

liulk (approved):

liulk:
Please run gofmt?

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -236,12 +236,16 @@ func (tc *testCase) verifyInstall(ctx context.Context, t *testing.T) {
 		default:
 			t.Fatalf("OS.Verify request failed: %v", err)
 		}
-
+		// when noreboot is set to false, the device returns "in-progress" before initiating the reboot
+		if got, want := r.GetActivationFailMessage(), "in-progress"; got == want {
```
xw-g, line 240: Here we don't know if are expecting "in-progress". So, to me it's just a normal condition that if we see it we sleep 1 min and then continue. Therefore, I think it make sense to just say `if r.GetActivationFailMessage() == "in-progress"`.

xw-g (dismissed):

xw-g:
Thanks @Hobbydos . After you have addressed all the comments, please help ping @jjing1122 who will drive the presubmit internally before merging this PR. Thanks!

Hobbydos:
> Please run gofmt?

gofmt should be ran by VSCode extension, running it manually, I didn't see any deltas?

Hobbydos:
> Thanks @Hobbydos . After you have addressed all the comments, please help ping @jjing1122 who will drive the presubmit internally before merging this PR. Thanks!

@jjing1122, I rewrote the if statement as per @xw-g's comments, and re-ran `gofmt` again.

Hobbydos:
@xw-g @jjing1122 Did this look OK in internal tests for merger?

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -257,29 +224,49 @@ func (tc *testCase) transferOS(ctx context.Context, t *testing.T, standby bool)
 // verifyInstall validates the OS.Verify RPC returns no failures and version numbers match the
 // newly requested software version.
 func (tc *testCase) verifyInstall(ctx context.Context, t *testing.T) {
-	r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})
-	if err != nil {
-		t.Fatalf("OS.Verify request failed: %s", err)
-	}
+	deadline := time.Now().Add(*timeout)
+	for time.Now().Before(deadline) {
```
alexmasi, line 228: important to note this "deadline" is not strict in the sense that it will not terminate actions in the for loop, it will only stop an additional loop from starting after the deadline.

if the intended behavior is for the timeout to be strict then use context.WithTimeout() https://pkg.go.dev/context#WithTimeout and defer the cancel(). Then have ocs.Verify graciously handle a cancelled context

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -257,29 +224,49 @@ func (tc *testCase) transferOS(ctx context.Context, t *testing.T, standby bool)
...
+		case codes.Unavailable:
+			t.Log("Reboot in progress.")
+			time.Sleep(time.Minute)
+			continue
+		default:
+			t.Fatalf("OS.Verify request failed: %v", err)
+		}
+		// when noreboot is set to false, the device returns "in-progress" before initiating the reboot
+		if r.GetActivationFailMessage() == "in-progress" {
+			time.Sleep(time.Minute)
```
alexmasi, line 241: maybe a log should be added here announcing the sleep like other logs in this func

feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -257,29 +224,49 @@ func (tc *testCase) transferOS(ctx context.Context, t *testing.T, standby bool)
...
-		t.Fatalf("OS.Verify request failed: %s", err)
-	}
+	deadline := time.Now().Add(*timeout)
+	for time.Now().Before(deadline) {
+		r, err := tc.osc.Verify(ctx, &ospb.VerifyRequest{})
+		switch status.Code(err) {
+		case codes.OK:
+		case codes.Unavailable:
+			t.Log("Reboot in progress.")
+			time.Sleep(time.Minute)
```
alexmasi, line 234: this duration of time.Minute is reused multiple times in this func, if this is intended to be the standard wait time for reboots or other conditions that desire a sleep, then create a var for this duration and reference the var in these locations in case it is deemed appropriate to change the standard wait time in the future

feature/gnoi/os/tests/osinstall/osinstall_test.go
liulk: I'm okay either way. Timeout in tests does not need to be too precise.

Using context.WithTimeout(), we probably don't need any other special handling. The default case for status.Code() will just t.Fatal which is probably fine.

jjing1122:
@Hobbydos can you please fix the new comments wrt go readability.  We will merge right after that.

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: Sounds good, made the changes locally, just firing a test-run on my DUT, upon success I'll raise a new commit.

feature/gnoi/os/tests/osinstall/osinstall_test.go
Hobbydos: same as above, I've made this change and will fire off a test run, if successful, I'll raise a commit.

Hobbydos:
> @Hobbydos can you please fix the new comments wrt go readability. We will merge right after that.

@jjing1122 added the requested changes & tested in my environment. Test Passed.

feature/gnoi/os/tests/osinstall/osinstall_test.go
alexmasi: This var does not need to be exported, also prefer `rebootWait` or similar to avoid renaming in case of increase to wait time.

Other than that this looks good to me for golang

Hobbydos: Sure, I actually had it as `rebootWait` at first, but thought it was a bit misleading. Let me do those quick changes, rebase and push.

jjing1122 (approved):

feature/gnoi/os/tests/osinstall/osinstall_test.go
jjing1122: commit has been raised

jjing1122 (approved):

PR #840, opened 2022-12-05
fix for gribi test

DanG100:

PR #728, opened 2022-11-01
OTG Test - TE-11.2: Backup NHG: Multiple NH

octpetre:
OTG conversion of **_TE-11.2: Backup NHG: Multiple NH_**
Output log found [here](https://drive.google.com/file/d/1RwoXXSC0UCy9Uqc8JsPmQSfa_nW-e9i3/view?usp=share_link)
Fixes #646

PR #841, opened 2022-12-05
github workflow add pull request rebase check

liulk:
#728 introduced a build-breaking change (fixed by #840) but it was not caught by pull request checks because the go.mod in the PR is out of date. To mitigate problems due to stale checks, I'm adding a rebase check that requires PR to be rebased if go.mod and checks are out of date.

### DIFF
--- a/.github/workflows/rebase_check.yml
+++ b/.github/workflows/rebase_check.yml
@@ -1,0 +1,43 @@
+name: Rebase Check
+on:
+  pull_request:
+    # Ignore the rebase check if the pull request touches the checked files.
+    paths-ignore:
+      - go.mod
+      - .github/CODEOWNERS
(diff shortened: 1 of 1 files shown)


bstoll (approved):

liulk:
I merged #842 before this on purpose to test that checking for outdated pull_request.yml works. It works but with the caveat that it's not automatically rechecked before merge, and I could not find a GitHub event with the merge trigger. But any push to the PR will trigger checks to rerun, which is probably fine.

Please reapprove.

bstoll:
I think the closest alternative is to copy the tasks that run on each PR to also run on the main branch.  This would only notify you after the fact that something is broken though.

bstoll (approved):

PR #847, opened 2022-12-06
RT-3.2: Multiple Rules for VRF Selection : changing VRFs and VRF enabled subinterface

avaneesh90:
1. Added Type configuration Type: oc.NetworkInstanceTypes_NETWORK_INSTANCE_TYPE_DEFAULT_INSTANCE
2. For dutPort2 and atePort2, unit 0 subinterface is configured without vlan and subinterfaces 1-3 are configured with vlan. Since subinterface 0 is not used, deleted the same.
3. Network instance with interface binding and sub interface configuration happens without vlan-id followed by vlan-id config for the subinterface. Edited to add vlan-config in the initial config
4. Added interface-ref with subinterface while applying vrf-selection-policy


ciscoranjns (commented):
" DISCARD  line breaks the code for us"
as discard  action is supported only under policy type "Policy_Type_PBR_POLICY" and not under "Policy_Type_VRF_SELECTION_POLICY

ciscoranjns (changes requested):
" DISCARD line breaks the code for us"
as discard action is supported only under policy type "Policy_Type_PBR_POLICY" and not under "Policy_Type_VRF_SELECTION_POLICY

feature/experimental/policy/policy_vrf_selection/ate_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
```diff
@@ -322,10 +319,15 @@ func testTrafficFlows(t *testing.T, args *testArgs, expectPass bool, flow ...*on
...
-func getL3PBRRule(args *testArgs, networkInstance string, index uint32, dscpset []uint8) *oc.NetworkInstance_PolicyForwarding_Policy_Rule {
+func getL3PBRRule(args *testArgs, networkInstance string, index uint32, dscpset []uint8, discard bool) *oc.NetworkInstance_PolicyForwarding_Policy_Rule {
 	r := oc.NetworkInstance_PolicyForwarding_Policy_Rule{}
 	r.SequenceId = ygot.Uint32(index)
-	r.Action = &oc.NetworkInstance_PolicyForwarding_Policy_Rule_Action{NetworkInstance: ygot.String(networkInstance)}
+	if networkInstance != "" {
+		r.Action = &oc.NetworkInstance_PolicyForwarding_Policy_Rule_Action{NetworkInstance: ygot.String(networkInstance)}
+	}
+	if discard {
+		r.Action = &oc.NetworkInstance_PolicyForwarding_Policy_Rule_Action{Discard: ygot.Bool(true)}
```
ciscoranjns, line 329: " DISCARD line breaks the code for us"
as discard action is supported only under policy type "Policy_Type_PBR_POLICY" and not under "Policy_Type_VRF_SELECTION_POLICY

sachendras: Is this a Cisco requirement or a Standard? Can you please point me to the yangdoc that says discard cant be used for "Policy_Type_VRF_SELECTION_POLICY"? Below is what I see in the [doc](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-policy-forwarding-policies-policy-rules-rule-action). May be I am missing something. If it is a Cisco issue then clearly the code needs to support the requirement.


```
/network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/action/config/
discard
description:
When this leaf is set to true, the local system should drop packets that match the rule.

nodetype: leaf (rw)

type: boolean

default: false
```

vosipchu: Sash, the right question should be (per my view) is whether it is a test requirement or not. There are many different leaves in OC, but it doesn't mean every vendor must support every leaf. Then, if it is not within a strict test requirement and there is already Vendor A passing test X, then every vendor that follows up has to make sure that all new PRs won't fail existing passes (again, unless there is a miss from the test expectation).  
If this happens, then the requestor should find a workaround.

robshakir: +1, I agree with @vosipchu here, I'm not quite sure that we require the explicit discard here. If we expect packets that are not matching a VRF selection criteria to be forwarded within the default network instance (which I think is the current expectation) then the discard seems like it would be actively harmful.

I think that this will break forwarding plain ol' IPv[46] packets if any implementation requires `DISCARD` here - which seems like it would break numerous things like traceroute/in-band management. Given this, it is probably even worth us extending the test case to cover that IP forwarding that doesn't match the VRF selection policy is forwarded.

Let's discuss together to check we're on the same page and we can confirm in this PR.

avaneesh90: I have removed the discard term in latest changes , please check

sachendras:
Replaced by pull/1055, pull/1056 and pull/1058

PR #849, opened 2022-12-06
proposal for P4RT-2.1 Mastership test

Hobbydos:


Hobbydos:
cc @xw-g @robshakir

feature/experimental/p4rt/tests/p4rt_mastership/p4rt_mastership.MD
```diff
@@ -0,0 +1,22 @@
+# P4RT-2.1: P4RT Mastership
+
+## Summary
+
+Validate the P4RT server handles master election and failover
+
+## Procedure
+
+* Enable P4RT on a single FAP by configuring an ID on the device and one or more interfaces.
```
Ankur19, line 9: Primary/Secondary over master/slave?

Ankur19 (changes requested):
Looks Good except a few NITs

feature/experimental/p4rt/tests/p4rt_mastership/p4rt_mastership.MD
Hobbydos: @Ankur19, Should we rename the test to P4RT election instead of mastership to keep the trend of inclusive language?

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,619 @@
...
+		}
+	}
+}
+
+// Test arbitration update from same client
+func TestArbitrationUpdate(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	configureDeviceId(t, dut)
+	configurePortId(t, dut)
+	testCases := []testArgs{
```
Ankur19, line 564: Single testArg. Doesn't need a table driven test and the loop over testCases.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,619 @@
...
+	dut := ondatra.DUT(t, "dut")
+	configureDeviceId(t, dut)
+	configurePortId(t, dut)
+	testCases := []mcTestArgs{
+		{
+			name:       "Primary_secondary_OK",
+			expectPass: true,
+			Items: [2]*testArgs{
+				{
+					desc:              "Primary connection",
```
Ankur19, line 488: The only variable I see are the "desc" and the "lowID". Can you just use the table for those and create a single testArg

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,619 @@
...
+}
+
+// Test Duplicate ElectionID
+func TestDuplicateElectionID(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	configureDeviceId(t, dut)
+	configurePortId(t, dut)
+	testCases := []testArgs{
+		{
+			desc:         "Primary connection",
```
Ankur19, line 433: Can we only have the values that are changing in the table here and then add the constant items later in `for test:= range testCases{}` loop

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,619 @@
...
+}
+
+// Test Primary/Secondary clients
+func TestPrimarySecondary(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	configureDeviceId(t, dut)
+	configurePortId(t, dut)
+	testCases := []testArgs{
+		{
+			desc:         "Primary connection",
```
Ankur19, line 382: Similar to the other comments below. Can we only add the variables in the table here.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,619 @@
...
+}
+
+// Test Primary Reconnect
+func TestPrimaryReconnect(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	configureDeviceId(t, dut)
+	configurePortId(t, dut)
+	testCases := []testArgs{
+		{
+			desc:         "Primary connection",
```
Ankur19, line 326: Same here

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,619 @@
...
+	handle.StreamChannelDestroy(&streamName)
+	handle.ServerDisconnect()
+}
+
+// Test Zero client with 0 electionID
+func TestZeroMaster(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	configureDeviceId(t, dut)
+	configurePortId(t, dut)
+	testCases := []testArgs{
```
Ankur19, line 288: No need of table here.

Ankur19 (changes requested):

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Hobbydos: I can do that, however if there are other tests to be added in the future we'll have to re-implement table driven test  + loop.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,617 @@
...
+}
+
+// Define Test Args Structure
+type testArgs struct {
+	desc              string
+	lowID             uint64
+	highID            uint64
+	deviceID          uint64
+	handle            *p4rt_client.P4RTClient
+	expectStatus      int32
```
robshakir, line 65: Generally we'd call these `inXXX` for input and `wantXXX` for expected values.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,617 @@
...
+	expectFinalStatus int32
+	expectPass        bool
+	expectWrite       bool
+	expectRead        bool
+}
+
+// configureDeviceId configures p4rt device-id on the DUT.
+func configureDeviceId(t *testing.T, dut *ondatra.DUTDevice) {
+	component := oc.Component{}
```
robshakir, line 74: Suggest to just define `component` here as a pointer since you never use it in a manner that requires it not to be a pointer.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,617 @@
...
+// Create client connection
+func clientConnection(t *testing.T, dut *ondatra.DUTDevice) *p4rt_client.P4RTClient {
+	clientHandle := p4rt_client.P4RTClient{}
+	if err := clientHandle.P4rtClientSet(dut.RawAPIs().P4RT().Default(t)); err != nil {
+		t.Fatalf("Could not initialize p4rt client: %v", err)
+	}
+	return &clientHandle
+}
+
+// Setup P4RT Arbitration Stream
```
robshakir, line 96: Can we define here what the return argument is? i.e., the arbitration status code.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,617 @@
...
+	handle.ServerDisconnect()
+}
+
+// Test Zero client with 0 electionID
+func TestZeroMaster(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	configureDeviceId(t, dut)
+	configurePortId(t, dut)
+	test := testArgs{
+
```
robshakir, line 293: (nit) remove blank line,

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,617 @@
...
+		desc:         pDesc,
+		lowID:        *ygot.Uint64(0),
+		highID:       highID,
+		handle:       clientConnection(t, dut),
+		deviceID:     deviceID,
+		expectPass:   false,
+		expectStatus: 0,
+	}
+	t.Run(test.desc, func(t *testing.T) {
```
robshakir, line 302: Do we expect to have >1 tests within this test case going forward? I guess the `t.Run` here is intended to just allow us to get a better failure report (since it seems like you couldn't just run this test case without the setup).

robshakir:
Thanks for this contribution -- generally LGTM, just some cleanup around the code that I think we need and to check some of the definitions.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Hobbydos: @robshakir , did you mean the return on line 93 or the comment from the function starting on line 96?

feature/experimental/p4rt/tests/p4rt_election/README.md
```diff
@@ -0,0 +1,22 @@
+# P4RT-2.1: P4RT Election
+
+## Summary
+
+Validate the P4RT server handles primary election and failover
```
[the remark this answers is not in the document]
Hobbydos: Pushed a commit with changes requested.

Hobbydos:
Went through and added suggested changes.

robshakir (commented):
This LGTM -- with one small pending question and one suggestion.

@xw-g -- do you want to take a look before we go ahead and merge?

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,626 @@
...
+			P4Info: &p4Info,
+			Cookie: &p4_v1.ForwardingPipelineConfig_Cookie{
+				Cookie: 159,
+			},
+		},
+	}); err != nil {
+		return errors.New("Errors seen when sending SetForwardingPipelineConfig.")
+	}
+
+	err = args.handle.Write(&p4_v1.WriteRequest{
```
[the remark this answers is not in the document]
Hobbydos: Made the suggested changes.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,625 @@
...
+	return true, nil
+}
+
+// Only Primary should be able to write
+func canWrite(t *testing.T, args *testArgs) (bool, error) {
+	pktIO := getGDPParameter(t)
+	writeErr := writeTableEntry(args, t, pktIO, false)
+	if writeErr != nil {
+		if args.wantWrite == false {
+			t.Logf("Expected P4Client error: Client write permission (highID %d, lowID %d): want %v, got %v", args.highID, args.lowID, args.wantWrite, false)
```
Ankur19, line 256: Logic error here. Should it not be args.wantWrite==true?

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,625 @@
...
+		highID:     inId0,
+		handle:     clientConnection(t, dut),
+		deviceID:   deviceId,
+		wantFail:   true,
+		wantStatus: 0,
+	}
+	t.Run(test.desc, func(t *testing.T) {
+		resp, err := streamP4RTArb(&test)
+		if err == nil && test.wantFail {
+			t.Errorf("Zero ElectionID (0,0) is allowed: %v", err)
```
Ankur19, line 313: Is this a requirement? Couldn't find anything related to this here: https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-updates

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Hobbydos: We're only hitting this block when the write actually fails (err != nil) and if args.wantWrite is false, then it's an expected error, hence the t.Logf

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Hobbydos: There was some debate on this, however the lack of clarification of the spec is what prompted for a decision. @robshakir, thoughts?  I checked on a few vendors and they considered (0,0) as invalid election ID as well.

The closest thing I found in the spec under `16.2. Client Arbitration Update` was: 
> The sender need not specify an election_id. If the election_id is not specified, the sender's election_id is considered lower than any election_id, and the sender will thus never become primary. This way, a controller can choose to be a backup controller, in order to avoid “flapping” (if a backup controller connects to the switch shortly before the actual primary controller, therefore becoming primary temporarily).

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,625 @@
...
+			wantFail:   false,
+			wantStatus: 0,
+		}, {
+			desc:       sDesc,
+			lowID:      inId100,
+			highID:     inId0,
+			handle:     clientConnection(t, dut),
+			deviceID:   deviceId,
+			wantFail:   true,
+			wantStatus: 0,
```
Ankur19, line 454: Should it be 3?

```
If the election_id is set and is already used by another controller (excluding the controller making the request) for the same (device_id, role), the P4Runtime server shall terminate the stream by returning an INVALID_ARGUMENT error.
```

error code for INVALID_ARGUMENTis 3

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: In master Arbitration, election ID zero need no be treated as a special case. The P4RT server should be able to accept the connection and client with election ID 0 may become primary, if its the only client getting connected.

rsindhujha:
In the below testcase(func TestPrimaryReconnect(t *testing.T)), the client is getting disconnected after every iteration. This means the connection should be successful for each iteration with election ID 100, 90 and 100 again. Since there is only one client exists within a given iteration, for iteration 2, the wantStatus should be 0 instead of 5?

https://github.com/openconfig/featureprofiles/blob/b389c9712e2b2a88758b11dd4cc1dac12bf0ef4c/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go#L327

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: For the below Testcase, sometimes we are observing that the testcase did not receive the status code 3, as the P4RT server already disconnects the streamChannel after sending the MasterArbitration Response. 

 func TestDuplicateElectionID(t *testing.T) { 

Reference from the specs (5.3.2.4)
If the election_id is set and is already used by another controller (excluding the controller making the request) for the same (device_id, role), the P4Runtime server shall terminate the stream by returning an INVALID_ARGUMENT error.

rsindhujha:
For the last testcase, the script is sending 2nd Arbitration Request with lower electionID over the same StreamChannel. This Arbitration Request should be accepted by the server, since there is no other client. In this case the expected Status code should be 0 and wantWrite True

https://github.com/openconfig/featureprofiles/blob/b389c9712e2b2a88758b11dd4cc1dac12bf0ef4c/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go#L567

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Hobbydos: > In the below testcase(func TestPrimaryReconnect(t *testing.T)), the client is getting disconnected after every iteration. This means the connection should be successful for each iteration with election ID 100, 90 and 100 again. Since there is only one client exists within a given iteration, for iteration 2, the wantStatus should be 0 instead of 5?
> 
> https://github.com/openconfig/featureprofiles/blob/b389c9712e2b2a88758b11dd4cc1dac12bf0ef4c/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go#L327

I think there might be a disconnect here, the server would keep track of the Primary Election ID, should a connection come in with anything < Primary Election ID - Even if the Primary is disconnected, the error MSG should be 5.

From Section 16.2 - https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-notification
> the P4Runtime server keeps track of the highest election_id that it has ever received. If a controller's election_id is equal to the highest election_id that the server has ever received, that controller is the primary. All other controllers are backups. Note that it is possible that all controllers are backups, and that there is no primary. There can be at most one primary, because for any given (device_id, role), each connected controller has a unique election_id.

And Section 5.4: https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-notification
>If there is a primary:
For the primary, status is OK (with status.code set to google.rpc.OK).
For all backup controllers, status is set to non-OK (with status.code set to google.rpc.ALREADY_EXISTS).
Otherwise, if there is no primary currently, for all backup controllers, status is set to non-OK (with status.code set to google.rpc.NOT_FOUND).

GRPC status codes for ref: https://developers.google.com/maps-booking/reference/grpc-api/status_codes

Hobbydos:
> In the below testcase(func TestPrimaryReconnect(t *testing.T)), the client is getting disconnected after every iteration. This means the connection should be successful for each iteration with election ID 100, 90 and 100 again. Since there is only one client exists within a given iteration, for iteration 2, the wantStatus should be 0 instead of 5?
> 
> https://github.com/openconfig/featureprofiles/blob/b389c9712e2b2a88758b11dd4cc1dac12bf0ef4c/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go#L327

> In the below testcase(func TestPrimaryReconnect(t *testing.T)), the client is getting disconnected after every iteration. This means the connection should be successful for each iteration with election ID 100, 90 and 100 again. Since there is only one client exists within a given iteration, for iteration 2, the wantStatus should be 0 instead of 5?
> 
> https://github.com/openconfig/featureprofiles/blob/b389c9712e2b2a88758b11dd4cc1dac12bf0ef4c/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go#L327

I think there might be a disconnect here, the server would keep track of the Primary Election ID, should a connection come in with anything < Primary Election ID - Even if the Primary is disconnected, the error MSG should be 5.

From Section 16.2 - https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-notification
> the P4Runtime server keeps track of the highest election_id that it has ever received. If a controller's election_id is equal to the highest election_id that the server has ever received, that controller is the primary. All other controllers are backups. Note that it is possible that all controllers are backups, and that there is no primary. There can be at most one primary, because for any given (device_id, role), each connected controller has a unique election_id.

And Section 5.4: https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-notification
>If there is a primary:
For the primary, status is OK (with status.code set to google.rpc.OK).
For all backup controllers, status is set to non-OK (with status.code set to google.rpc.ALREADY_EXISTS).
Otherwise, if there is no primary currently, for all backup controllers, status is set to non-OK (with status.code set to google.rpc.NOT_FOUND).

GRPC status codes for ref: https://developers.google.com/maps-booking/reference/grpc-api/status_codes

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Hobbydos: I'm currently looking into this case, as you mention the streamChannel gets disconnected right after the MasterArbitration Response, making it challenging to capture the Arbitration response status code. Might have to look at changing the client, but it's a WIP.

rsindhujha: Thanks @Hobbydos

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Hobbydos: @rsindhujha, IIRC there was a conversation with @robshakir where we had concluded that zero election ID was not to become primary even if it was the only client getting connected.  Perhaps Rob can share thoughts here.

xw-g:
> This LGTM -- with one small pending question and one suggestion.
> 
> @xw-g -- do you want to take a look before we go ahead and merge?

no, I'll let you drive this one. we already have 3 reviewers on this one. That's good enough :)

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,625 @@
...
+					DeviceId: streamParameter.DeviceId,
+					ElectionId: &p4_v1.Uint128{
+						High: streamParameter.ElectionIdH,
+						Low:  streamParameter.ElectionIdL,
+					},
+				},
+			},
+		}); err != nil {
+			return 0, errors.New("Errors seen when sending ClientArbitration message.")
+		}
```
Ankur19, line 129: You can now get the StreamTermErr after sendinding Msg.

```
if _, _, arbErr := client.StreamChannelGetArbitrationResp(&streamName, 1); arbErr != nil {
  if err := p4rtutils.StreamTermErr(client.StreamTermErr); err != nil {
	  return err
  }
  return fmt.Errorf("errors seen in ClientArbitration response: %v", arbErr)
}
```

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Ankur19: @Hobbydos please refer to other tests to see how to capture Stream Termination Errors and then update to the correct status codes to validate

Hobbydos: Thanks @Ankur19  - I plan to revisit this shortly. It seems like I'll have to change how the p4rt node name is passed, too.
Ref:

Tests should get the P4RT Node Name by walking the Components OC tree. Components of type INTEGRATED_CIRCUIT should have child Components of type PORT. These PORT Components can be mapped to currently reserved Interfaces using the hardware-port leaf in the Interfaces tree. Such an implementation already exists in p4rtutils library: p4rtutils.P4RTNodesByPort().

If P4RT Node Names cannot be resolved by walking the Components tree, use deviation flag --deviation_explicit_p4rt_node_component and pass the node names through args --arg_p4rt_node_name_1, --arg_p4rt_node_name_2.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Hobbydos: I tried to look at `StreamTermErr` but it seemed like it was `nil`. Is it possible that the stream gets destroyed before reaching the code? 
Here's a view from DLV in VSCode of the args
image
 
Code Snippet (don't mind the `return 3`, I was putting it there for debug purposes.
```go
func getRespCode(args *testArgs) (int32, error) {
	handle := args.handle
	if handle != nil {
		// Grab Arb Response to look at status code
		_, arbResp, arbErr := handle.StreamChannelGetArbitrationResp(&streamName, 1)
		if arbErr != nil {
			if err := p4rtutils.StreamTermErr(args.handle.StreamTermErr); err != nil {
				return 3, err
			}
			return 0, fmt.Errorf("Errors seen in ClientArbitration response: %v", arbErr)
		}
		// Handle exception when GetCode is empty.
		respCode := arbResp.Arb.GetStatus().GetCode()
		return respCode, nil
	}
	return 0, errors.New("Missing Client")
}
```

Ankur19: I think it is because of how you initialize `P4RTClient`.

* If you do `p4rt_client.P4RTClient{}` then `StreamTermErr` channel is not initialized.

The proper way would be using the method `p4rt_client.NewP4RTClient()`.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,625 @@
...
+
+// configurePortId configures p4rt port-id on the DUT.
+func configurePortId(t *testing.T, dut *ondatra.DUTDevice) {
+	portName := dut.Port(t, "port1").Name()
+	gnmi.Replace(t, dut, gnmi.OC().Interface(portName).Id().Config(), portId)
+}
+
+// Create client connection
+func clientConnection(t *testing.T, dut *ondatra.DUTDevice) *p4rt_client.P4RTClient {
+	clientHandle := p4rt_client.P4RTClient{}
```
Ankur19, line 97: Change to `p4rt_client.NewP4RTClient()` method. This would initialize `StreamTermErr`

Hobbydos: Thanks @Ankur19, changes worked. Just working on a few more things and should be in a decent place to push something for this week.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Hobbydos: @Ankur19 The requested changes have been added

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,636 @@
...
+// Obtain status Code from Arbitration
+func getRespCode(args *testArgs) (int32, error) {
+	handle := args.handle
+	if handle != nil {
+		// Grab Arb Response to look at status code
+		_, arbResp, arbErr := handle.StreamChannelGetArbitrationResp(&streamName, 1)
+		if arbErr != nil {
+			if err := p4rtutils.StreamTermErr(args.handle.StreamTermErr); err != nil {
+				if strings.Contains(err.Error(), "InvalidArgument") {
+					return 3, err
```
Ankur19, line 151: Please use status and code package here. 

https://pkg.go.dev/google.golang.org/grpc/status

Ankur19 (changes requested):

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Hobbydos: sounds good, will add

Hobbydos: Just to make sure I'm not going off the deep-end and adhering to the standard, you mean something like this, correct?

```go
				if status.Code(err) == codes.InvalidArgument {
					return int32(codes.InvalidArgument), err
				}
```

Ankur19: No need to validate `InvalidArgument` here. Just return whatever error was returned by the server. 

```
If err != nil {
   return int32(status.Code(err))
}
```

Hobbydos: SG - Thanks for clarification

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -0,0 +1,636 @@
...
+
+// Obtain status Code from Arbitration
+func getRespCode(args *testArgs) (int32, error) {
+	handle := args.handle
+	if handle != nil {
+		// Grab Arb Response to look at status code
+		_, arbResp, arbErr := handle.StreamChannelGetArbitrationResp(&streamName, 1)
+		if arbErr != nil {
+			if err := p4rtutils.StreamTermErr(args.handle.StreamTermErr); err != nil {
+				if err != nil {
```
Ankur19, line 150: err would always be nil here. No need to check again. 

```
if err := p4rtutils.StreamTermErr(args.handle.StreamTermErr); err != nil {
				return int32(status.Code(err)), err
}
```

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
Ankur19: This looks somewhat relevant: https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-zero-as-reserved-value

Ankur19 (dismissed):
LGTM!

xw-g:
@liulk is the rundata checking failure a blocker to this merge? Thanks.

xw-g:
@Hobbydos can you help run `go run ./tools/addrundata --fix` (asked @liulk offline). Thanks!

Hobbydos:
> @Hobbydos can you help run `go run ./tools/addrundata --fix` (asked @liulk offline). Thanks!

@xw-g Ran and committed - cc @liulk

xw-g:
Since it's a new test, I"m skipping the internal presumit run. Will merge upon the github check passes.

xw-g:
@Ankur19 and @robshakir gave me LGTM offline.

xw-g (approved):

PR #850, opened 2022-12-06
TE-3.1: Fix checking for nexthop

mojiiba:
Function [testRecursiveIPv4Entry](https://github.com/openconfig/featureprofiles/blob/5890cc99fb3eec7a4738a5bf264611ee9fb00daa/feature/gribi/ate_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go#L274) tries to verify that hierarchical gribi entries  198.51.100.0/24 (A) -> 203.0.113.1/32 (B) -> 192.0.2.6 (C) through AFT Telemetry. The nexthop for A (198.51.100.0/24) is 203.0.113.1 (B) without recursive resolution  and is C (192.0.2.6) with recursive resolution. 

The test first checks the nexthop for A and then for C. However, the check of the nexthop for A assumes that the device returns the recursively resolved nexthop via AFT.  This pull fixes this issue by checking if the returned nexthop is B when the check for the C fails.

Please note that since the testRecursiveIPv4Entry already checks the nexthop for B, this changes does not causes any issues for verifying that A can be resolved via C recursively.

### DIFF
--- a/feature/gribi/ate_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go
+++ b/feature/gribi/ate_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go
@@ -308,7 +308,7 @@ func testRecursiveIPv4Entry(t *testing.T, args *testArgs) {
 		nh := gnmi.Get(t, args.dut, gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance).Afts().NextHop(nhIndexInst).State())
 		// for devices that return  the nexthop with resolving it recursively, e.g., for a->b->c the device returns c
 		if got, want := nh.GetIpAddress(), atePort2.IPv4; got != want {
-			// for devices that return the next hop without resolving it recursively e.g., for a->b->c the device returns b
+			// for devices that return the nexthop without resolving it recursively e.g., for a->b->c the device returns b
 			if got, want := nh.GetIpAddress(), ateIndirectNH; got != want {
 				t.Errorf("next-hop is incorrect: got %v, want %v", got, want)
 			}
--- a/feature/gribi/ate_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go
+++ b/feature/gribi/ate_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go
@@ -273,7 +273,7 @@ func deleteRecursiveIPv4Entry(t *testing.T, args *testArgs) {
 	)
 }
 
-// testRecursiveIPv4Entry verifies recursive IPv4 Entry for 198.51.100.0/24 -> 203.0.113.1/32 -> 192.0.2.6.
+// testRecursiveIPv4Entry verifies recursive IPv4 Entry for 198.51.100.0/24 (a) -> 203.0.113.1/32 (b) -> 192.0.2.6 (c).
 // The IPv4 Entry is verified through AFT Telemetry and Traffic.
 func testRecursiveIPv4Entry(t *testing.T, args *testArgs) {
 	setupRecursiveIPv4Entry(t, args)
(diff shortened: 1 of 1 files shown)


mojiiba:
@vosipchu FYI

mojiiba:
@jjing1122 could you please review this pr?

vosipchu (approved):

feature/gribi/ate_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go
```diff
@@ -302,8 +302,12 @@ func testRecursiveIPv4Entry(t *testing.T, args *testArgs) {
 			t.Errorf("next-hop index is incorrect: got %v, want %v", got, want)
 		}
 		nh := gnmi.Get(t, args.dut, gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance).Afts().NextHop(nhIndexInst).State())
+		// for devices that return  the nexthop with resolving it recursively, e.g., for a->b->c the device returns c
 		if got, want := nh.GetIpAddress(), atePort2.IPv4; got != want {
-			t.Errorf("next-hop is incorrect: got %v, want %v", got, want)
+			// for devices that return the nexthop without resolving it recursively e.g., for a->b->c the device returns b
+			if got, want := nh.GetIpAddress(), ateIndirectNH; got != want {
```
xw-g, line 308: Seems like it only validates that a->b, but does not validate b->c?  

Shall we use this PR (or at least add a TODO here) to deliver some holistic/generic solution?  Maybe a recursive helper function? something like (pseudo code) the following, which should be handle whether a vendor OS returns the flatten next hop or not, right? 
```
func recursiveNextHop(destIP, aftEntries) nextHop {
	nextHopIP, isFound := aftEntries.nexthop(destIP)
	if !isFound {
		return destIP
	}
	return recursiveNextHop(nextHopIP, aftEntries)
}
```

mojiiba: the check for b->c is done after the check for a->b https://github.com/openconfig/featureprofiles/blob/5890cc99fb3eec7a4738a5bf264611ee9fb00daa/feature/gribi/ate_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go#L317.

concerning the helper, I agree with that we need a proper helper function here. Will add TODO for now.

xw-g: a TODO sounds good to me. 

can you help add a comment here saying that b->c is validated below so that other readers later doesn't have the same question? Thanks.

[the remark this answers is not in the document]
mojiiba: done

mojiiba: done

thesrinath (approved):

feature/gribi/ate_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go
```diff
@@ -306,15 +307,20 @@ func testRecursiveIPv4Entry(t *testing.T, args *testArgs) {
 			t.Errorf("next-hop index is incorrect: got %v, want %v", got, want)
 		}
 		nh := gnmi.Get(t, args.dut, gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance).Afts().NextHop(nhIndexInst).State())
-		if got, want := nh.GetIpAddress(), atePort2.IPv4; got != want {
-			t.Errorf("next-hop is incorrect: got %v, want %v", got, want)
+		// for devices that return  the nexthop with resolving it recursively. For a->b->c the device returns c
+		if got := nh.GetIpAddress(); got != atePort2.IPv4 {
```
xw-g, line 311: minor: how about `if got := nh.GetIpAddress(); got != atePort2.IPv4 && got != ateIndirectNH`?  (you can merge the comment together too.)

xw-g (approved):

PR #862, opened 2022-12-07
gNOI-3.3: Add SkipStandbyManagementInterfaceCheck

mojiiba:
We have an issue for generating telemetry event for the backup management interface if the streaming start before the standby rp is ready. I wonder if we can have this deviation until the issue is fixed.


mojiiba:
this is not need anymore so closing the pull

mojiiba:
we found out the fix for this easier than we thought, so closing this pull try to provide the fix as soon as possible.

PR #863, opened 2022-12-07
added deviations to gNOI-5.1

Hobbydos:
Creating this PR as request from @xw-g for further discussions

### DIFF
--- a/feature/gnoi/system/tests/ping_test/ping_test.go
+++ b/feature/gnoi/system/tests/ping_test/ping_test.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"testing"
 
+	"github.com/openconfig/featureprofiles/internal/deviations"
 	"github.com/openconfig/featureprofiles/internal/fptest"
 	spb "github.com/openconfig/gnoi/system"
 	tpb "github.com/openconfig/gnoi/types"
@@ -110,7 +111,11 @@ func TestGNOIPing(t *testing.T) {
 	if len(ipv6Addrs) == 0 {
 		t.Fatalf("Failed to get a valid IPv6 loopback address: %+v", ipv6Addrs)
 	}
-
+	if *deviations.ExplicitInterfaceInDefaultVRF {
+		if dut.Vendor() == ondatra.NOKIA {
+			fptest.AssignToNetworkInstance(t, dut, lbIntf, "mgmt", 0)
+		}
+	}
 	commonExpectedIPv4Reply := &spb.PingResponse{
 		Source:   ipv4Addrs[0].GetIp(),
 		Time:     minimumPingTime,
--- a/feature/gnoi/system/tests/ping_test/ping_test.go
+++ b/feature/gnoi/system/tests/ping_test/ping_test.go
@@ -112,7 +112,7 @@ func TestGNOIPing(t *testing.T) {
 		t.Fatalf("Failed to get a valid IPv6 loopback address: %+v", ipv6Addrs)
 	}
 	if *deviations.ExplicitInterfaceInDefaultVRF {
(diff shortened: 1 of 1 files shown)


Hobbydos:
cc @xw-g @robshakir

feature/gnoi/system/tests/ping_test/ping_test.go
```diff
@@ -110,7 +111,11 @@ func TestGNOIPing(t *testing.T) {
 	if len(ipv6Addrs) == 0 {
 		t.Fatalf("Failed to get a valid IPv6 loopback address: %+v", ipv6Addrs)
 	}
-
+	if *deviations.ExplicitInterfaceInDefaultVRF {
+		if dut.Vendor() == ondatra.NOKIA {
```
robshakir, line 115: I don't think that we need this `dut.Vendor` check here, since the flags can be enabled explictly for specific test runs.

robshakir (changes requested):

feature/gnoi/system/tests/ping_test/ping_test.go
Hobbydos: That's fair, but the only discrepancy here is that there's an expectation the interface to be part of mgmt NI. If no other vendor require that, I can remove the vendor flag.

feature/gnoi/system/tests/ping_test/ping_test.go
```diff
@@ -110,7 +111,9 @@ func TestGNOIPing(t *testing.T) {
 	if len(ipv6Addrs) == 0 {
 		t.Fatalf("Failed to get a valid IPv6 loopback address: %+v", ipv6Addrs)
 	}
-
+	if *deviations.ExplicitInterfaceInDefaultVRF {
+		fptest.AssignToNetworkInstance(t, dut, lbIntf, "mgmt", 0)
```
xw-g, line 115: "mgmt" should be the default VRF? This test case currently does not test ping from non-default VRF, right?

feature/gnoi/system/tests/ping_test/ping_test.go
xw-g: @Hobbydos, I don't see any part in this test expects the ping from `mgmt` NI. (it's a TODO from the MD though). If so, I think you just need to put the interface in the default VRF, right?

feature/gnoi/system/tests/ping_test/ping_test.go
Hobbydos: It could be default VRF if your connection entrypoint is via default VRF and the gNOI server is operational in that VRF. In my lab I'm using 'mgmt', but it's a variable, if in your bindings/textproto file your connection IP to gNOI is through the default VRF, I can change this to be default. The test-case simply pings from the obtained connection endpoint, I think it doesn't look at the NI, it's just which NI the client connected from

xw-g: I think there are 2 moving parts: the gNOI wanted behavior (in terms of specifying the VRF) and different OS implementations. If the source IP is not specified, I do feel like that the current gNOI seems allow different OS implementation choose their own VRF or source IP. 

This test currently only focuses on ping in the default VRF. Maybe let it keep it that way?  

BTW, `ExplicitInterfaceInDefaultVRF` should be only about the adding interface in the default VRF, not arbitrary VRF.

Hobbydos: changed to default VRF

xw-g:
LGTM.

Because this PR change is gated by an approved deviation flag, and the usage is approved. I'll merge it without other Quattro team member approval. 

I have some other concerns about the test itself. But I'll follow up offline.

xw-g (approved):

PR #864, opened 2022-12-08
TE-3.2: fix migration to use election id handshake

mojiiba:
The upgrade to do automatic election id sync missed to upgrade this test.

### DIFF
--- a/feature/gribi/ate_tests/weighted_balancing_test/port_flap_rebalanced_test.go
+++ b/feature/gribi/ate_tests/weighted_balancing_test/port_flap_rebalanced_test.go
@@ -117,17 +117,10 @@ func TestPortFlap(t *testing.T) {
 		WithPersistence()
 	c.Start(ctx, t)
 	defer c.Stop(t)
-	defer func() {
-		// Flush all entries after test.
-		if err := gribi.FlushAll(c); err != nil {
-			t.Errorf("Cannot flush: %v", err)
-		}
-	}()
 	c.StartSending(ctx, t)
 	if err := awaitTimeout(ctx, c, t, time.Minute); err != nil {
 		t.Fatalf("Await got error during session negotiation: %v", err)
 	}
-
 	gribi.BecomeLeader(t, c)
 
 	// Flush all entries before test.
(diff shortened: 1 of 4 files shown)


feature/gribi/ate_tests/weighted_balancing_test/port_flap_rebalanced_test.go
```diff
@@ -116,16 +117,21 @@ func TestPortFlap(t *testing.T) {
 		WithPersistence()
 	c.Start(ctx, t)
 	defer c.Stop(t)
+	defer func() {
```
liulk, line 120: Since we already flush before test, why do we need to flush after test?

liulk (changes requested):
Also update otg_tests?

https://github.com/openconfig/featureprofiles/blob/main/feature/gribi/otg_tests/weighted_balancing_test/port_flap_rebalanced_test.go

feature/gribi/ate_tests/weighted_balancing_test/port_flap_rebalanced_test.go
mojiiba: just copying what is done for other tests. I also think flushing after test is not necessary, assuming all gribi tests do the same. will remove that and update the otg version as well.

liulk: Should be fine even if the test does not clean up after itself. The binding is capable of doing a flush but it has to be requested in the binding.

https://github.com/openconfig/featureprofiles/blob/82732de3b989fa7c8399903e02ed2534f01ee6b3/topologies/binding/reset.go#L120

mojiiba: great, was not aware of this.

mojiiba: @liulk  addressed your comments, also applied the same missing change to TE-11.1

mojiiba: @liulk could u please merge this pr?

PR #865, opened 2022-12-08
Fixes for P4RT-5.2

singh-prem:
1. Removed following line from description as its not valid for traceroute packet-out scenario
   Install P4RT table entries required for traceroute.
2. Changed submit_to_egress_pipeline to submit_to_ingress
3. Removed Ethernet Headers from packet as the packet is subjected to FIB lookup in submit_to_ingress scenario before sending out (the packet needs to be an L3 packet).
4. TTL/Hoplimit=0 packets should be dropped by the device.


prinikasn (commented):
/gcbrun

prinikasn (approved):

PR #871, opened 2022-12-09
Uplift rt1.2 for route_policy under neighbor

singh-prem:
1. Added global address families
2. Added route-policy configuration under neighbor (under deviation if route-policy is not supported under peer group)
3. Fixed verifyPortsUp to check status of ports used in the test.
4. added two deviations

### DIFF
--- a/feature/bgp/policybase/ate_tests/route_installation_test/route_installation_test.go
+++ b/feature/bgp/policybase/ate_tests/route_installation_test/route_installation_test.go
@@ -167,6 +167,7 @@ func bgpCreateNbr(localAs, peerAs uint32, policy string) *oc.NetworkInstance_Pro
 	global.RouterId = ygot.String(dutDst.IPv4)
 	global.As = ygot.Uint32(localAs)
 	global.GetOrCreateAfiSafi(oc.BgpTypes_AFI_SAFI_TYPE_IPV4_UNICAST).Enabled = ygot.Bool(true)
+	global.GetOrCreateAfiSafi(oc.BgpTypes_AFI_SAFI_TYPE_IPV6_UNICAST).Enabled = ygot.Bool(true)
 
 	// Note: we have to define the peer group even if we aren't setting any policy because it's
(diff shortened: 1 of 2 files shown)


thesrinath (approved):

feature/bgp/policybase/ate_tests/route_installation_test/route_installation_test.go
```diff
@@ -181,13 +184,21 @@ func bgpCreateNbr(localAs, peerAs uint32, policy string) *oc.NetworkInstance_Pro
 			nv4.PeerGroup = ygot.String(peerGrpName)
 			nv4.PeerAs = ygot.Uint32(nbr.as)
 			nv4.Enabled = ygot.Bool(true)
-			nv4.GetOrCreateAfiSafi(oc.BgpTypes_AFI_SAFI_TYPE_IPV4_UNICAST).Enabled = ygot.Bool(true)
+			af4 := nv4.GetOrCreateAfiSafi(oc.BgpTypes_AFI_SAFI_TYPE_IPV4_UNICAST)
+			af4.Enabled = ygot.Bool(true)
+			if *deviations.RoutePolicyUnderPeerGroup {
```
xw-g, line 189: The deviation name says "policy under peer group", but the actually code says "policy under address family"?

xw-g (changes requested):

feature/bgp/policybase/ate_tests/route_installation_test/route_installation_test.go
singh-prem: HI @xw-g, thanks for the review. The above code protected under deviation is a workaround for not supporting route-policy under peer-group. The this condition is paired with a previous condition in code:
`
if policy != "" && !*deviations.RoutePolicyUnderPeerGroup {
		pg.GetOrCreateApplyPolicy().ImportPolicy = []string{policy}
	}
`
If we are deviated, then we skip the route-policy under peer-group but apply under address family.

vosipchu (approved):

thesrinath (approved):

PR #892, opened 2022-12-13
TE-15.1: Removing not-required Static-route protocol config

trathod1:
Considering there is no static-route configured by the test, removing `GetOrCreateProtocol`  for `STATIC`. As without route configuration `static-routes` container is not configured.


trathod1:
@xw-g @robshakir

xw-g (approved):

xw-g:
@bstoll I think it makes sense that we don't need static route in the non-default VRF, as part of setting the pre-test env. WDYT?

PR #896, opened 2022-12-13
OTG Test - TE-6.2: Route Removal In Non Default VRF

### LINKED ISSUE #743: OTG: TE-6.2: Route Removal In Non Default VRF

octpetre:
OTG conversion for **_TE-6.2: Route Removal In Non Default VRF_**
Fixes #743 
Test log link [**here**](https://drive.google.com/file/d/1XXNTZvzXGjazCR-gnxyxakVAbl0ZO3CD/view?usp=share_link)


michaelore (approved):

feature/gribi/otg_tests/route_removal_non_default_vrf_test/route_removal_non_default_vrf_test.go
```diff
@@ -0,0 +1,435 @@
...
+	gnmi.Replace(t, dut, gnmi.OC().NetworkInstance(nonDefaultVRF).Config(), nonDefaultNI)
+}
+
+// networkInstance creates an OpenConfig network instance with the specified name
+func networkInstance(t *testing.T, name string) *oc.NetworkInstance {
+	d := &oc.Root{}
+	ni := d.GetOrCreateNetworkInstance(name)
+	ni.Description = ygot.String("Non Default routing instance created for testing")
+	ni.Type = oc.NetworkInstanceTypes_NETWORK_INSTANCE_TYPE_L3VRF
+	ni.Enabled = ygot.Bool(true)
```
michaelore, line 347: I think you need to manually merge in some changes made to the ate_test version. In particular https://github.com/openconfig/featureprofiles/commit/b47bcfc517747439ab734566a2771acb2435eeec , which removes references to fields that no longer exist and prevent successful compilation.

Apologies for our delay in review that let this get out of sync.

feature/gribi/otg_tests/route_removal_non_default_vrf_test/route_removal_non_default_vrf_test.go
```diff
@@ -0,0 +1,441 @@
...
+func testTraffic(t *testing.T, ate *ondatra.ATEDevice, config gosnappi.Config) int64 {
+	ate.OTG().StartTraffic(t)
+	time.Sleep(15 * time.Second)
+	ate.OTG().StopTraffic(t)
+	otgutils.LogFlowMetrics(t, ate.OTG(), config)
+
+	flowMetric := gnmi.Get(t, ate.OTG(), gnmi.OTG().Flow("Flow").State())
+	txPackets := flowMetric.GetCounters().GetOutPkts()
+	rxPackets := flowMetric.GetCounters().GetInPkts()
+	lossPct := int64((txPackets - rxPackets) * 100 / txPackets)
```
michaelore, line 370: Pardon me, I noticed one more thing.

Can you add a separate check for txPackets being zero? Then we can have a clearer error message in case of division by zero.

michaelore: Also, I'd like for the lossPct checks at the callsites to be clearer... It's not clear what the return value of "testTraffic" is unless we go look at the function.

I'm assuming you changed this as using the LossPct() telemetry doesn't work (if that does work, that would be a good solution)

Could you split the lossPct calculation into a separate test helper with something like "lossPct" in the name?

octpetre: I've added the check for 0 txPackets and renamed the call to **getTrafficPacketLoss**

michaelore: Hm, "get" makes it sound like it just takes a GNMI Get (or maybe a little bit more work) to get the packet loss. We also generate traffic and block for 15 seconds...

I think a slightly different name that makes it sound like more work is done such as computeTrafficPacketLoss would be okay.

See also guidance at https://google.github.io/styleguide/go/decisions#getters

As another option, it may be good to split into two functions so we can have clear names for both steps. Something like:

func generateTraffic(t *testing.T, ate *ondatra.ATEDevice, config gosnappi.Config) {
	t.Helper()
	ate.OTG().StartTraffic(t)
	time.Sleep(15 * time.Second)
	ate.OTG().StopTraffic(t)
	otgutils.LogFlowMetrics(t, ate.OTG(), config)
}

func getTrafficPacketLoss(t *testing.T, ate *ondatra.ATEDevice) int64 {
	t.Helper()
	ate.OTG().StartTraffic(t)
	time.Sleep(15 * time.Second)
	ate.OTG().StopTraffic(t)
	otgutils.LogFlowMetrics(t, ate.OTG(), config)

	flowMetric := gnmi.Get(t, ate.OTG(), gnmi.OTG().Flow("Flow").State())
	txPackets := flowMetric.GetCounters().GetOutPkts()
	if txPackets == 0 {
		t.Fatal("Got 0 txPackets. Want >0")
	}
	...
}

michaelore: Also, whether splitting the function are not, I think these should be marked as test helpers with t.Helper(). See the example in the prior comment.

Also, t.Fatal is better here so we don't go forward testing with an invalid "0" return value.

See guidance on test helpers at https://google.github.io/styleguide/go/decisions#test-helpers

michaelore (changes requested):

feature/gribi/otg_tests/route_removal_non_default_vrf_test/route_removal_non_default_vrf_test.go
octpetre: Split the call into 2 functions and renamed to computeLossPct. Added t.Fatal in case of 0 tx packets

michaelore (approved):
Thanks!

PR #899, opened 2022-12-13
rundata dut add Nokia support

liulk:


LimeHat:
Hi Likai @liulk ,
You will be able to fetch vendor & model from /components/component[name=Chassis]/state but it's not exposed in the current build you guys have. Should be available in the next build.
Let me check about the sw version, will update you soon.

Part. no is not a random string, 3HE.... is a part numbering scheme we use at Nokia (each corresponds to a specific chassis type or a component)

E.g.
```
    "updates": [
      {
        "Path": "openconfig-platform:components/component[name=Chassis]/state",
        "values": {
          "openconfig-platform:components/component/state": {
            "description": "7250 IXR-10e",
            "empty": false,
            "id": "Chassis",
            "mfg-date": "10132021",
            "mfg-name": "Nokia",
            "name": "Chassis",
            "oper-status": "openconfig-platform-types:ACTIVE",
            "part-no": "3he16997AA",
            "removable": false,
            "serial-no": "[REDACTED]",
            "type": "openconfig-platform-types:CHASSIS"
          }
        }
      }
    ]
  }
]
```

internal/rundata/dut.go
```diff
@@ -39,8 +39,8 @@ type dutInfo struct {
 
 // setFromComponentChassis sets dutInfo from the first component of type CHASSIS.
 //
-//   - vendor from mfg-name (Arista, Cisco).
-//   - model from either description (Cisco, Juniper) or part-no (Arista).
+//   - vendor from mfg-name (Arista, Cisco, newer Nokia).
```
xw-g, line 42: nit: remove "newer"? so that we don't need to remember to update it later (next build will support it).

internal/rundata/dut.go
```diff
@@ -81,8 +81,13 @@ func (di *dutInfo) setFromComponentChassis(ctx context.Context, y components.Y)
 	}
 
 	if desc := component.Description; desc != nil {
-		if di.model == "" || strings.HasPrefix(*desc, "Cisco ") {
-			di.model = *component.Description
+		switch {
+		case strings.HasPrefix(di.vendor, "Cisco"):
+			fallthrough
+		case strings.HasPrefix(di.vendor, "Nokia"):
```
xw-g, line 87: Per the sample output (https://github.com/openconfig/featureprofiles/pull/899#issuecomment-1349986107), `Description` does not contain Nokia, instead it has the actually chassis model name?

internal/rundata/dut.go
```diff
@@ -120,30 +125,47 @@ func (di *dutInfo) setFromComponentOS(ctx context.Context, y components.Y) {
 	}
 }
 
-// setFromLLDP sets dutInfo vendor from LLDP system-description (Juniper).  This is less
-// reliable because LLDP config can be changed.
+// setFromLLDP sets dutInfo vendor (Juniper, older Nokia) and osver (older Nokia) from
```
xw-g, line 128: Maybe we don't need to support the `older Nokia`, as the needed feature will be available in next build soon?

internal/rundata/dut.go
liulk: The "Nokia" prefix is checked on di.vendor which comes from mfg-name, not the description.

xw-g (approved):

PR #900, opened 2022-12-13
TE-11.1: Single Backup NHG test

bstoll:
This PR modifies the TE-11.1 Single Backup NHG test plan and code to cover the use case in more detail.

* Traffic should be able to ingress a VRF and the backup NHG should trigger a lookup in a different table.  Previous test plan had traffic ingress at default table and forward into a VRF.
* Update AFT telemetry checks to allow more time for the telemetry to become correct.  There can be some delay between a GRIBI change and when the change is seen in GNMI AFT.
* Add deviation to support traffic ingress via a VRF selection policy if required.
* Add deviation to skip hierarchical backup next hop group test.  If a DUT has a default route present in the default table, the primary path is always valid because it will utilize the default route.  This test requires the DUT to have no default route present to trigger backup NHG use.

### DIFF
--- a/feature/experimental/backup_nhg/ate_tests/backup_nhg_test/README.md
+++ b/feature/experimental/backup_nhg/ate_tests/backup_nhg_test/README.md
@@ -9,28 +9,41 @@ containing a single NH.
 
 *   Connect ATE port-1 to DUT port-1, ATE port-2 to DUT port-2, and ATE port-3
     to DUT port-3.
-*   TODO: Create a non-default VRF, VRF-3, which includes DUT port-3.
-*   TODO: Create a static default route in the VRF-3, pointing to ATE port-3.
+*   Create a non-default VRF, VRF-1. Assign DUT port-1 to VRF-1.
+*   All GRIBI NHGs and NHs will be installed to the default VRF.
 *   Connect gRIBI client to DUT with persistence `PRESERVE`, redundancy
     `SINGLE_PRIMARY`, with election ID 1.
-*   Install an IPv4Entry in the default VRF pointing to ATE port-2 for prefix
-    198.51.100.0/24, with a backup nexthop group pointing to ATE port-3 (TODO:
-    change ATE port-3 to VRF-3).
+
+To validate GRIBI backup next hop group behavior:
+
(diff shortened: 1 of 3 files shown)


feature/experimental/backup_nhg/ate_tests/backup_nhg_test/README.md
```diff
@@ -9,28 +9,31 @@ containing a single NH.
...
     *   Interface DUT port-2 is disabled.
-*   Remove all previously installed IPv4Entry routes. Create an entry for
-    198.51.100.0/24 with a next-hop of 192.0.2.254/32. Inject a second entry
-    with 192.0.2.254/32 resolved to ATE port-2. Specify a backup NHG pointing to
-    ATE port-3 (TODO: change ATE port-3 to VRF-3) for the 198.51.100.0/24
-    entry’s NHG.
-    *   Remove the entry for 192.0.2.254/32, and ensure that traffic is
-        forwarded to ATE port-3 for destinations in 198.51.100.0/24.
+
+To validate hierarchical backup next hop group behavior:
```
xw-g, line 29: let's investigate the default 0/0 solution/impact.

bstoll: I think the solution we discussed will work, so I removed the deviation/skip feature for the hierarchical test.

feature/experimental/backup_nhg/ate_tests/backup_nhg_test/README.md
```diff
@@ -9,28 +9,31 @@ containing a single NH.
 
 *   Connect ATE port-1 to DUT port-1, ATE port-2 to DUT port-2, and ATE port-3
     to DUT port-3.
-*   TODO: Create a non-default VRF, VRF-3, which includes DUT port-3.
-*   TODO: Create a static default route in the VRF-3, pointing to ATE port-3.
+*   Create a non-default VRF, BACKUP-NHG-1.  Assign DUT port-1 to BACKUP-NHG-1.
```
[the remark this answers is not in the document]
bstoll: I updated this to be VRF-1.

PR #910, opened 2022-12-15
Guard ipv4 enabled when IPv4MissingEnabled is set

kjahed:
Enabled should not be configured for IPv4 subinterfaces when -deviation_ipv4_missing_enabled is used.


vosipchu (approved):

dplore:
assigning @liulk for functional test style and golang review

liulk (approved):
Thanks for propagating the deviation to other tests!

PR #912, opened 2022-12-15
TE-2.1: gRIBI IPv4 Entry : Adding ipv6 entry for this testcase

avaneesh90:
TE-2.1: gRIBI IPv4 Entry : Adding ipv6 attribute entry for this testcase


xw-g:
Can you share some insight of why to add the IPv6 addresses? I see it's not part of the README? Thanks.

feature/gribi/ate_tests/ipv4_entry_test/ipv4_entry_test.go
```diff
@@ -57,36 +57,48 @@ var (
...
 	}
 	atePort3 = attrs.Attributes{
 		Name:    "atePort3",
 		Desc:    "ATE Port 3",
 		IPv4:    "192.0.2.10",
+		IPv6:    "2001:db8::192:0:2:a",
 		IPv4Len: 30,
+		IPv6Len: 126,
 	}
 )
```
sachendras, line 103: What happens if we just remove the IPv6 addresses here and instead just enable IPv6 under the interface? Will that pass the test?

avaneesh90: I have removed the configuring ipv6 address and added enabling ipv6 in the interface

sachendras (approved):

sachendras:
Internal Google agreement in b/264809673

PR #920, opened 2022-12-19
Create README.md for DP 1.10 Mixed strict priority traffic and WRR traffic test case

sezhang2:
Convert DP 1.10 Mixed strict priority traffic and WRR traffic test case to README


dplore (approved):
LGTM

PR #938, opened 2022-12-21
gnoi 5.3 test case commit with changes to feature gnoi directory

nraghavendran:
editing gnoi test case to use component chassis and moving to gnoi/tests folder


sezhang2 (approved):

nraghavendran:
merging the changes

PR #944, opened 2023-01-03
change minTraceroutePktSize

Hobbydos:
ref: https://github.com/openconfig/featureprofiles/pull/687

### DIFF
--- a/feature/gnoi/system/tests/traceroute_test/README.md
+++ b/feature/gnoi/system/tests/traceroute_test/README.md
@@ -33,22 +33,8 @@ source, destination, max_ttl, do_not_fragment and L4Protocols.
         *   Set to 1
         *   TODO: Set to 255
     *   Do_not_fragment: Check the following cases when DF bit is:
-
-        *   Set: *Intermediate hop router should have link MTU set to less than
-            the traceroute packet size.
-
         *   TODO: Unset
-
-    *   Size:
-
-        *   Set to min packet size of 64, ethernet packet size of 1512, max mtu
-            of jumbo frame 9202, and value slightly bigger than the egress
-            interface MTU of a transit router to test do_not_fragment.
-        *   TODO: verify these for vlan tagged vs untagged packets. May need +4
-            bytes
-
     *   L4Protocol: set as:
-
         *   ICMP
         *   TCP
         *   UDP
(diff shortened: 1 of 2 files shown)


xw-g:
@bstoll I think this PR itself it's ok because I don't see that the test currently specifies the size. The default value would be upto the system/OS implementation. Raising up the bar here seems fine to me. Also reference https://github.com/openconfig/featureprofiles/pull/687

My question is more about the test itself. The test plan talks about different packet size (not TODO), but I don't see it in the code. Also, I don't see gNOI API supports packet size. Shall we completely remove the mentioning of packet size at all from this test? Thoughts?

bstoll:
I agree that in the current version of this RPC spec it doesn't make a lot of sense to look into the packet size.  It would be useful to be able to specify the packet size but that would be a change to the GNOI spec first before we can test for it.

I think we could completely remove the packet size from this version and tie it as a new feature requiring GNOI spec to be updated first.

xw-g:
Thanks, @bstoll. SGTM. 

@Hobbydos can you help use this PR to remove all the packet size related code and in the README too? Thanks. 

cc: @vcheny @sachendras @vosipchu. We plan to use this PR to remove packet size related tests (README part and code part), because gNOI API currently does not even support it.  See https://github.com/openconfig/featureprofiles/pull/944#issuecomment-1372598410 and https://github.com/openconfig/featureprofiles/pull/944#issuecomment-1374006566 for more context. 

It should have 0 impact on passed tests. Ping me if any objection, otherwise I"ll merge the PR upon my own LGTM.

Hobbydos:
> Thanks, @bstoll. SGTM.
> 
> @Hobbydos can you help use this PR to remove all the packet size related code and in the README too? Thanks.
> 
> cc: @vcheny @sachendras @vosipchu. We plan to use this PR to remove packet size related tests (README part and code part), because gNOI API currently does not even support it. See [#944 (comment)](https://github.com/openconfig/featureprofiles/pull/944#issuecomment-1372598410) and [#944 (comment)](https://github.com/openconfig/featureprofiles/pull/944#issuecomment-1374006566) for more context.
> 
> It should have 0 impact on passed tests. Ping me if any objection, otherwise I"ll merge the PR upon my own LGTM.

Removed pktSize associations to test file and readme file, PTAL and let me know if anything needs changing.

Hobbydos:
@xw-g  @jjing1122 Does this PR look good for merging? Once confirmed, I'll add deviation to this test.

xw-g (approved):

PR #939, opened 2022-12-22
Mass push for NI Deviation

Hobbydos:
This PR brings forth changes for assigning interfaces in the Default network instance.

PR #948, opened 2023-01-06
NI Deviation requiring testing.TB interface reference update.

trathod1:
Followup changes based on - https://github.com/openconfig/featureprofiles/pull/939#discussion_r1062924708


xw-g:
@liulk I'm not familiar with our choice of T and TB, maybe you can help offer some opinions?

BTW: the usage of the deviation is approved and Quattro team is ok with it.

liulk (changes requested):
Can we change fptest.AssignToNetworkInstance to just take testing.TB, not *testing.T? Also if not already, make sure it calls t.Helper() the first thing.

liulk:
To provide some rationale: *testing.T can be converted to a testing.TB but not vice versa, so helpers in a shared package should only accept testing.TB. This is to avoid having large scale changes like this PR.

trathod1:
> Can we change fptest.AssignToNetworkInstance to just take testing.TB, not *testing.T? Also if not already, make sure it calls t.Helper() the first thing.

Updated deviation as per comment. Restored testing.TB argument in session.go and setup_test.go as before.
I understand, these changes would be applicable to [`fptest.SetPortSpeed` ](https://github.com/nokia/featureprofiles/blob/99ea33270e5e6d3ca3e97278605413ac40ef5740/internal/fptest/portspeed.go#L44)deviation also. Will update it in separate PR.

trathod1:
@liulk Does this look fine ? Will raise a PR with similar changes for fptest.SetPortSpeed based on it.

xw-g:
The deviation usage is LGTMed, waiting for @liulk to LGTM the TB changes.

liulk (approved):

PR #953, opened 2023-01-06
fixes for isis RT 2.1

manan-patel:
Fixes:

1: addressed comment for protocol enable flag
-https://github.com/openconfig/featureprofiles/pull/745#issuecomment-1314620480
2: Removed isis instance config and status check call
-if the instance configuration is needed, we need to change it to digits 1-65535, and multi-instance flag has to be enabled on ATE side for parameters to match and neighborship to establish
3: Added deviation flags for
-Global level required
-Missing isis interface enable flag
-Missing isis hello padding adaptive mode
-Isis single topology 
-Missing isis enable flag for af, afi and level at global mode
4: In TestTraffic tc, with no link metric set, prefixes advertised with wrong metric and hence traffic will fail


feature/experimental/isis/ate_tests/internal/session/session.go
```diff
@@ -106,17 +106,32 @@ func ProtocolPath() *networkinstance.NetworkInstance_ProtocolPath {
 func addISISOC(dev *oc.Root, areaAddress, sysID, ifaceName string) {
 	inst := dev.GetOrCreateNetworkInstance(*deviations.DefaultNetworkInstance)
 	prot := inst.GetOrCreateProtocol(PTISIS, ISISName)
-	prot.Enabled = ygot.Bool(true)
+	if !*deviations.ISISprotocolEnabledNotRequired {
+		prot.Enabled = ygot.Bool(true)
+	}
 	isis := prot.GetOrCreateIsis()
 	glob := isis.GetOrCreateGlobal()
-	glob.Instance = ygot.String(ISISName)
```
robshakir, line 112: Discussed with @vosipchu -- this looks fine to remove, per the discussion in https://github.com/openconfig/public/pull/710. Since multi-instance is not being used this leaf is not required.

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
```diff
@@ -60,23 +60,38 @@ func TestBasic(t *testing.T) {
 	}
 	isisRoot := session.ISISPath()
 	port1ISIS := isisRoot.Interface(ts.DUTPort1.Name())
-	if err := check.Equal(isisRoot.Global().Instance().State(), session.ISISName).AwaitFor(time.Second, ts.DUTClient); err != nil {
```
robshakir, line 63: What is this replaced by? Is there an alternate check that we should have that determines that IS-IS is running?

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
```diff
@@ -60,23 +60,38 @@ func TestBasic(t *testing.T) {
...
-		t.Fatalf("IS-IS failed to configure: %v", err)
-	}
 	// There might be lag between when the instance name is set and when the
 	// other parameters are set; we expect the total lag to be under 5s
 	deadline := time.Now().Add(time.Second * 5)
 
 	t.Run("read_config", func(t *testing.T) {
-		for _, vd := range []check.Validator{
+		data := []check.Validator{
```
robshakir, line 68: Suggest to call this something clearer than `data`. It's not really data but rather a set of checks that we are running.

feature/experimental/isis/ate_tests/internal/session/session.go
```diff
@@ -106,17 +106,32 @@ func ProtocolPath() *networkinstance.NetworkInstance_ProtocolPath {
...
 	glob.Net = []string{fmt.Sprintf("%v.%v.00", areaAddress, sysID)}
 	glob.GetOrCreateAf(oc.IsisTypes_AFI_TYPE_IPV4, oc.IsisTypes_SAFI_TYPE_UNICAST).Enabled = ygot.Bool(true)
 	glob.GetOrCreateAf(oc.IsisTypes_AFI_TYPE_IPV6, oc.IsisTypes_SAFI_TYPE_UNICAST).Enabled = ygot.Bool(true)
+	if *deviations.IsisAfMetricStyleWideLevelRequired {
+		level := isis.GetOrCreateLevel(2)
+		level.MetricStyle = 2
```
robshakir, line 119: Please use the named enumeration for this (this is an enum type [doc](https://github.com/openconfig/public/blob/master/release/models/isis/openconfig-isis-types.yang#L306)) - it's named `oc.IsisTypes_MetricStyle_WIDE_METRIC`

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
manan-patel: At the global container, we configure the net and address family which is validated under the "read_config" section along with level-capability. None of the other leaf seems helpful in determining IS-IS is running.

My suggestion would be to move the "net" check here.

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
manan-patel: Will fix it

feature/experimental/isis/ate_tests/internal/session/session.go
manan-patel: Will fix it

prinikasn (commented):
/gcbrun

PR #954, opened 2023-01-06
LLDP disabled change

Hobbydos:
Changed the return value if `!present`, when `!lldpEnabled`. Both `intf` and `present` shouldn't be defined here.


xw-g:
Current changes LGTM. Let me work on the presubmit.

xw-g:
b/264909809 internal bug. @vosipchu LGTMed offline.

xw-g (approved):

PR #955, opened 2023-01-06
added nokia to gRIBIDaemons

Hobbydos:
added process name to gRIBIDaemons for Nokia


xw-g:
merging this one with one LGTM because it's only affecting NOKIA.

PR #956, opened 2023-01-06
Update README.md

sezhang2:
Added the test topology info.

### DIFF
--- a/feature/gnoi/system/tests/ping_test/README.md
+++ b/feature/gnoi/system/tests/ping_test/README.md
@@ -6,6 +6,11 @@ Validate ping request and response work on the device, with parameters covering
 VRF, source-IP address and different packet sizes. L3 Protocol used will be
 ICMP, which is default.
 
+## Topology
+
+*   DUT
+    * Note: The test may require the use of ATE for new tests in future.  
+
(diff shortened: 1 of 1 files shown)


xw-g (commented):
Maybe we can add a bit more wording to specifically call out that currently the test only ping itself for the purpose of exercising the gNOI API? 

Also, can you help update the code comment to keep it in sync? Thanks. 
e.g.https://github.com/openconfig/featureprofiles/blob/4f5daf1845581067c9a443c2b6b74d4d71bc22e6/feature/gnoi/system/tests/ping_test/ping_test.go#L77 still says the topology is DUT-ATE.

sezhang2:
> Maybe we can add a bit more wording to specifically call out that currently the test only ping itself for the purpose of exercising the gNOI API?
> 
> Also, can you help update the code comment to keep it in sync? Thanks. e.g.
> 
> https://github.com/openconfig/featureprofiles/blob/4f5daf1845581067c9a443c2b6b74d4d71bc22e6/feature/gnoi/system/tests/ping_test/ping_test.go#L77
> 
> still says the topology is DUT-ATE.

Done

PR #957, opened 2023-01-09
Populate rundata in all feature tests.

liulk:


liulk:
Going to abandon this PR for now and split rundata to its own generated file, as discussed.

PR #958, opened 2023-01-09
gNOI-3.4: ensure that gnoi reboot status request is sent for the active rp when the device is dual-rp

mojiiba:
Our current implementation saves the context of the gnoi reboot with delay for the active rp. Thus, the component name for status request should be the active rp.


prinikasn (commented):
/gcbrun

feature/gnoi/system/tests/chassis_reboot_status_and_cancel_test/chassis_reboot_status_and_cancel_test.go
```diff
@@ -206,3 +188,16 @@ func TestCancelReboot(t *testing.T) {
...
 }
+
+func getSubCompPath(t *testing.T, dut *ondatra.DUTDevice) *tpb.Path {
+	t.Helper()
+	controllerCards := components.FindComponentsByType(t, dut, oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_CONTROLLER_CARD)
+	if len(controllerCards) == 0 {
+		t.Fatal("No controller card components found in DUT.")
+	}
+	activeRP := controllerCards[0]
+	if len(controllerCards) == 2 {
```
prinikasn, line 199: Would there be a possible scenario where controllerCards is greater than 2?

mojiiba: I am not aware of a device with more than two rps, so the length should be 1 or two.

vosipchu: +1, we will have devices with 2 control modules only

prinikasn (approved):

PR #961, opened 2023-01-10
Add P4RT Deviation, Args and Implementation Guide

Ankur19:


feature/experimental/p4rt/README.md
```diff
@@ -10,12 +10,26 @@ This document specifies the requirements for p4rt test implementation.
     1.  WBB P4 Protobuf file:
         https://github.com/openconfig/featureprofiles/blob/main/feature/experimental/p4rt/wbb.p4info.pb.txt
 
-3.  The client should make use of Ondatra Raw API
-    `dut.RawAPIs().P4RT().Default(t)`
+3.  Tests should instantiate new P4RT clients using
+    `p4rt_client.NewP4RTClient()` function. This instantiates the
+    `StreamTermErr` channel required to investigate errors when the p4rt stream
+    terminates.
```
liulk, line 16:
```suggestion
3.  Tests should create new P4RT clients using the
    `p4rt_client.NewP4RTClient()` function, which sets up the
    `StreamTermErr` channel required to check errors when the p4rt stream
    terminates.
```

Just a nit to use shorter words because they are easier to read.

feature/experimental/p4rt/README.md
```diff
@@ -10,12 +10,26 @@ This document specifies the requirements for p4rt test implementation.
...
 
-4.  Tests should get the P4RT Node Name by walking the Components OC tree.
+4.  Tests should make use of Ondatra Raw API `dut.RawAPIs().P4RT().Default(t)`
+    during client instantiation.
+
+5.  To avoid indefinite blocking during Get response
+    `client.StreamChannelGetArbitrationResp()`, tests should make use of
+    `p4rtutils.StreamTermErr()` function which returns an Error if the p4rt
+    stream has been terminated.
```
liulk, line 24: Let's explain why the indefinite blocking happens, namely that GetArbitrartion() blocks on the cond.Wait() after the cstream Go routine exits, which could be fixed by replacing cond.Wait() with a sleep (or a select on StreamTermErr channel).

Let's make it clear that the suggested workaround here does not fix the underlying issue. Use a code example to illustrate what the workaround looks like.

feature/experimental/p4rt/README.md
```diff
@@ -10,12 +10,26 @@ This document specifies the requirements for p4rt test implementation.
...
+6.  Tests should get the P4RT Node Name by walking the Components OC tree.
     Components of type `INTEGRATED_CIRCUIT` should have child Components of type
     `PORT`. These PORT Components can be mapped to currently reserved Interfaces
     using the `hardware-port` leaf in the Interfaces tree. Such an
     implementation already exists in `p4rtutils` library:
     `p4rtutils.P4RTNodesByPort()`.
+
+7.  If P4RT Node Names cannot be resolved by walking the Components tree, use
+    deviation flag `deviation_explicit_p4rt_node_component` and pass the node
+    names through args `arg_p4rt_node_name_1`, `arg_p4rt_node_name_2`.
```
liulk, line 35:
```suggestion
7.  If P4RT Node Names cannot be resolved by walking the Components tree, use
    deviation flag `--deviation_explicit_p4rt_node_component` and pass the node
    names through args `--arg_p4rt_node_name_1`, `--arg_p4rt_node_name_2`.
```

liulk (changes requested):

liulk (changes requested):
As discussed: update go.mod to point to the latest p4rt library with the concurrency fixes, then update the implementor guide and tests to reflect the fact we no longer need to sleep.

liulk (approved):

PR #968, opened 2023-01-11
Adding 2 tests for the simultaneous CLI + OC in SetRequest().

averasko:
Add a QoS CLI/replace + OC/update in a single `SetRequest`

SetRequest contains:
* replace: "cli" current running-config + additional QoS config.
* update: OC Qos config.

----

Notes:
1. CLI is used to retrieve running-config instead of using a GetRequest/SubscribeRequest.
2. "banner motd" is removed from the retrieved running-config before applying the SetRequest, since otherwise there is an error.


sachendras:
Can we please add some context to the readme to explain more about the changes?

averasko:
done.

sachendras (approved):

feature/system/gnmi/cliorigin/ate_tests/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -39,7 +39,7 @@ required in the emergency case that is covered by this requirement.
     *   `origin: ""` (openconfig, default origin) setting the DUT port-1 string
         value at `/interfaces/interface/config/description` to `"from oc"`.
 
-*   Validate that DUT port-1 description is `"from oc"`.
+*   Validate that DUT port-1 description is `"from cli"` as a CLI value should take precedence over an OpenConfig value if both are defined as per [Mixing Schemas in gNMI](https://github.com/openconfig/reference/blob/master/rpc/gnmi/mixed-schema.md).
```
wenovus, line 42: Where in the document does it say this precedence? I scanned the doc but couldn't find it.

feature/system/gnmi/cliorigin/ate_tests/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -55,3 +55,15 @@ required in the emergency case that is covered by this requirement.
         ```
 
 *   Validate that DUT port-1 description is `"from cli"`.
+
+## Interdependent Case
+
+There latter two test cases cover setting interdependent CLI and OC configuration in the same request (OC requires CLI to be applied first in order to make sense).
```
wenovus, line 61: There latter two test cases -> TestQoSDependentCLIThenOC

wenovus (commented):
LGTM from a

feature/system/gnmi/cliorigin/ate_tests/mixed_oc_cli_origin_support_test/README.md
wenovus: Removed interface description tests and so also removed this.

feature/system/gnmi/cliorigin/ate_tests/mixed_oc_cli_origin_support_test/README.md
wenovus: Done.

wenovus:
@dplore I removed the interface description tests along with the README description for it.
I'm still confused about the "inter-dependency" that's described in the new test cases, namely
```go
	testQoSWithCLIAndOCUpdates(t, testCase{
		cliConfig: `
qos traffic-class 1 name target-group-BE1
qos tx-queue 1 name BE1
	`,
		queueName:        "BE1",
		forwardGroupName: "target-group-BE1",
		reorderFunc:      swap,
		okToFail:         true,
	})
```
Looking at it superficially, `queueName` and `forwardGroupName` are the same between the CLI and OC, so I don't see how they conflict with another, and so `okToFail=true` doesn't make sense to me.

**Questions:**
1. Do we want to keep the "common case" behaviour in the README (which I've deleted)? If so, perhaps we should also keep some of the interface description tests?
2. Is my understanding of the Qos tests correct? If so, do we actually want a conflicting QoS configuration test to be added? And if so, would should the behaviour be?

Thanks

wenovus:
@dplore @liulk This is ready for review.

SetRequest contains:
* replace: "cli" current running-config + additional QoS config.
* update: OC Qos config.

NOTE: This test is currently failing. If I change to CLI/update (of just the additional config) + OC/update, which was the original version of the test, then the test passes.

feature/experimental/system/gnmi/cliorigin/tests/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	if got := gotFG.GetOutputQueue(); got != tCase.queueName {
+		t.Errorf("Get(DUT forwarding group output queue): got %v, want %v", got, tCase.queueName)
+	}
+}
+
+func TestQoSDependentCLI(t *testing.T) {
+	testQoSWithCLIAndOCUpdates(t, testCase{
+		cliConfig: `qos traffic-class 0 name target-group-BE0
+qos tx-queue 0 name BE0`,
```
dplore, line 137: The value here is Arista EOS specific.  Other vendors may have 0 or more lines of cliConfig_test to be added to their `runningConfig` cli which will be added in followup PR

dplore (changes requested):
Can you also move to the experimental folder to indicate the status of this test is not passing on any DUT at this time.  

Also one comment to address

wenovus:
I am closing this PR to open a new one due to CLA issues.

feature/experimental/system/gnmi/cliorigin/tests/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
wenovus: I've added a vendor switch statement for handling different vendor CLIs in the re-opened PR: https://github.com/openconfig/featureprofiles/pull/1104

wenovus:
> Can you also move to the experimental folder to indicate the status of this test is not passing on any DUT at this time.
> 
> Also one comment to address

Reopened PR at https://github.com/openconfig/featureprofiles/pull/1104
The test is already moved to experimental, just the path is really long so it's hidden by the leading `...`

PR #972, opened 2023-01-11
TE 3.3 Order of replace operations for vrf and interface 

mojiiba:
Our current implementation does not allow replacing interfaces' configuration when they are assigned to a VRF.This pull change the order of replaces in test to address this.


prinikasn (commented):
/gcbrun

prinikasn (commented):
/gcbrun

PR #973, opened 2023-01-12
TE-3.3 increase the traffic distribution deviation 

mojiiba:
This pull increases the traffic distribution deviation from 0.5 to 1 same as TE-3.2 (https://github.com/openconfig/featureprofiles/blob/b03a56fd2537b160aa8f25c7e06adadb59e7a053/feature/gribi/ate_tests/weighted_balancing_test/weighted_balancing_test.go#L106).


mojiiba:
> Looks like deviation value is used as the absolute margin when comparing traffic weight distribution. If the weight values are close to 1 like below, an absolute margin as 1 seems indicating a large deviation which may not be desired....
> 
> ```
> wantWeights := map[string]float64{
>         "1": 1.25,
>         "2": 1.875,
>     }
> ```
> 
> In addition, should we consider comparing the traffic ratio instead? e.g. got below should be considered matching the want.
> 
> ```
> wantWeights := map[string]float64{
>         "1": 1.25,
>         "2": 1.875,
>     }
> got := map[string]float64{
>         "1": 2.5,
>         "2": 3.75,
>     }
> ```

I think wantWeights is a misleading variable name, otherwise wantWeights contains the ratio percentage as you suggest. So the comparison is based on the percentage (ratio) and 1 means 1% which is not a big deviation and are used in other tests. to clarify the test compare the distribution in two places: 
1- L465 in which the ratio (wantWeights) are [6.25, 18.75, 30.45]=100%
2- L552 as you referred in the comment which ratio (wantWeights) are [1.25, 1.875, 6.05,..... , 6.05 ] (16 *6.05 + 1.25 + 1.875)~= 100%

I can rename wantWeights to wantRatio if you agree.

mojiiba:
> Looks like deviation value is used as the absolute margin when comparing traffic weight distribution. If the weight values are close to 1 like below, an absolute margin as 1 seems indicating a large deviation which may not be desired....
> 
> ```
> wantWeights := map[string]float64{
>         "1": 1.25,
>         "2": 1.875,
>     }
> ```
> 
> In addition, should we consider comparing the traffic ratio instead? e.g. got below should be considered matching the want.
> 
> ```
> wantWeights := map[string]float64{
>         "1": 1.25,
>         "2": 1.875,
>     }
> got := map[string]float64{
>         "1": 2.5,
>         "2": 3.75,
>     }
> ```

concerning the deviation value, larger than .05% is needed based on the values that are set as the weights. To clarify this, devices often support different levels of granularity (2^n bucket) for traffic distribution.  If the sum of weights for a certain next-hop group is set to values other than (2^n) the distribution ratio will be rounded inevitably which will lead to deviation. To give u an example, let us assume a device supports a granularity of 64 buckets, and the weights for the next hops nh1 and nh2 are set to 2 and 3 (as it is done in the test). In this case, the expectation is to have the traffic distribution of 40% and 60% for the nh1 and nh2. However, since the 5 is not devisable by 64, the values will be rounded up as shown below which results in a deviation of around.

Screenshot 2023-01-12 at 4 38 16 PM

Similarly, the deviation will be 0.62, 0.15, 0.39 for devices with 32, 128, and 512 buckets. 
Thus, we either need to increase the deviation to 1 or change the test and make sure the sum of the weights for all nhgs is divisible by 2^n.  Please let me know which one you agree with. It is also possible to get the granularity as the test argument and calculate the expected deviation based on the granularity and the weights which I personally this will make the test complex unnecessarily.

Also. I do not follow your second comments. the wantWeights includes the ratio and its value at the comparison point (line 552) is  [1.25 1.875 6.05 6.05 6.05 6.05 6.05 6.05 6.05 6.05 6.05 6.05 6.05 6.05 6.05 6.05 6.05 6.05].

vcheny:
ok, the `wantWeights` is the traffic distribution ratio in percentage instead of the weight values. So, `deviation` as 1 below means the max deviation in  absolute margin is 1. In other words, we consider test fail if the actual traffic ratio in percentage has >1% different from the expected ones. 
```
cmp.Diff(wantWeights, got, cmpopts.EquateApprox(0, deviation))
```
This looks good to me. @vosipchu @sachendras @xw-g to comment on the acceptable deviation.

vcheny (approved):

thesrinath (approved):

PR #977, opened 2023-01-13
Create an experimental mixed_sp_wrr_traffic_test.go - part 1

sezhang2:
Added QoS traffic from 2 input interfaces

Test case:

- https://github.com/openconfig/featureprofiles/blob/main/feature/qos/ate_tests/mixed_sp_wrr_traffic_test/README.md

### DIFF
--- a/feature/experimental/qos/ate_tests/mixed_sp_wrr_traffic_test/mixed_sp_wrr_traffic_test.go
+++ b/feature/experimental/qos/ate_tests/mixed_sp_wrr_traffic_test/mixed_sp_wrr_traffic_test.go
@@ -1,0 +1,253 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mixed_sp_wrr_traffic_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/openconfig/featureprofiles/internal/deviations"
+	"github.com/openconfig/featureprofiles/internal/fptest"
+	"github.com/openconfig/ondatra"
+	"github.com/openconfig/ondatra/gnmi"
+	"github.com/openconfig/ondatra/gnmi/oc"
+	"github.com/openconfig/ygot/ygot"
+)
+
+type trafficData struct {
+	trafficRate float64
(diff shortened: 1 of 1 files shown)


nraghavendran (approved):

feature/experimental/qos/ate_tests/mixed_sp_wrr_traffic_test/mixed_sp_wrr_traffic_test.go
```diff
@@ -0,0 +1,250 @@
...
+		WithDefaultGateway("198.51.100.2")
+	intf3 := top.AddInterface("intf3").WithPort(ap3)
+	intf3.IPv4().
+		WithAddress("198.51.100.5/31").
+		WithDefaultGateway("198.51.100.4")
+	top.Push(t).StartProtocols(t)
+
+	var trafficFlows map[string]*trafficData
+
+	switch dut.Vendor() {
```
dplore, line 89: @sezhang2 why do we have more flows for ARISTA vs. Juniper and Cisco which appear to be the same except for queue assignment?   Different queue numberings per vendor seems ok?  But shouldn't the traffic on the wire be the same across all vendor cases?

sezhang2: Yes, the traffic would same for all vendors based on our test plan.
  - https://github.com/openconfig/featureprofiles/blob/main/feature/qos/ate_tests/mixed_sp_wrr_traffic_test/README.md

dplore: Ok, but the traffic flows are slightly different for arista for some reason?  trafficrate and framesize for example are not the same here:

https://github.com/openconfig/featureprofiles/pull/977/files/d6663d697873a778d75f3e2bddef48c8d9b5bc6e#diff-7ad47340a83785ccaedf86016c00d364191da968a47ee2a468cabda638549e23R92

as here:
https://github.com/openconfig/featureprofiles/pull/977/files/d6663d697873a778d75f3e2bddef48c8d9b5bc6e#diff-7ad47340a83785ccaedf86016c00d364191da968a47ee2a468cabda638549e23R107

sezhang2: > Ok, but the traffic flows are slightly different for arista for some reason? trafficrate and framesize for example are not the same here:
> 
> https://github.com/openconfig/featureprofiles/pull/977/files/d6663d697873a778d75f3e2bddef48c8d9b5bc6e#diff-7ad47340a83785ccaedf86016c00d364191da968a47ee2a468cabda638549e23R92
> 
> as here: https://github.com/openconfig/featureprofiles/pull/977/files/d6663d697873a778d75f3e2bddef48c8d9b5bc6e#diff-7ad47340a83785ccaedf86016c00d364191da968a47ee2a468cabda638549e23R107

Ack. The traffic flow will have same values as described in the test plan in the next commit.

PR #978, opened 2023-01-13
Changed DUTInfo struct and NewDUTInfo for export

lgomez9:
Need to export the DUTInfo functionality into a different folder. As part of the export, also changed the NewDUTInfo function to take a gnmi client instead of components.Y.


internal/rundata/dut.go
```diff
@@ -27,22 +28,39 @@ import (
...
 )
 
-// dutInfo retrieves the vendor, model, and OS version from the device from various
+// DUTInfo retrieves the vendor, model, and OS version from the device from various
 // OpenConfig paths.
-type dutInfo struct {
+type DUTInfo struct {
 	vendor string
 	model  string
 	osver  string
```
liulk, line 40: Let's just export the fields so we don't need to add the accessor functions.

Also, Go initialism would export `osver` as `OSVer`.

liulk (changes requested):

internal/rundata/dut.go
lgomez9: Done!

liulk:
The rundata check is addressed in #979

liulk (approved):

PR #981, opened 2023-01-17
P4RT node name mapping for Cisco 8xxx

kjahed:
This PR changes explicitP4RTNodes to compute p4rt node names for CISCO devices based on the model number when -deviation_explicit_p4rt_node_component is set.

### DIFF
--- a/feature/experimental/p4rt/internal/p4rtutils/p4rtutils.go
+++ b/feature/experimental/p4rt/internal/p4rtutils/p4rtutils.go
@@ -195,7 +195,7 @@ func explicitP4RTNodes() map[string]string {
 
 // inferP4RTNodesCisco infers the P4RT node name from the port name and device model
 // for Cisco devices.
-func inferP4RTNodesCisco(t *testing.T, dut *ondatra.DUTDevice) map[string]string {
+func inferP4RTNodesCisco(t testing.TB, dut *ondatra.DUTDevice) map[string]string {
 	t.Helper()
 	npus := []int{0, 1, 2}
 	pranges := []int{11, 23, 35}
@@ -219,7 +219,7 @@ func inferP4RTNodesCisco(t *testing.T, dut *ondatra.DUTDevice) map[string]string
 
 // P4RTNodesByPort returns a map of <portID>:<P4RTNodeName> for the reserved ondatra
 // ports using the component and the interface OC tree.
-func P4RTNodesByPort(t *testing.T, dut *ondatra.DUTDevice) map[string]string {
+func P4RTNodesByPort(t testing.TB, dut *ondatra.DUTDevice) map[string]string {
 	t.Helper()
 	if *deviations.ExplicitP4RTNodeComponent {
 		if dut.Vendor() == ondatra.CISCO {
--- a/feature/experimental/p4rt/internal/p4rtutils/p4rtutils.go
+++ b/feature/experimental/p4rt/internal/p4rtutils/p4rtutils.go
@@ -21,11 +21,17 @@
 package p4rtutils
 
 import (
+	"fmt"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


kjahed:
@vosipchu

feature/experimental/p4rt/internal/p4rtutils/p4rtutils.go
```diff
@@ -182,7 +188,40 @@ func ACLWbbIngressTableEntryGet(infoList []*ACLWbbIngressTableEntryInfo) []*p4_v
 	return updates
 }
 
-func explicitP4RTNodes() map[string]string {
+func explicitP4RTNodes(t *testing.T, dut *ondatra.DUTDevice) map[string]string {
+	t.Helper()
+	if dut.Vendor() == ondatra.CISCO {
```
liulk, line 193: Let's factor out the Cisco specific portion to its own function and leave the original function true to its intended purpose.

```go
// inferP4RTNodesCisco infers the P4RT node name from the port name and device model
// for Cisco devices.
func inferP4RTNodesCisco(...) map[string]string {
  ...
}
```

This should be called from P4RTNodesByPort() below.

feature/experimental/p4rt/internal/p4rtutils/p4rtutils.go
```diff
@@ -194,7 +233,7 @@ func explicitP4RTNodes() map[string]string {
 func P4RTNodesByPort(t *testing.T, dut *ondatra.DUTDevice) map[string]string {
 	t.Helper()
 	if *deviations.ExplicitP4RTNodeComponent {
-		return explicitP4RTNodes()
+		return explicitP4RTNodes(t, dut)
```
liulk, line 236:
```suggestion
		if dut.Vendor() == ondatra.CISCO {
			return inferP4RTNodesCisco(t, dut)
		}
		return explicitP4RTNodes(t, dut)
```

But reusing deviations.ExplicitP4RTNodeComponent also means that Cisco will not be able to specify P4RT node names by args. @vosipchu does this warrant a new deviation, or is it okay to repurpose an existing deviation?

liulk (changes requested):

feature/experimental/p4rt/internal/p4rtutils/p4rtutils.go
vbhan-cisco: @liulk 
  Based on our discussion with @vosipchu he wanted a way to get the mapping b/w NPU to I/f without passing it via arg  since  ATE may be connected to different NPU's on different setups

feature/experimental/p4rt/internal/p4rtutils/p4rtutils.go
kjahed: Thanks for the suggestion. Refactored.

vosipchu (dismissed):

prinikasn (commented):
/gcbrun

feature/experimental/p4rt/internal/p4rtutils/p4rtutils.go
```diff
@@ -189,11 +193,38 @@ func explicitP4RTNodes() map[string]string {
...
+// inferP4RTNodesCisco infers the P4RT node name from the port name and device model
+// for Cisco devices.
+func inferP4RTNodesCisco(t *testing.T, dut *ondatra.DUTDevice) map[string]string {
+	t.Helper()
+	npus := []int{0, 1, 2}
+	pranges := []int{11, 23, 35}
+	res := make(map[string]string)
+	dist := dut.Model() == "" || strings.HasPrefix(dut.Model(), "CISCO-88")
+	for _, p := range dut.Ports() {
+		if dist {
```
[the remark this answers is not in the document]
kjahed: I agree. The reliable way to do this is to look at component tree, but that adds 3-5 minutes to the test's running time, since the test can't access the rundata/dutInfo information that is collected prior to the test execution. I'm not sure if there are any plans to export that. For now, we will rely on the information in the testbed file.

PR #984, opened 2023-01-17
RT-5.1 modify checks for counters 

mojiiba:
This pull removes the check for checking interface counter since gNMI-1.11: Telemetry: Interface Packet Counters.


thesrinath (commented):
/gcbrun

thesrinath (approved):

PR #999, opened 2023-01-20
OC 1.1 increase timeout for await call

mojiiba:
This pull increase the timeout for await call from 30 second to one minute. otherwise, it is possible the first telemetry for the related path in test to be failed.


prinikasn (commented):
/gcbrun

prinikasn (approved):

PR #1012, opened 2023-01-25
TE-15.1

mojiiba:
This pull adds:
- election id sync for TE-15.1 to allow consecutive runs of the tests without resetting the gribi server
- bump gribigo version to include recent tests
- add skip flag for MPLS tests


vosipchu (dismissed):

feature/gribi/ate_tests/gribigo_compliance_test/gribigo_compliance_test.go
```diff
@@ -95,13 +97,32 @@ func shouldSkip(tt *compliance.TestSpec) string {
...
+		return "This RequiresMPLS test is skipped by --skip_mpls"
 	}
 	return moreSkipReasons[tt.In.ShortName]
 }
 
+func syncElectionID(t *testing.T, dut *ondatra.DUTDevice) {
+	t.Helper()
+	clientA := &gribi.Client{
+		DUT:         dut,
+		Persistence: true,
```
xw-g, line 110: If your goal of setting `PERSISTENT` mode is to make sure most current implementations can support the negotiation, then probably you want to set the redundancy mode too, right? (to `SINGLE_PRIMARY`).

feature/gribi/ate_tests/gribigo_compliance_test/gribigo_compliance_test.go
```diff
@@ -95,13 +97,32 @@ func shouldSkip(tt *compliance.TestSpec) string {
...
+	}
+	t.Log("Establish gRIBI client connection with PERSISTENCE set to True")
+	if err := clientA.Start(t); err != nil {
+		t.Fatalf("gRIBI Connection for clientA could not be established")
+	}
+	clientA.BecomeLeader(t)
```
xw-g, line 116: You don't need to become a leader to learn the current election ID, right? (In fact, `BecomeLeader()` also calls `LearnElectionID` to understand the base).

So, seems like this `clientA.BecomelLeader` is not needed?

xw-g (changes requested):

feature/gribi/ate_tests/gribigo_compliance_test/gribigo_compliance_test.go
mojiiba: yes, the start function does this by default: https://github.com/openconfig/featureprofiles/blob/4ac3c512ce9f4974dac9892faf380d0592e3ffad/internal/gribi/gribi.go#L118

feature/gribi/ate_tests/gribigo_compliance_test/gribigo_compliance_test.go
mojiiba: thanks for pointing this out, fixed.

xw-g (approved):

PR #1014, opened 2023-01-25
Update telemetry_basic_check_test.go

sezhang2:
Deleted QoS counter update test case to avoid the extra maintenance efforts as new QoS tests have been created and have more coverage.


sezhang2:
Close this one and open a new one to fix the rebase and build failure

PR #1015, opened 2023-01-26
TE-11.1: Backup NHG Single NH : Fix migration to use election id handshake in func TestIndirectBackupNexthopGroup

stthummuri:
Added fix migration to use election id handshake in func TestIndirectBackupNexthopGroup. It only exists in func TestDirectBackupNexthopGroup which is also needed in func TestDirectBackupNexthopGroup

Reference PR #864


michaelore (approved):
Thanks!

PR #1016, opened 2023-01-26
Update TE-14.1 gribi scaling

LimeHat:
Updated TE-14.1 gribi scaling to fix how deviation ExplicitInterfaceInDefaultVRF is handled to support all subinterfaces:
1. function pushConfig() merged into configureInterfaceDUT()
2. handling of ExplicitInterfaceInDefaultVRF deviation is moved from pushConfig()/configureInterfaceDUT() into configureSubinterfaceDUT() function (because each subinterface has to be assigned into a network instance)

Additionally:

3. added NoMixOfTaggedAndUntaggedSubinterfaces deviation support to the OTG test.
4. removed unused protocol static configuration line


xw-g:
I think we want to wait on https://github.com/openconfig/featureprofiles/pull/1067, some OS implementation requires unit0 be there.

LimeHat:
> I think we want to wait on #1067, seems like some OS implementation requires untagged unit0 be there.

Thanks for the feedback @xw-g . I looked at #1067 and it seems they're actually implementing the same thing but with deviation instead of direct test modification -> as far as I understand Juniper's implementation also does not allow that mix.

xw-g:
> > I think we want to wait on #1067, seems like some OS implementation requires untagged unit0 be there.
> 
> Thanks for the feedback @xw-g . I looked at #1067 and it seems they're actually implementing the same thing but with deviation instead of direct test modification -> as far as I understand Juniper's implementation also does not allow that mix.
> 
> So while it can be done this way, I don't think it has to be? Let me know what you think

Direct change without deviation is making a strong assumption that all other OS implementation does not need unit0 (which seems not true). Therefore, we prefer to go with deviation for now.

LimeHat:
We could use tagged unit0 as well, in principle.

Ok, I'll wait for the deviation PR to be merged & will reduce the scope of this PR just to fix ExplicitInterfaceInDefaultVRF assignment logic then. Thank you Xiao.

xw-g:
Thanks! I prefer to splitting this PR.

LimeHat:
@xw-g i updated the pr to accommodate for the new deviation

xw-g:
changes seems good to me. I'll carry the presubmit test and then merge if no issue.

MarcCharlebois:
@xw-g any progress on the presubmit tests and merge efforts?

xw-g:
@LimeHat can you help resolve the `This branch is out-of-date with the base branch`? Thanks. I'm working on the presubmit now.

LimeHat:
@xw-g done. This message will reappear if any other PR will be merged into master before this one, but it is not blocking unless there's a merge conflict AFAIK?

xw-g:
@LimeHat will refactor this PR per our offline discussion.

LimeHat:
Main functional changes are split & moved out into PRs #1235 , #1236 and #1238

PR #1025, opened 2023-01-26
Update telemetry_basic_check_test.go

sezhang2:
Deleted QoS counter update test case to avoid the extra maintenance efforts as new QoS tests have been created and have more coverage.

This test coverage is now provided by https://github.com/openconfig/featureprofiles/tree/main/feature/experimental/qos/ate_tests/qos_output_queue_counters_test

### DIFF
--- a/feature/gnmi/ate_tests/telemetry_basic_check_test/README.md
+++ b/feature/gnmi/ate_tests/telemetry_basic_check_test/README.md
@@ -59,17 +59,6 @@ following features:
 
     *   Check some counters are updated correctly.
 
-*   QoS counters
-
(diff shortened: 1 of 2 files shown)


dplore (commented):
Can you reference the new tests which replace the deleted code here?  One way to do this is to edit the first comment in this PR with links to the relevant code.

sezhang2:
> Can you reference the new tests which replace the deleted code here? One way to do this is to edit the first comment in this PR with links to the relevant code.

The test case is replaced by QoS counter test  under
 - https://github.com/openconfig/featureprofiles/tree/main/feature/experimental/qos/ate_tests/qos_output_queue_counters_test

dplore (approved):

PR #1028, opened 2023-01-27
tools/addrundata outputs listing in CSV and TestTracker JSON

liulk:


tools/addrundata/json.go
```diff
@@ -0,0 +1,419 @@
...
+			children = append(children, child) // Passthrough JSON-only testsuites.
+			continue
+		}
+		js.merge(o)
+		children = append(children, o)
+		delete(todos, sec)
+	}
+
+	// Update the todos that were missing from the JSON.
+	for _, sec := range todos.sortedKeys() {
```
Ankur19, line 183: Why do we need to sort the keys here? running `json.Marshal` sorts the keys already.

Ankur19 (approved):
LGTM!

tools/addrundata/json.go
liulk: First, these children are objects in a list, so they appear in the same order in the slice that we have to sort first.

Secondly, builtin sorting is done using string comparison ("TE-10.1" < "TE-2.1"), but the sorting we do here is version sort ("TE-2.1" < "TE-10.1").

PR #1029, opened 2023-01-27
RT IS-IS 2.2 Changes to IS-IS LSP Updates test

arvbaska1:
Made below changes to testcase

- Get the initial database-overloads leaf counter value, after config is pushed to the DUT & before IS-IS adjacency is established.
- Checking the counter value after set overload-bit by incrementing the variable by 1.

The testcase run is dependant on session.go changes done in PR https://github.com/openconfig/featureprofiles/pull/953


vosipchu (approved):

feature/experimental/isis/ate_tests/lsp_updates_test/lsp_updates_test.go
```diff
@@ -62,7 +73,7 @@ func TestOverloadBit(t *testing.T) {
 		GetOrCreateOverloadBit().SetBit = ygot.Bool(true)
 	ts.PushDUT(context.Background())
 	// TODO: Verify the link state database once device support is added.
-	if err := check.Equal[uint32](overloads.State(), 1).AwaitFor(time.Second*10, ts.DUTClient); err != nil {
+	if err := check.Equal(overloads.State(), uint32(olVal+1)).AwaitFor(time.Second*10, ts.DUTClient); err != nil {
```
xw-g, line 76: nit: the overload bit can either be 1 or 0, right? Not sure why we want `+1` here (instead of just saying 1)?

xw-g (approved):
LGTM, but just one minor.

feature/experimental/isis/ate_tests/lsp_updates_test/lsp_updates_test.go
arvbaska1: This leaf /network-instances/network-instance/protocols/protocol/isis/levels/level/system-level-counters/state/database-overloads of type uint32 returns number of times is-is database is in overload state.
Here we are checking that after we set the overload bit the counter value should be initial value +1.

thesrinath (approved):

PR #848, opened 2022-12-06
TE-14.1 Fix election id issue

mojiiba:
The last update to the test missed to update the election id for modify requests. This simply fixes that.
Passing log:
```
********************************************

=== RUN   TestScaling
    gribi_scaling_test.go:422: 
        *** Fetching gRIBI client for SF-D...
        
        
    gribi_scaling_test.go:308: Config for nonDefaultNI at /network-instances/network-instance[name=vrf1]:
        {
          "openconfig-network-instance:config": {
            "enabled": true,
            "enabled-address-families": [
              "openconfig-types:IPV4",
              "openconfig-types:IPV6"
            ],
            "name": "vrf1",
            "type": "openconfig-network-instance-types:L3VRF"
          },
          "openconfig-network-instance:name": "vrf1"
        }
2022/12/06 13:33:09 Test output "TestScaling Config for nonDefaultNI at /network-instances/network-instance[name=vrf1]" is discarded without -outputs_dir.  Please specify -outputs_dir to keep it.
    gribi_scaling_test.go:308: Config for nonDefaultNI at /network-instances/network-instance[name=vrf2]:
        {
          "openconfig-network-instance:config": {
            "enabled": true,
            "enabled-address-families": [
              "openconfig-types:IPV4",
              "openconfig-types:IPV6"
            ],
            "name": "vrf2",
            "type": "openconfig-network-instance-types:L3VRF"
          },
          "openconfig-network-instance:name": "vrf2"
        }
2022/12/06 13:33:09 Test output "TestScaling Config for nonDefaultNI at /network-instances/network-instance[name=vrf2]" is discarded without -outputs_dir.  Please specify -outputs_dir to keep it.
    gribi_scaling_test.go:308: Config for nonDefaultNI at /network-instances/network-instance[name=vrf3]:
        {
          "openconfig-network-instance:config": {
            "enabled": true,
            "enabled-address-families": [
              "openconfig-types:IPV4",
              "openconfig-types:IPV6"
            ],
            "name": "vrf3",
            "type": "openconfig-network-instance-types:L3VRF"
          },
          "openconfig-network-instance:name": "vrf3"
        }
2022/12/06 13:33:10 Test output "TestScaling Config for nonDefaultNI at /network-instances/network-instance[name=vrf3]" is discarded without -outputs_dir.  Please specify -outputs_dir to keep it.
    gribi_scaling_test.go:429: Put port TenGigE0/0/0/4/0 into vrf vrf1
[...]

PR #634, opened 2022-10-11
TE-14.1 (gRIBI Scaling): fix invalid nhg and a deprecated leaf issues

mojiiba:
This pull fixes the following two issues for TE-14.1: (gRIBI Scaling)
- the tests adds ipv4entry with invalid nhg
- the test uses deprecated leaf to configure vlan id.

PR #1031, opened 2023-01-27
Update OTG gribi_scaling_test to match the ATE version of the test.

bstoll:
PR #848, #600, #715, and #634 updated the ATE version of this test without updating the OTG version.


gforeal:
copybara and local test run confirmed the  "remove base routing" specific issue is not happening anymore!

gforeal (approved):
copybara and local test run confirmed the "remove base routing" specific issue is not happening anymore!
LGTM

PR #1033, opened 2023-01-27
gNOI-3.2: Updating Watch function timeout duration for interface OperStatus

trathod1:
After linecard reboot, sometimes interface takes more than 5 minutes to come up. Increasing watch function timeout duration to avoid intermittent script failure.


xw-g:
LGTM, since it's just relax the arbitrary max wait time (I think still within reasonable range) and does not increase the execution time of the happy path on other testbed.

vosipchu:
Agreed, Xiao. LGTM

xw-g (approved):

PR #933, opened 2022-12-21
OTG Test: P4RT-3.2: Google Discovery Protocol: PacketOut

ANISH-GOTTAPU:
OTG Test: P4RT-3.2: Google Discovery Protocol: PacketOut [here](https://github.com/openconfig/featureprofiles/blob/main/feature/experimental/p4rt/ate_tests/google_discovery_protocol_packetout_test/google_discovery_protocol_packetout_test.go)

resolves #882

PR #1037, opened 2023-01-30
Revert "OTG Test: P4RT-3.2: Google Discovery Protocol: PacketOut"

michaelore:
Reverts openconfig/featureprofiles#933

Looks like this made it through without rundata.go, and it's breaking checks for other PRs.

See https://github.com/openconfig/featureprofiles/actions/runs/4044245397/jobs/6954152729


Ankur19:
Adding fix here: https://github.com/openconfig/featureprofiles/pull/1039

We can skip the revert.

Ankur19:
Closing since rundata is fixed.

PR #1040, opened 2023-01-30
Gnoi-5.2 : add deviations for l4 protocol and fragmenation

mojiiba:
This pull adds two deviations concerning devices that:
- support UDP as l4 protocol (TraceRouteL4ProtocolUDP)
- do not support fragmentation (TraceRouteFragmentation)

It also set minTracerouteRTT to zero sincethe a loopback is used as the traceroute destination. Further, we add sleep before each tests since our device does not allow traceroute flooding.


feature/gnoi/system/tests/traceroute_test/traceroute_test.go
```diff
@@ -230,6 +243,20 @@ func TestGNOITraceroute(t *testing.T) {
...
+				if tc.defaultL4Protocol {
+					tc.traceRequest.L4Protocol = spb.TracerouteRequest_UDP
+				}
+				if tc.traceRequest.L4Protocol != spb.TracerouteRequest_UDP {
+					t.Skip("Test is skiped due to the TraceRouteL4ProtocolUDP deviatiopn")
+				}
+			}
+			if *deviations.TraceRouteFragmentation {
+				if tc.traceRequest.DoNotFragment {
+					t.Skip("Test is skiped due to the TraceRouteFragmentation deviatiopn")
```
prinikasn, line 257: deviation

prinikasn (commented):
/gcbrun

prinikasn (approved):

PR #1041, opened 2023-01-30
OC 1.2

mojiiba:
This pull simply removes the call to reset dut since this is not implemented by most vendors. Also, the reset dut is already called by FP test runner before running each test.


feature/networkinstance/ate_tests/defaults_test/defaults_test.go
```diff
@@ -134,7 +134,6 @@ func TestDefaultAddressFamilies(t *testing.T) {
 	for _, tc := range cases {
 		t.Run(tc.desc, func(t *testing.T) {
 			// Reset DUT config.
```
guoshiuan, line 136: Remove the comment also.

vosipchu (approved):

xw-g:
internal b/267202852 for presubmit effort.

PR #1045, opened 2023-01-31
Modify BGP keepalive and holdtime values to accommodate different vendors

cprabha:
Hi,

Default values of BGP keepalive/holdtimer vary across vendors. Some vendors have default keepalive/holdtime values as 60s/180s while some have it as 30s/90s. 

This PR changes the keepalive/holdtime values configured explicitly on the DUT to 25s/75s to ensure that the test works consistently across all vendors.

Thanks,
Prabha


Wanmathy (changes requested):
Vendor default keep alive interval may be different so expecting 30 as keep alive will not work unless peer device keep alive interval is set to 30. We're not passing peer device keep alive interval for peer dut in this case. Thanks.

cprabha:
> Vendor default keep alive interval may be different so expecting 30 as keep alive will not work unless peer device keep alive interval is set to 30. We're not passing peer device keep alive interval for peer dut in this case. Thanks.

@Wanmathy , 
Some vendors  have default keepalive/holdtimer value as 60/180 and other vendors have it as 30/90. To verify the negotiation and have consistent result , I have changed the value to 25/75 in the dut now.

Wanmathy (dismissed):

singh-prem:
@cprabha 25/75 looks to be on aggressive side of the timers. Is it okay to keep it something realistic that would be used in production?

xw-g:
internal discussion b/267870303

sachendras:
> 
Given that it is just one peering for the functional test, this setting seems reasonable that will pass the test for all vendors. This is also because, the DUT is configured for a static value while the peer is left to its default in the functional test.

For production, we will have a more realistic definition and we will determine internally if we need to have this hard set at both ends or we leave these timers to their default at both ends.

sachendras:
Google internal LGTMs in b/267870303

sachendras (approved):

PR #1048, opened 2023-02-02
Bump ygnmi & ondatra versions

wenovus:


greg-dennis:
This PR should no longer be necessary. I just merged https://github.com/openconfig/featureprofiles/pull/1049, which bumps ygnmi to the same version as this PR and Ondatra to the next-higher patch version.

PR #1051, opened 2023-02-02
Adding deviation for RT-6.1

agvarghe-cisco:
Adding deviation for RT-6.1 for allowing lldp interfaces config to override the global config

Test expects that DUT does not generate any LLDP messages irrespective of the configuration of lldp/interfaces/interface/config/enabled (TRUE or FALSE) on any interface.In another word, they expect the lldp/config/enable to switch on/off lldp globally. However, XR allow interfaces config to override the global config (LLDP message will be generated even its turned off globally but enabled on a interface).This deviation is for checking neighbours are seen when  lldp is disabled globally but enabled on interface.


feature/lldp/tests/core_lldp_tlv_population_test/core_lldp_tlv_population_test.go
```diff
@@ -143,9 +144,17 @@ func verifyNodeTelemetry(t *testing.T, node, peer gnmi.DeviceOrOpts, nodePort, p
 				return true
 			}
 			gotLen = len(intf.Neighbor)
-			return gotLen == 0
+			if *deviations.LLDPInterfaceConfigOverrideGlobal {
```
prinikasn, line 147: if block ends with a return statement, so drop this else and outdent its block

https://google.github.io/styleguide/go/decisions#indent-error-flow

agvarghe-cisco: Done

xw-g:
1. We need an internal bug for deviation, this will be for tracking purpose. @coveralls  
2. @agvarghe-cisco can you help put more words in the description to talk about the motivations of this deviation? Right now, it only says what it does (not why it does it).
3. @agvarghe-cisco  Seems it's hard for reviewers to correlate the meaning of the deviation to why it affects the number of expected neighbors. Maybe some comments there too? Thanks.

aaronmillisor:
@xw-g, IOS XR has two ways of configuring LLDP.
1. Globally
2. Per-interface

This allows for two types of operations. One is an opt-in mechanism (globally turned off, but turn on per interface), or for an opt-out mechanism (globally enabled, but turn it off per interface).

Cisco has mapped the global lldp configuration to the global lldp leaf, which opts in all interfaces even if LLDP is not explicitly configured on a specific interface.

The test as written expects that a device only sends lldp messages if both the global lldp configuration _and_ per-interface configuration is set. The semantics of the test do not match IOS XR's implementation of LLDP (i.e. XR has more configuration options than are present in OC).

This test is enabling LLDP on an interface (opt in mode), while disabling it at the global level, and expects no LLDP messages to be sent on the interface. Because the test is globally enabling LLDP, while not enabling it at the interface level, XR will still create an adjacency.

xw-g:
@vosipchu can you help drive the clarification on our expectation? Thanks.

vosipchu:
This is based on our internal discussion (b/268524109):

1) We start with Global LLDP enabled:false and Interfaces local LLDP enabled:false.
    - this means no LLDP is allowed to work on any port (whether enabled is true or false)
2) As soon as there is a need to start LLDP, as a protocol, it means that:
    - Global LLDP enabled:true
    - Activating per Interface local LLDP enabled:true
    - All other interfaces stay in local LLDP enabled:false with no LLDP running there.

@aaronmillisor let me know if there is anything else missing for you?

prinikasn (commented):
/gcbrun

vosipchu (approved):

xw-g:
@vosipchu is this comment, https://github.com/openconfig/featureprofiles/pull/1051#issuecomment-1432217401, a confirmation that existing test plan is correct (and we don't need to update our test plan). Or, are you suggesting to update the test plan (and then update the code) per your comment? Thanks.

prinikasn (approved):

PR #1058, opened 2023-02-02
Added interface-ref for binding vrf-selection-policy

mohanasm:
Added deviation and code for attaching vrf-selection-policy to interface using interface-ref


sachendras:
Google internal confirmation in b/267783676

sachendras (approved):

sachendras (approved):

PR #1059, opened 2023-02-03
Update RT-1.5 BGP prefix limit

LimeHat:
This PR introduces 2 changes:
1) route policy is always added and configured on a per-neighbor basis if RoutePolicyUnderNeighborAfiSafi deviation is not enabled,
2) explicitly enables/disables both address families for each BGP neighbor (ipv4 en. / ipv6 dis. for v4 peers, v4 dis., v6 en. for v6 peers)


LimeHat:
superseded by #1171

PR #1064, opened 2023-02-03
Modified config of member link before configuring aggregate interface 

mohanasm:
1.  Some vendors require the aggregated ethernet members to be configured prior to configuring the aggregated ethernet bundle.
2.  It is mandatory to specify interface type while configuring interfaces.


sachendras (approved):

sachendras:
b/264674518 for Google's internal confirmation

PR #1081, opened 2023-02-08
Update cisco kne topology

mojiiba:
This pull updates the kne topology by adding services, updating the interface name and constraint. This can be used to run otg tests.


alexmasi:
Once https://github.com/openconfig/kne/pull/303 is tested and merged, those changes should be synced here and I will approve. Thanks

topologies/kne/cisco_ixia.textproto
```diff
@@ -32,13 +32,8 @@ nodes: {
     os: "ios-xr"
     config: {
         file: "cisco.config"
-        image: "8000e:latest"
```
greg-dennis, line 35: Why was the image removed? It's convenient for us to always have an image specified, even if that means we rewrite it on import.

mojiiba: I did it based on the @alexmasi  comment on the kne pr, I am ok with reverting the change if u both agree.

alexmasi: Yes Moji, for this repo we prefer to leave the image string explicitly in the proto (but thats not the case for the kne repo examples)

Sorry about any confusion there

mojiiba: sure, done

alexmasi (approved):

thesrinath (approved):

PR #1084, opened 2023-02-08
Updated README.md to align with test plan

mohanasm:
Updated test step for ATE port-2 and DUT port-2 config as in test plan


mohanasm:
CLA error. Will open new PR.

PR #1067, opened 2023-02-04
Added deviation for adding vlan for all subinterfaces

mohanasm:
Changes:
  For dutPort2 and atePort2, unit 0 subinterface is configured without vlan(vlan-id 0). Added deviation to configure vlan for all subinterfaces to avoid mix of tagged and untagged subinterfaces.

PR #1086, opened 2023-02-08
Aligning README.md with test plan for TE-3.3

mohanasm:
Updated test step for ATE port-2 and DUT port-2 as in test plan.

### DIFF
--- a/feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
+++ b/feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
@@ -13,20 +13,7 @@ Configure ATE and DUT:
 
 *   Create a non-default VRF (VRF-1) that includes DUT port-1.
 
-*   For DUT port-2 interface, create a subinterface with Index 0 and IPv4
-    address 192.0.2.5, not in a VLAN.
-
-*   For ATE port-2, create a subinterface with IPv4 address 192.0.2.6 and
-    Default Gateway of 192.0.2.5, not in a VLAN.
-
-*   Repeat for 18 more subinterfaces in a VLAN configuration:
-
-    *   For DUT port-2, subinterfaces indices 1...18 with VLAN IDs 1...18 and
-        corresponding IPv4 addresses 192.0.2.9 ... 192.0.2.73
-
-    *   For ATE port-2, subinterfaces with VLAN IDs 1...18 and corresponding
-        IPv4 addresses 192.0.2.10 ... 192.0.2.79 with default gateways of
-        192.0.2.9 ... 192.0.2.78
+*   Create 18 L3 sub-interfaces (each with their own VLAN and IP) under DUT port-2 and corresponding 18 L3 sub-interfaces on ATE port-2.
 
 Test case for basic hierarchical weight:
 
--- a/feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
+++ b/feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
@@ -13,7 +13,7 @@ Configure ATE and DUT:
 
 *   Create a non-default VRF (VRF-1) that includes DUT port-1.
 
-*   Create 18 L3 sub-interfaces (each with their own VLAN and IP) under DUT port-2 and corresponding 18 L3 sub-interfaces on ATE port-2.
+*   Create 18 L3 sub-interfaces (each with their own VLAN and IP) under DUT port-2 and corresponding 18 L3 sub-interfaces on ATE port-2. 
 
 Test case for basic hierarchical weight:
 


liulk:
Sometimes when we implement a test plan, we discover ambiguities that we later try to clarify in the README.md. In this case, it looks like you're removing the clarification to make it ambiguous again.

README.md should be the source of truth that supersedes the test plan. FYI, @bstoll.

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
```diff
@@ -13,20 +13,13 @@ Configure ATE and DUT:
...
 
-*   For DUT port-2 interface, create a subinterface with Index 0 and IPv4
-    address 192.0.2.5, not in a VLAN.
+*   On DUT port-2 and ATE port-2 create 18 L3 sub-interfaces each as below:
 
-*   For ATE port-2, create a subinterface with IPv4 address 192.0.2.6 and
-    Default Gateway of 192.0.2.5, not in a VLAN.
+    *   On DUT port-2, create subinterfaces with indices 1 to 18 mapped to VLAN IDs 1 to 18 and corresspoding IPv4 addresses 192.0.2.5, 192.0.2.9, ... , 192.0.2.73 respectively.
+    
+    * On ATE port-2, create subinterfaces with indices 1 to 18 mapped to VLAN IDs 1 to 18 and corresponding IPv4 addresses 192.0.2.6, 192.0.2.10, ... , 192.0.2.74 and default gateways as 192.0.2.5, 192.0.2.9, ..., 192.0.2.73 respectively
```
liulk, line 20:
```suggestion
    *   On ATE port-2, create subinterfaces with indices 1 to 18 mapped to VLAN IDs 1 to 18 and corresponding IPv4 addresses 192.0.2.6, 192.0.2.10, ... , 192.0.2.74 and default gateways as 192.0.2.5, 192.0.2.9, ..., 192.0.2.73 respectively
```

liulk (approved):

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
```diff
@@ -13,20 +13,13 @@ Configure ATE and DUT:
 
 *   Create a non-default VRF (VRF-1) that includes DUT port-1.
 
-*   For DUT port-2 interface, create a subinterface with Index 0 and IPv4
-    address 192.0.2.5, not in a VLAN.
+*   On DUT port-2 and ATE port-2 create 18 L3 sub-interfaces each as below:
 
-*   For ATE port-2, create a subinterface with IPv4 address 192.0.2.6 and
-    Default Gateway of 192.0.2.5, not in a VLAN.
+    *   On DUT port-2, create subinterfaces with indices 1 to 18 mapped to VLAN IDs 1 to 18 and corresspoding IPv4 addresses 192.0.2.5, 192.0.2.9, ... , 192.0.2.73 respectively.
```
liulk, line 18:
```suggestion
    *   On DUT port-2, create subinterfaces with indices 1 to 18 mapped to VLAN IDs 1 to 18 and corressponding IPv4 addresses 192.0.2.5, 192.0.2.9, ... , 192.0.2.73 respectively.
```

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
```diff
@@ -13,20 +13,13 @@ Configure ATE and DUT:
 
 *   Create a non-default VRF (VRF-1) that includes DUT port-1.
 
-*   For DUT port-2 interface, create a subinterface with Index 0 and IPv4
-    address 192.0.2.5, not in a VLAN.
+*   On DUT port-2 and ATE port-2 create 18 L3 sub-interfaces each as below:
```
liulk, line 16:
```suggestion
*   On DUT port-2 and ATE port-2 create 18 L3 sub-interfaces each with a /30 subnet as below:
```

LimeHat:
Is the deviation introduced in #1067 not needed for these cases anymore?

Based on our conversation with @xw-g earlier I was under impression that it was required.

sachendras:
> #1067

This test in its updated format wouldnt need that deviation. However, we will continue to maintain the other tests as it is so the implementation that do not allow a mix of tagged and untagged frames on a interface fix their implementation.
In this particular case though, if any implementation is not supporting all tagged VLANs on an interface then they should start a deviation w/ an expectation to fix it at some point.

LimeHat:
Thanks @sachendras .
Is there a general guidance on how to distinguish between test cases (with multiple subinterfaces) where this deviation should be used, and where it is not needed (or, in other words, what tests can be refactored to use tagged only subinterfaces vs a mix of tagged & untagged)?

sachendras:
> 

There is no strict requirements on one versus the other as far as the VLAN requirements on an interfaces is concerned. It is how the initial test was written. Intent is to just make sure, among the entire suite of tests, we have atleast one of each kind so we ensure that both setups need to be supported by all implementations atleast at some point. At this point we dont see a need to have a dedicated test for just checking this functionality. We may add in future.

liulk (approved):

sachendras (dismissed):

mohanasm:
Raised https://github.com/openconfig/featureprofiles/pull/1239 to track this.

PR #1095, opened 2023-02-09
TE-3.6 fix test issue for creation of static route

mojiiba:
The test unnecessarily does getconfig for network-instance that can fail if there it is not configured. This pull fixes the issue.


guoshiuan:
b/268560755 has the internal test result.

guoshiuan (approved):

prinikasn (approved):

PR #1097, opened 2023-02-10
Remove the recently new added config deleting operation

gforeal:
…config

### DIFF
--- a/feature/experimental/isis/ate_tests/internal/session/session.go
+++ b/feature/experimental/isis/ate_tests/internal/session/session.go
@@ -256,7 +256,7 @@ func (s *TestSession) PushDUT(ctx context.Context) error {
 		return fmt.Errorf("configuring network instance: %w", err)
 	}
 	dutConf := s.DUTConf.GetOrCreateNetworkInstance(*deviations.DefaultNetworkInstance).GetOrCreateProtocol(PTISIS, ISISName)
-	_, err = ygnmi.Replace(ctx, s.DUTClient, ProtocolPath().Config(), dutConf)
+	_, err := ygnmi.Replace(ctx, s.DUTClient, ProtocolPath().Config(), dutConf)
 	if err != nil {
 		return fmt.Errorf("configuring ISIS: %w", err)
 	}
--- a/feature/experimental/isis/ate_tests/internal/session/session.go
+++ b/feature/experimental/isis/ate_tests/internal/session/session.go
@@ -263,6 +263,9 @@ func (s *TestSession) PushDUT(ctx context.Context) error {
 		return fmt.Errorf("deleting ISIS config before configuring test config: %w", err)
 	}
 
+	// After deleting, Wait few seconds here before config the protocol again
+	time.Sleep(5 * time.Second)
+
 	// Configure ISIS test config on DUT
 	_, err = ygnmi.Replace(ctx, s.DUTClient, ProtocolPath().Config(), dutConf)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


mojiiba:
I think the delete should be removed in this code since replace is supposed to replace the existing config and there is not need for delete if replace works as executed.

cprabha:
> I think the delete should be removed in this code since replace is supposed to replace the existing config and there is not need for delete if replace works as executed.

@mojiiba    As counters are being checked before adjacency, delete is required to get consistent counter values without any stale counter.

mojiiba:
> > I think the delete should be removed in this code since replace is supposed to replace the existing config and there is not need for delete if replace works as executed.
> 
> @mojiiba As counters are being checked before adjacency, delete is required to get consistent counter values without any stale counter.

@cprabha i understand your point. But should not also replace clean the counter if the replace is at the container (protocol) level?

gforeal:
Before PR/792, the code actually just doing that - the Replace only.

gforeal:
For Arista EOS, the current delete then replace been handled in the back-end just as-is - first deconfig then config but back to back!

gforeal:
Updated the PR to remove the recently added new deleting operation, thus the sleep is also not necessary anymore, plz review again!

nagarushirao:
pls update the PR description to match the fix

gforeal:
Cisco is the one hard to satisfy so far;
revert the Delete (thus no delay needed anymore) will pass for Arista but fails with Cisco.

gforeal:
By reverting the Delete operation, the test is passing with Arista but still flaky with Cisco mainly in the TestAuthentication section; Cisco team has agreed to address that issue with new PR.

Thanks all for the comments and review!

rszarecki:
What's Juniper view?

gforeal:
The test doesn't have a pass record for Juniper yet, looks it's failing with:

=== RUN   TestBasic/adjacency_state/state/area-address
    third_party/openconfig/featureprofiles/feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go:243: /network-instances/network-instance[name=DEFAULT]/protocols/protocol[identifier=ISIS][name=DEFAULT]/isis/interfaces/interface[interface-id=et-1/0/0]/levels/level[level-number=2]/adjacencies/adjacency[system-id=6401.0001.0000]/state/area-address: got []string{"49.0001", "49.0002"}, want []string{"49.0002", "49.0001"} (deadline exceeded)

cprabha:
@rszarecki  We prefer keep delete , it cleans up device before beginning with test.  That way it will be inline with other tests like bgp. We can add 5 secs if it is required by other vendors.

gforeal:
@cprabha the change in this PR doesn't have negative impact to Juniper, the tests this change is about are actually passing in Juniper, the current only failure is in the above TestBasic/adjacency_state/state/area-address;

Cisco already stated that the Delete is not supported, and the back to back delete then config has negative impact to Arista.

I'm not sure for which vendor or platform the added Delete is to target to? could you please share if any insight, thanks!

vosipchu:
+1 for @gforeal . Delete is not supported for Cisco and we won't be able to move forward with that. One way or another, Replace will have to be used.

rszarecki (commented):
LGTM

rszarecki (approved):

vosipchu:
LGTM for cisco

jjing1122 (approved):

sachendras:
LGTM

PR #1104, opened 2023-02-10
Add a QoS CLI/replace + OC/update in a single `SetRequest` to experimental

wenovus:
Reopened from https://github.com/openconfig/featureprofiles/pull/968
The old interface mixed test is removed.

This covers Arista only for now.

SetRequest contains:
* replace: "cli" current running-config + additional QoS config.
* update: OC Qos config.

----

Note: CLI is used to retrieve running-config instead of using a GetRequest/SubscribeRequest.


wenovus:
@dplore comments from https://github.com/openconfig/featureprofiles/pull/968 are addressed, PTAL.

dplore (approved):

wenovus:
Thanks for the review!

PR #1105, opened 2023-02-10
RT-1.2: Updating BGP peer config to have V4 and V6 routes to respective neighbor 

trathod1:
*  When IPv4/IPv6 (AfiSafi) is enabled for all neighbors. The IPv6 routes are received from the ATE over IPv4 and IPv6 peer.
*  The IPv6 routes received over IPv4 peer are considered active based on the route selection over NextHop (::ffff:192.0.2.6), as all the other parameter for the IPv6 routes is same between the IPv4 and IPv6 peer.
*  These IPv6 routes show up as Active under the IPv4 peer, but in RT-1.2 script expects these active routes under IPv6 peer in "verifyPrefixesTelemetryV6" function.
*  Thus to make sure script meets test requirement of ipv6 routes are exchanged over ipv6 neighbor, and similarly ipv4 routes over ipv4 neighbor, we need to disable inheritance of the other not-required protocol family under neighbor configuration.


xw-g:
Can you help add a few words describing the motivation of why making the suggestions (is it required by the README test plan)?

LimeHat:
Hi Xiao @xw-g 

This change is needed to align with the inheritance model discussed in openconfig/public#774 -> in the current code both IPv4 and IPv6 are enabled on a protocol level, so in a per-neighbor config we want to disable inheritance of the other protocol family (i.e. keep v4 peers enabled with ipv4-unicast afi/safi and v6 peers with ipv6-unicast afi/safi only)

We also introduce the same logic in another pr #1059

trathod1:
> Can you help add a few words describing the motivation of why making the suggestions (e.g. the current code does not meet the test plan README, or is it just about some vendor OS behaviors?). Thanks.

Updated PR description with more details and also updated README file with test specifics.

MarcCharlebois:
@xw-g, was there any updates on the pre-submit/merge efforts?

xw-g (approved):

xw-g:
b/271889146 finished the internal presubmit test. merging it now.

PR #1106, opened 2023-02-10
infer p4rt nodes nokia

Ankur19:


Ankur19:
@Hobbydos @trathod1 Does this look good for infering the P4RT Node Name in Nokia?

LimeHat:
@Ankur19 in the current generation of 36x400 linecards we use two asics per card, so the second index can be 0 (for ports 1-18) or 1 (19-36)

Ankur19:
> @Ankur19 in the current generation of 36x400 linecards we use two asics per card, so the second index can be 0 (for ports 1-18) or 1 (19-36)

Thanks for the info! will update accordingly.

LimeHat:
Thanks, looks good. I think this should cover the immediate need, and later we can revisit if needed.

One nit: comment above the function references a wrong function name (// inferP4RTNodesCisco)

feature/experimental/p4rt/internal/p4rtutils/p4rtutils.go
```diff
@@ -193,6 +194,40 @@ func explicitP4RTNodes() map[string]string {
...
+		port, err := strconv.Atoi(m[2])
+		if err != nil {
+			t.Fatalf("Error generating P4RT Node Name: %v", err)
+		}
+		if port > 18 {
+			res[p.ID()] = node(fpc, "1")
+			continue
+		}
+		res[p.ID()] = node(fpc, "0")
+	}
```
liulk, line 220:
```suggestion

var nokiaPortNameRE = regexp.MustCompile("ethernet-([0-9]+)/([0-9]+)")

// inferP4RTNodesNokia infers the P4RT node name from the port name for Nokia devices.
func inferP4RTNodesNokia(t testing.TB, dut *ondatra.DUTDevice) map[string]string {
	res := make(map[string]string)
	for _, p := range dut.Ports() {
		m := nokiaPortNameRE.FindStringSubmatch(p.Name())
		if len(m) != 3 {
			continue
		}

		fpc := m[1]
		port, err := strconv.Atoi(m[2])
		if err != nil {
			t.Fatalf("Error generating P4RT Node Name: %v", err)
		}
		asic := 0
		if port > 18 {
			asic = 1
		}
		res[p.ID()] = fmt.Sprintf("SwitchChip%s/%d", fpc, asic)
	}
```

jjing1122 (approved):

PR #1114, opened 2023-02-11
Remove ssh from tls_authentication_over_grpc_test

greg-dennis:
We don't require SSH support and test of whether authentication works "over grpc" should stick to testing over grpc


bstoll:
The README.md should be updated to reflect this change in the plan as well.  It sounds like this test validates the following:

* TODO: Set the user and password using AAA CLI on the device, configured to authenticate using TACACS+ with a fallback of the local database.
* Configure local user on the DUT using GNMI
* Ensure GNMI Get/Set requests succeed using the local credentials.
* Ensure GNMI Get/Set requests fail with invalid credentials.

The test plan is removing:
* Console credential verification
* SSH credential verification

My opinion is also that SSH and console access are still relevant tests - when a network device breaks, SSH and console access are still going be relevant for troubleshooting for quite a while.  I think they should be left and marked at least as TODO items.

PR #1119, opened 2023-02-14
Tweak p4rt utils for nokia

LimeHat:
Hi @Ankur19,
I added a small change on top of #1106 - if a user provided explicit P4RT node names, they will be used instead of inferred names.
Otherwise we default to inferred.
This allows us to use linecards with different interface/asic layouts internally


feature/experimental/p4rt/internal/p4rtutils/p4rtutils.go
```diff
@@ -259,6 +259,9 @@ func P4RTNodesByPort(t testing.TB, dut *ondatra.DUTDevice) map[string]string {
 		case ondatra.CISCO:
 			return inferP4RTNodesCisco(t, dut)
 		case ondatra.NOKIA:
+			if *args.P4RTNodeName1 != "" && *args.P4RTNodeName2 != "" {
```
Ankur19, line 262: Suggest moving this to the `inferP4RTNodesNokia()` function.

LimeHat: Done, thx

Ankur19 (dismissed):
LGTM!

PR #1122, opened 2023-02-14
RT1.4 Changes

agvarghe-cisco:
- Deviation for route-policy configuration under bgp peer-group
- Check for prefix installed before traffic test
- Add Config Lookup between delete and Replace to make sure config is deleted
- Deviation for bgp state to be active after ACL deny policy


feature/bgp/gracefulrestart/ate_tests/bgp_graceful_restart_test/bgp_graceful_restart_test.go
```diff
@@ -398,13 +453,19 @@ func TestTrafficWithGracefulRestartSpeaker(t *testing.T) {
...
+		if *deviations.RoutePolicyUnderPeerGroup {
+			configureRoutePolicy(t, dut, "ALLOW", oc.RoutingPolicy_PolicyResultType_ACCEPT_ROUTE)
+		}
 	})
 
 	// Configure BGP+Neighbors on the DUT
 	t.Run("configureBGP", func(t *testing.T) {
 		t.Log("Configure BGP with Graceful Restart option under Global Bgp")
 		dutConfPath := gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance).Protocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP")
 		gnmi.Delete(t, dut, dutConfPath.Config())
```
mojiiba, line 465: we convinced google team that delete is unnecessary in ISIS cases and replace should do what delete is trying to do + update. Same case here no point for delete + replace here, let's simply remove the delete and see how it goes. We can return the delete if they disagree.

mojiiba (commented):
i just have one comment, thanks a lot

feature/bgp/gracefulrestart/ate_tests/bgp_graceful_restart_test/bgp_graceful_restart_test.go
agvarghe-cisco: Removed delete operation and kept only Replace here

feature/bgp/gracefulrestart/ate_tests/bgp_graceful_restart_test/bgp_graceful_restart_test.go
```diff
@@ -162,30 +172,58 @@ func bgpWithNbr(as uint32, nbrs []*bgpNeighbor) *oc.NetworkInstance_Protocol {
...
 	bgpgr := g.GetOrCreateGracefulRestart()
 	bgpgr.Enabled = ygot.Bool(true)
 	bgpgr.RestartTime = ygot.Uint16(grRestartTime)
 	bgpgr.StaleRoutesTime = ygot.Uint16(grStaleRouteTime)
 
-	pg := bgp.GetOrCreatePeerGroup(peerGrpName)
+	pg := bgp.GetOrCreatePeerGroup(peerv4GrpName)
 	pg.PeerAs = ygot.Uint32(ateAS)
-	pg.PeerGroupName = ygot.String(peerGrpName)
+	pg.PeerGroupName = ygot.String(peerv4GrpName)
```
ming2004, line 185: This line seems redundant since line#183 has set this.

feature/bgp/gracefulrestart/ate_tests/bgp_graceful_restart_test/bgp_graceful_restart_test.go
```diff
@@ -162,30 +172,58 @@ func bgpWithNbr(as uint32, nbrs []*bgpNeighbor) *oc.NetworkInstance_Protocol {
...
+
+	pgv6 := bgp.GetOrCreatePeerGroup(peerv6GrpName)
+	pgv6.PeerAs = ygot.Uint32(ateAS)
+	pgv6.PeerGroupName = ygot.String(peerv6GrpName)
```
ming2004, line 189: This line seems redundant since line#187 has set this.

thesrinath (approved):

PR #1123, opened 2023-02-14
 Bringing down the tolerance to Zero and sending accurate prefixes

RanjithansCisco:
BGP RT 1.5 FP

1)Modified the traffic/flow to send appropriate prefix value for each test case

- First TC (Prefix -1) checks traffic /flow for 200 prefixes instead of 199., hence we see 0.5% of tolerance. So have modified the flow to send appropriate prefix value , similarly for every case via code

2) Tolerance value brought to zero to ensure zero traffic failure


LimeHat:
So if the expected loss is zero and you propose to use zero tolerance for this test, what is the point in introducing the global argument?

`BGPTrafficTolerance = flag.Int("arg_bgp_tolerance_value", 0, "Allowed tolerance for BGP traffic flow while comparing for pass or fail condition.")`

mojiiba:
> So if the expected loss is zero and you propose to use zero tolerance for this test, what is the point in introducing the global argument?
> 
> `BGPTrafficTolerance = flag.Int("arg_bgp_tolerance_value", 0, "Allowed tolerance for BGP traffic flow while comparing for pass or fail condition.")`

we added it in cases other vendors want to  use that. However, the correct implementation should pass without tolerance.

LimeHat:
> However, the correct implementation should pass without tolerance.

Shouldn't be this a deviation in this case?

mojiiba:
> > However, the correct implementation should pass without tolerance.
> 
> Shouldn't be this a deviation in this case?

sure, we will move this to deviation file

prinikasn (commented):
/gcbrun

prinikasn (approved):

PR #1127, opened 2023-02-15
gNOI-3.2: add check on linecard emptiness

LimeHat:
Current version of the test selects the last removable linecard to reboot, however it ignores the empty flag (so it tries to reboot an unpopulated card in some cases). This PR introduces an additional check to skip cards that are not populated

Additionally, fixed one typo in a log message and renamed one variable to avoid name collision with an imported package.


LimeHat:
Superseded by #1174

PR #1131, opened 2023-02-15

saarvind6481:


sezhang2 (changes requested):
Can you explain why rename queue names?

Please also mention that why vendor specific workarounds are needed? Is there any bug to track the workaround?

aaronmillisor:
> Can you explain why rename queue names?
> 
> Please also mention that why vendor specific workarounds are needed? Is there any bug to track the workaround?

At least part of the problem really stems from a deficiency in the OC-QoS model. Silicon One ASIC's, similar to many others in the industry, have a concept called a 'traffic-class' which also exists in many other asics. This traffic-class is a numeric identifier.

In IOS XR's configuration model, this traffic class has a [tight coupling to a specific priority value in the case of strict priority schedulers](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/qos/70x/configuration/guide/b-qos-cg-8k-70x/traffic_management_overview.html#concept_izx_bhw_m3b).

The missing traffic-class equivalent has been a long-standing concern for this model, however it now appears to be getting addressed through the introduction of queue-identifiers in [pull 772 for OC-QoS](https://github.com/openconfig/public/pull/772).

The way that this missing functionality is addressed seems to be varying across implementors. In XR we are allocating the traffic-class in the order they appear in config to satisfy the ordering requirements of the scheduling model. In order to get this to show up properly via OC, we are using queue names that are defined lexically ordered according to the desired scheduling policy.

Alternatives that implementors may use could include fixed queue names with an out of band association, or magic fixed queue names. Each approach has downsides.

We expect this ordering issue to be addressed after the merge of PR772.

Separately, QoS scheduling models are likely to vary across hardware platforms, and should be expected to change according to the asic/os implementation even when OC is used.

dplore:
Operationally I'd like to keep the queue names the same. But we can use CLI
commands to map the traffic class identifiers to the open config queue
names.

The OC model solution for this specific scenario will be merged soon.

guoshiuan:
Internal test result: b/269634520

prinikasn (commented):
/gcbrun

prinikasn (approved):

feature/experimental/qos/ate_tests/qos_output_queue_counters_test/qos_output_queue_counters_test.go
```diff
@@ -108,12 +112,13 @@ func TestQoSCounters(t *testing.T) {
...
-			"flow-nc1": {frameSize: 1000, trafficRate: 1, dscp: 56, queue: "7"},
-			"flow-af4": {frameSize: 400, trafficRate: 3, dscp: 32, queue: "4"},
-			"flow-af3": {frameSize: 300, trafficRate: 2, dscp: 24, queue: "3"},
-			"flow-af2": {frameSize: 200, trafficRate: 2, dscp: 16, queue: "2"},
-			"flow-af1": {frameSize: 1100, trafficRate: 1, dscp: 8, queue: "0"},
-			"flow-be1": {frameSize: 1200, trafficRate: 1, dscp: 0, queue: "1"},
+			"flow-nc1": {frameSize: 700, trafficRate: 7, dscp: 56, queue: "a_NC1"},
+			"flow-af4": {frameSize: 400, trafficRate: 4, dscp: 32, queue: "b_AF4"},
+			"flow-af3": {frameSize: 1300, trafficRate: 3, dscp: 24, queue: "c_AF3"},
+			"flow-af2": {frameSize: 1200, trafficRate: 2, dscp: 19, queue: "d_AF2"},
```
sezhang2, line 118: why change DSCP from 16 to 19?

sezhang2 (dismissed):
Per the discussion with [aaronmillisor](https://github.com/aaronmillisor) yesterday, please add the bug# for the workaround.

Please also address the comment about DSCP change as well.

sezhang2 (commented):
Does the test pass in your test bed with this PR?

guoshiuan:
> Does the test pass in your test bed with this PR?

Internal test result: b/269634520

feature/experimental/qos/ate_tests/qos_output_queue_counters_test/qos_output_queue_counters_test.go
saarvind6481: I have resubmitted the PR with DSCP 16

guoshiuan:
We (Sean, Viktor, Cisco team and I) had a discussion (b/268403096#comment9). Viktor has filed bugs to track the explanation of Cisco implementation, and we decided to move forward with this PR.

guoshiuan (approved):

prinikasn (approved):

PR #1133, opened 2023-02-15
Update TE-3.3

xw-g:
Changes:
* Implement the TODO.
* Update the 2nd level weight requirement.


manan-patel:
fixed under https://github.com/openconfig/featureprofiles/pull/1417

PR #1139, opened 2023-02-16
gNOI-5.1: create loopback interface

kjahed:
Create loopback interface in case a loopback interface does not exist.


marcushines (approved):

feature/gnoi/system/tests/ping_test/ping_test.go
```diff
@@ -47,6 +49,16 @@ const (
 	minimumStdDev = 1
 )
 
+var (
+	lo10Intf = attrs.Attributes{
+		Desc:    "lo10",
```
sachendras, line 54: Juniper loopback interfaces are lo0.0, lo0.1... lo0.10 etc. Can we please update accordingly?

Also, we will need to account for the following
1. Only one loopback interface is allowed per VRF.
    -  Hence, the code should check if a loopback already exits, if yes use that IP address.
2. Only lo0.0 is allowed in default VRF
    - If the test uses default vrf and there doesnt exist a loopback as per bullet#1 above, create lo0.0
    - If the test is a different VRF than default, then create lo0.1 or lo0.2 etc

For Viz:
@mohanasm @sgirishj

PR #1144, opened 2023-02-16
Update TE-11.2

xw-g:
Changes: 
* We'd like the traffic goes from one VRF to another VRF, instead of all happening in the DEFAULT VRF.


prinikasn:
These asks have been addressed in https://github.com/openconfig/featureprofiles/pull/1181. Closing this PR as requested.

Issue #1103, opened 2023-02-10
go src file modification


cprabha:
Hi,

I have requirement to verify of array of strings in unordered fashion and It needs some changes in /usr/local/go/src/reflect/deepequal.go file.

Please let me know how to open github pull request for go src files .

This change is needed to support RT 2.1 test.
Currently we have check.Equal , which validates only ordered verification of slice elements.

Similar to check.equal we need code to validate slice elements in unordered fashion.

Thanks,
Prabha

cprabha:
do we need to sort slices and then call deepEqual api ?
But got and want are both of type T.  Is there a function to sort slices of type T ? please could you provide some refs.

cprabha:
@xw-g , please could you check below code to checker.go  .

// Compare slices in un ordered way.
func UnOrderedEqual[T any, QT ygnmi.SingletonQuery[T]](query QT, want T) Validator {
	return Predicate(query, fmt.Sprintf("want %#v", want), func(got T) bool {

		v1 := reflect.ValueOf(got)
		v2 := reflect.ValueOf(want)
		if v1.Len() != v2.Len() {
			return false
		}
		if !v1.IsValid() || !v2.IsValid() {
			return v1.IsValid() == v2.IsValid()
		}
		if v1.Type() != v2.Type() {
			return false
		}
		found := 0
		for i := 0; i < v1.Len(); i++ {
			for j := 0; j < v2.Len(); j++ {
				if v1.Index(i).String() == v2.Index(j).String() {
					found++
					break
				}
			}
		}
		// If found all slices of v1 found will be v.len()
		return found == v1.Len()
	})
}

cprabha:
@xw-g ,  I have opened pull request for this change. Please could you check.

Thanks,
Prabha

PR #1145, opened 2023-02-16
RT 2.1 - base_adjacencies_test.go

cprabha:
Hi,

Area address telemetry contains both DUT and ATE Area addresses.  Test expects DUT Area Address and ATE Area address to be in specific order.
Added modification to verify the list of AreaAddress in unordered way. Also added the unordered list verification code in check.go

Issue Ref:  https://github.com/openconfig/featureprofiles/issues/1103

Thanks,
Prabha


cprabha:
@xw-g , please could review the changes for un ordered list verification. thanks

internal/check/check.go
```diff
@@ -423,3 +423,32 @@ func NotPresent[T any, QT ygnmi.SingletonQuery[T]](query QT) Validator {
 		return nil
 	})
 }
+
+// UnOrderedEqual is to compare slices in un ordered way.
+func UnOrderedEqual[T any, QT ygnmi.SingletonQuery[T]](query QT, want T) Validator {
+	return Predicate(query, fmt.Sprintf("want %#v", want), func(got T) bool {
+
+		v1 := reflect.ValueOf(got)
```
liulk, line 431: Rather than rolling our own comparison, let's just use the cmp library with the cmp.SortSlices option.

https://pkg.go.dev/github.com/google/go-cmp/cmp#Equal
https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#SortSlices

internal/check/check.go
```diff
@@ -423,3 +423,32 @@ func NotPresent[T any, QT ygnmi.SingletonQuery[T]](query QT) Validator {
 		return nil
 	})
 }
+
+// UnOrderedEqual is to compare slices in un ordered way.
+func UnOrderedEqual[T any, QT ygnmi.SingletonQuery[T]](query QT, want T) Validator {
```
liulk, line 428: UnOrderedEqual -> UnorderedEqual

liulk (changes requested):
Sorry, I somehow missed this.

internal/check/check.go
cprabha: updated, thank you

internal/check/check.go
cprabha: updated function name. Thank you

internal/check/check.go
```diff
@@ -423,3 +425,11 @@ func NotPresent[T any, QT ygnmi.SingletonQuery[T]](query QT) Validator {
 		return nil
 	})
 }
+
+// UnorderedEqual function is used to compare slices in un ordered way.
+func UnorderedEqual[T any, QT ygnmi.SingletonQuery[T]](query QT, want T) Validator {
+	return Predicate(query, fmt.Sprintf("want %#v", want), func(got T) bool {
+		// Sort slices to compare them in unorderd way.
+		return cmp.Equal(got, want, cmpopts.SortSlices(func(a, b string) bool { return a < b }))
```
liulk, line 433: Can we make `func(a, b string) bool { return a < b }` part of the argument to be supplied by caller?

cprabha: updated, please check

internal/check/check.go
```diff
@@ -423,3 +425,11 @@ func NotPresent[T any, QT ygnmi.SingletonQuery[T]](query QT) Validator {
...
 	})
 }
+
+// UnorderedEqual function is used to compare slices in un ordered way.
+func UnorderedEqual[T any, QT ygnmi.SingletonQuery[T]](query QT, want T, sortSliceCompFunc interface{}) Validator {
+	return Predicate(query, fmt.Sprintf("want %#v", want), func(got T) bool {
+		// Sort slices to compare them in unorderd way.
+		return cmp.Equal(got, want, cmpopts.SortSlices(sortSliceCompFunc))
+	})
+}
```
liulk, line 435: Sorry yesterday I was in a hurry to leave the office, so I didn't have time to give a more detailed comment.

Since these functions use generics, let's avoid the any interface by making the type more precise. Here the item is T, so the sorting is between two T's, but the got and want are []T.

```suggestion
// UnorderedEqual function is used to compare a slice of T in an unordered way.
func UnorderedEqual[T any, QT ygnmi.SingletonQuery[[]T]](query QT, want []T, less func(a, b T) bool) Validator {
	return Predicate(query, fmt.Sprintf("want %#v", want), func(got []T) bool {
		// Sort slices to compare them in unorderd way.
		return cmp.Equal(got, want, cmpopts.SortSlices(less))
	})
}
```

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
```diff
@@ -208,16 +208,18 @@ func TestBasic(t *testing.T) {
...
 		// There are about 16 RPCs executed in quick succession in this block.
 		// Increasing the wait-time value to accommodate this.
 		deadline = time.Now().Add(time.Minute)
 		adj := port1ISIS.Level(2).Adjacency(systemID)
+		sortSliceCompFunc = func(a, b string) bool { return a < b }
 		for _, vd := range []check.Validator{
 			check.Equal(adj.AdjacencyState().State(), oc.Isis_IsisInterfaceAdjState_UP),
 			check.Equal(adj.SystemId().State(), systemID),
-			check.Equal(adj.AreaAddress().State(), []string{session.ATEAreaAddress, session.DUTAreaAddress}),
+			check.UnorderedEqual(adj.AreaAddress().State(), []string{session.ATEAreaAddress, session.DUTAreaAddress}, sortSliceCompFunc),
```
liulk, line 222: Just inline `func (a, b string) bool { return a < b }` in the argument here and remove `var sortSliceCompFunc` above.

liulk (changes requested):

internal/check/check.go
cprabha: updated, thank you.

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
cprabha: updated, thank you

liulk (approved):

sachendras (approved):

PR #1154, opened 2023-02-17
TE-11.2, remove AFT check on NH viability

xw-g:
In Openconfig AFT model, it's not yet defined/concluded that if the viability of a NH should be represented by the existence of the NH, or should it be indicated by a new leaf container. Therefore, this PR suggests relax the AFT checking on NH now.

### DIFF
--- a/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
+++ b/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
@@ -282,23 +282,19 @@ func testIPv4BackUpSwitch(ctx context.Context, t *testing.T, args *testArgs) {
 	BaseFlow := createFlow(t, args.ate, args.top, "BaseFlow")
 
 	// validate programming using AFT
-	aftCheck(t, args.dut, dstPfx, []string{"192.0.2.6", "192.0.2.10"})
+	aftCheck(t, args.dut, dstPfx)
 	// Validate traffic over primary path port2, port3
 	validateTrafficFlows(t, args.ate, BaseFlow, false, []string{"port2", "port3"})
 
 	//shutdown port2
 	flapinterface(t, args.ate, "port2", false)
 	defer flapinterface(t, args.ate, "port2", true)
-	// validate programming using AFT
-	aftCheck(t, args.dut, dstPfx, []string{"192.0.2.10"})
 	// Validate traffic over primary path port3
 	validateTrafficFlows(t, args.ate, BaseFlow, false, []string{"port3"})
 
 	//shutdown port3
 	flapinterface(t, args.ate, "port3", false)
 	defer flapinterface(t, args.ate, "port3", true)
-	// validate programming using AFT
-	aftCheck(t, args.dut, dstPfx, []string{"192.0.2.14"})
 	// validate traffic over backup
 	validateTrafficFlows(t, args.ate, BaseFlow, false, []string{"port4"})
 }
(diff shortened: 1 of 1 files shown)


feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -282,23 +282,19 @@ func testIPv4BackUpSwitch(ctx context.Context, t *testing.T, args *testArgs) {
 	BaseFlow := createFlow(t, args.ate, args.top, "BaseFlow")
 
 	// validate programming using AFT
-	aftCheck(t, args.dut, dstPfx, []string{"192.0.2.6", "192.0.2.10"})
```
robshakir, line 285: Replace this with a `TODO`? It seems like we should say that when we have concluded exactly what to mark here then we should come back and add the check.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -282,23 +282,19 @@ func testIPv4BackUpSwitch(ctx context.Context, t *testing.T, args *testArgs) {
...
+	aftCheck(t, args.dut, dstPfx)
 	// Validate traffic over primary path port2, port3
 	validateTrafficFlows(t, args.ate, BaseFlow, false, []string{"port2", "port3"})
 
 	//shutdown port2
 	flapinterface(t, args.ate, "port2", false)
 	defer flapinterface(t, args.ate, "port2", true)
-	// validate programming using AFT
```
robshakir, line 292: Ditto re: TODO.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -282,23 +282,19 @@ func testIPv4BackUpSwitch(ctx context.Context, t *testing.T, args *testArgs) {
...
-	aftCheck(t, args.dut, dstPfx, []string{"192.0.2.10"})
 	// Validate traffic over primary path port3
 	validateTrafficFlows(t, args.ate, BaseFlow, false, []string{"port3"})
 
 	//shutdown port3
 	flapinterface(t, args.ate, "port3", false)
 	defer flapinterface(t, args.ate, "port3", true)
-	// validate programming using AFT
-	aftCheck(t, args.dut, dstPfx, []string{"192.0.2.14"})
```
robshakir, line 301: same TODO.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -364,19 +360,4 @@ func aftCheck(t testing.TB, dut *ondatra.DUTDevice, prefix string, expectedNH []
 	if got := len(aftNHG.NextHop); got < 1 && aftNHG.BackupNextHopGroup == nil {
 		t.Fatalf("Prefix %s reachability didn't switch to backup path", prefix)
 	}
-	if len(aftNHG.NextHop) != 0 {
```
robshakir, line 367: Is there a reason to remove this code entirely, shouldn't it pass with other cases?

xw-g: Because the entire block of codes says: `for each streamed next hop, it has to be in the expectedNH`, **not** `for each IP in the expectedNHs, it has to be in the streamed next hops`. Basically the code blocks is defining that only the viable NHs can be streamed in the AFT. (which is not a conclusion yet, see the PR description).

Also because the way the code is structured is kinda of confusing. So, I just removed it, instead of refactor it into `for each streamed next hop, it has to be in the expectedNH`.

I added a TODO here same as above.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
xw-g: Done.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
xw-g: Done.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
xw-g: Done.

mojiiba (approved):

robshakir (approved):

PR #1155, opened 2023-02-18
TE-1.1: refactor usage of ExplicitPortSpeed deviation

LimeHat:
This PR updates how ExplicitPortSpeed deviation is handled in this test case.

In the current version the deviation uses a separate gnmi set update request that is called after an interface is configured with gnmi set replace.
The direct consequence of this approach is that a DUT interface undergoes through an up-down-up state transition during the test run (since configureDUT() function is called twice, L256 & L280): 
https://github.com/openconfig/featureprofiles/blob/1e17e8a2091cb6854f5376726186e63f8f32cfd3/feature/interface/staticarp/ate_tests/static_arp_test/static_arp_test.go#L256 https://github.com/openconfig/featureprofiles/blob/1e17e8a2091cb6854f5376726186e63f8f32cfd3/feature/interface/staticarp/ate_tests/static_arp_test/static_arp_test.go#L280
This state transition is problematic with current ATE capabilities/test logic since ATE does not own the "poisoned" mac, but only uses a filter to track and count data plane frames. Thus ATE is unable to process control plane information destined to the poisoned mac (neighbor advertisement) after the interface state change.

In order to resolve this:
1) we introduce additional function to fptest package that just returns an interface speed in OC-compatible format, instead of doing a direct gNMI update of the interface speed
2) we move ExplicitPortSpeed deviation logic from configureDUT() function to configInterfaceDUT() function, so all interface settings are processed & sent to a DUT together in a single gnmi set request.


jjing1122:
Internal test result: b/270225344

jjing1122 (approved):

PR #1160, opened 2023-02-20
Adding Deviation for gRIBI state under network instance.

MarcCharlebois:
Adding deviation for Nokia since it's required to have gRIBI enabled in the Network instance. Using CLI Origin to push the config since it's only accessible in native models.

This code is a Contribution to the OpenConfig Feature Profiles project (“Work”) made under the Google Software Grant and Corporate Contributor License Agreement (“CLA”) and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia’s intellectual property are granted for any other purpose. This code is provided on an “as is” basis without any warranties of any kind.


internal/fptest/networkinstance.go
```diff
@@ -43,3 +45,29 @@ func AssignToNetworkInstance(t testing.TB, d *ondatra.DUTDevice, i string, ni st
...
 	}
 }
+
+// EnableGribiOnNetworkInstance enables Gribi protocol under network instance.
+func EnableGribiUnderNetworkInstance(t testing.TB, d *ondatra.DUTDevice, ni string) {
+	t.Helper()
+	if ni == "" {
+		t.Fatalf("Network instance not provided for protocol definition")
+	}
+	cliCmd := fmt.Sprintf("/set network-instance %s protocols gribi admin-state enable", ni)
```
robshakir, line 55: Marc -- do you prefer to do this in the SRLinux model vs. CLI? Per our discussion given that this is behind a deviation this seems OK (and we should figure out the plan here), but I'm not sure this _must_ be CLI :-)

xw-g, line 55: @trathod1 since this is a very Nokia CLI specific, do you mind do a pre condition check here? e.g. `if tc.vendor != "Nokia", t.Fatlf ()`

MarcCharlebois: @robshakir, if we're allowed to use the native model for this deviation, I'd prefer that. I'll work on a commit to migrate it. cc @xw-g

MarcCharlebois: @xw-g I'll add this in the commit for the migration from CLI origin to Native model.

MarcCharlebois: Added the safeguard, I decided to do a no-op instead of `t.Fatalf`. This way the code won't be executed by other vendors and the test won't fail if someone mistakenly uses this deviation

MarcCharlebois:
@xw-g  have you had a chance to review this PR with the changes?

internal/fptest/networkinstance.go
```diff
@@ -43,3 +45,50 @@ func AssignToNetworkInstance(t testing.TB, d *ondatra.DUTDevice, i string, ni st
...
+						JsonIetfVal: []byte(`"enable"`),
+					},
+				},
+			}},
+		}
+		gnmiClient := d.RawAPIs().GNMI().Default(t)
+		if _, err := gnmiClient.Set(context.Background(), gpbSetRequest); err != nil {
+			t.Fatalf("Enabling Gribi on network-instance %s failed with unexpected error: %v", ni, err)
+		}
+	} else {
```
xw-g, line 90: nit, prefers just to return instead of using else

xw-g:
working on @MarcCharlebois can you help resolve the merge conflict? Thanks.

xw-g:
just need to resolve the merge conflict and we are ready to go.

MarcCharlebois:
> just need to resolve the merge conflict and we are ready to go.

Fixed nit + merge conflict.

internal/fptest/networkinstance.go
```diff
@@ -43,3 +45,50 @@ func AssignToNetworkInstance(t testing.TB, d *ondatra.DUTDevice, i string, ni st
 		gnmi.Update(t, d, gnmi.OC().NetworkInstance(ni).Config(), netInst)
 	}
 }
+
+// EnableGribiOnNetworkInstance enables Gribi protocol under network instance.
+func EnableGribiUnderNetworkInstance(t testing.TB, d *ondatra.DUTDevice, ni string) {
+	t.Helper()
+	if ni == "" {
+		t.Fatalf("Network instance not provided for protocol definition")
```
xw-g, line 53: call out it's gRIBI.

internal/fptest/networkinstance.go
```diff
@@ -43,3 +45,50 @@ func AssignToNetworkInstance(t testing.TB, d *ondatra.DUTDevice, i string, ni st
...
+					},
+				},
+			}},
+		}
+		gnmiClient := d.RawAPIs().GNMI().Default(t)
+		if _, err := gnmiClient.Set(context.Background(), gpbSetRequest); err != nil {
+			t.Fatalf("Enabling Gribi on network-instance %s failed with unexpected error: %v", ni, err)
+		}
+	default:
+		t.Logf("Ignored deviation_explicit_gribi_under_network_instance as it is vendor specific.")
```
xw-g, line 92: t.Fatalf("the current vendor %s does not support the balbla deviation flag %s")

xw-g:
b/270975260 has LGTM for this deviation. All logic in this PR is gated by the deviation, therefore merging this PR with my LGTM onlyl.

xw-g (approved):

xw-g (approved):

PR #1164, opened 2023-02-21
Update TE-11.1

xw-g:
Changes: 
* Change the topology to have traffic flow through VRFs, instead of all happening in the DEFAULT VRF.
* Remove the non hierarchical scenario. Make all tests happening under the same hierarchical route resolution scenario.


manan-patel:
TE 11.1 is currently having 2 TC's

case 1: TestDirectBackupNexthopGroup
case 2: TestIndirectBackupNexthopGroup

As per the instructions in the readme, we are only mapping to case 2 with some additional steps, should we update the TC's with provided instructions?

PR #1167, opened 2023-02-22
QoS: Rollback PR1100 for sequence# change in scheduler policy

sezhang2:
Rollback https://github.com/openconfig/featureprofiles/pull/1100 for sequence# change in scheduler policy


dplore:
Explanation of the intent here is a given sequence id may have multiple queues assigned.  Strict priority queues are assigned to  sequence `0` and WRR queues are assigned to sequence `1`.

dplore (approved):

PR #1171, opened 2023-02-22
RT-1.5: Updating neighbor configuration for missing routing-policy and specific afi-safi config

trathod1:
This PR introduces 2 changes:

1. route policy is always added and configured on a per-neighbor basis if RoutePolicyUnderNeighborAfiSafi deviation is not enabled,

2. explicitly enables/disables both address families for each BGP neighbor (ipv4 en. / ipv6 dis. for v4 peers, v4 dis., v6 en. for v6 peers). This change is needed to align with the inheritance model discussed in Remove the default of "enabled" of BGP AFI-SAFI to support BGP inheritance model public#774 -> in the current code both IPv4 and IPv6 are enabled on a protocol level, so in a per-neighbor config we want to disable inheritance of the other protocol family.

Reference- https://github.com/openconfig/featureprofiles/pull/1059


feature/bgp/prefixlimit/ate_tests/bgp_prefix_limit_test/bgp_prefix_limit_test.go
```diff
@@ -130,9 +130,7 @@ func configureDUT(t *testing.T, dut *ondatra.DUTDevice) {
 		fptest.AssignToNetworkInstance(t, dut, p1, *deviations.DefaultNetworkInstance, 0)
 		fptest.AssignToNetworkInstance(t, dut, p2, *deviations.DefaultNetworkInstance, 0)
 	}
-	if *deviations.RoutePolicyUnderNeighborAfiSafi {
-		configureRoutePolicy(t, dut, rplName, rplType)
-	}
+	configureRoutePolicy(t, dut, rplName, rplType)
```
robshakir, line 133: Discussed that this ensures that the route policy itself is created regardless of the state of the deviation.

xw-g:
changes seems good to me. pending internal presubmit tests and then merge if no issues.

MarcCharlebois:
@xw-g any progress on the pre-submit tests?

xw-g:
b/272085856 is the internal presubmit effort.

xw-g (approved):

PR #1172, opened 2023-02-22
OC-1.2: Update to support ExplicitPortSpeed & ExplicitInterfaceInDefaultVRF deviations

trathod1:
In this pr we introduce two changes:

1. when ExplicitInterfaceInDefaultVRF is used, we need to unassign interfaces from the default vrf before they can be assigned to a custom vrf. unassignPort() function is added for this purpose, which is called only when ExplicitInterfaceInDefaultVRF is true
2. added fptest.SetPortSpeed() function calls when ExplicitPortSpeed is true

Reference - https://github.com/openconfig/featureprofiles/pull/1161


feature/networkinstance/ate_tests/defaults_test/defaults_test.go
```diff
@@ -131,12 +137,22 @@ func TestDefaultAddressFamilies(t *testing.T) {
...
+				fptest.SetPortSpeed(t, dutP1)
+				fptest.SetPortSpeed(t, dutP2)
+			}
 			d := &oc.Root{}
 			// Assign two ports into the network instance.
-			assignPort(t, d, dut.Port(t, "port1").Name(), tc.niName, dutPort1)
-			assignPort(t, d, dut.Port(t, "port2").Name(), tc.niName, dutPort2)
+			assignPort(t, d, dutP1.Name(), tc.niName, dutPort1)
+			assignPort(t, d, dutP2.Name(), tc.niName, dutPort2)
+			if *deviations.ExplicitInterfaceInDefaultVRF {
```
xw-g, line 152: Let's be more specific `if *deviations.ExplicitInterfaceInDefaultVRF && tc.niName == *deviation.DefaultNetworkInstance`, so that at least this PR changes would be immune to possible test cases ordering changes? What do you think?

xw-g:
Looks good except a minor comment.

feature/networkinstance/ate_tests/defaults_test/defaults_test.go
LimeHat: @xw-g in fact, the current proposal does not depend on the ordering: we need to unassign a port before we can assign to a new VRF in all cases (default->non-default, or non-default->default).

`if *deviations.ExplicitInterfaceInDefaultVRF && tc.niName == *deviation.DefaultNetworkInstance` will work, but it will achieve the inverse target: only default-> non-default test order will work.

trathod1: We would need to `unassignPort` from current vrf, whenever we more ports across vrf, so we need this `unassignPort` call even when `tc.niName == "xyz"`.  Considering that, in future if testcase order changes, current code should work fine.

MarcCharlebois:
@xw-g Anything else required to drive this to merge?

LimeHat:
@xw-g i updated the PR based on our conversation, please take a look.
I can also remove the defers and just move unassign functions to the bottom, don't have a strict preference here

Additionally, I copied the changes from #1041 to the OTG version of the test that wasn't updated

xw-g:
b/272049151 internal presubmit effort.

xw-g (approved):

PR #1179, opened 2023-02-23
Deleting zero Election ID case for P4RT-2.1

rsindhujha:
As per comments from google in last week meeting, deleting the below testcase 

// Test Zero client with 0 electionID
func TestZeroMaster(t *testing.T) {
	dut := ondatra.DUT(t, "dut")
	configureDeviceId(t, dut)
	configurePortId(t, dut)
	test := testArgs{
		desc:       pDesc,
		lowID:      inId0,
		highID:     inId0,
		handle:     clientConnection(t, dut),
		deviceID:   deviceId,
		wantFail:   true,
		wantStatus: 3,
	}
	t.Run(test.desc, func(t *testing.T) {
		resp, err := streamP4RTArb(&test)
		if err == nil && test.wantFail {
			t.Errorf("Zero ElectionID (0,0) is allowed: %v", err)
			removeClient(test.handle)
		} else {
			t.Logf("Zero ElectionID (0,0) connection failed as expected: %v", err)
		}
		// Validate status code
		if resp != test.wantStatus {
			t.Errorf("Incorrect status code received: want %d, got %d", test.wantStatus, resp)
		}
	})
}


Ankur19 (dismissed):
LGTM!

greg-dennis (approved):
Approved for Go quality only

LimeHat:
Hey Ankur @Ankur19 , Rob @robshakir,
Could you please provide a more detailed comment on this change?

I think initially this was discussed in #849 and the conclusion was that (0,0) is invalid.
Does that imply that (0,0) now is considered valid? If so, what is the expected behavior (perhaps accept the connection & always treat the client as a non-primary)? Or both behaviors are considered valid?

robshakir:
Sorry, I had missed the conversation in #849. 

There are four cases here I think:

* "specified" - the `Uint128` is set with one of the fields `high` or `low` set to a non-zero value. In this case, the `election_id` is clearly valid and >0.
* "not specified" - the `Uint128` is not set _at all_ -- since this is a message, we can tell that the `election_id` field is set to nothing.
* "specified as zero" - the `Uint128` is set explicitly to `low: 0, high: 0`.
* "non-nil uint128" - the `Uint128` message is set, but the fields are not explicitly set.

In gRIBI, we decided that the two ambiguous cases should be treated as invalid to have consistent treatment of the zero value throughout the specification. The "not specified" case is invalid in `SINGLE_PRIMARY` mode too. 

This text in the P4RT spec in §16.2 is relevant IMHO:

> The sender need not specify an election_id. If the election_id is not specified, the sender's election_id is considered lower than any election_id, and the sender will thus never become primary. This way, a controller can choose to be a backup controller, in order to avoid “flapping” (if a backup controller connects to the switch shortly before the actual primary controller, therefore becoming primary temporarily).

In this case, we're saying that the `election_id` is not specified. I interpret this to _at least_ mean the "not specified" case above. I'd consider the last two ("specified as zero", and "non-nil uint128") to come under the behaviour in §8.3:

> p4runtime.proto uses proto3 syntax, and so it does not allow not specifying a scalar data type, such as a uint32. Therefore, we usually reserve value 0 for those fields to mean unset. In particular, 0 is not a valid P4 object ID and it is an error to specify 0 for any P4 object ID in a non-read request towards the switch, such as in a WriteRequest or a SetForwardingPipelineConfigRequest.

since this explicit says zero means _unset_ then the last two cases should likely be valid and just treated the same as the `election_id` message not being set at all.

I'm going to ping this response to one of the folks who wrote the p4rt spec and see if they agree with this analysis.

[0]: code to reproduce the checks above:

```golang
package main

import (
	"fmt"

	log "github.com/golang/glog"
	"github.com/golang/protobuf/proto"

	p4pb "github.com/p4lang/p4runtime/go/p4/v1"
)

func main() {
	hasElectionID := func(u *p4pb.MasterArbitrationUpdate) bool {
		return u.GetElectionId() != nil
	}

	nonZeroElectionID := func(u *p4pb.MasterArbitrationUpdate) bool {
		return u.GetElectionId() != nil && !(u.GetElectionId().GetHigh() == 0 && u.GetElectionId().GetLow() == 0)
	}

	marshalUnmarshal := func(u *p4pb.MasterArbitrationUpdate) *p4pb.MasterArbitrationUpdate {
		b, err := proto.Marshal(u)
		if err != nil {
			log.Exitf("cannot marshal protobuf: %v", err)
		}
		m := &p4pb.MasterArbitrationUpdate{}
		if err := proto.Unmarshal(b, m); err != nil {
			log.Exitf("cannot unmarshal protobuf: %v", err)
		}
		return m
	}

	u := &p4pb.MasterArbitrationUpdate{}
	fmt.Printf("not specified: %v %v\n", hasElectionID(marshalUnmarshal(u)), nonZeroElectionID(marshalUnmarshal(u)))

	v := &p4pb.MasterArbitrationUpdate{ElectionId: &p4pb.Uint128{Low: 0, High: 1}}
	fmt.Printf("specified: %v %v\n", hasElectionID(marshalUnmarshal(v)), nonZeroElectionID(marshalUnmarshal(v)))

	w := &p4pb.MasterArbitrationUpdate{ElectionId: &p4pb.Uint128{Low: 0, High: 0}}
	fmt.Printf("specified as zero: %v %v\n", hasElectionID(marshalUnmarshal(w)), nonZeroElectionID(marshalUnmarshal(w)))

	x := &p4pb.MasterArbitrationUpdate{ElectionId: &p4pb.Uint128{}}
	fmt.Printf("non-nil uint128: %v %v\n", hasElectionID(marshalUnmarshal(x)), nonZeroElectionID(marshalUnmarshal(x)))
}
```

Output:

```
[15:22] src/robjs/p4elec                                                                                                                                                                                                                             
▶ go run main.go
not specified: false false
specified: true true
specified as zero: true false
non-nil uint128: true false
```

LimeHat:
Thanks Rob. Personally, I agree with your reading of the spec.

>I'm going to ping this response to one of the folks who wrote the p4rt spec and see if they agree with this analysis.

Sounds good. We'll wait for an update and update our implementation if needed to treat 0,0 as 'not specified'.

Another question. If we define 0,0 to be equal to 'not specified', perhaps this test scenario should be updated, but not removed? Or at least marked as a todo item to test the behavior.

robshakir:
I agree -- given that there is some discussion here, that would imply me that it is worth having a test to ensure that the behaviour we've decided on is encoded into a test.

Ankur19:
> I agree -- given that there is some discussion here, that would imply me that it is worth having a test to ensure that the behaviour we've decided on is encoded into a test.

I am in the process of clarifying the Test Case as part of #1192 . Will also add the Zero Election Id test case and validation requirements there.

robshakir:
@Ankur19 - thanks. Have you already clarified with the p4rt specification owners what the expectation is for zero election ID?

Ankur19:
> @Ankur19 - thanks. Have you already clarified with the p4rt specification owners what the expectation is for zero election ID?

No, I haven't yet.

robshakir:
> > @Ankur19 - thanks. Have you already clarified with the p4rt specification owners what the expectation is for zero election ID?
> 
> No, I haven't yet.

@Ankur19 I added you to the email thread :-)

Ankur19:
> @Ankur19 , Should i modify the zero election test, based on your README update #1192 ?

We are still waiting for some clarification from the P4 team. Can you keep it as is for now?

rsindhujha:
> > @Ankur19 , Should i modify the zero election test, based on your README update #1192 ?
> 
> @rsindhujha We are still waiting for some clarification from the P4 team. Can you keep it as is for now?

Sure, got it. Thankyou!

LimeHat:
@Ankur19 just to confirm, implementations should treat (0,0) as 'not specified' from now on, correct?

What about the other case Rob mentioned - `"non-nil uint128" - the Uint128 message is set, but the fields are not explicitly set.` - the same applies?

Ankur19:
> @Ankur19 just to confirm, implementations should treat (0,0) as 'not specified' from now on, correct?
> 
> What about the other case Rob mentioned - `"non-nil uint128" - the Uint128 message is set, but the fields are not explicitly set.` - the same applies?

They are the same and imply `unset` election_id. 

```
p4runtime.proto uses proto3 syntax, and so it does not allow not specifying a scalar data type, such as a uint32. Therefore, we usually reserve value 0 for those fields to mean unset. In particular, 0 is not a valid P4 object ID and it is an error to specify 0 for any P4 object ID in a non-read request towards the switch, such as in a WriteRequest or a SetForwardingPipelineConfigRequest.
```

rsindhujha:
> > > > @Ankur19 , Should i modify the zero election test, based on your README update #1192 ?
> > > 
> > > 
> > > @rsindhujha We are still waiting for some clarification from the P4 team. Can you keep it as is for now?
> > 
> > 
> > Sure, got it. Thankyou!
> 
> @rsindhujha The Zero ElectionID expectation in #1192 can now be implemented.
Hi @Ankur19 . This is taken care now.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -316,21 +316,27 @@ func TestZeroMaster(t *testing.T) {
...
 	t.Run(test.desc, func(t *testing.T) {
 		resp, err := streamP4RTArb(&test)
-		if err == nil && test.wantFail {
-			t.Errorf("Zero ElectionID (0,0) is allowed: %v", err)
+		if err != nil && test.wantFail {
+			t.Errorf("Zero ElectionID (0,0) is rejected by P4RT server: %v", err)
 			removeClient(test.handle)
 		} else {
-			t.Logf("Zero ElectionID (0,0) connection failed as expected: %v", err)
+			t.Logf("Zero ElectionID (0,0) connection success as expected: %v", err)
```
Ankur19, line 330: Remove the `else` statement here and add a `return` inside the if statement.

greg-dennis (approved):

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: @Ankur19 ,
Does this look ok?

`t.Run(test.desc, func(t *testing.T) {
		resp, err := streamP4RTArb(&test)
		if err != nil && test.wantFail {
			t.Errorf("Zero ElectionID (0,0) is rejected by P4RT server: %v", err)
		} 
		// Validate status code
		if resp != test.wantStatus {
			t.Errorf("Incorrect status code received: want %d, got %d", test.wantStatus, resp)
		}
		// Validate the response for Read/Write for zero election ID
		validateRWResp(t, &test)
		// Disconnect Primary
		removeClient(test.handle)
	}`

rsindhujha: With t.Errorf do we need a separate return statement?

Ankur19:
```
t.Run(test.desc, func(t *testing.T) {
		resp, err := streamP4RTArb(&test)
		if err != nil && test.wantFail {
			t.Errorf("Zero ElectionID (0,0) is rejected by P4RT server: %v", err)
			removeClient(test.handle)
                        return
		} 
		t.Logf("Zero ElectionID (0,0) connection success as expected: %v", err)
                ....
```

rsindhujha: Thanks i have taken care of this.

Ankur19 (dismissed):
LGTM

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -316,21 +316,27 @@ func TestZeroMaster(t *testing.T) {
...
 	}
 	t.Run(test.desc, func(t *testing.T) {
 		resp, err := streamP4RTArb(&test)
-		if err == nil && test.wantFail {
-			t.Errorf("Zero ElectionID (0,0) is allowed: %v", err)
+		if err != nil && test.wantFail {
 			removeClient(test.handle)
-		} else {
-			t.Logf("Zero ElectionID (0,0) connection failed as expected: %v", err)
+			t.Errorf("Zero ElectionID (0,0) is rejected by P4RT server: %v", err)
```
greg-dennis, line 328: it doesn't make sense why there should be an error here. The if-condition says this is the case when there is an error but we are _expecting_ a failure.

Ankur19: The `wantFail` variable is not required at all. There is only one test case and it is required to pass.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -316,21 +316,27 @@ func TestZeroMaster(t *testing.T) {
...
+			return
 		}
+		t.Logf("Zero ElectionID (0,0) connection success as expected: %v", err)
```
Ankur19, line 331: Also this is not required here. There won't be any `err` when this line is reached.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: i will remove the wantFail and the err in my next commit. Thanks

rsindhujha: this is done!

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: Removed wantFail and err

LimeHat:
Hi @Ankur19 @robshakir 
One more question regarding server behavior with (0,0) & `not specified` election-ids:

Should a server allow for an 'unlimited' number of those clients? Or only a single client w/o an election id should be allowed (per device-id)?
I think that rules outlined in [section 5.3 of the spec](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-updates) imply the former, e.g.
> If the election_id is set and is already used by another controller for the same (device_id, role), the P4Runtime server shall terminate the stream by returning an INVALID_ARGUMENT error.

but want to confirm with you.

Ankur19:
> Hi @Ankur19 @robshakir One more question regarding server behavior with (0,0) & `not specified` election-ids:
> 
> Should a server allow for an 'unlimited' number of those clients? Or only a single client w/o an election id should be allowed (per device-id)? I think that rules outlined in [section 5.3 of the spec](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-updates) imply the former, e.g.
> 
> > If the election_id is set and is already used by another controller for the same (device_id, role), the P4Runtime server shall terminate the stream by returning an INVALID_ARGUMENT error.
> 
> but want to confirm with you.

Hi @LimeHat Zero electionID is considered as `unset` so the `If election_id is set ...` scenario should not be considered. It is expected that multiple clients/controllers can connect to the P4RT server with electionID zero/unset.

LimeHat:
Thx for the confirmation

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -323,21 +323,26 @@ func TestZeroMaster(t *testing.T) {
...
+			t.Errorf("Zero ElectionID (0,0) is rejected by P4RT server: %v", err)
+			return
 		}
+		t.Logf("Zero ElectionID (0,0) connection success as expected")
 		// Validate status code
 		if resp != test.wantStatus {
 			t.Errorf("Incorrect status code received: want %d, got %d", test.wantStatus, resp)
 		}
+		// Validate the response for Read/Write for zero election ID
+		validateRWResp(t, &test)
```
Ankur19, line 343: This should verify Get and Set of `ForwardingPipelineConfig`

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -323,21 +323,26 @@ func TestZeroMaster(t *testing.T) {
 		highID:     inId0,
 		handle:     clientConnection(t, dut),
 		deviceID:   deviceId,
-		wantFail:   true,
-		wantStatus: 3,
+		wantWrite:  false,
```
greg-dennis, line 326: wantWrite and wantRead are unused. If they remain unused, please delete

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -323,21 +323,26 @@ func TestZeroMaster(t *testing.T) {
...
 	}
 	t.Run(test.desc, func(t *testing.T) {
 		resp, err := streamP4RTArb(&test)
-		if err == nil && test.wantFail {
-			t.Errorf("Zero ElectionID (0,0) is allowed: %v", err)
+		if err != nil {
 			removeClient(test.handle)
-		} else {
-			t.Logf("Zero ElectionID (0,0) connection failed as expected: %v", err)
+			t.Errorf("Zero ElectionID (0,0) is rejected by P4RT server: %v", err)
```
greg-dennis, line 334: instead of t.Errorf followed by return, just use t.Fatalf

greg-dennis (changes requested):

Ankur19:
@rsindhujha @LimeHat Sorry for the back and forth with the definition of `unset`.

We found some inconsistencies in the definition and made some updates to the spec here: https://github.com/p4lang/p4runtime/pull/418/files

As per the spec, `unset` is when the electionID is not set, and only that is allowed to have multiple clients. However, if electionID is set to (0, 0) then it is considered as SET and only a single client would be allowed for such a scenario.


For more context. Look at tests here:

Test1: https://github.com/sonic-net/sonic-pins/blob/main/p4rt_app/tests/arbitration_test.cc#L202

Test2: https://github.com/sonic-net/sonic-pins/blob/main/p4rt_app/tests/arbitration_test.cc#L214

LimeHat:
Thank @Ankur19 . Does that also imply that, in principle, (0,0) controller could become a primary (if it is a very first controller connected to the server)?

Ankur19:
> https://github.com/sonic-net/sonic-pins/blob/main/p4rt_app/tests/arbitration_test.cc#L202

Yes. Look at the test case here for reference: https://github.com/sonic-net/sonic-pins/blob/main/p4rt_app/tests/arbitration_test.cc#L202

Ankur19:
> Thank @Ankur19 . Does that also imply that, in principle, (0,0) controller could become a primary (if it is a very first controller connected to the server)?

Yes. Look at the test case here: https://github.com/sonic-net/sonic-pins/blob/main/p4rt_app/tests/arbitration_test.cc#L202

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: Updated the Testcase for two clients with unset electionId. Please check

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: Thanks @greg-dennis. These variables not used anymore.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: Used t.Fatalf wherever necessary in latest commit .

greg-dennis (changes requested):

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -312,33 +312,83 @@ func removeClient(handle *p4rt_client.P4RTClient) {
...
+					t.Fatalf("Errors while sending Arbitration Request with unset Election ID")
+				}
+			}
+			time.Sleep(1 * time.Second)
+			// Validate status code
+			resp, err := getRespCode(&test)
+			if resp != test.wantStatus {
+				t.Fatalf("Incorrect status code received: want %d, got %d", test.wantStatus, resp)
+			}
+			if err != nil {
```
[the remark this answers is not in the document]
rsindhujha: Included the err in t.Errorf

sachendras (dismissed):

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -312,33 +312,83 @@ func removeClient(handle *p4rt_client.P4RTClient) {
...
+			deviceID:   deviceId,
+			wantStatus: 5,
+		},
 	}
-	t.Run(test.desc, func(t *testing.T) {
-		resp, err := streamP4RTArb(&test)
-		if err == nil && test.wantFail {
-			t.Errorf("Zero ElectionID (0,0) is allowed: %v", err)
+
+	// Connect 2 clients to same deviceId with unset electionId
```
greg-dennis, line 333: go style says comments should be sentence style: starting with a capital and ending with a period. Please make that true here and with the other comments below.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -312,33 +312,83 @@ func removeClient(handle *p4rt_client.P4RTClient) {
...
+			resp, err := getRespCode(&test)
+			if resp != test.wantStatus {
+				t.Fatalf("Incorrect status code received: want %d, got %d", test.wantStatus, resp)
+			}
+			if err != nil {
+				t.Errorf("Errors seen in ClientArbitration response: %v", err)
+			}
+			t.Logf("Arbitration response status code is as expected for unset ElectionId")
+			// GetForwardingPipeline
+			_, err = test.handle.GetForwardingPipelineConfig(&p4_v1.GetForwardingPipelineConfigRequest{
```
greg-dennis, line 363: Again, these comments should be sentence style like "// Get the forwarding pipeline configuration." and so on below.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -312,33 +312,83 @@ func removeClient(handle *p4rt_client.P4RTClient) {
...
+					Cookie: &p4_v1.ForwardingPipelineConfig_Cookie{
+						Cookie: 159,
+					},
+				},
+			}); err == nil {
+				t.Errorf("SetForwardingPipelineConfig accepted for unset Election ID: %v", err)
+			}
+
+			// Disconnect Primary
 			removeClient(test.handle)
```
Ankur19, line 389: The `removeClient()` needs to be called after both the testCases have run.

We are also testing that multiple clients can connect at the same time when electionID is unset.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -312,33 +312,83 @@ func removeClient(handle *p4rt_client.P4RTClient) {
...
+	clients := []testArgs{
+		{desc: pDesc,
+			handle:     clientConnection(t, dut),
+			deviceID:   deviceId,
+			wantStatus: 5,
+		},
+		{desc: sDesc,
+			handle:     clientConnection(t, dut),
+			deviceID:   deviceId,
+			wantStatus: 5,
```
Ankur19, line 329: Can you use `codes.NotFound` which is internally a uint32 instead of hardcoding to `5`.

This will make the test expectations clearer.

https://pkg.go.dev/google.golang.org/grpc@v1.53.0/codes

Ankur19 (commented):
Few minor changes and we can merge.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -312,33 +312,83 @@ func removeClient(handle *p4rt_client.P4RTClient) {
...
-	test := testArgs{
-		desc:       pDesc,
-		lowID:      inId0,
-		highID:     inId0,
-		handle:     clientConnection(t, dut),
-		deviceID:   deviceId,
-		wantFail:   true,
-		wantStatus: 3,
+	clients := []testArgs{
+		{desc: pDesc,
```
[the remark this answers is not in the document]
rsindhujha: updated new line and indent

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: All comments are taken care in sentence style. Please check

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: Thanks. I moved removeClient() out of the for loop.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: Updated with codes.NotFound instead of hardcoded value.

Ankur19 (approved):
LGTM!

Ankur19:
> This is not related to the changes in this PR but shouldn't we delete the table entry on successful write? According to the [specs](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-write-rpc):
> 
> > INSERT: Inserts the given P4 entity in the entity container. The entity field always specifies the full state of the P4 entity. If the entity already exists, an ALREADY_EXISTS error is returned, and the existing entity remains unchanged.
> 
> This will cause some tests like TestPrimaryReconnect to fail. I can submit a PR after this is merged if needed.

Hi @kjahed Right! Another possible bug here is that we are returning hardcoded status codes in some places instead of just returning the status code received from the response. 

Would appreciate if you fix that as well in your PR.

xw-g:
@LimeHat Do you already receive all the clarification and are you ok to proceed this PR? Thanks.

sachendras:
> 

I will wait until EOD for a response.

LimeHat:
@xw-g yes i think we are clear, thank you for asking.

Also thanks to Ankur & Rob for the comments.

greg-dennis (approved):
Approving for Go quality only

PR #1181, opened 2023-02-23
Te11.2 update

manan-patel:
Addressed

https://github.com/openconfig/featureprofiles/pull/1154/files
https://github.com/openconfig/featureprofiles/pull/1144/files


feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -160,12 +164,29 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) *ondatra.ATETopology {
 	return top
 }
 
+// Configure Network instance
+func configNetworkInstance(t *testing.T, dut *ondatra.DUTDevice, vrfname string, intfname ...string) {
+	d := &oc.Root{}
+	ni := d.GetOrCreateNetworkInstance(vrfname)
+	ni.Type = oc.NetworkInstanceTypes_NETWORK_INSTANCE_TYPE_L3VRF
+	if len(intfname) != 0 {
```
xw-g, line 172: Your func() API claims it multiple interface names, so your func() should either do a `for loop` to support it, or gate on `if len(intfName) == 1` and error out if it's not. I think a `for loop` is better. (otherwise why you want your func API supports the list of interface? :) )

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -262,45 +283,49 @@ func testIPv4BackUpSwitch(ctx context.Context, t *testing.T, args *testArgs) {
```
xw-g, line 283: Can you help also update the comment of this func so that it's more aligned with the README? Thanks.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -262,45 +283,49 @@ func testIPv4BackUpSwitch(ctx context.Context, t *testing.T, args *testArgs) {
 
 	const (
 		// Next hop group adjacency identifier.
-		NHGID = 100
+		NHGID1, NHGID2 = 100, 200
```
xw-g, line 286: Does it need to be exportable?

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -262,45 +283,49 @@ func testIPv4BackUpSwitch(ctx context.Context, t *testing.T, args *testArgs) {
...
+		BackupNHGID = 500
 
-		NH1ID, NH2ID, NH3ID = 1001, 1002, 1003
+		NH1ID, NH2ID, NH3ID, NH4ID = 1001, 1002, 1003, 1004
 	)
-	t.Logf("Program a backup pointing to ATE port-4 via gRIBI")
-	args.client.AddNH(t, NH3ID, atePort4.IPv4, *deviations.DefaultNetworkInstance, fluent.InstalledInRIB)
-	args.client.AddNHG(t, BackupNHGID, map[uint64]uint64{NH3ID: 10}, *deviations.DefaultNetworkInstance, fluent.InstalledInRIB)
+	t.Logf("Program a backup pointing to VRF-B via gRIBI")
+	args.client.AddNH(t, NH3ID, "", *deviations.DefaultNetworkInstance, fluent.InstalledInFIB, &gribi.NHOptions{VrfName: "VRF-B"})
```
xw-g, line 293: vrfB instead of "VRF-B"

xw-g (changes requested):
Thanks for the contribution! Overall it's good, except a few code comment. 

Also, can you help update the OTG tests? Thanks.

xw-g:
I think it's good to add another pair of eyes for review so that I don't miss anything. @LimeHat can you also help do a quick review (focus on aligning with the README)? Thanks.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -160,12 +164,29 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) *ondatra.ATETopology {
...
+	}
+	gnmi.Replace(t, dut, gnmi.OC().NetworkInstance(vrfname).Config(), ni)
+}
+
 // configureDUT configures port1, port2, port3 and port4 on the DUT.
 func configureDUT(t *testing.T, dut *ondatra.DUTDevice) {
 	d := gnmi.OC()
 
 	p1 := dut.Port(t, "port1")
-	gnmi.Replace(t, dut, d.Interface(p1.Name()).Config(), dutPort1.NewOCInterface(p1.Name()))
```
LimeHat, line 168: What is the reason for this change?

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -160,12 +164,29 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) *ondatra.ATETopology {
...
 }
 
+// Configure Network instance
+func configNetworkInstance(t *testing.T, dut *ondatra.DUTDevice, vrfname string, intfname ...string) {
+	d := &oc.Root{}
+	ni := d.GetOrCreateNetworkInstance(vrfname)
+	ni.Type = oc.NetworkInstanceTypes_NETWORK_INSTANCE_TYPE_L3VRF
+	if len(intfname) != 0 {
+		intf := ni.GetOrCreateInterface(intfname[0])
+		intf.Interface = ygot.String(intfname[0])
```
LimeHat, line 174: Missing subinterface ref

internal/gribi/gribi.go
```diff
@@ -218,6 +218,14 @@ func (c *Client) AddNH(t testing.TB, nhIndex uint64, address, instance string, e
 			NH = NH.WithNextHopNetworkInstance(opt.VrfName)
 		}
 		c.fluentC.Modify().AddEntry(t, NH)
+	case "":
```
LimeHat, line 221: Reserve a new keyword here, instead of ""? E.g. "VRF-ONLY"

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -324,20 +349,20 @@ func createFlow(t *testing.T, ate *ondatra.ATEDevice, top *ondatra.ATETopology,
...
-	got := gnmi.Get(t, ate, flowPath.LossPct().State())
-	if drop {
-		if got != 100 {
-			t.Fatalf("Traffic passing for flow %s got %f, want 100 percent loss", flow.Name(), got)
-		}
-	} else {
-		if got > 0 {
-			t.Fatalf("LossPct for flow %s got %f, want 0", flow.Name(), got)
-		}
+	val, _ := gnmi.Watch(t, ate, flowPath.LossPct().State(), 5*time.Minute, func(val *ygnmi.Value[float32]) bool {
```
LimeHat, line 357: Why wait for 5 minutes after the traffic has been stopped?

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -324,20 +349,20 @@ func createFlow(t *testing.T, ate *ondatra.ATEDevice, top *ondatra.ATETopology,
...
+		return val.IsPresent()
+	}).Await(t)
+	lossPct, present := val.Val()
+	if !present {
+		t.Fatalf("Could not read loss percentage for flow %q from ATE.", flow.Name())
+	}
+	if lossPct > 0.5 {
```
LimeHat, line 364: Why 0.5% instead of 0%?

LimeHat:
Currently all traffic is generated with a single src. IP & a single dst. IP

https://github.com/openconfig/featureprofiles/blob/9e4444ebf3fe1b4dcc4d7829ba329329730f1617/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go#L341

Would it make sense to use the full dst subnet (/24) for LB purposes? Two primary next-hops are currently being provisioned with 80/20 weights
https://github.com/openconfig/featureprofiles/blob/9e4444ebf3fe1b4dcc4d7829ba329329730f1617/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go#L299
https://github.com/openconfig/featureprofiles/blob/9e4444ebf3fe1b4dcc4d7829ba329329730f1617/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go#L38-L40

(also const "dstPfx" on L40 is currently unused in ATE test)

mojiiba:
@xw-g  @LimeHat  appreciate the feedback. Manan is off, so I or someone else from our team will address the comments.

xw-g:
> Currently all traffic is generated with a single src. IP & a single dst. IP
> 
> https://github.com/openconfig/featureprofiles/blob/9e4444ebf3fe1b4dcc4d7829ba329329730f1617/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go#L341
> 
> Would it make sense to use the full dst subnet (/24) for LB purposes? Two primary next-hops are currently being provisioned with 80/20 weights
> 
> https://github.com/openconfig/featureprofiles/blob/9e4444ebf3fe1b4dcc4d7829ba329329730f1617/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go#L299
> 
> 
> https://github.com/openconfig/featureprofiles/blob/9e4444ebf3fe1b4dcc4d7829ba329329730f1617/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go#L38-L40
> 
> (also const "dstPfxMask" on L40 is unused in ATE test)

just fyi, @manan-patel I'm ok with not validating the LB (80/20), because it's not the focus of this test (and we have other test to cover it).

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
mojiiba: also, the watch here is for traffic generator and we want to ensure test does not fail in case ATE is slow.

LimeHat:
Yeah i'm not saying we should validate 80/20 specifically, however with just a single flow you are unlikely to validate that the second next hop is active at all

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
LimeHat: Yes but we don't switch to backup with traffic in-flight. All modify operations are performed after the traffic is stopped.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
LimeHat: I'm just surprised by the fact that ATE could take up to 5 minutes to generate packet loss statistics? Is that the case?

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
LimeHat: > Actually I'm fine with removing this line. Is it a concern ? @LimeHat
Essentially the question is "why we are replacing L168 with L189"
I.e. the typical order of operations is 1) configure interface (usually with a replace but I'm not against an update either) 2) assign it to a network instance

Not sure why both changes were needed here?

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
LimeHat: In most test cases Get seems to be sufficient for this purpose, in a couple cases i see a 1-minute timer.. so If there's no strict need for 5 minute i'd prefer to keep tests consistent and use up to 1 minute or Get

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -36,6 +38,8 @@ const (
 	dstPfx        = "203.0.113.0/24"
 	dstPfxMin     = "203.0.113.0"
 	dstPfxMask    = "24"
+	vrf1          = "VRF-A"
```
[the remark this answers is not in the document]
agvarghe-cisco: @ xw-g 
Can you be more elaborate , In READE vrfs mentioned are VRF-A and VRF-B

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
agvarghe-cisco: intfname is used as an optional argument (variadic args)  in function configNetworkInstance , such that same method can used to configure vrf with and without interface config

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
agvarghe-cisco: This is the order of operation done here 
1) Replace on port1 with vrf1 (assign it to a network instance)
2) Update on port1 with v4/v6 ip and other configs . Doing a replace instead of this update will remove previous network instance configs

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
mojiiba: we are trying deal with flakyness (not a normal flow), otherwise you are right and get work for most of the time, but it also fails intermittently. if the ATE is fast, then the await call with return soon and will not wait for 5 minutes (this is the a safe upper bound). Concerning the consistency we are going to convert all ATE call to use await to deal with flakiness.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
LimeHat: @agvarghe-cisco 
Could you please implement this in the following way:
1) configure physical interface (gnmi Replace)
2) configure a subinterface w/o IP configuration
3) assign the subinterface to a vrf
4) update subinterface with IP configuration (gnmi update)
(you can combine multiple steps in a single gnmi call, e.g. 1-2 or 2-3 or 1-3)
An example of such config order can be found in test case RT-3.2

cc @xw-g

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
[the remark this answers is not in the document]
xw-g: I meant that it would be more convenient to name the variable as `vrfA` and `vrfB` so that readers won't have to keep the `vrf1 is VRF-A` and `vrf2 is VRF-B` mapping in their head. :)

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
xw-g: My point is that your func API is not consistent is with what it does, e.g. if it's a list why you simply just do `intfname[0]`?

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
xw-g: sgtm.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
xw-g: no, there is no modify operation here after the initial setup. here is about local repair. that's why I think we start with a relaxed bar (0.5%) is ok.

xw-g:
@mojiiba any ETA for next review? I think overall this PR is good, just some minor things to address. Thanks!

internal/gribi/gribi.go
mojiiba: done

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
mojiiba: decreased to two minutes, since i noticed when there is a traffic loss, the call wait for 5 minutes because of the check for predicate.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
mojiiba: add checks for interface status and remove the loss threshold to 0

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
mojiiba: done

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
LimeHat: Thanks @mojiiba. Looks good!

agvarghe-cisco:
@xw-g ,

Please review .Addressed the changes

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -324,20 +385,20 @@ func createFlow(t *testing.T, ate *ondatra.ATEDevice, top *ondatra.ATETopology,
 }
 
 // validateTrafficFlows verifies that the flow on ATE and check interface counters on DUT
-func validateTrafficFlows(t *testing.T, ate *ondatra.ATEDevice, flow *ondatra.Flow, drop bool, dPort []string) {
+func validateTrafficFlows(t *testing.T, ate *ondatra.ATEDevice, flow *ondatra.Flow, dPort []string) {
```
LimeHat, line 388: Looks like `dPort []string` argument is not used in this function.
Could you please check if we have dst. interface validation logic in here?

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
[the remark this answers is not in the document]
xw-g: @agvarghe-cisco agree?

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
xw-g: agree?

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
xw-g: Actually, @LimeHat and @mojiiba you brought a good point. Reading the code again, it stops the traffic before the flipping the port status. 

What we wanted is that
1. Start traffic.
2. flip the port status.
3. Validate the traffic moved between ports. 

@mojiiba can your team help update the code? Thanks.

xw-g:
> @xw-g ,
> 
> Please review .Addressed the changes

Thanks @agvarghe-cisco I see there are still several comments not addressed. Can you help address them? Thanks.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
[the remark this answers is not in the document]
agvarghe-cisco: Changed to vrfA and vrfB

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
agvarghe-cisco: updated

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
agvarghe-cisco: changed

xw-g:
Also, note that https://github.com/openconfig/featureprofiles/pull/1154 is merged.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
LimeHat: Looks good now. Thanks

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
LimeHat: Looks good now. thanks.

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -27,15 +27,20 @@ import (
...
 	"github.com/openconfig/ondatra/gnmi"
+	"github.com/openconfig/ondatra/gnmi/oc"
 	"github.com/openconfig/ygnmi/ygnmi"
+	"github.com/openconfig/ygot/ygot"
 )
 
 const (
 	ipv4PrefixLen = 30
 	ipv6PrefixLen = 126
 	dstPfx        = "203.0.113.0/24"
```
xw-g, line 38: This should be /32.

LimeHat:
Thanks for the changes and updates guys, looks good

Just two more small requests from me:
1) please remove this line (assignment of port1 to DEFAULT vrf under deviation flag; since this port is now assigned to vrf1)
https://github.com/openconfig/featureprofiles/blob/8a61c08bf7dbbc09d7155017f3d9c0f600fd8ede/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go#L234
2) please add the following deviation code to `configureDUT()` function (any place inside the function is fine, e.g. at the very end is ok):
https://github.com/openconfig/featureprofiles/blob/8a61c08bf7dbbc09d7155017f3d9c0f600fd8ede/feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go#L205

```
	if *deviations.ExplicitGRIBIUnderNetworkInstance {
		fptest.EnableGRIBIUnderNetworkInstance(t, dut, *deviations.DefaultNetworkInstance)
		fptest.EnableGRIBIUnderNetworkInstance(t, dut, vrf1)
		fptest.EnableGRIBIUnderNetworkInstance(t, dut, vrf2)
	}
```

Additionally, I think there was a comment by @xw-g that went unnoticed regarding our earlier conversation about traffic and packet loss:
https://github.com/openconfig/featureprofiles/pull/1181#discussion_r1150873178

manan-patel:
@LimeHat There is a comment by @xw-g requesting to change the prefix to /32, indeed we have the following options
1: Have /32 prefix and advertise only 1 prefix via ATE
2: Have /32 mask and enhance AddIPv4 call to handle scale config of prefixes
3: Have /24 mask

LimeHat:
@manan-patel i'm ok with the current prefixes (single /32), that's fine.
I'm refering to the other comment (linked above):
```
Actually, @LimeHat and @mojiiba you brought a good point. Reading the code again, it stops the traffic before the flipping the port status.

What we wanted is that

Start traffic.
flip the port status.
Validate the traffic moved between ports.
```

xw-g:
Thanks @manan-patel for addressing all the comment. Please ping me when it's ready for another review. Thanks!

manan-patel:
@xw-g  TC is ready for another review

guoshiuan:
b/271139038 has the internal test results.

guoshiuan (dismissed):

xw-g (approved):

internal/gribi/gribi.go
```diff
@@ -218,6 +218,14 @@ func (c *Client) AddNH(t testing.TB, nhIndex uint64, address, instance string, e
 			NH = NH.WithNextHopNetworkInstance(opt.VrfName)
 		}
 		c.fluentC.Modify().AddEntry(t, NH)
+	case "VRFOnly":
+		NH := fluent.NextHopEntry().
```
xw-g, line 222: lower case, because you are not planning to export this variable?

manan-patel: Addressed the comment

feature/gribi/ate_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -245,113 +299,150 @@ func TestBackup(t *testing.T) {
...
 //   - Connect ATE port-4 to DUT port-4.
-//   - Connect a gRIBI client to the DUT and inject an IPv4Entry for 203.0.113.0/24 pointing to a NextHopGroup containing:
+//   - Create a L3 routing instance (vrfA), and assign DUT port-1 to vrfA.
+//   - Create a L3 routing instance (vrfB) that includes no interface.
+//   - Connect a gRIBI client to the DUT and inject an IPv4Entry for 203.0.113.1/32 pointing to a NextHopGroup containing:
 //   - Two primary next-hops:
 //   - 2: to ATE port-2
 //   - 3: to ATE port-3.
 //   - A backup NHG containing a single next-hop:
 //   - 4: to ATE port-4.
```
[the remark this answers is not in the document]
manan-patel: Updated the statements to reflect the same steps as indicated in the readme

guoshiuan (approved):

PR #1185, opened 2023-02-24
For interface configuration, Arista prefers config Vrf first then the…

gforeal:
… IP address

### DIFF
--- a/feature/gribi/ate_tests/route_removal_non_default_vrf_test/route_removal_non_default_vrf_test.go
+++ b/feature/gribi/ate_tests/route_removal_non_default_vrf_test/route_removal_non_default_vrf_test.go
@@ -97,12 +97,27 @@ func TestRouteRemovalNonDefaultVRFFlush(t *testing.T) {
 	ctx := context.Background()
 
 	dut := ondatra.DUT(t, "dut")
+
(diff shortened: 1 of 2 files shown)


gforeal:
the fix constrains the impact to Arista only while all other platforms still with the config sequence & workflow as-is.
have tested works!

feature/gribi/ate_tests/route_removal_non_default_vrf_test/route_removal_non_default_vrf_test.go
```diff
@@ -97,12 +97,27 @@ func TestRouteRemovalNonDefaultVRFFlush(t *testing.T) {
 	ctx := context.Background()
 
 	dut := ondatra.DUT(t, "dut")
+
+	// For interface configuration, Arista prefers config Vrf first then the IP address
+	switch dut.Vendor() {
+	case ondatra.ARISTA:
+		configureNetworkInstance(t, dut)
```
robshakir, line 104: Please can we explain why this is order dependent only for a single vendor? If it's to do with references being validated then shouldn't it be generic? If it's not, then this should likely be behind a deviation (I generally don't expect that we have `switch dut.Vendor()` without a deviation unless it's for a naming specific thing).

rszarecki (approved):

PR #1188, opened 2023-02-24
Power range update as per standards

jyotigup-cisco:
PR contains
1. Modification of mini. power range as per standards and observed on physical platform
_Fiber optic sources may vary from -20dBm to +20dBm and receiver power may go as low as -40dBm._

2. Correction of "if check"


prinikasn (commented):
/gcbrun

prinikasn (approved):

rszarecki (approved):

PR #1191, opened 2023-02-24
TE 3.3 hierarchical_weight_resolution_test - Increasing Traffic Flows.

cprabha:
Hi,

This pull request is created to increase traffic flows from 256 to 1022.

Thanks,
Prabha


cprabha:
> What is the reasoning for using 1022 instead of 1024?
Thought of excluding prefix and broadcast address, but we can send traffic to them as well.
I have updated script to use 1024 now.

greg-dennis (approved):
approving Go quality only

sachendras (approved):

PR #1192, opened 2023-02-24
Update README for P4RT-2.1

Ankur19:


rsindhujha:
Hi @Ankur19 ,

Per the specifications, if we do not allow SetForwardingPipelineConfig for zero election ID, then both Read Write RPC are expected to fail. Can you please clarify?


8.3. Zero as Reserved Value

p4runtime.proto uses proto3 syntax, and so it does not allow not specifying a scalar data type, such as a uint32. Therefore, we usually reserve value 0 for those fields to mean unset. In particular, 0 is not a valid P4 object ID and it is an error to specify 0 for any P4 object ID in a non-read request towards the switch, such as in a WriteRequest or a SetForwardingPipelineConfigRequest. 

13. Read RPC

If the Read is attempted before ForwardingPipelineConfig has been set, the server must return a FAILED_PRECONDITION error.

Ankur19:
> ForwardingPipelineConfig

Hi @rsindhujha Sorry for the confusion here. 

By `read` I was implying any Get request to the server. Even though we can not successfully make a `ReadRequest` to the server, we should be able to make a `GetForwardingPipelineConfigRequest` to the server which is also considered a data read.

I will clarify the `README` on this.

Ankur19:
@rsindhujha @LimeHat @kjahed Are you good with the updates here?

LimeHat:
Looks good to me!

rsindhujha:
> @rsindhujha @LimeHat @kjahed Are you good with the updates here?

@Ankur19 , Looks good. Thanks.

kjahed:
LGTM!

Issue #1002, opened 2023-01-23
gNMI-1.10 - Telemetry Basic Check: Use of hardcoded component names and parent<->child relationships


sgirishj:
https://github.com/openconfig/featureprofiles/blob/7061d7918e7a3131492083a4195b94f4a9e6ba90/feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go#L325

The test 'TestComponentParent' in the above references script appears to be making assumptions regarding the below:
- that components in the device's inventory are named a certain way (e.g. Fabric, Supervisor, etc.)
- that the parent<->child relationships between components/sub-components in the device adhere to what is defined in the test (e.g. 'Chassis' is the parent of 'Fabric' and 'Fabric' is the parent of 'FabricChip')

The OC model around system component inventory defines/enforces neither of the above. Furthermore, names and parent<->child relationships are typically vendor/platform/implementation specific.

@dplore , @greg-dennis - request your inputs on this.

Thanks!

dplore:
This code should be refactored to use the OpenConfig component identities instead of component name strings.

Regarding the parent relationships, this test is targeted for modular form factor devices where these components are expected to be identified by the DUT.    The test should be refactored to use the yang identities instead of the name strings which are vendor specific.  ie: the hierarchy should be defined like: `CHASSIS` <- `FABRIC` <- `INTEGRATED_CIRCUIT`.  

Agreed that the openconfig-platform model does not require this relationship and I acknowledge that all devices may not have these components.  We should clarify this test is intended for modular form factor devices either through documentation or perhaps a parameter such as `modular_form_factor`) which when set requires the parent relationships to exist.  It is then up to the caller to assert if the DUT is required to behave as a modular_form_factor device or not.  When testing a DUT such as a single INTEGRATED_CIRCUIT device which is not expected to have a `FABRIC` component, the invoker would not set this parameter and the test would skip validating the relevant component identities and leafs.

What do you think?

earies:
+1 to not using component name strings anywhere for assumptions - this should be a featureprofile design guideline to not happen again as we've seen this a few times already.  Conclusion - it is up to each implementation to formulate these keys (as are various other mostly r/o keys elsewhere in the model set)

[...]

PR #1195, opened 2023-02-25
gNMI-1.10: Replacing component parent<->child relationship check with chassis association check

sgirishj:
Component hierarchies could vary widely across vendor implementations and platforms. Also, OpenConfig does not require/enforce specific parent<->child relationships. The requirement around this test mainly is to be able to map various components in the device back to the top level chassis component (without placing any constraints on the number and type of intermediate nodes that may be present in the tree).

Modifying the test to fit the above by:
- Removing the check for the type of the parent component to be of a specific statically defined value
- Replacing it with: start from a given component, traverse up the component's hierarchy tree one node a time, looking for the chassis component; the test would pass if the chassis component is found in the tree; it would fail if the chassis component is not to be found

Also:
- Added checks to detect circular references between parent<->child nodes that could lead to the test getting stuck in an infinite loop
- Removed fabricChip from TestComponentParent test as per the discussion in https://github.com/openconfig/featureprofiles/issues/1002


feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -395,40 +395,36 @@ func TestComponentParent(t *testing.T) {
...
 	for _, tc := range cases {
 		t.Run(tc.desc, func(t *testing.T) {
-			t.Logf("Found card list for Component Type %v : %v", tc.componentType, cardList[tc.desc])
+			t.Logf("Found Component list for Type %v : %v", tc.componentType, cardList[tc.desc])
 			if len(cardList[tc.desc]) == 0 {
-				t.Fatalf("Get Card list for %q: got 0, want > 0", dut.Model())
+				t.Fatalf("Get Component list for %q: got 0, want > 0", dut.Model())
 			}
 			// Validate parent component
 			for _, card := range cardList[tc.desc] {
```
greg-dennis, line 403: Please rename "card" and "cardList" to "comp" and "compList" to align with your better logging here that these are generally components, not cards

greg-dennis (changes requested):

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -395,40 +395,36 @@ func TestComponentParent(t *testing.T) {
...
-				t.Logf("Validate card %s", card)
+				t.Logf("Validate component %s", card)
 				cardName := card
```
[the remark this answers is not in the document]
sgirishj: Local variable 'cardName' is getting set to 'parentName' later within the 'for' loop. Wont be able to use the loop index variable 'card' in place of the local variable 'cardName'. Will rename to card->comp though. Is that ok?
Line #422: cardName = parentName

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -395,40 +395,36 @@ func TestComponentParent(t *testing.T) {
...
-							}
-						} else {
-							if got == tc.parent {
-								t.Logf("Got expected parent for card %s", card)
-								break
-							}
+						if got == chassisType {
+							foundChassis = true
+							t.Logf("Found chassis component in the hierarchy tree of component %s", card)
+							break
```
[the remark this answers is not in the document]
sgirishj: When we get out of the inner 'for' loop, we need to know if the loop was exited due to the 'break' ('pass' scenario) or because we reached the top of the component tree and didn't find what we were looking for ('fail' scenario), so that we could log the appropriate 'fail' message. Had used the foundChassis flag for this purpose.

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -395,40 +395,35 @@ func TestComponentParent(t *testing.T) {
 
 	for _, tc := range cases {
 		t.Run(tc.desc, func(t *testing.T) {
-			t.Logf("Found card list for Component Type %v : %v", tc.componentType, cardList[tc.desc])
-			if len(cardList[tc.desc]) == 0 {
-				t.Fatalf("Get Card list for %q: got 0, want > 0", dut.Model())
+			t.Logf("Found Component list for Type %v : %v", tc.componentType, compList[tc.desc])
```
greg-dennis, line 398: lowercase "component"

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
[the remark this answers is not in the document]
greg-dennis: I misread and forgot this validation is happening for every component in the list, not just one. Just leave as is.

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -395,40 +395,35 @@ func TestComponentParent(t *testing.T) {
...
 						}
-
 					}
-					parentName := gnmi.Get(t, dut, gnmi.OC().Component(val).Name().State())
-					cardName = parentName
+					// Not reached chassis yet; go one level up
+					compName = gnmi.Get(t, dut, gnmi.OC().Component(val).Name().State())
+				}
+				if !foundChassis {
+					t.Fatalf("Chassis component NOT found in the hierarchy tree of component %s", comp)
```
greg-dennis, line 424: Change to t.Errorf, so you get failures for multiple components in a single test run

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -395,40 +395,35 @@ func TestComponentParent(t *testing.T) {
...
+				t.Logf("Validate component %s", comp)
+				compName := comp
+				foundChassis := false
 				for {
-					parent := gnmi.Lookup(t, dut, gnmi.OC().Component(cardName).Parent().State())
+					parent := gnmi.Lookup(t, dut, gnmi.OC().Component(compName).Parent().State())
 					val, present := parent.Val()
 					if !present {
-						t.Fatalf("Parent not present for %q: got %v, want true", card, parent.IsPresent())
+						t.Fatalf("Parent not present for %q: got %v, want true", comp, parent.IsPresent())
```
greg-dennis, line 411: I think you can just break here instead. In fact, I think you _need_ to break to ever realistically get the "not found" message below.

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -395,40 +395,35 @@ func TestComponentParent(t *testing.T) {
...
 					} else {
```
greg-dennis, line 412: no need to put this logic in an else-lock, since the !present case always exits the loop

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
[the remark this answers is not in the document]
greg-dennis: It doesn't matter that cardName is being set to parentName later on. You can still set the loop variable to a new value. It would be equivalent to what you have now. If you don't like the style of doing that, please rename "cardName" to "curr"

greg-dennis (changes requested):

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
sgirishj: Not sure I understood right. Isn't the t.Fatalf already achieving that (breaking the loop + failing and aborting the current test run when the specified condition is met)?

greg-dennis: I think the Fatal here means:

1. You will _never_ see your "Chassis component NOT found in the hierarchy tree of component" error message below. If the chassis component is not in the tree, it will continue looking through the parents until it gets no parent and then fail here instead.
2. Because it is t.Fatal instead of t.Error, none of the other components will be checked. 

Let's fix both of these. My suggestion is one way to do. Another way that requires even less code is to put your error message below here instead ("t.Errorf("Chassis component NOT found in the hierarchy tree of component %s", comp)"), followed by a break. Then just remove the "if !foundChassis {" below.

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -395,40 +395,34 @@ func TestComponentParent(t *testing.T) {
...
+					got := gnmi.Get(t, dut, gnmi.OC().Component(val).Type().State())
+					if got == chassisType {
+						foundChassis = true
+						t.Logf("Found chassis component in the hierarchy tree of component %s", comp)
+						break
 					}
-					parentName := gnmi.Get(t, dut, gnmi.OC().Component(val).Name().State())
-					cardName = parentName
+					// Not reached chassis yet; go one level up
+					curr = gnmi.Get(t, dut, gnmi.OC().Component(val).Name().State())
```
greg-dennis, line 420: is this equivalent to "curr = parent"?

greg-dennis (changes requested):

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
sgirishj: Practically, yes. I had tried it earlier, so we could avoid performing the Get() operation here and it had worked fine. 
Decided later on to leave the original code as is since it helps to validate:
- that the state/name leaf of the parent node is implemented and it is
- implemented correctly so as to be in sync with the state/parent leaf of the node's child

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -395,40 +395,35 @@ func TestComponentParent(t *testing.T) {
...
-				cardName := card
-				for {
-					parent := gnmi.Lookup(t, dut, gnmi.OC().Component(cardName).Parent().State())
+			// Validate parent component.
+			for _, comp := range compList[tc.desc] {
+				t.Logf("Validate component %s", comp)
+				visited := make(map[string]bool)
+				for curr := comp; ; {
+					if visited[curr] {
+						t.Errorf("Component %s already visited; likely loop detected in the hierarchy.", curr)
```
greg-dennis, line 408: I would just drop the word "likely"

greg-dennis (dismissed):
approving for Go quality only

greg-dennis (approved):

sachendras (approved):

PR #1204, opened 2023-03-02
gNMI-1.4: Telemetry: Inventory - Removed fabricChip from the list of components; added vendor specific deviations

stthummuri:
1. Removed fabricChip from the list of components as per guidance from the issue - https://github.com/openconfig/featureprofiles/issues/1002; 

2. Added three vendor specific deviations


feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -440,119 +412,97 @@ func ValidateComponentState(t *testing.T, dut *ondatra.DUTDevice, cards []string
...
-
 		if p.parent != "" {
-			parent := gnmi.Get(t, dut, component.Parent().State())
-			t.Logf("Hardware card %s parent: %s", card, parent)
-			if parent != p.parent {
-				t.Errorf("component.Parent().Get(t) for %q): got %v, want %v", card, parent, p.parent)
+			parent := gnmi.Lookup(t, dut, gnmi.OC().Component(card).Parent().State())
+			val, present := parent.Val()
+			if !present {
+				t.Errorf("Chassis component NOT found in the hierarchy tree of component %s", card)
```
greg-dennis, line 496: can use %q for the card, as you do in the error below

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -440,119 +412,97 @@ func ValidateComponentState(t *testing.T, dut *ondatra.DUTDevice, cards []string
...
+				break
+			}
+			got := gnmi.Get(t, dut, gnmi.OC().Component(val).Type().State())
+			if got == chassisType {
+				t.Errorf("component.Parent().Get(t) for %q): got %v, want %v", card, got, p.parent)
```
greg-dennis, line 501: delete extraneous closing parent ")" after %q

greg-dennis (approved):

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
stthummuri: > can use %q for the card, as you do in the error below

Done

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
stthummuri: > delete extraneous closing parent ")" after %q

Done

greg-dennis (approved):
approving for Go quality only

LimeHat:
I don't think you can completely eliminate regex-based checks in this scenario.
E.g.
1) You need to differentiate between fabricChip & switchChip components
2) You can't make an assumption that every SENSOR component is a temperature sensor

stthummuri:
> 

fabricChipType and switchChipType are now being retrieved from oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_INTEGRATED_CIRCUIT and temperature sensor with oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_SENSOR which was how it was earlier with regexp. Now it is being retrieved through oc

LimeHat:
> fabricChipType and switchChipType are now being retrieved from oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_INTEGRATED_CIRCUIT

How do you distinguish between the two? They have different validations in the current test. How do you even check that both are present? They are the same oc type.

> temperature sensor with oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_SENSOR which was how it was earlier with regexp

It wasn't the same with regex because there was a match on the component name. Component type sensor may mean different sensor types (e.g. voltage sensor) in addition to temp. sensors.

sgirishj:
> > fabricChipType and switchChipType are now being retrieved from oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_INTEGRATED_CIRCUIT
> 
> How do you distinguish between the two? They have different validations in the current test. How do you even check that both are present? They are the same oc type.
> 

Yes, thats a valid point. Actually there is no OC type defined for Fabric Chip that could differentiate it from Switch Chip. Following along the guidance given in https://github.com/openconfig/featureprofiles/issues/1002#issuecomment-1451316954, Fabric Chip should be dropped from the list of components to be verified in this test.

stthummuri:
> I don't think you can completely eliminate regex-based checks in this scenario. E.g.
> 
> 1. You need to differentiate between fabricChip & switchChip components
> 2. You can't make an assumption that every SENSOR component is a temperature sensor
Done. Added check for temperature based sensors

LimeHat:
Thanks @stthummuri @sgirishj 

I see in the latest you're also introducing per-vendor check lists. While I'm not, strictly speaking, against it, I would propose to

1. Move those changes into a separate PR,
2. Discuss this approach vs using a deviation (like it was done in #1248 for a similar test and CPU mfgname state path)

LimeHat:
Also I just noticed there's a new PR #1277 that has similar objectives and also introduces a few deviations in this area

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -261,59 +240,66 @@ func TestHardwarecards(t *testing.T) {
...
-		parent:                "",
-		pType:                 componentType["switchchip"],
-	}
-
-	r, err := regexp.Compile(regexpPattern)
-	if err != nil {
-		t.Fatalf("Cannot compile regular expression: %v", err)
+	var cardFields properties
+	switch dut.Vendor() {
+	case ondatra.JUNIPER:
```
greg-dennis, line 265: tests should not be vendor-specific without a documented deviation

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -133,11 +130,11 @@ func TestHardwarecards(t *testing.T) {
 			rrValidation:          false,
 			operStatus:            activeStatus,
 			parent:                "",
-			pType:                 componentType["chassis"],
+			pType:                 chassisType.String(),
```
greg-dennis, line 133: Could you change the parent and pType fields to be of type oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT?

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -509,50 +483,48 @@ func ValidateComponentState(t *testing.T, dut *ondatra.DUTDevice, cards []string
...
 		if p.parent != "" {
-			parent := gnmi.Get(t, dut, component.Parent().State())
-			t.Logf("Hardware card %s parent: %s", card, parent)
-			if parent != p.parent {
-				t.Errorf("component.Parent().Get(t) for %q): got %v, want %v", card, parent, p.parent)
+			parent := gnmi.Lookup(t, dut, gnmi.OC().Component(card).Parent().State())
+			val, present := parent.Val()
+			if !present {
+				t.Errorf("Chassis component NOT found in the hierarchy tree of component %q", card)
+				break
```
greg-dennis, line 519: You really want to break out of the loop and not check any of the other cards? And do you not want to check pType in this case? A "continue" here in place of the break would check the remaining cards. Or just rework the if-statement so that the subsequent Get is skipped -- then you would still check pType and the remaining cards.

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -350,86 +336,77 @@ func TestSwitchChip(t *testing.T) {
...
-			t.Logf("component.Id().Lookup(t) for %q: got false, want true", sensor)
-		} else {
-			t.Logf("TempSensor %s Id: %s", sensor, gnmi.Get(t, dut, component.Id().State()))
-		}
+		r := regexp.MustCompile("Temp|TEMP")
+		if r.Match([]byte(sensor)) {
+			t.Logf("Validate card %s", sensor)
+			component := gnmi.OC().Component(sensor)
+
+			if !gnmi.Lookup(t, dut, component.Id().State()).IsPresent() {
```
greg-dennis, line 351: I know this predated this PR, but all these if-statements are doing an unnecessary Get. in the else clause. Better to just do:

if val, ok := gnmi.Lookup(t, dut, component.Id().State()).Val(); ok {
  t.Logf("TempSensor %s Id: %s", sensor, val)
else {
  // Just log the result using Logf instead of Errorf.
  t.Logf("component.Id().Lookup(t) for %q: got false, want true", sensor)
}

To avoid the extra Get call

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -112,14 +108,15 @@ func TestMain(m *testing.M) {
 
 func TestHardwarecards(t *testing.T) {
 	dut := ondatra.DUT(t, "dut")
-
 	cases := []struct {
 		desc          string
 		regexpPattern string
 		cardFields    properties
+		componentType oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT
+		parent        oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT
```
greg-dennis, line 116: where is the parent field used?

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -112,14 +108,15 @@ func TestMain(m *testing.M) {
...
 	}{{
 		desc:          "Chassis",
-		regexpPattern: "^Chassis",
+		componentType: chassisType,
```
greg-dennis, line 119: it looks like every componentType used is the same as the card pType. Drop the componentType field, too?

greg-dennis (changes requested):

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
stthummuri: Done

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
stthummuri: Removed parent

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
stthummuri: Done

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
stthummuri: Done

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
stthummuri: Done. Added new deviations

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -20,32 +20,29 @@ import (
...
-	"fabricchip":  "INTEGRATED_CIRCUIT",
-	"linecard":    "LINECARD",
-	"fan":         "FAN",
-	"powersupply": "POWER_SUPPLY",
-	"supervisor":  "CONTROLLER_CARD",
-	"switchchip":  "INTEGRATED_CIRCUIT",
-	"transceiver": "TRANSCEIVER",
-	"tempsensor":  "SENSOR",
-}
+const (
```
greg-dennis, line 33: Let's drop all these constants and just inline the (longer oc constants). They create a double-maintenance burden and can add confusion. As one example, fabricChipType and switchChipType are both set to the same constant, which obscures the intention of the test.

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
greg-dennis: pType still needs to be changed to be of type oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
greg-dennis: I still see the componentType field in some tests

greg-dennis (changes requested):

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
stthummuri: Changed pType's type. Could not change parent's type to oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT since the field is being set to empty for certain components in order to skip the check for that leaf; and if it is of type oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT, I am unable to set it to an empty value.

greg-dennis: Use oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT for the parent type and use oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_UNSET as the empty value. (That's the zero value it will default to if you just don't specify the field).

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
stthummuri: Done

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
stthummuri: Removed all

greg-dennis (approved):
Approving Go quality only

stthummuri:
> Actually, there's a static analysis failure due to the order of imports. Please fix and then everything's all set.

Done. Thank you!

greg-dennis (approved):
Approving for Go quality only

stthummuri:
> Thanks @stthummuri @sgirishj
> 
> I see in the latest you're also introducing per-vendor check lists. While I'm not, strictly speaking, against it, I would propose to
> 
> 1. Move those changes into a separate PR,
> 2. Discuss this approach vs using a deviation (like it was done in [gnmi 1.10: add deviations MissingInterfacePhysicalChannel, MissingInterfaceHardwarePort, and MissingCPUMfgName #1248](https://github.com/openconfig/featureprofiles/pull/1248) for a similar test and CPU mfgname state path)

Done

stthummuri:
> I don't think you can completely eliminate regex-based checks in this scenario. E.g.
> 
> 1. You need to differentiate between fabricChip & switchChip components
> 2. You can't make an assumption that every SENSOR component is a temperature sensor

1. fabricChip component has been removed.
2. For temperature sensors, a regexp check has been added

sachendras (dismissed):

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -186,12 +146,11 @@ func TestHardwarecards(t *testing.T) {
 			fwVerValidation:       false,
 			rrValidation:          false,
 			operStatus:            activeStatus,
-			parent:                "",
-			pType:                 componentType["fan"],
+			parent:                oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_CHASSIS,
```
mojiiba, line 149: these are vendor dependent values and rather than this we should check for eventual parent that should be chasis. We did the same in telemetry basic check

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -204,12 +163,11 @@ func TestHardwarecards(t *testing.T) {
 			fwVerValidation:       false,
 			rrValidation:          false,
 			operStatus:            activeStatus,
-			parent:                componentParent["linecard"],
-			pType:                 componentType["linecard"],
+			parent:                oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_CHASSIS,
```
mojiiba, line 166: same comments stand for all componnets

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -260,32 +216,27 @@ func TestHardwarecards(t *testing.T) {
...
 	}}
 
 	for _, tc := range cases {
 		t.Run(tc.desc, func(t *testing.T) {
-			r, err := regexp.Compile(tc.regexpPattern)
-			if err != nil {
-				t.Fatalf("Cannot compile regular expression: %v", err)
-			}
-			cards := findMatchedComponents(t, dut, r)
+			cards := components.FindComponentsByType(t, dut, tc.cardFields.pType)
```
mojiiba, line 226: running this call for every component type is expensive and will make the test so slow. Please cach the values in the first run and use it for the subsequent calls. U may copy the code from here https://github.com/openconfig/featureprofiles/pull/1277

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -298,149 +249,137 @@ func TestSwitchChip(t *testing.T) {
...
+	sensors := components.FindComponentsByType(t, dut, oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_SENSOR)
+	t.Logf("Found Sensor list: %v", sensors)
 	if len(sensors) == 0 {
 		t.Fatalf("Get TempSensor list for %q: got 0, want > 0", dut.Model())
 	}
 
 	for _, sensor := range sensors {
-		t.Logf("Validate card %s", sensor)
-		component := gnmi.OC().Component(sensor)
+		r := regexp.MustCompile("Temp|TEMP")
```
mojiiba, line 311: this is vendor dependent check and i have already addressed in a way all vendors can use please check https://github.com/openconfig/featureprofiles/pull/1277

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -450,37 +389,37 @@ func ValidateComponentState(t *testing.T, dut *ondatra.DUTDevice, cards []string
...
 				t.Errorf("component.Name().Get(t) for %q): got empty string, want non-empty string", card)
 			}
 		}
-
 		if p.partNoValidation {
-			partNo := gnmi.Get(t, dut, component.PartNo().State())
-			t.Logf("Hardware card %s PartNo: %s", card, partNo)
-			if partNo == "" {
-				t.Errorf("component.PartNo().Get(t) for %q): got empty string, want non-empty string", card)
+			if !(*deviations.SwitchChipPartNoUnsupported && p.pType == oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_INTEGRATED_CIRCUIT) {
```
mojiiba, line 408: i have discussed this with google team and we may drop this check without a deviation

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -490,10 +429,22 @@ func ValidateComponentState(t *testing.T, dut *ondatra.DUTDevice, cards []string
 		}
 
 		if p.mfgDateValidation {
-			mfgDate := gnmi.Get(t, dut, component.MfgDate().State())
-			t.Logf("Hardware card %s mfgDate: %s", card, mfgDate)
-			if mfgDate == "" {
-				t.Errorf("component.MfgName().Get(t) for %q): got empty string, want non-empty string", card)
+			if !(*deviations.SupervisorMfgDateUnsupported && p.pType == oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_CONTROLLER_CARD) {
```
mojiiba, line 432: i have discussed this with google team and MfgDate is optional, so we can drop the check rather than having deviation

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -490,10 +429,22 @@ func ValidateComponentState(t *testing.T, dut *ondatra.DUTDevice, cards []string
...
+				mfgDate := gnmi.Get(t, dut, component.MfgDate().State())
+				t.Logf("Hardware card %s mfgDate: %s", card, mfgDate)
+				if mfgDate == "" {
+					t.Errorf("component.MfgName().Get(t) for %q): got empty string, want non-empty string", card)
+				}
+			}
+		}
+
+		if p.fwVerValidation {
```
mojiiba, line 441: same as above

internal/deviations/deviations.go
```diff
@@ -208,6 +208,16 @@ var (
 	NetworkInstanceTableDeletionRequired = flag.Bool("deviation_network_instance_table_deletion_required", false,
 		"Set to true for device requiring explicit deletion of network-instance table, default is false")
 
+	SwitchChipIDUnsupported = flag.Bool("deviation_switch_chip_id_unsupported", false, "Device does not support id leaf for SwitchChip components. Set this flag to skip checking the leaf.")
```
prinikasn, line 211: Please use this new format for introducing new deviations: https://github.com/openconfig/featureprofiles/pull/1358.

We will be adding a check for this soon, all new flags should not be exported (define them in lowercase) and accessed from tests through a public accessors (ref in PR 1358)

prinikasn (changes requested):
Please use this new format for introducing new deviations: https://github.com/openconfig/featureprofiles/pull/1358.

We will be adding a check for this soon, all new flags should not be exported (define them in lowercase) and accessed from tests through a public accessors (ref in PR 1358)

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
sachendras: I would rather prefer this to be a vendor specific switch case inside the script rather than a runtime argument which may open up possibilities for test failures due to user error.

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
sachendras: @vosipchu is there a Google bug capturing this conversation?

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
sachendras: @vosipchu is there a Google bug capturing this conversation?

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
sachendras: @vosipchu is there a Google bug capturing this conversation?

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
sachendras: LGTM for removal

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
sachendras: LGTM for removal

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
sachendras: LGTM for removal

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
sachendras: All for efficiency, thanks @mojiiba

sgirishj:
Closed by mistake. Reopening.

rszarecki (dismissed):

internal/deviations/deviations.go
stthummuri: Done

prinikasn (commented):
LGTM from deviations code perspective

thesrinath:
> Please use this new format for introducing new deviations: #1358.
> 
> We will be adding a check for this soon, all new flags should not be exported (define them in lowercase) and accessed from tests through a public accessors (ref in PR 1358)

Prinika has given LGTM for deviations. Hence closing.

thesrinath (dismissed):

thesrinath (approved):

PR #1207, opened 2023-03-02
P4RT-5.1 - wait for packets

kjahed:
Currently, the test iterates a fixed number of times trying to fetch at least one packet from the p4rt client to pass. The test fails if the packets are delayed. This PR changes the test to align with P4RT-7.1:
- Query ATE for tx packets count
- Try to receive the expected packets for a duration


Ankur19:
Hi @kjahed This is not required now as `StreamChannelGetPackets()` can be used which was merged yesterday: https://github.com/cisco-open/go-p4/pull/15

I am going to update the tests accordingly.

kjahed:
Very timely. I'll close this one then.

PR #1213, opened 2023-03-04
update gNMI-1.13

LimeHat:
This PR introduces a few changes to gnmi-1.13 Telemetry: Optics Power and Bias Current test

1) we refactor `findTransceiverName()` function that returns transceiver names to use switch/case statements instead of the previous if-then logic
2) we correct `findTransceiverName()` function logic for Nokia (to return capitalized "Ethernet" name instead of lowercase "ethernet")
3) check on L145 now exits with t.Fatal instead of t.Skip.
This test ("TestOpticsPowerUpdate") lookups a single transceiver based on the static name received from `findTransceiverName()` function, hence it was possible to get a "PASS" on the test even when an incorrect transceiver name was returned by the function and all test logic was skipped, e.g.:
```
    optics_power_and_bias_current_test.go:146: component.MfgName().Lookup(t).IsPresent() for "ethernet-1/8-transceiver" is false. skip it

--- PASS: TestOpticsPowerUpdate (5.26s)
    --- SKIP: TestOpticsPowerUpdate/Check_initial_input_and_output_optics_powers_are_OK (3.25s)
    --- SKIP: TestOpticsPowerUpdate/Check_output_optics_power_is_very_small_after_interface_is_disabled (0.10s)
    --- SKIP: TestOpticsPowerUpdate/Check_output_optics_power_is_normal_after_interface_is_re-enabled (1.90s)
PASS

```
4. a bit of cleanup: removed unused constants and a comment about topology & gnmic, fixed a typo in a comment.

---
<sup>


feature/platform/tests/optics_power_and_bias_current_test/optics_power_and_bias_current_test.go
```diff
@@ -30,29 +30,16 @@ import (
...
 )
 
 func TestMain(m *testing.M) {
 	fptest.RunTests(m)
 }
 
-// Topology:
-//   ate:port1 <--> port1:dut:port2 <--> ate:port2
-//
-//  Sample CLI command to get telemetry using gmic:
```
xw-g, line 49: Any reason removing this comment? Seems helpful. 

nit: yes you can update the link to: https://github.com/openconfig/gnmic

xw-g (commented):
logic looks good, just some code comment.

feature/platform/tests/optics_power_and_bias_current_test/optics_power_and_bias_current_test.go
LimeHat: I think it may be more relevant for README? I.e. that's just a description on how to do gnmi call manually, and there's nothing specific about it?
And topology part above is just incorrect i think, we don't use port2 for example.

feature/platform/tests/optics_power_and_bias_current_test/optics_power_and_bias_current_test.go
```diff
@@ -179,28 +166,22 @@ func TestOpticsPowerUpdate(t *testing.T) {
...
+		suffix := " transceiver"
 		interfaceSplit := strings.Split(interfaceName, "/")
 		interfaceSplitres := interfaceSplit[:len(interfaceSplit)-1]
-		transceiverName = strings.Join(interfaceSplitres, "/") + name
-		if dut.Vendor() == ondatra.NOKIA {
-			transceiverName = interfaceName + name
-		}
+		return strings.Join(interfaceSplitres, "/") + suffix, nil
+	case ondatra.CISCO:
+		return "", nil
```
xw-g: umm...this changed the original business logic, right? It should return `interfaceName`

feature/platform/tests/optics_power_and_bias_current_test/optics_power_and_bias_current_test.go
```diff
@@ -179,28 +166,22 @@ func TestOpticsPowerUpdate(t *testing.T) {
...
+	case ondatra.JUNIPER:
+		return "", nil
+	case ondatra.NOKIA:
+		suffix := "-transceiver"
+		interfaceName = strings.Replace(interfaceName, "ethernet", "Ethernet", 1)
+		return interfaceName + suffix, nil
 	}
-	return transceiverName, nil
+	return "", fmt.Errorf("no transceiver interface available for DUT vendor %v", dut.Vendor())
```
xw-g, line 186: To make a consistent `switch` syntax, let's move this to the `default` case. Also, I think we can remove the Cisco and Juniper case, and just make `default` to return the `interfaceName` (no error on not specified vendor).

feature/platform/tests/optics_power_and_bias_current_test/optics_power_and_bias_current_test.go
LimeHat: Yep fixed a few minutes before you left the comment :) sorry about that

feature/platform/tests/optics_power_and_bias_current_test/optics_power_and_bias_current_test.go
```diff
@@ -143,7 +129,7 @@ func TestOpticsPowerUpdate(t *testing.T) {
 
 			component := gnmi.OC().Component(transceiverName)
 			if !gnmi.Lookup(t, dut, component.MfgName().State()).IsPresent() {
-				t.Skipf("component.MfgName().Lookup(t).IsPresent() for %q is false. skip it", transceiverName)
+				t.Fatalf("component.MfgName().Lookup(t).IsPresent() for %q is false", transceiverName)
```
xw-g, line 132: I would actually suggest to remove this whole `if` block, because It's not the focus of this test, and is not a must to have by the test (to validate the transceiver power).

feature/platform/tests/optics_power_and_bias_current_test/optics_power_and_bias_current_test.go
```diff
@@ -179,28 +165,19 @@ func TestOpticsPowerUpdate(t *testing.T) {
...
-			ondatra.NOKIA:   "-transceiver",
-		}
-	)
-	transceiverName := interfaceName
-	name, ok := transceiverMap[dut.Vendor()]
-	if !ok {
-		return "", fmt.Errorf("No transceiver interface available for DUT vendor %v", dut.Vendor())
-	}
-	if name != "" {
+	switch dut.Vendor() {
```
cprabha, line 170: @LimeHat , please could you add transceiver name for JUNIPER.  Here is the code for the same.

case ondatra.JUNIPER:
		interfaceSplit := strings.Split(interfaceName, "/")
		interfaceSplit1 := strings.Split(interfaceSplit[0], "-")
		transceiverName := "FPC" + interfaceSplit1[1] + ":PIC" + interfaceSplit[1] + ":PORT" + interfaceSplit[2] + ":Xcvr0"
		return transceiverName, nil

PR #1219, opened 2023-03-06
Update experimental qos_output_queue_counters_test.go with more checks on QoS counters

sezhang2:


feature/experimental/qos/ate_tests/qos_output_queue_counters_test/qos_output_queue_counters_test.go
```diff
@@ -136,52 +136,84 @@ func TestQoSCounters(t *testing.T) {
...
+		counters["dutQosPktsAfterTraffic"][data.queue] = gnmi.Get(t, dut, gnmi.OC().Qos().Interface(dp2.Name()).Output().Queue(data.queue).TransmitPkts().State())
+		counters["dutQosOctetsAfterTraffic"][data.queue] = gnmi.Get(t, dut, gnmi.OC().Qos().Interface(dp2.Name()).Output().Queue(data.queue).TransmitOctets().State())
+		counters["dutQosDroppedPktsAfterTraffic"][data.queue] = gnmi.Get(t, dut, gnmi.OC().Qos().Interface(dp2.Name()).Output().Queue(data.queue).DroppedPkts().State())
+		counters["dutQosDroppedOctetsAfterTraffic"][data.queue] = gnmi.Get(t, dut, gnmi.OC().Qos().Interface(dp2.Name()).Output().Queue(data.queue).DroppedOctets().State())
+		t.Logf("ateInPkts: %v, txPkts %v, Queue: %v", counters["ateInPkts"][data.queue], counters["dutQosPktsAfterTraffic"][data.queue], data.queue)
 
 		lossPct := gnmi.Get(t, ate, gnmi.OC().Flow(trafficID).LossPct().State())
-		if lossPct >= 1 {
-			t.Errorf("Get(traffic loss for queue %q: got %v, want < 1", data.queue, lossPct)
+		t.Logf("Get flow %q: lossPct: %.2f%% or rxPct: %.2f%%, want: %.2f%%\n\n", data.queue, lossPct, 100.0-lossPct, 100.0)
```
dplore, line 181: Should we define want as a static?  Where want  = 100.0?  (or since we are tolerating some loss, the something less than 100?)

feature/experimental/qos/ate_tests/qos_output_queue_counters_test/qos_output_queue_counters_test.go
```diff
@@ -136,52 +136,84 @@ func TestQoSCounters(t *testing.T) {
...
+		if got, want := 100.0-lossPct, float32(100.0); got < want-2.0 || got > want+2.0 {
```
dplore, line 182: 102 make sense?  Would it be simpler to statically assign want to be the minimum value we want and  the comparison be got >= want

Also why do we tolerate any packet loss?  It is not clear to me if these flows expect loss?

feature/experimental/qos/ate_tests/qos_output_queue_counters_test/qos_output_queue_counters_test.go
sezhang2: Yes, There should be no traffic loss for under-subscribed case. Updated the code to check if the throughput is 100.0%.

Test rerun passed.

feature/experimental/qos/ate_tests/qos_output_queue_counters_test/qos_output_queue_counters_test.go
sezhang2: Replied in the previous comment.
   _There should be no traffic loss for under-subscribed case. Updated the code to check if the throughput is 100.0%._

dplore (approved):

PR #1226, opened 2023-03-07
Rework presubmit to run KNE tests for all vendors

alexmasi:


bstoll:
This change SGTM if we can figure out what is going on with the cisco_8000e (or exclude it for now until we do).  I made a mistake on my first GCP CPU quota request and am waiting on my second request still.

alexmasi:
explicitly enabling nested virt fixed the issue, ready for review

bstoll (approved):

PR #1227, opened 2023-03-07
RT-2.1 and RT-2.2, Enabling "MissingValueForDefaults" deviation for default-value checks after adjacency

trathod1:
All the default values are checked via MissingValueForDefaults deviation before the adjacency, Same is required for expected unchanged counters after adjacency checks. Considering, these default-values will be updated only when it changes.


jjing1122:
@trathod1 please rebase

jjing1122:
filed internal presubmit bug b/272600746

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
```diff
@@ -183,7 +183,7 @@ func TestBasic(t *testing.T) {
 				EqualToDefault(sysCounts.AuthFails().State(), uint32(0)),
 				EqualToDefault(sysCounts.AuthTypeFails().State(), uint32(0)),
 				EqualToDefault(sysCounts.CorruptedLsps().State(), uint32(0)),
-				check.Present[uint32](sysCounts.DatabaseOverloads().State()),
+				EqualToDefault(sysCounts.DatabaseOverloads().State(), uint32(0)),
```
arvbaska1, line 186: database-overloads leaf does’nt have any default value in the model & this leaf can be set by different ways [set-overload bit, after reload, config  etc] based on Implementation.
So it’s not correct to assume value is either 0 or nil , and it can be ≥ 0 as well.

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
```diff
@@ -325,16 +325,16 @@ func TestBasic(t *testing.T) {
...
-				check.Equal(sysCounts.ExceedMaxSeqNums().State(), uint32(0)),
-				check.Equal(sysCounts.IdLenMismatch().State(), uint32(0)),
-				check.Equal(sysCounts.LspErrors().State(), uint32(0)),
-				check.Equal(sysCounts.MaxAreaAddressMismatches().State(), uint32(0)),
-				check.Equal(sysCounts.OwnLspPurges().State(), uint32(0)),
-				check.Equal(sysCounts.SeqNumSkips().State(), uint32(0)),
+				EqualToDefault(sysCounts.AuthFails().State(), uint32(0)),
+				EqualToDefault(sysCounts.AuthTypeFails().State(), uint32(0)),
+				EqualToDefault(sysCounts.CorruptedLsps().State(), uint32(0)),
+				EqualToDefault(sysCounts.DatabaseOverloads().State(), uint32(0)),
```
arvbaska1, line 331: Same as previous comment.

jjing1122:
@trathod1 As we discussed offline, please make the changes to keep the previous Present check, but guard it with the missing defaults if-statement (so no check if the deviation is enabled). Feel free to assign back when it is ready to review.

trathod1:
@jjing1122 Updated `DatabaseOverloads` check as discussed, it is now guarded based on MissingValueForDefaults deviation.

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
trathod1: Updated code to perform `check.Present` only when `MissingValueForDefaults` is marked false.

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
trathod1: Updated code to perform `check.Present` only when `MissingValueForDefaults` is marked false.

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
```diff
@@ -49,6 +49,17 @@ func EqualToDefault[T any](query ygnmi.SingletonQuery[T], val T) check.Validator
 	return check.Equal(query, val)
 }
 
+// CheckPresence check for the leaf presense only when MissingValueForDefaults
+// deviation is marked false.
+func CheckPresence[T any](query ygnmi.SingletonQuery[T]) check.Validator {
```
aaronmillisor, line 54: Do you mind explaining the benefits of using a generic function for this?

trathod1: @aaronmillisor Initially this function was made to check for `DatabaseOverloads` which carries value of `uint32` type. But that makes this function very specific for this leaf only. I think, better approach is to keep the function generation for all leaf, just like `EqualToDefault` is generic.

PR #1228, opened 2023-03-07
P4RT: PacketOut: Add Ethernet header to TraceRoute packet

Ankur19:


dplore:
@vbhan-cisco can you validate that these changes should work on your platform?  Based on interpretation of p4rt it seems that specifying the L2 header should be permitted.  We expect a p4rt controller can set the L2 header.

rsindhujha:
hi @Ankur19 
While we add Metadata Id 1 to p4v1.PacketOut, shoudn't we set a value as "0" , instead of "unset"?  For the case of GDP/LLDP packetout we add the p4rt port id number.  
Else we can skip sending this metadata Id 1 for traceroute.

Ankur19:
> > hi @Ankur19 While we add Metadata Id 1 to p4v1.PacketOut, shoudn't we set a value as "0" , instead of "unset"? For the case of GDP/LLDP packetout we add the p4rt port id number. Else we can skip sending this metadata Id 1 for traceroute.
> 
> Hi @rsindhujha I think we need to set all the metadata fields that are defined in P4Info.
> 
> From: https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-packet-i_o
> 
> ```
> Indeed, when a P4Runtime client (or server) generates a PacketOut (or PacketIn) message, it needs to populate the metadata field with as many values as in ControllerPacketMetadata.metadata for the packet-out (or packet-in) case.
> ```
> 
> pins tests set all three fields: https://github.com/sonic-net/sonic-pins/blob/main/p4rt_app/tests/packetio_test.cc#L112

Also agree that we need to set `"0"` for `egress_port` instead of `unset` which is the default.

rsindhujha:
> > > hi @Ankur19 While we add Metadata Id 1 to p4v1.PacketOut, shoudn't we set a value as "0" , instead of "unset"? For the case of GDP/LLDP packetout we add the p4rt port id number. Else we can skip sending this metadata Id 1 for traceroute.
> > 
> > 
> > Hi @rsindhujha I think we need to set all the metadata fields that are defined in P4Info.
> > From: https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-packet-i_o
> > ```
> > Indeed, when a P4Runtime client (or server) generates a PacketOut (or PacketIn) message, it needs to populate the metadata field with as many values as in ControllerPacketMetadata.metadata for the packet-out (or packet-in) case.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > pins tests set all three fields: https://github.com/sonic-net/sonic-pins/blob/main/p4rt_app/tests/packetio_test.cc#L112
> 
> Also agree that we need to set `"0"` for `egress_port` instead of `unset` which is the default. I will update accordingly.

Thanks Ankur.

amrindrr (approved):

PR #1233, opened 2023-03-08
Update mixed_sp_wrr_traffic_test.go

saarvind6481:
added deviation flags


sezhang2 (dismissed):
Why is deviation_qos_dropped_octets added? dropped octets is a required feature support for QoS and vendor neutral counter.

vosipchu:
@sezhang2 Sean, that's the purpose, we start with a deviation and when support comes (or we agree it is not a req), we will remove this.

sezhang2:
> @sezhang2 Sean, that's the purpose, we start with a deviation and when support comes (or we agree it is not a req), we will remove this.

SGTM.

sezhang2 (commented):
Per the discussion, deviation flag is used a workaround to skip the check.

feature/experimental/qos/ate_tests/mixed_sp_wrr_traffic_test/mixed_sp_wrr_traffic_test.go
```diff
@@ -597,12 +612,13 @@ func TestMixedSPWrrTraffic(t *testing.T) {
 				if dutOctetCounterDiff < ateOctetCounterDiff {
 					t.Errorf("Get dutOctetCounterDiff for queue %q: got %v, want >= %v", data.queue, dutOctetCounterDiff, ateOctetCounterDiff)
 				}
-
-				ateDropOctetCounterDiff := (counters["ateOutPkts"][data.queue] - counters["ateInPkts"][data.queue]) * uint64(data.frameSize)
-				dutDropOctetCounterDiff := counters["dutQosDroppedOctetsAfterTraffic"][data.queue] - counters["dutQosDroppedOctetsBeforeTraffic"][data.queue]
-				t.Logf("Queue %q: ateDropOctetCounterDiff: %v dutDropOctetCounterDiff: %v", data.queue, ateDropOctetCounterDiff, dutDropOctetCounterDiff)
-				if dutDropOctetCounterDiff < ateDropOctetCounterDiff {
-					t.Errorf("Get dutDropOctetCounterDiff for queue %q: got %v, want >= %v", data.queue, dutDropOctetCounterDiff, ateDropOctetCounterDiff)
+				if *deviations.QOSDroppedOctets != false {
```
prinikasn, line 615: shouldnt this check be if deviation_qos_dropped_octets == false, so that test continues with original behavior? And for Cisco we set the flag to true? 

Please let me know if I am missing something here.

internal/deviations/deviations.go
```diff
@@ -196,6 +196,8 @@ var (
 
 	BGPMD5RequiresReset = flag.Bool("deviation_bgp_md5_requires_reset", false, "Device requires a BGP session reset to utilize a new MD5 key")
 
+	QOSDroppedOctets = flag.Bool("deviation_qos_dropped_octets", false, "QOS Dropped octets stats for interface")
```
xw-g, line 199: What does this mean? Can you help add more context?

prinikasn (commented):
/gcbrun

prinikasn (commented):
/gcbrun

prinikasn (approved):

internal/deviations/deviations.go
aaronmillisor: The leaf 'dropped octets' is a new leaf added recently for these tests, this is not supported on some platforms and so the check for the leaf is is skipped if this is set to true.

PR #1234, opened 2023-03-08

saarvind6481:
added deviations for droppedoctects


sezhang2 (dismissed):
Why is deviation_qos_dropped_octets added? dropped octets is a required feature support for QoS and vendor neutral counter.

vosipchu:
@sezhang2 Sean, that's the purpose, we start with a deviation and when support comes (or we agree it is not a req), we will remove this.

sezhang2:
> @sezhang2 Sean, that's the purpose, we start with a deviation and when support comes (or we agree it is not a req), we will remove this.

SGTM. Is there is a bug to track the drop counter support?

prinikasn (commented):
/gcbrun

prinikasn (approved):

PR #1236, opened 2023-03-08
TE-14.1: remove unused protocol static from VRF config

LimeHat:
Static routes are not currently used in this test, thus no need to configure the protocol


xw-g:
b/272383411 contains the presubmit effort.

xw-g (approved):

PR #1238, opened 2023-03-08
TE-14.1: fix vrf assignment

LimeHat:
This PR implements proper vrf assignment in TE-14.1 test scenario and refactors usage of some DUT config functions. 
- moved DUT config steps from main `TestScaling()` fn into a new, dedicated `configureDUT()` function.
- default vrf config steps are also moved from `TestScaling()` into `createVrf()`
- `configureInterfaceDUT` is now used to configure only physical/ethernet interface, w/o subif and IP config
- `createSubifDUT()` is used for immediate creation of a subinterface w/o an IP config
- `configureL3SubifDUT()` is used to perform a gnmi.Update against the subif to add IPv4 config (after the subif is assigned to a vrf when using non-default VRFs)
- `generateSubIntfPair()` is split into two separate functions for ATE and DUT subinterface config (`configureATESubIfs` & `configureDUTSubIfs`)

Background info:

Currently, the upstream version of the test is supposed to perform an assignment of DUT port#1 to VRF1 after configuration by using the following code:
https://github.com/openconfig/featureprofiles/blob/1189b075e193be4cc11471237dd58ae26a089ed1/feature/gribi/ate_tests/gribi_scaling/gribi_scaling_test.go#L438

https://github.com/openconfig/featureprofiles/blob/1189b075e193be4cc11471237dd58ae26a089ed1/feature/gribi/ate_tests/gribi_scaling/gribi_scaling_test.go#L366-L371

However this only creates logical struct and prints a log message.

Actual config push via gNMI is supposedly performed via pushConfig() function call:
https://github.com/openconfig/featureprofiles/blob/1189b075e193be4cc11471237dd58ae26a089ed1/feature/gribi/ate_tests/gribi_scaling/gribi_scaling_test.go#L441

But pushconfig performs gNMI replace only against /interface/config/ level of configuration.
https://github.com/openconfig/featureprofiles/blob/1189b075e193be4cc11471237dd58ae26a089ed1/feature/gribi/ate_tests/gribi_scaling/gribi_scaling_test.go#L312-L317

Hence network-instance config is not updated and port#1 is never assigned to VRF1.

---
<sup>


xw-g:
pinged you offline, we need more debugging.

LimeHat:
Note: this may require an adjustment for deviation_interface_config_vrf_before_address introduced #1185

LimeHat:
I took another stab at it with fixing the assignment & doing some refactoring at the same time.

I think this should work for all vendors w/o using the deviation, now the order of operations is:

1) configure all VRFs w/o any interface bindings
2) configure just a Ethernet interface (gnmi replace against /interfaces/interface/config)
3) create a subif (gnmi update against /interfaces/interface/subinterfaces/subinterface/config) w/o IP addressing info
4) bind the subinterface to a vrf
5) update subif with IP address info (gnmi update against /interfaces/interface/subinterfaces/subinterface/config)

Updated the PR description above as well.

xw-g:
> I took another stab at fixing the assignment & doing some refactoring at the same time.
> 
> I think this should work for all vendors w/o using the deviation, now the order of operations is:
> 
> 1. configure all VRFs w/o any interface bindings
> 2. configure just an Ethernet interface (gnmi replace against /interfaces/interface/config)
> 3. create a subif (gnmi update against /interfaces/interface/subinterfaces/subinterface/config) w/o IP addressing info
> 4. bind the subinterface to a vrf
> 5. update subif with IP address info (gnmi update against /interfaces/interface/subinterfaces/subinterface/config)
> 
> Updated the PR description above as well.

@aaronmillisor  thoughts?

LimeHat:
@bkreddy143 please don't introduce any changes to my branch including merging. thank you.

xw-g:
@mojiiba can you help take a look at this PR? I can not tell why it's failing to assign the interface to the VRF. 

```
	createSubifDUT(t, d, dut, dp1, 0, 0)
	fptest.AssignToNetworkInstance(t, dut, dp1.Name(), vrf1, 0)
```

feature/gribi/ate_tests/gribi_scaling/gribi_scaling_test.go
```diff
@@ -289,87 +289,66 @@ func pushDefaultEntries(t *testing.T, args *testArgs, nextHops []string) []strin
...
+	vrfs := []string{*deviations.DefaultNetworkInstance, vrf1, vrf2, vrf3}
+	createVrf(t, dut, vrfs)
+
+	// configure Ethernet interfaces first
+	configureInterfaceDUT(t, d, dut, dp1, "src")
+	configureInterfaceDUT(t, d, dut, dp2, "dst")
+
+	// configure an L3 subinterface without vlan tagging under DUT port#1 and assign it to vrf1
+	createSubifDUT(t, d, dut, dp1, 0, 0)
+	fptest.AssignToNetworkInstance(t, dut, dp1.Name(), vrf1, 0)
```
mojiiba, line 306: This pr introduce a breakage for us. can u please add a deviation in AssignToNetworkInstance to use the interface id exactly as interface name for subinterface zero. 

```
	if si==0 & *deviations.SubInterfaceIDZero{
		netInstIntf.Id = ygot.String(intf.GetName()))
	} else {
		  netInstIntf.Id = ygot.String(intf.GetName() + "." + fmt.Sprint(si))
	}

LimeHat: Ok. That would be the new deviation right?

Thanks for the feedback @mojiiba

xw-g: Thanks @mojiiba. I don't see the deviation existing. Do you plan to create one? 

Or, do you prefer us to follow the same way as you did in https://github.com/openconfig/featureprofiles/pull/1181/files#diff-22fbd408bbd0345dd7d80fd32959110220f47eda63668a9891b92c31d46ba23bR212 ?

LimeHat: Added a `deviation_subif_id_zero` as suggested above, please take a look

mojiiba: thanks a lot!

mojiiba, line 306: also please move fptest.AssignToNetworkInstance(t, dut, dp1.Name(), vrf1, 0) to the end of function. note that there is a deviation InterfaceConfigVrfBeforeAddress that can be used to call fptest.AssignToNetworkInstance either at the beginning or end of the function.

LimeHat: Same comment here as I left in the other pr:

Could you elaborate on this?
Previously I've been told that IP addresses have to be configured after interfaces are assigned to VRFs in some implementations.. Is that not the case anymore?

mojiiba: I think it is still the case for other vendors.  that is why it safe to use InterfaceConfigVrfBeforeAddress.

xw-g:
We plan to use the same approach as https://github.com/openconfig/featureprofiles/pull/1302#issuecomment-1498242756

xw-g:
Same as https://github.com/openconfig/featureprofiles/pull/1302#issuecomment-1505711520, @mojiiba @cprabha can you help review the same approach in this PR too so there is no surprise? Thanks.

cprabha (approved):
It looks good.

mojiiba (approved):

mojiiba:
thanks, LGTM.

xw-g (approved):

xw-g:
Change of test plan, reviewed by multi vendors, merging.

Issue #1044, opened 2023-01-31
RT-1.1 TestParameters: For the case "Test the iBGP session establishment: Neighbor AS", 'local-as' number needs to be different than 'nbrLocalAS'


arvbaska1:
In the RT 1.1 test for In the RT 1.1 test for TestEstablishAndDisconnect:

For the below case for the dutconf

{
			name:    "Test the iBGP session establishment: Neighbor AS",
			**dutConf: bgpCreateNbr(&bgpTestParams{localAS: dutAS2, peerAS: ateAS2, nbrLocalAS: dutAS2}),**
			ateConf: configureATE(t, &bgpTestParams{localAS: ateAS2, peerIP: dutIP}, connInternal),
		}

The localAS (the Global AS) number used is same as for nbrLocalAS (per-neighbor local-as).

As per RFC https://www.rfc-editor.org/rfc/rfc7705#section-3.3.
  "When the "Local AS" capability is used, a
   local ASN will be provided in the configuration that is different
   from the globally configured ASN of the BGP router."

So the localAS should be set to a different value e.g dutAS instead of dutAS2.

Issue #1043, opened 2023-01-31
RT-1.1 TestEstablishAndDisconnect: After BGP cease notification message sent from ATE BGP state verification is for ACTIVE instead of IDLE


arvbaska1:
In the RT 1.1 test for TestEstablishAndDisconnect:

- A cease BGP notification message with error code 6 subcode 6 is sent from ATE to DUT.
	t.Log("Send Cease Notification from ATE to DUT")
	ate.Actions().NewBGPPeerNotification().WithCode(6).WithSubCode(6).WithPeers(bgpPeer).Send(t)

- After this BGP state is expected to be in ACTIVE state.
	t.Log("Verify BGP session state : ACTIVE")
	gnmi.Await(t, dut, nbrPath.SessionState().State(), time.Second*60, oc.Bgp_Neighbor_SessionState_ACTIVE)

As per RFC https://www.rfc-editor.org/rfc/rfc4271.html#section-8.2.2, under the ‘Established State’ part,

"If the local system receives a NOTIFICATION message (Event 24 or
      Event 25) or a TcpConnectionFails (Event 18) from the underlying
      TCP, the local system:

        - sets the ConnectRetryTimer to zero,

        - deletes all routes associated with this connection,

        - releases all the BGP resources,

        - drops the TCP connection,

        - increments the ConnectRetryCounter by 1,

        - changes its state to Idle."

   Event 25: NotifMsg

         Definition: An event is generated when a NOTIFICATION message
                     is received and the error code is anything but
                     "version error"
           Status:     Mandatory
     
So this check here  needs to be modified to be IDLE instead of ACTIVE.

PR #1241, opened 2023-03-09
RT-1.1 Changes to base_bgp_session_parameters_test

arvbaska1:
Made below changes to the testcase:

- As per https://github.com/openconfig/featureprofiles/issues/1043, When BGP is in Established state & it receives a notification message BGP should go into IDLE state. Made changes to check for Not Established BGP state instead of Active.

- For TestEstablishAndDisconnect case "Test the iBGP session establishment: Neighbor AS", changed localAS (the Global AS) numberto be different than for nbrLocalAS (per-neighbor local-as). https://github.com/openconfig/featureprofiles/issues/1044

- Added a deviation for TestPassword to skip the mismatched password.


guoshiuan:
Internal test results can be found at b/272595336.

guoshiuan (approved):

prinikasn (approved):
/gcbrun

PR #1245, opened 2023-03-09
addrundata list simplified JSON rundata

liulk:
Also renamed -list=json to -list=testtracker.

### DIFF
--- a/tools/addrundata/list_json.go
+++ b/tools/addrundata/list_json.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"encoding/json"
+	"io"
+	"path/filepath"
+)
+
+// listJSON writes the testsuite as a JSON map, mapping from the test package directory to
+// the test rundata.
+//
+// Example:
+//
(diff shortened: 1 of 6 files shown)


tools/addrundata/list_json.go
```diff
@@ -0,0 +1,54 @@
...
+//
+//	{
+//	  "feature/subfeature/otg_tests/foo_test": {
+//	    "test.plan_id": "XX-1.1",
+//	    "test.description": "Foo Functional Test",
+//	    "test.uuid": "123e4567-e89b-42d3-8456-426614174000",
+//	  },
+//	  ...
+//	}
+func listJSON(w io.Writer, featuredir string, ts testsuite) error {
```
greg-dennis, line 22: maybe camelCase: featureDir, testSuite

liulk: In these cases, I prefer to treat them as compound nouns to avoid mistaking the first word as a verb.

PR #1248, opened 2023-03-09
gnmi 1.10: add deviations MissingInterfacePhysicalChannel, MissingInterfaceHardwarePort, and MissingCPUMfgName

mojiiba:
This pull adds three deviations MissingInterfacePhysicalChannel, MissingInterfaceHardwarePort, and MissingCPUMfgName to skips the checks for interface/physicalchannel, interface/hardwareport, component/MfgName accordingly.

Also, two of the subtests for QoS counter and LACP member assume the device already has base config for QoS and LACP. So, we add two arguments to skip the test when related base config is not loaded.


internal/args/args.go
```diff
@@ -30,4 +30,6 @@ var (
 	P4RTNodeName2           = flag.String("arg_p4rt_node_name_2", "", "The P4RT Node Name for the second FAP. Test that reserves ports in two different FAPs should configure this P4RT Node in addition to the Node defined in P4RTNodeName1. The value will only be used if deviation ExplicitP4RTNodeComponent is applied.")
 	FullConfigReplaceTime   = flag.Duration("arg_full_config_replace_time", 0, "Time taken for gNMI set operation to complete full configuration replace. Expected duration is in nanoseconds. Expectation is not checked when value is 0.")
 	SubsetConfigReplaceTime = flag.Duration("arg_subset_config_replace_time", 0, "Time taken for gNMI set operation to modify a subset of configuration. Expected duration is in nanoseconds. Expectation is not checked when value is 0.")
+	QoSBaseConfigPresent    = flag.Bool("arg_qos_baseconfig_present", false, "QoS Counter subtest in gNMI-1.10 requires related base config to be loaded. Use this flag to skip the when base config is not loaded.")
```
guoshiuan, line 33: The default value should be true, and I will change it for Cisco tests.

internal/args/args.go
```diff
@@ -30,4 +30,6 @@ var (
...
+	LACPBaseConfigPresent   = flag.Bool("arg_lacp_baseconfig_present", false, "LACP subtest in gNMI-1.10 requires related base config to be loaded. Use this flag to skip the test when base config is not loaded.")
```
guoshiuan, line 34: The default value should be true, and I will change it for Cisco tests.

internal/args/args.go
mojiiba: done

internal/args/args.go
mojiiba: done

guoshiuan:
b/272600177 has the internal test results.

guoshiuan (approved):

xw-g:
Can we have internal bugs to track the deviations? Thanks.

guoshiuan:
> Can we have internal bugs to track the deviations? Thanks.

Done. Please see them in the internal bug.

prinikasn (approved):
/gcbrun

PR #1249, opened 2023-03-10
AFT get flakyness: replace aft get with await to avoid flakiness due to early get

mojiiba:
A few of FP tests do aft get right after programing (without FIB ACK) or switchover. These early get call can be flaky. This pull replaces them with await to avoid the flakiness.


feature/gribi/ate_tests/base_leader_election_test/base_leader_election_test.go
```diff
@@ -220,10 +220,9 @@ func testIPv4LeaderActiveChange(ctx context.Context, t *testing.T, args *testArg
 	args.clientB.AddIPv4(t, ateDstNetCIDR, nhgIndex, *deviations.DefaultNetworkInstance, "", fluent.InstalledInRIB)
 
 	// Verify the entry for 198.51.100.0/24 is active through AFT Telemetry.
+	t.Log("Verify the entry for 198.51.100.0/24 is active through AFT Telemetry.")
```
guoshiuan, line 223: Please use the var `ateDstNetCIDR` and avoid using hard-coded IP address to increase maintainability.

feature/gribi/ate_tests/base_leader_election_test/base_leader_election_test.go
```diff
@@ -247,10 +246,9 @@ func testIPv4LeaderActiveChange(ctx context.Context, t *testing.T, args *testArg
 	args.clientA.AddIPv4(t, ateDstNetCIDR, nhgIndex+2, *deviations.DefaultNetworkInstance, "", fluent.InstalledInRIB)
 
 	// Verify the entry for 198.51.100.0/24 is active through AFT Telemetry.
+	t.Log("Verify the entry for 198.51.100.0/24 is active through AFT Telemetry.")
```
guoshiuan, line 249: Please use the var ateDstNetCIDR and avoid using hard-coded IP address to increase maintainability.

feature/gribi/ate_tests/supervisor_failure_test/supervisor_failure_test.go
```diff
@@ -383,12 +383,10 @@ func TestSupFailure(t *testing.T) {
 	}
 
 	// Verify the entry for 203.0.113.0/24 is active through AFT Telemetry.
+	t.Log("Verify the entry for 203.0.113.0/24 is active through AFT Telemetry.")
```
guoshiuan, line 386: use ateDstNetCIDR instead of hardcoding the address.

feature/gribi/otg_tests/base_leader_election_test/base_leader_election_test.go
```diff
@@ -268,10 +268,9 @@ func testIPv4LeaderActiveChange(ctx context.Context, t *testing.T, args *testArg
 	args.clientB.AddIPv4(t, ateDstNetCIDR, nhgIndex, *deviations.DefaultNetworkInstance, "", fluent.InstalledInRIB)
 
 	// Verify the entry for 198.51.100.0/24 is active through AFT Telemetry.
+	t.Log("Verify the entry for 198.51.100.0/24 is active through AFT Telemetry.")
```
guoshiuan, line 271: ateDstNetCIDR

feature/gribi/otg_tests/base_leader_election_test/base_leader_election_test.go
```diff
@@ -293,10 +292,9 @@ func testIPv4LeaderActiveChange(ctx context.Context, t *testing.T, args *testArg
 	args.clientA.AddIPv4(t, ateDstNetCIDR, nhgIndex+2, *deviations.DefaultNetworkInstance, "", fluent.InstalledInRIB)
 
 	// Verify the entry for 198.51.100.0/24 is active through AFT Telemetry.
+	t.Log("Verify the entry for 198.51.100.0/24 is active through AFT Telemetry.")
```
guoshiuan, line 295: ateDstNetCIDR

feature/gribi/otg_tests/supervisor_failure_test/supervisor_failure_test.go
```diff
@@ -369,12 +369,10 @@ func TestSupFailure(t *testing.T) {
 	}
 
 	// Verify the entry for 203.0.113.0/24 is active through AFT Telemetry.
+	t.Log("Verify the entry for 203.0.113.0/24 is active through AFT Telemetry.")
```
guoshiuan, line 372: ateDstNetCIDR

LimeHat:
Hey @mojiiba , could you elaborate on a few things

- why RIB ACK is not enough in this case
- do you really need to wait up to 2 minutes? shouldn't a few seconds be enough?
- it seems that with the current usage of gnmi.Await you will not fail the tests even if AFT entries are not found after 2 minutes?

mojiiba:
> Hey @mojiiba , could you elaborate on a few things
> 
>     * why RIB ACK is not enough in this case
> 
>     * do you really need to wait up to 2 minutes? shouldn't a few seconds be enough?

Hey @LimeHat  thanks for the comments.
- AFT is supposed to provide an abstract view of the FIB (https://github.com/openconfig/public/blob/53e66659f5ca430f89d707b1be52a9649e4b86b5/release/models/aft/openconfig-aft.yang#L34), that is why if we try AFT before getting FIB ack, there is a chance for getting no data that will result in the test failure. 
- Right now, the timeout values for await calls in FP tests are somehow arbitrarily including this one. Otherwise, you are right, and assuming AFT telemetry is event-based, we should get the event soon after the entry is installed in **FIB**. But setting the timeout to a few seconds is not necessary since the test is a functional (not the bench-marking) test and having a larger timeout value to ensure the stability of the test is much more important than having a small timeout value that results in flakiness. Anyway, I decreased the timeout value to 1 minute which is the upper bound for fib ack timeout, but anything less than that requires changing the test to use FIB ack to prevent flakiness.

LimeHat:
Fair enough. Although, even if we assume that you populate AFT entries only after you program the FIB, I'd argue that even functional tests must have some upper limits on what is acceptable (FIB programming that takes tens of seconds w/o much load or scale sounds questionable?), and indeed you're right that many of the current limits are quite arbitrary. But that's not my call.

What about the switchover case? There's no new fib programming there + a built-in delay of 30 seconds (at least):
https://github.com/openconfig/featureprofiles/blob/39c72c9400c4e67817abb1b1385e0ff797c5e62a/feature/gribi/ate_tests/supervisor_failure_test/supervisor_failure_test.go#L352-L356

Does it take more than that to synchronize AFT telemetry state with the underlying RIB/FIB state after performing a switchover?

mojiiba:
> Fair enough. Although, even if we assume that you populate AFT entries only after you program the FIB, I'd argue that even functional tests must have some upper limits on what is acceptable (FIB programming that takes tens of seconds w/o much load or scale sounds questionable?), and indeed you're right that many of the current limits are quite arbitrary. But that's not my call.
> 
> What about the switchover case? There's no new fib programming there + a built-in delay of 30 seconds (at least):
> 
> https://github.com/openconfig/featureprofiles/blob/39c72c9400c4e67817abb1b1385e0ff797c5e62a/feature/gribi/ate_tests/supervisor_failure_test/supervisor_failure_test.go#L352-L356
> 
> Does it take more than that to synchronize AFT telemetry state with the underlying RIB/FIB state after performing a switchover?

Right now test allows 900 seconds for the swicth-over to be complete (https://github.com/openconfig/featureprofiles/blob/39c72c9400c4e67817abb1b1385e0ff797c5e62a/feature/gribi/ate_tests/supervisor_failure_test/supervisor_failure_test.go#L365). However, it does it by checking for the current data-time that can be processed quickly (less than a minute) after initiating switch-over. Replying to data-time telemetry does not mean the switch-over is completed specially for processes  (e.g., bgp, gribi) that require data synchronization. So, at some point we need to discuss and change the test to use a reliable way to check if the switch-over is completed or not (but probably this is not the call we can make). 

So, concerning the timeout for await, the timeout can be set  maxSwitchoverTime - uint64(time.Since(startSwitchover).Seconds()), however that will result in waiting more that 800 seconds.  That is why I use two minutes that seems a safe upper bound number to prevent flaky results. Later we may revisit the test and increase/decrease the timeout based on the HLD requirements.

LimeHat:
>. So, at some point we need to discuss and change the test to use a reliable way to check if the switch-over is completed or not

Perhaps state-synced flag would be the solution for this issue in the longer run. https://github.com/openconfig/public/blob/53e66659f5ca430f89d707b1be52a9649e4b86b5/release/models/aft/openconfig-aft.yang#L244

mojiiba:
> > . So, at some point we need to discuss and change the test to use a reliable way to check if the switch-over is completed or not
> 
> Perhaps state-synced flag would be the solution for this issue in the longer run. https://github.com/openconfig/public/blob/53e66659f5ca430f89d707b1be52a9649e4b86b5/release/models/aft/openconfig-aft.yang#L244

yes, agree, this seems a solution.

guoshiuan (approved):

vosipchu (approved):

rszarecki (approved):

prinikasn (approved):
/gcbrun

PR #1175, opened 2023-02-22
RT-3.2: add missing instances of InterfaceEnabled deviation

trathod1:
This PR adds missing instances of InterfaceEnabled deviation:

1. ipv6 subif config in ATE version of the test
2. interface, ipv4 subif, ipv6 subif in the OTG version of the test

Reference - https://github.com/openconfig/featureprofiles/pull/1136

"This code is a Contribution to the OpenConfig Feature Profiles project ("Work") made under the Google Software Grant and Corporate Contributor License Agreement ("CLA") and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia's intellectual property are granted for any other purpose. This code is provided on an "as is" basis without any warranties of any kind."

PR #1250, opened 2023-03-10
New deviation flag for include the IP config enabled explicitly when …

gforeal:
…create vlan sub interface

With this new deviation flag, the IP config enabled element could be turned on of off depends on vendor and/or platform whatever applicable.


LimeHat:
hey @gforeal 
Could you elaborate a bit on how is this compared to the current use of deviations `deviations.InterfaceEnabled` & `deviations.IPv4MissingEnabled`?
E.g. changes made in #1175 for RT-3.2 case.

PR #1252, opened 2023-03-10
Add a test case for gNOI FABRIC component reboot.

lifuzhang1108:


rohit-rp:
These changes are already merged as party of https://github.com/openconfig/featureprofiles/pull/1701
Closing the PR

PR #1253, opened 2023-03-11
RT-1.1: Adding deviation for delete of /network-instances/network-instance/tables

trathod1:
Adding deviation `NetworkInstanceTableDeletionRequired` to allow deletion of `/network-instances/network-instance/tables` during `bgpClearConfig` only when deviation is marked true.


feature/experimental/bgp/ate_tests/base_bgp_session_parameters/base_bgp_session_parameters_test.go
```diff
@@ -106,11 +106,13 @@ func bgpClearConfig(t *testing.T, dut *ondatra.DUTDevice) {
 	resetBatch := &gnmi.SetBatch{}
 	gnmi.BatchDelete(resetBatch, gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance).Protocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP").Config())
 
-	tablePath := gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance).TableAny()
-	for _, table := range gnmi.LookupAll[*oc.NetworkInstance_Table](t, dut, tablePath.Config()) {
-		if val, ok := table.Val(); ok {
-			if val.GetProtocol() == oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP {
-				gnmi.BatchDelete(resetBatch, gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance).Table(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, val.GetAddressFamily()).Config())
+	if !*deviations.NoBGPNetworkInstanceTable {
```
xw-g, line 109: Seems like we don't even configure `table` when doing the BGP configuration in this test.

Sounds like we can simply remove the whole related BGP configurations (instead of doing this way using the unimplemented table, therefore we don't even need this deviation)? It does not change the focus of this test, right? 

@trathod1 @mojiiba thoughts?

trathod1: True, script doesn't seem to configure the `table` anywhere. It is is only referred for delete operation during the `bgpClearConfig` function which was recently added.

LimeHat: @xw-g this change was introduced in #1206, maybe you can check with @bstoll on the reasoning?
In our opinion indeed tables (if they are used) should be deleted automatically together with the protocol config.

mojiiba: echo what @LimeHat. the test does not configure tables, and if we introduce a deviation here, that should be for the devices that require deleting tables explicitly.

xw-g: SGTM that we just simply remove the related BGP configs which does the same thing. @bstoll any objection?

bstoll: The documentation for [tables](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-tables) indicates these paths should be created when a protocol is enabled:

> An implementation is expected to create entries within this list when the relevant protocol context is enabled. i.e., when a BGP instance is created with IPv4 and IPv6 address families enabled, the protocol=BGP, address-family=IPv4 table is created by the system.

If the device populates tables at the time of the protocol being configured, they would result in orphans when later removing the protocol.  We have to remove the tables if they exist when removing the protocol.

I think the deviation as proposed would work, because (based on the linked documentation) this path should exist, but I was really hoping we wouldn't need deviations in this function since it's just for cleanup.  Another option might be to use [testt.CaptureFatal](https://github.com/openconfig/testt/blob/main/testt.go#L44) on the LookupAll and not have a deviation - the idea is that if the device does not populate the tables path we would not try to send a Delete on it.  We don't care if the lookup returns no results because that's really what we wanted in the first place.

xw-g: Routing tables are usually created by the protocols, not by the configs. @dplore I think the intent of the `table` here is more for streaming not for configuration, right? (at least for the BGP use case).

Also I'm not aware of any vendor's BGP implementations that supports, or requires, operators to manually clear the  a BGP table via configs. cc: @vosipchu @sachendras

internal/deviations/deviations.go
```diff
@@ -200,4 +200,7 @@ var (
 
 	SkipBGPTestPasswordMismatch = flag.Bool("deviation_skip_bgp_test_password_mismatch", false,
 		"Skip BGP TestPassword mismatch subtest if value is true, Default value is false")
+
+	NoBGPNetworkInstanceTable = flag.Bool("deviation_no_bgp_network_instance_table", false,
```
xw-g, line 204: @trathod1 is it better to name it as: `NetworkInstanceTableDeletionNotRequired`? (Because it is about deletion).

trathod1: @xw-g Current operation is specific to deletion. Considering, it is always going to be for deletion, sure I can name it as you mentioned. I was earlier thinking to keeping it based on entity (if table exists or not) instead of operation (config/delete/update).

feature/experimental/bgp/ate_tests/base_bgp_session_parameters/base_bgp_session_parameters_test.go
xw-g: @dplore in short, IMHO, if a vendor OS implicitly configures the BGP table when configuring BGP protocol, then it shouldn't require explicitly deleting the BGP table when removing the BGP protocol configs. 

In other words, this PR should be the standard way and does not need a deviation flag. Existing behavior (requires explicitly deleting the BGP table) should have a deviation flag.

WDYT?

dplore: network-instance/tables are required to be created _and deleted_ by the network operating system.  An OC client should not need to create or delete these tables.   Any other behavior is a deviation from the requirement.  

The OC yang description only mentions creation.  I will propose and update to the OC yang description to include deletion.

xw-g: With the recommendation from @dplore, @trathod1 can you help refactor this PR so that it's a deviation (require a deviation flag to be set to true) to delete the table? Thanks.

trathod1: Sure. I will push the commit with changes. I will keep the deviation name as `NetworkInstanceTableDeletionRequired`.

trathod1: @xw-g Committed changes as per discussion.

xw-g:
b/273978406 carries the internal presubmit effort.

xw-g:
b/273977750 is the internal bug for tracking the approval.

xw-g:
This PR only introduced a deviation, and did not change the testing logic. Since b/273977750 approved the deviation creation, I"m merging with my own LGTM

xw-g (approved):

PR #1258, opened 2023-03-13
P4RT: Configure port id for needed interfaces only

kjahed:
Currently, P4RT tests configure port ID for all interfaces found in the binding file. This causes the tests to fail if the base configuration does not configure these interfaces (for e.g., due to missing interface type). This PR changes those tests to configure the port id only for the interfaces used in the test.


Ankur19:
Fix static analysis by deleting the `sortPorts` function.

Also need to make similar changes for the otg_tests

feature/experimental/p4rt/ate_tests/base_p4rt/base_p4rt_test.go
xw-g, line 1: @LimeHat can you help review this file? Thanks.

prinikasn (approved):

PR #1260, opened 2023-03-13
OC-1.2: Setting DefaultNetworkInstance on the device

stthummuri:
Network instance type is set to "DEFAULT".
It is mandatory for user to specify type of network instance. For DEFAULT network instance type is "DEFAULT"


greg-dennis (approved):
Approving for Go quality only

sachendras (approved):

PR #1261, opened 2023-03-13
Add ability to read rundata from metadata.textproto

greg-dennis:


internal/rundata/rundata.go
```diff
@@ -125,6 +138,21 @@ func Properties(ctx context.Context, resv *binding.Reservation) map[string]strin
 	return m
 }
 
+// readMetadataProto reads and unmarshals the metadata proto file.
+// If a metadata proto does not exist, returns a nil metadata and nil error.
+func readMetadataProto() (*mpb.Metadata, error) {
+	const metadataFilename = "metadata.textproto"
+	if _, err := os.Stat(metadataFilename); err != nil {
```
liulk, line 145: While assuming that metadata proto exists at  ./metadata.textproto may work for third_party_go_test, this won't work for open source developers because they typically run a test at repo root like this:

```sh
go test ./feature/subfeature/otg_tests/foo_test
```

The test might also pick up the wrong metadata file if it happens to run from the wrong directory.

The primary mechanism should be use the source file path to find the metadata. We can factor out the sourceFullPath from these lines:

https://github.com/openconfig/featureprofiles/blob/bd74c0dc59bb5866677084f3298d20191ee0946c/internal/rundata/local.go#L154-L168

A secondary mechanism for Bazel could be to use os.Args[0] to find the test executable path, and replace the executable's basename with metadata.textproto. This should work for both go_test and third_party_go_test.

internal/rundata/rundata.go
```diff
@@ -125,6 +138,21 @@ func Properties(ctx context.Context, resv *binding.Reservation) map[string]strin
...
+		return nil, nil
```
liulk, line 146: I worry that swallowing the error from os.Stat is a little too subtle and might end up swallowing unintended errors as well. Can we just use os.ReadFile and return any error to the caller? The caller would then use os.IsNotExist to explicitly ignore that particular error.

internal/rundata/rundata.go
```diff
@@ -100,6 +104,15 @@ func topology(resv *binding.Reservation) string {
...
 // Properties builds the test properties map representing run data.
 func Properties(ctx context.Context, resv *binding.Reservation) map[string]string {
+	switch md, err := readMetadataProto(); {
+	case err != nil:
+		log.Errorf("Error reading metadata proto: %v", err)
+	case md != nil:
+		TestUUID = md.GetUuid()
+		TestPlanID = md.GetPlanId()
+		TestDescription = md.GetDescription()
+	}
```
liulk, line 114: Let's factor this part out to its own function.

Also, precedence wise, we should always honor the explicit setting in code, so only use the proto if the code does not set it.

```go
if TestUUID == "" {
  TestUUID = md.GetUuid()
}
```

etc.

Otherwise we might end up in a situation where proto is present but missing uuid, plan ID, description, the code sets them, but they are still emptied out. Could be confusing to developers.

liulk (changes requested):

internal/rundata/rundata.go
greg-dennis: Actually, it will work if someone runs `go test ./feature/subfeature/otg_tests/foo_test`. Go test always uses the test binary (the test package path) as the current working directory.

internal/rundata/rundata.go
greg-dennis: Good idea, done.

internal/rundata/rundata.go
greg-dennis: Since I wanted the metadata textproto to take precedence, I didn't mind so much that a blank metadata would overwrite the rundata_test. As a compromise, I've changed it so that a field takes precedence when it is nonempty.

internal/rundata/rundata_test.go
```diff
@@ -115,6 +117,56 @@ func TestTopology(t *testing.T) {
...
+	TestUUID = ""
+	TestPlanID = ""
+	TestDescription = ""
+
+	metadataText := fmt.Sprintf(`
+uuid: "%s"
+plan_id: "%s"
+description: "%s"
+`, wantUUID, wantPlanID, wantDescription)
+	if err := os.WriteFile("metadata.textproto", []byte(metadataText), 0644); err != nil {
```
liulk, line 134: As discussed yesterday, please only write to t.TempDir() and os.Chdir() there for testing. Avoid clobbering the source tree.

greg-dennis: Fixed yesterday but forgot to push my commits. Done.

liulk (approved):

PR #1277, opened 2023-03-15
 gNMI-1.4: refactor test to use type rather than regex, add deviations and args

mojiiba:
This pull 
- change the test to use component type rather regex to find the components of interest when it is possible.
- it still uses name patterns for switchchip and temp sensor since there is not specific component type for them.
- remove the immediate parent check and ensure the chassis is the eventual parent of all components. 
- remove the software validation for supervisor since according to oc model the software version is only for software components.
- add few deviations.
- removed the skip for software module tests as it is supported by some device. deviations can be added for non-compliant devices.


feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -282,36 +288,99 @@ func TestHardwarecards(t *testing.T) {
...
+	}
+	if len(componentsByType) != 0 {
+		return componentsByType
+	}
+	componentsByType = make(map[string][]string)
+	components := gnmi.GetAll(t, dut, gnmi.OC().ComponentAny().State())
+	for compName := range componentType {
+		for _, c := range components {
+			if strings.Contains(c.GetName(), "Fan") {
+				t.Logf("Component %s type is missing from telemetry", c.GetName())
```
guoshiuan, line 313: Is this log message correct?

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -282,36 +288,99 @@ func TestHardwarecards(t *testing.T) {
...
+			}
+
+			if c.GetType() == nil {
+				t.Logf("Component %s type is missing from telemetry", c.GetName())
+				continue
+			}
+			t.Logf("Component %s has type: %v", c.GetName(), c.GetType())
+			if v := c.GetType(); v == componentType[compName] {
```
guoshiuan, line 321: Why not just "c.GetType() == componentType[compName]"?

guoshiuan, line 321: Can we have an early return here for "c.GetType() != componentType[compName]", and revise the following code with:
switch compName {
case "SwitchChip":
  if *args.SwitchChipNamePattern != "" && !isCompNameExpected(t, c.GetName(), *args.SwitchChipNamePattern) {
    continue
  }
}

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
mojiiba: thanks, it was fixed

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
mojiiba: fixed

mojiiba: thanks for the suggestion, addressed

internal/args/args.go
```diff
@@ -24,13 +24,17 @@ import (
...
+	SubsetConfigReplaceTime = flag.Duration("arg_subset_config_replace_time", 0, "Time taken for gNMI set operation to modify a subset of configuration. Expected duration is in nanoseconds. Expectation is not checked when value is 0.")
+	QoSBaseConfigPresent    = flag.Bool("arg_qos_baseconfig_present", true, "QoS Counter subtest in gNMI-1.10 requires related base config to be loaded. Use this flag to skip the when base config is not loaded.")
+	LACPBaseConfigPresent   = flag.Bool("arg_lacp_baseconfig_present", true, "LACP subtest in gNMI-1.10 requires related base config to be loaded. Use this flag to skip the test when base config is not loaded.")
+	TempSensorNamePattern   = flag.String("arg_temp_sensor_name_pattern", "^[0-9]/RP[0-9]/CPU[0-9].*TEMP.*", "There is no component type specifically for temperature sensors. So, we use the name pattern to find them.")
+	SwitchChipNamePattern   = flag.String("arg_switchchip_name_pattern", "^[0-9]/[0-9]/CPU[0-9]-NPU[0-9]", "There is no component type specifically for SwitchChip components. So, we use the name pattern to find them.")
```
LimeHat, line 36: Would it make sense to use a switch statement on dut.vendor() instead of an argument?
Also, we shouldn't be using vendor-specific defaults for common arguments IMO.

mojiiba: addressed your comment about the default values. missed to clean out when pushed the code.
with switch each vendor should open a separate pr. this way we can all use it. Also, not all vendors require patterns for all components type. That is why i used args that eventually can be removed when OC component model is improved. I can still open to make the change if you have prefer that way.

guoshiuan (approved):

PR #1278, opened 2023-03-16
Update gNMI-1.12 mixed-origin test to use `SetRequest{replace on "openconfig" origin, update on "cli" origin}`

wenovus:
* Changed test from {replace on "cli", update on "openconfig"} to {replace on "openconfig" origin, update on "cli" origin}
* Added test case for subtree replace in addition to root-level replace for "openconfig" origin.
* Added a sanity check with just {replace on "openconfig"}: `"Step 3: Test that replacing device with current config is accepted and is a no-op."`


rcdomigan (dismissed):
I'm shaky on the local style, but the constructed SetRequest is similar to what we're building in our configuration generator (one OC replace and one CLI update)

liulk:
Some points to consider:

 1. The original gNMI-1.12 just tests a simple mixed origin CLI/OC by setting the interface description. I think there is value in keeping that test just to see if the device can process such SetRequest. It can also test dependency, such as CLI depending on something OC configured, and OC depending on something CLI configured, using a much simpler mechanism (e.g. putting a loopback in a non-default VRF).
 2. Let's be explicit in the test plan about which part of OC depends on which part of CLI, and vice versa. The test plan should be concrete enough so that if the test code is deleted, someone else could write a new test with the same spec.
 3. I feel that the more substantial QoS config testing should probably be part of DP. Currently, the test has to use mixed origin just because QoS is not completely modeled in OC yet. In the future, once OC models QoS completely, the test would not need mixed origin anymore and will just be a regular DP test.

wenovus:
> Some points to consider:
> 
> 1. The original gNMI-1.12 just tests a simple mixed origin CLI/OC by setting the interface description. I think there is value in keeping that test just to see if the device can process such SetRequest. It can also test dependency, such as CLI depending on something OC configured, and OC depending on something CLI configured, using a much simpler mechanism (e.g. putting a loopback in a non-default VRF).
I think a good-enough mixed-origin test here would be an example of what we care about in our workflows, and in this case QoS is one of them. While a more permanent test is more ideal, I don't think it's high priority for us to implement. I've put a TODO for changing this to a truly permanent example of a mutual dependency that must be modelled partially in CLI even as OC modelling continues to mature.

> 2. Let's be explicit in the test plan about which part of OC depends on which part of CLI, and vice versa. The test plan should be concrete enough so that if the test code is deleted, someone else could write a new test with the same spec.
Done.

> 3. I feel that the more substantial QoS config testing should probably be part of DP. Currently, the test has to use mixed origin just because QoS is not completely modeled in OC yet. In the future, once OC models QoS completely, the test would not need mixed origin anymore and will just be a regular DP test.
See 1. If there is a truly permanent example of OC+CLI where the CLI part CANNOT be expressed in OC, then I think it would be a better test case. Otherwise I think the current test case is sufficient. Even if this existed I would keep it as a TODO since in the short term the QoS test is better. The current QoS test doesn't exercise this mixed-origin case and instead sets them separately.

feature/system/gnmi/cliorigin/ate_tests/mixed_oc_cli_origin_support_test/mixed_oc_cli_origin_support_test.go
```diff
@@ -124,22 +124,35 @@ func testQoSWithCLIAndOCUpdates(t *testing.T, dut *ondatra.DUTDevice, tCase test
 	}
 }
 
-func TestQoSDependentCLI(t *testing.T) {
+func TestQoSDependentCLIFullReplace(t *testing.T) {
+	// TODO: Skipping this test case until we can figure out how to signal
+	// different subtests' test results.
```
liulk, line 129: As discussed, change to: skipping this test because it is not expected to pass right now.

feature/system/gnmi/cliorigin/ate_tests/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+```
+
+## Procedure
+
+1.  Delete the queue in OpenConfig in case it is still there, and check that it
+    is no longer present.
+2.  Retrieve currently-running OpenConfig and CLI configs.
+3.  Validate that device can accept root replace of current OC config without
```
liulk, line 41: Clarify that this is currently skipped?

feature/system/gnmi/cliorigin/ate_tests/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,61 @@
+# gNMI-1.12: Mixed OpenConfig/CLI Origin
+
+## Summary
+
+Ensure that both CLI and OC configuration can be pushed to the device within the
+same `SetRequest`, with OC config as a `replace` operation and the CLI as an
+`update` operation. Note that this implies stale CLI config may remain after the
+`SetRequest` operation.
+
+The configuration used in this test is a QoS configuration wherein the
```
liulk, line 10: Move this paragraph and the config to the procedures section?

feature/system/gnmi/cliorigin/ate_tests/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,61 @@
...
+  }
+  val:  {
+    ascii_val:  "qos traffic-class 0 name target-group-TEST\nqos tx-queue 0 name TEST"
+  }
+}
+```
+
+## Procedure
+
+1.  Delete the queue in OpenConfig in case it is still there, and check that it
```
liulk, line 38:
```suggestion
1.  Delete the `TEST` queue in OpenConfig in case it is still there, and check that it
```

feature/system/gnmi/cliorigin/ate_tests/mixed_oc_cli_origin_support_test/README.md
```diff
@@ -0,0 +1,61 @@
...
+  val:  {
+    json_ietf_val:  "{\n  full config omitted \n}"
+  }
+}
+update:  {
+  path:  {
+    origin:  "cli"
+  }
+  val:  {
+    ascii_val:  "qos traffic-class 0 name target-group-TEST\nqos tx-queue 0 name TEST"
```
liulk, line 31: Clarify that only Arista CLI is supported by the test, and Cisco and Juniper TODO?

wenovus:
@liulk Done, PTAL

PR #652, opened 2022-10-18
gNMI-1.10 - Changes for Vendor specific component names

agvarghe-cisco:
Changes for Vendor specific component names

PR #1279, opened 2023-03-16
gnmi-1.10: update TestSoftwareVersion() in ATE version of the test

LimeHat:
This PR updates ATE version of the test to retrieve SW versions from control modules instead of CPUs
This is inline with the current version of OTG test:
https://github.com/openconfig/featureprofiles/blob/16fcb63df34ce78cc81d820eb1faa22546bf92b6/feature/gnmi/otg_tests/telemetry_basic_check_test/telemetry_basic_check_test.go#L396-L398

In README, however, it is mentioned that SW version should be retrieved from Linecads, so we may need to update the code to do that instead. Or update the readme.
Or we can do both, retrieve SW versions from linecards and control modules.

https://github.com/openconfig/featureprofiles/blob/16fcb63df34ce78cc81d820eb1faa22546bf92b6/feature/gnmi/otg_tests/telemetry_basic_check_test/README.md#L88-L91

---
<sup>


feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -446,7 +446,7 @@ func TestComponentParent(t *testing.T) {
 func TestSoftwareVersion(t *testing.T) {
 	dut := ondatra.DUT(t, "dut")
 
-	cards := components.FindComponentsByType(t, dut, cpuType)
+	cards := components.FindComponentsByType(t, dut, supervisorType)
```
mojiiba, line 449: this may not help as well, since supervisorType is a hardware component and according to oc model software version is only for software components:

leaf software-version {
      type string;
      description
        "For software components such as operating system or other
        software module, this is the version of the currently
        running software.";
    }

LimeHat: It may be implied that SW component (NOS) is running on top of a supervisor module and its version is reported. And supervisor modules program linecard accordingly during a boot/upgrade procedure, using a specific SW version as well.

In this case, control module#1 can report a different sw version from control module#2 under some circumstances (e.g.  during an upgrade of a system with redundant control modules) and so on.

But I'll leave it up to openconfig experts to decide.

xw-g: @dplore can you help provide some clarify from OC? Thanks. 

The current README: https://github.com/openconfig/featureprofiles/blob/main/feature/gnmi/ate_tests/telemetry_basic_check_test/README.md

xw-g: Current README seems questionable, we will file a README PR to clarify things later.

mojiiba:
I also have a general comment here (not specific to this pr). so far most our focus was on ATE tests. So, any update should go from ATE to OTG. 
FYI @vosipchu

xw-g:
@mojiiba thanks for raising the concern. This PR definitely is not about updating ATE based on OTG. 

Check the https://github.com/openconfig/featureprofiles/pull/1279#discussion_r1137983967, it is more about that our README (test plan) is not clear/questionable, and we will take time to confirm/clarify it. Once that's done, the README is the goal the both ATE and OTG test should be aligned with.

LimeHat:
@mojiiba 
>I also have a general comment here (not specific to this pr). so far most our focus was on ATE tests. So, any update should go from ATE to OTG.

I'm not sure to whom this comment was addressed, but in addition to what Xiao said, this PR was opened only for ATE because OTG logic is already in-line with the changes that were proposed here (I mentioned that in the PR description above).

The discrepancy in the test logic between ATE & OTG was introduced, if I'm not mistaken, by one of your colleagues in #652

xw-g:
An [internal bug](https://b.corp.google.com/issues/274004168#comment12) has concluded that for this specific test we want to update the test plan to check for `/system/state/software-version`, which should indicate the active/primary OS of the CONTROLLER CARD.

@LimeHat can you help update the test plan and your code? Thanks!

xw-g:
> An [internal bug](https://b.corp.google.com/issues/274004168#comment12) has concluded that for this specific test we want to update the test plan to check for `/system/state/software-version`, which should indicate the active/primary OS of the CONTROLLER CARD.
> 
> @LimeHat can you help update the test plan and your code? Thanks!

Actually, let me close this PR first. I'll file a PR to update the test plan. And then we can pull changes into the PR.

xw-g:
Effort will continue in https://github.com/openconfig/featureprofiles/pull/1313

PR #1289, opened 2023-03-16
gNOI-3.2: add logging to simplify failure analysis

mojiiba:
This pull adds a logging to show the failed interfaces after line card reload.


feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
```diff
@@ -216,6 +218,12 @@ func TestLinecardReboot(t *testing.T) {
 		return true
 	})
 	if val, ok := watch.Await(t); !ok {
+		for port, status := range upInterfaces {
```
guoshiuan, line 221: Nit: I think "port" should be named as "intf", and "status" should be named as "up"?

mojiiba: done

guoshiuan:
b/274004398 has the internal test results.

guoshiuan (approved):

feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go
```diff
@@ -210,12 +212,21 @@ func TestLinecardReboot(t *testing.T) {
...
+			} else {
+				upInterfaces[port] = true
 			}
 		}
 		return true
 	})
 	if val, ok := watch.Await(t); !ok {
+		for intf, up := range upInterfaces {
+			if !up {
+				gnmi.Get(t, dut, gnmi.OC().Interface(intf).State())
```
sgirishj, line 226: Hi @mojiiba, what is the purpose of this Get()? The value obtained is not being logged anyway. And if for some reason no value is present at the path, Get() will fail the test fatally thereby skipping the logging at line 227.

mojiiba: when we enable -v 2, it will write the interface data in the log that can be used for troubleshooting. note that this get will happen when the test is going to fail, so it will not add a new failure case to the test.

sgirishj: Got it, thanks!

PR #1297, opened 2023-03-17
gNMI-1.3: Use WatchAll for verifying metric changes in ate

agvarghe-cisco:
- Use WatchAll instead of GetAll for verifying metric changes in ate


guoshiuan (approved):
b/274120483 has the internal test results.

prinikasn (approved):

PR #1272, opened 2023-03-15
Update RT-3.2 Test Plan

xw-g:
Updated the expectation of case#3.

PR #1301, opened 2023-03-17
rt-3.2: update case#3 to skip the test if PF config is rejected

LimeHat:
Test case#3 (duplicated DSCP values) updated according to the new README (#1272).

We introduce a new bool flag "rejectable" that indicates that the test case can be skipped if PF config is rejected by a DUT

---
<sup>

### DIFF
--- a/feature/experimental/policy/policy_vrf_selection/ate_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
+++ b/feature/experimental/policy/policy_vrf_selection/ate_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
@@ -17,6 +17,7 @@
 package policy_based_vrf_selection_test
 
 import (
+	"github.com/openconfig/testt"
 	"strconv"
 	"testing"
 	"time"
@@ -424,6 +425,7 @@ func TestPBR(t *testing.T) {
 		policy       *oc.NetworkInstance_PolicyForwarding
 		passingFlows []*ondatra.Flow
 		failingFlows []*ondatra.Flow
+		rejectable   bool
 	}{
 		{
 			name: "RT3.2 Case1",
@@ -437,6 +439,7 @@ func TestPBR(t *testing.T) {
 				getIPinIPFlow(args, dstEndPointVlan10, "ipinipd10", 10),
 				getIPinIPFlow(args, dstEndPointVlan20, "ipinipd20", 20),
 				getIPinIPFlow(args, dstEndPointVlan30, "ipinipd30", 30)},
+			rejectable: false,
 		},
 		{
 			name: "RT3.2 Case2",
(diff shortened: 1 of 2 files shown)


feature/experimental/policy/policy_vrf_selection/ate_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
```diff
@@ -503,8 +509,15 @@ func TestPBR(t *testing.T) {
...
-			gnmi.Update(t, dut, gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance).PolicyForwarding().Config(), tc.policy)
-
+			if tc.rejectable {
+				if errMsg := testt.CaptureFatal(t, func(t testing.TB) {
+					gnmi.Update(t, dut, gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance).PolicyForwarding().Config(), tc.policy)
+				}); errMsg != nil {
+					t.Skipf("Skipping test case %q, PolicyForwarding config was rejected with an error: %s", tc.name, *errMsg)
+				}
+			} else {
+				gnmi.Update(t, dut, gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance).PolicyForwarding().Config(), tc.policy)
```
xw-g, line 519: I think you can write something like the following, so you would avoid duplicated code of the `gnmi.Update()`, WDTY?

```
errMsg := testt.CaptureFatal(gnmi.Update(blablabla))
if errMsg != nil {
    if tc.rejectable {
        t.Skipf(blablabla)
    }
    t.Fatal(blablabla)
}
```

feature/experimental/policy/policy_vrf_selection/ate_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
```diff
@@ -437,6 +439,7 @@ func TestPBR(t *testing.T) {
 				getIPinIPFlow(args, dstEndPointVlan10, "ipinipd10", 10),
 				getIPinIPFlow(args, dstEndPointVlan20, "ipinipd20", 20),
 				getIPinIPFlow(args, dstEndPointVlan30, "ipinipd30", 30)},
+			rejectable: false,
```
xw-g, line 442: Golang boolean by default is false, no need to call it out explicitly.

feature/experimental/policy/policy_vrf_selection/ate_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
LimeHat: Indeed, let me do that. thx

LimeHat:
Thanks for the review @xw-g , updated the code accordingly.

PR #1302, opened 2023-03-20
te-3.3: update interface config / vrf binding logic

LimeHat:
This PR refactors ConfigureDUT() logic to create subinterfaces first before they are assigned to network-instances, followed by L3 subif configuration
---
<sup>
</sup>

### DIFF
--- a/feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
+++ b/feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
@@ -247,14 +247,6 @@ func (a *attributes) configSubinterfaceDUT(t *testing.T, intf *oc.Interface) {
 		} else {
 			s.GetOrCreateVlan().GetOrCreateMatch().GetOrCreateSingleTagged().VlanId = ygot.Uint16(uint16(i))
 		}
-		//ip := a.ip(uint8(i))
-		//s4 := s.GetOrCreateIpv4()
-		//if *deviations.InterfaceEnabled && !*deviations.IPv4MissingEnabled {
-		//	s4.Enabled = ygot.Bool(true)
-		//}
-		//s4a := s4.GetOrCreateAddress(ip)
-		//s4a.PrefixLength = ygot.Uint8(a.IPv4Len)
-		//t.Logf("Adding DUT Subinterface with ID: %d, Vlan ID: %d and IPv4 address: %s", i, i, ip)
 	}
 }
 
--- a/feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
+++ b/feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
@@ -95,8 +95,9 @@ var (
 			IPv4:    dutPort2IPv4(0),
 			IPv4Len: ipv4PrefixLen,
 		},
-		numSubIntf: 18,
-		ip:         dutPort2IPv4,
+		numSubIntf:      18,
+		networkInstance: *deviations.DefaultNetworkInstance,
+		ip:              dutPort2IPv4,
 	}
 
 	atePort2 = attributes{
@@ -230,10 +231,13 @@ func awaitTimeout(ctx context.Context, c *fluent.GRIBIClient, t testing.TB, time
 // for Subinterface(1) = dutPort.ip(1) = dutPort.ip + 4
 func (a *attributes) configSubinterfaceDUT(t *testing.T, intf *oc.Interface) {
 	t.Helper()
-
+	if a.numSubIntf == 0 {
+		s := intf.GetOrCreateSubinterface(0)
+		if *deviations.InterfaceEnabled {
+			s.Enabled = ygot.Bool(true)
+		}
+	}
 	for i := uint32(1); i <= a.numSubIntf; i++ {
-		ip := a.ip(uint8(i))
-
 		s := intf.GetOrCreateSubinterface(i)
 		if *deviations.InterfaceEnabled {
 			s.Enabled = ygot.Bool(true)
@@ -243,6 +247,31 @@ func (a *attributes) configSubinterfaceDUT(t *testing.T, intf *oc.Interface) {
 		} else {
 			s.GetOrCreateVlan().GetOrCreateMatch().GetOrCreateSingleTagged().VlanId = ygot.Uint16(uint16(i))
 		}
+		//ip := a.ip(uint8(i))
+		//s4 := s.GetOrCreateIpv4()
+		//if *deviations.InterfaceEnabled && !*deviations.IPv4MissingEnabled {
+		//	s4.Enabled = ygot.Bool(true)
+		//}
+		//s4a := s4.GetOrCreateAddress(ip)
+		//s4a.PrefixLength = ygot.Uint8(a.IPv4Len)
+		//t.Logf("Adding DUT Subinterface with ID: %d, Vlan ID: %d and IPv4 address: %s", i, i, ip)
+	}
+}
+
+func (a *attributes) configL3SubifDUT(t *testing.T, intf *oc.Interface) {
+	if a.numSubIntf == 0 {
+		s := intf.GetOrCreateSubinterface(0)
+		ip := a.ip(uint8(0))
+		s4 := s.GetOrCreateIpv4()
+		if *deviations.InterfaceEnabled && !*deviations.IPv4MissingEnabled {
+			s4.Enabled = ygot.Bool(true)
+		}
+		s4a := s4.GetOrCreateAddress(ip)
+		s4a.PrefixLength = ygot.Uint8(a.IPv4Len)
+	}
+	for i := uint32(1); i <= a.numSubIntf; i++ {
+		s := intf.GetOrCreateSubinterface(i)
+		ip := a.ip(uint8(i))
 		s4 := s.GetOrCreateIpv4()
 		if *deviations.InterfaceEnabled && !*deviations.IPv4MissingEnabled {
 			s4.Enabled = ygot.Bool(true)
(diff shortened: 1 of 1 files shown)


feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
```diff
@@ -243,6 +247,23 @@ func (a *attributes) configSubinterfaceDUT(t *testing.T, intf *oc.Interface) {
...
 		}
+	}
+}
+
+func (a *attributes) configL3SubifDUT(t *testing.T, intf *oc.Interface) {
+	if a.numSubIntf == 0 {
+		s := intf.GetOrCreateSubinterface(0)
+		ip := a.ip(uint8(0))
+		s4 := s.GetOrCreateIpv4()
+		if *deviations.InterfaceEnabled && !*deviations.IPv4MissingEnabled {
```
jjing1122, line 258: this can get merged to reduce code duplication

feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
```diff
@@ -243,6 +247,23 @@ func (a *attributes) configSubinterfaceDUT(t *testing.T, intf *oc.Interface) {
 		} else {
 			s.GetOrCreateVlan().GetOrCreateMatch().GetOrCreateSingleTagged().VlanId = ygot.Uint16(uint16(i))
 		}
+	}
+}
+
+func (a *attributes) configL3SubifDUT(t *testing.T, intf *oc.Interface) {
```
jjing1122, line 253: Line 253-274 can be  merged to reduce code duplication

feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
```diff
@@ -259,51 +280,55 @@ func (a *attributes) configInterfaceDUT(t *testing.T, d *ondatra.DUTDevice, p *o
...
-
-	if a.numSubIntf > 0 {
-		i.Description = ygot.String(a.Desc)
-		i.Type = oc.IETFInterfaces_InterfaceType_ethernetCsmacd
-		if *deviations.InterfaceEnabled {
-			i.Enabled = ygot.Bool(true)
-		}
-	} else {
-		i = a.NewOCInterface(p.Name())
+	i.Description = ygot.String(a.Desc)
```
jjing1122, line 283: Is Line 283- Line 286 valid for a.numSubIntf==0?

LimeHat: I don't see why they wouldn't be valid (it's just a basic Ethernet interface config -> description/type/speed/enabled), I would assume it would (and should) be applicable for all Ethernet interfaces regardless of number of subinterfaces.

But if there are any knows issues or limitations - please le me know

feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
LimeHat: Updated

feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
LimeHat: Updated per our conversation.

feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
```diff
@@ -270,40 +287,50 @@ func (a *attributes) configInterfaceDUT(t *testing.T, d *ondatra.DUTDevice, p *o
...
+
+func (a *attributes) assignSubifsToNetworkInstance(t *testing.T, d *ondatra.DUTDevice, p *ondatra.Port) {
+	niName := a.networkInstance
+
+	if niName != *deviations.DefaultNetworkInstance || *deviations.ExplicitInterfaceInDefaultVRF {
+		if a.numSubIntf == 0 {
+			fptest.AssignToNetworkInstance(t, d, p.Name(), niName, 0)
+		} else {
+			for i := uint32(1); i <= a.numSubIntf; i++ {
+				fptest.AssignToNetworkInstance(t, d, p.Name(), niName, i)
```
mojiiba, line 332: This pr introduce a breakage for us. can u please add a deviation in AssignToNetworkInstance to use the interface id exactly as interface name for subinterface zero. 

```
	if si==0 & *deviations.SubInterfaceIDZero{
		netInstIntf.Id = ygot.String(intf.GetName()))
	} else {
		  netInstIntf.Id = ygot.String(intf.GetName() + "." + fmt.Sprint(si))
	}

File: feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go [hunk 5]
```diff
@@ -270,40 +287,50 @@ func (a *attributes) configInterfaceDUT(t *testing.T, d *ondatra.DUTDevice, p *o
...
 
-	a.configSubinterfaceDUT(t, i)
-	intfPath := gnmi.OC().Interface(p.Name())
-	gnmi.Update(t, d, intfPath.Config(), i)
 	if *deviations.ExplicitPortSpeed {
-		fptest.SetPortSpeed(t, p)
+		i.GetOrCreateEthernet().PortSpeed = fptest.GetIfSpeed(t, p)
 	}
+
+	a.configSubinterfaceDUT(t, i)
```
mojiiba, line 294: can u please use the following code 
```
	a.configSubinterfaceDUT(t, i)
	intfPath := gnmi.OC().Interface(p.Name())
	a.configL3SubifDUT(t, i)
	gnmi.Replace(t, d, intfPath.Config(), i)
	fptest.LogQuery(t, "DUT", intfPath.Config(), gnmi.GetConfig(t, d, intfPath.Config()))
	a.assignSubifsToNetworkInstance(t, d, p)
	
```

LimeHat: Could you elaborate on this?

Previously I've been told that IP addresses have to be configured after interfaces are assigned to VRFs in some implementations.. Is that not the case anymore?

mojiiba: there is deviation InterfaceConfigVrfBeforeAddress but that is not used by us.

LimeHat: Okay, that's good.
So, just to settle it once and for all,
1) you're ok with fully-configured subinterface before the vrf assignment, right?
2) do you need an address (or full subinterface config) before the assignment? I.e. the currently proposed scheme will not work for you? Or it doesn't matter

(sorry for so many questions; it's just we've been changing it back and forth while trying to accommodate all vendor requirements, so want to get maximum clarity)

LimeHat: @mojiiba any feedback on the above two questions?

as an additional point, in the upstream main version of the test, currently the same order is used (L3 addressing is configured after VRF assignment). So I'm not entirely sure why are you asking to change this now in this PR (unless it also breaks something for you)

xw-g: @mojiiba can you help give us some hint? Thanks.

xw-g:
@LimeHat meanwhile, let's update the README to use VRF selection policy. Thanks!

feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
mojiiba: currently, we do not allow gnmi replace on the assigned interfaces to non-default vrfs. so, either you need to configure interfaces and then assign them to non-default vrf or use gnmi.update instead of replace.  so to answer your question we are ok with option 1.

LimeHat: > so, either you need to configure interfaces and then assign them to non-default vrf or use gnmi.update instead of replace.

but the code was fully aligned with that?
the order of operations was:
1) gnmi.Replace (to create a subif)
2) assign
3) gnmi.Update (to add l3 info)

anyhow, switched to another method based on conversation with @xw-g.

LimeHat:
@xw-g updated based on our conversation offline; please take a look.
Thanks!

feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
```diff
@@ -270,51 +269,98 @@ func (a *attributes) configInterfaceDUT(t *testing.T, d *ondatra.DUTDevice, p *o
...
-				Interface:    ygot.String(p.Name()),
-				Subinterface: ygot.Uint32(0),
-			})
+		if *deviations.ExplicitGRIBIUnderNetworkInstance {
+			fptest.EnableGRIBIUnderNetworkInstance(t, d, vrf)
 		}
 	}
+
+	// configure PBF in DEFAULT vrf
+	defNIPath := gnmi.OC().NetworkInstance(*deviations.DefaultNetworkInstance)
```
mohanasm, line 318: Can you add the config to specify the instance type for *deviations.DefaultNetworkInstance, like below ?
	gnmi.Replace(t, d, defNIPath.Type().Config(), oc.NetworkInstanceTypes_NETWORK_INSTANCE_TYPE_DEFAULT_INSTANCE)

mohanasm (changes requested):

feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
```diff
@@ -270,51 +269,98 @@ func (a *attributes) configInterfaceDUT(t *testing.T, d *ondatra.DUTDevice, p *o
...
+}
+
+// configurePBF returns a fully configured network-instance PF struct
+func configurePBF() *oc.NetworkInstance_PolicyForwarding {
+	d := &oc.Root{}
+	ni := d.GetOrCreateNetworkInstance(*deviations.DefaultNetworkInstance)
+	pf := ni.GetOrCreatePolicyForwarding()
+	vrfPolicy := pf.GetOrCreatePolicy(policyName)
+	vrfPolicy.SetType(oc.Policy_Type_VRF_SELECTION_POLICY)
+	vrfPolicy.GetOrCreateRule(1).GetOrCreateIpv4().SourceAddress = ygot.String(dutPort1.IPv4 + "/32")
```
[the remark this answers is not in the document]
xw-g: yes, I agree you both. Seems like original code is a typo there.

feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
xw-g, line 1: I think we need to copy and paste your changes to OTG too?

xw-g:
@mojiiba @cprabha this is a change of the test plan. 

The motivation is to use VRF selection policy to avoid the conflict of how different vendors assign interfaces to non-default VRF. (At least not a priority to address the challenge now). 

The changes looks good to me. Can you help review? Thanks!.

cprabha:
> @mojiiba @cprabha this is a change of the test plan.
> 
> The motivation is to use VRF selection policy to avoid the conflict of how different vendors assign interfaces to non-default VRF. (At least not a priority to address the challenge now).
> 
> The changes looks good to me. Can you help review? Thanks!.

@xw-g will check and update . Thank you

feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
LimeHat: There's no corresponding version of this test for OTG I believe

cprabha (approved):
It looks good.

mojiiba:
Thanks! LGTM

xw-g (approved):

PR #1303, opened 2023-03-20
Expand error checking and unit tests of reading test case data.

greg-dennis:
Prepare transition from parsedData to metadata proto step 1.


tools/addrundata/case.go
```diff
@@ -31,18 +31,13 @@ func (tc *testcase) read(testdir string) error {
...
 		return fmt.Errorf("could not glob: %w", err)
 	}
-	for _, testpath := range testpaths {
-		if err := readFile(testpath, tc.readPackage); err != nil {
-			return fmt.Errorf("could not detect test package: %w", err)
-		}
-		if tc.pkg != "" {
-			break
-		}
+	if len(testpaths) == 0 {
```
liulk, line 34: Previously, we allow a test package to have no code and just README.md, and rundata_test.go would be generated with an assumed package name that is taken from the basename of the testdir.

Now it would be impossible to generate a rundata_test.go for a new test without at least some dummy code, which seems like a regression.

Revert all the changes in this function? Feel free to add a comment about assumed package name.

tools/addrundata/case.go
```diff
@@ -84,7 +82,14 @@ func (tc *testcase) readPackage(r io.Reader) error {
 func (tc *testcase) check() []error {
 	var errs []error
 
-	if tc.markdown.hasData {
+	if !tc.existing.hasData {
```
liulk, line 85: I agree the changes here look cleaner.

tools/addrundata/case.go
greg-dennis: Where is the logic to infer the package name?

greg-dennis: behavior reverted

liulk (approved):

PR #1308, opened 2023-03-20
Pin openconfig/public repo when validating paths using a bash file.

wenovus:
This seems the cleanest way to do this. While having all the dependencies in `go.mod` would be nice, it is not possible to have the `require` in `go.mod` without making openconfig/public have a `go.mod` file, and even if we could do that, `go mod vendor` will download all dependencies for the module, and that doesn't seem right just to maintain a view of a non-Go repo.

### DIFF
--- a/tools/clone_oc_public.sh
+++ b/tools/clone_oc_public.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+git clone https://github.com/openconfig/public.git $1
+cd $1
+branch=$(git tag -l "$2*" | sort -V | tail -1)
+git checkout $branch
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 openconfig_public:
-	git clone https://github.com/openconfig/public.git openconfig_public
+	tools/clone_oc_public.sh openconfig_public v1
 
 .PHONY: validate_paths
 validate_paths: openconfig_public proto/feature_go_proto/feature.pb.go


liulk:
Looks like the bash script tries to detect the latest version matching "v1*" is that right? Right now I see there is just "v1.0.0". I think you could also alias that tag to latest-stable or something, and we can just do `git clone -b latest-stable ...` here.

Another option: didn't we use to point a git submodule to openconfig/public? Was that ruled out?

https://git-scm.com/book/en/v2/Git-Tools-Submodules

wenovus:
> Looks like the bash script tries to detect the latest version matching "v1*" is that right? Right now I see there is just "v1.0.0". I think you could also alias that tag to latest-stable or something, and we can just do `git clone -b latest-stable ...` here.
> 
> Another option: didn't we use to point a git submodule to openconfig/public? Was that ruled out?
> 
> https://git-scm.com/book/en/v2/Git-Tools-Submodules

One concern I have about the stable idea is that once released, all new CI workflows would fail. Do we want to have those disruptions? I think a better way might be to add a very simple GitHub action that errors out when the major version is behind. I can work on that in a separate PR if we agree that's the right approach.

I think submodules is not necessary because we're not changing any files in openconfig/public, or even depending on them at all except for the path checker script. My understanding of submodules is that it's useful when we're developing between two repos, but if all the paths come from Ondatra I don't think it's very useful for that.

liulk:
> One concern I have about the stable idea is that once released, all new CI workflows would fail. Do we want to have those disruptions? I think a better way might be to add a very simple GitHub action that errors out when the major version is behind. I can work on that in a separate PR if we agree that's the right approach.

If the issue is that `latest-stable` isn't always clear it should be tied to v1, then let's have `v1-latest` that only points to the latest v1.* release but will never be v2.

By the way, just realized that `v1*` pattern will accidentally pick up `v10`, `v11`, etc.

> I think submodules is not necessary because we're not changing any files in openconfig/public, or even depending on them at all except for the path checker script. My understanding of submodules is that it's useful when we're developing between two repos, but if all the paths come from Ondatra I don't think it's very useful for that.

git submodule is just a way to recursively clone/checkout another repo elsewhere that is always pinned to a specific commit. It doesn't matter if there is active development or not.

wenovus:
> > One concern I have about the stable idea is that once released, all new CI workflows would fail. Do we want to have those disruptions? I think a better way might be to add a very simple GitHub action that errors out when the major version is behind. I can work on that in a separate PR if we agree that's the right approach.
> 
> If the issue is that `latest-stable` isn't always clear it should be tied to v1, then let's have `v1-latest` that only points to the latest v1.* release but will never be v2.
> 
> By the way, just realized that `v1*` pattern will accidentally pick up `v10`, `v11`, etc.
> 
Git doesn't have good support for floating tags -- once a `latest` tag has been created and pulled down to a local repo, then when it's forced-updated upstream, even after pulling it's not updated at the local repo. I think this means this will not be implemented for the openconfig/public repo.

For `v1*` good catch, although the `sort -V` flag is actually just to handle this issue so I think it's alright.

> > I think submodules is not necessary because we're not changing any files in openconfig/public, or even depending on them at all except for the path checker script. My understanding of submodules is that it's useful when we're developing between two repos, but if all the paths come from Ondatra I don't think it's very useful for that.
> 
> git submodule is just a way to recursively clone/checkout another repo elsewhere that is always pinned to a specific commit. It doesn't matter if there is active development or not.

I think what I want is to make this as less manual as possible. I believe git submodule doesn't update automatically if a new minor public repository comes out, so I'm still leaning towards the github actions solution.

tools/clone_oc_public.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+git clone https://github.com/openconfig/public.git $1
+cd $1
+branch=$(git tag -l "$2.*" | sort -V | tail -1)
```
liulk, line 5:
```suggestion
git clone https://github.com/openconfig/public.git "$1"
cd "$1"
branch="$(git tag -l "$2.*" | sort -V | tail -1)"
```

tools/clone_oc_public.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
liulk, line 1:
```suggestion
#!/bin/bash
# Usage: clone_oc_public.sh local_dir version_prefix
```

tools/clone_oc_public.sh
```diff
@@ -0,0 +1,6 @@
...
+set -e
+git clone https://github.com/openconfig/public.git $1
+cd $1
+branch=$(git tag -l "$2.*" | sort -V | tail -1)
+git checkout $branch
```
liulk, line 6:
```suggestion
git checkout "$branch"
```

liulk (approved):

PR #1309, opened 2023-03-21
RT-5.1: Updating mac-address comparison to ignore-case

trathod1:
To compare actual mac-address value and ignore upper/lower case, updating both `want` and `got` mac-address to upper-case before `ygot.Diff` is called.


jjing1122:
filed internal presubmit b/274525500

jjing1122 (approved):

PR #1312, opened 2023-03-22
Change parsedData from a mostly pass-by-value type to a pass-by-pointer

greg-dennis:
This more easily paves the way for using the metadata proto, which will be pass-by-pointer, in place of the parsedData type.


tools/addrundata/rundata.go
```diff
@@ -15,51 +15,52 @@ type parsedData struct {
...
 	testUUID        string
-	hasData         bool
 }
 
 // markdownRE matches the heading line: `# XX-1.1: Foo Functional Test`
 var markdownRE = regexp.MustCompile(`#(.*?):(.*)`)
 
 // fromMarkdown reads parsedData from README.md
-func (pd *parsedData) fromMarkdown(r io.Reader) error {
+func fromMarkdown(r io.Reader) (*parsedData, error) {
```
liulk, line 24: nit: since this is now a top-level function, let's rename the functions to better disambiguate them

* fromMarkdown -> parseMarkdown
* fromCode -> parseCode

greg-dennis: Good call, done.

liulk (approved):

PR #1318, opened 2023-03-23
Update README.md for gNMI-1.10

sachendras:
Updating test to provide more clarity on the correctness check for hardware-port leaf under interfaces. Goal is to ensure that the leaf is of type oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_CHASSIS_PORT and has an ancestor of type CHASSIS.


dplore (approved):
Suggestion for additional clarification

dplore (approved):

PR #1326, opened 2023-03-24
addrundata writes out metadata.proto files when fixing rundata

greg-dennis:


tools/addrundata/case.go
```diff
@@ -17,31 +16,26 @@ import (
...
-	if err != nil {
-		return fmt.Errorf("could not glob: %w", err)
-	}
-	for _, testpath := range testpaths {
-		if err := readFile(testpath, tc.readPackage); err != nil {
-			return fmt.Errorf("could not detect test package: %w", err)
-		}
-		if tc.pkg != "" {
-			break
+	if err := readFile(filepath.Join(testdir, "rundata_test.go"), tc.readCode); err != nil {
```
liulk, line 30: Suggest use switch to avoid double nesting?

```go
switch err := readFile(...) {
case err == nil:
  tc.deprecated = true
case os.IsNotExist(err):
  // This is fine.
default:
  return fmt.Errorf(...)
}

File: tools/addrundata/case.go [hunk 2]
```diff
@@ -174,22 +162,19 @@ var errNoop = errors.New("already up to date")
...
 // write commits the fixed rundata to the filesystem.
 func (tc *testcase) write(testdir string) error {
-	if tc.pkg == "" {
-		return errors.New("missing test package name")
-	}
 	if tc.fixed == nil {
 		return errors.New("test case was not fixed")
 	}
-	if proto.Equal(tc.existing, tc.fixed) {
+	if proto.Equal(tc.existing, tc.fixed) && !(*checkDeprecated && tc.deprecated) {
```
liulk, line 168:
```suggestion
	ok := !tc.deprecated || !*checkDeprecated
	if ok && proto.Equal(tc.existing, tc.fixed) {
```

tools/addrundata/rundata.go
```diff
@@ -92,29 +103,22 @@ func parseInit(sc *bufio.Scanner) (*mpb.Metadata, error) {
 	return nil, errors.New("func init() was not terminated")
 }
 
-var tmpl = template.Must(template.New("rundata_test.go").Parse(
-	`// Code generated by go run tools/addrundata; DO NOT EDIT.
-package {{.Package}}
-
-import "github.com/openconfig/featureprofiles/internal/rundata"
+var tmpl = template.Must(template.New("metadata.textproto").Parse(
```
liulk, line 106: We really have to allow pass-through of fields other than those listed here.

tools/addrundata/rundata.go
```diff
@@ -92,29 +103,22 @@ func parseInit(sc *bufio.Scanner) (*mpb.Metadata, error) {
...
+	`# proto-file: proto/metadata.proto
```
liulk, line 107: For proto-file, use fully qualified path github.com/openconfig/featureprofiles/proto/metadata.proto?

liulk (changes requested):

tools/addrundata/case.go
greg-dennis: Done

tools/addrundata/case.go
greg-dennis: "ok" seemed a little cryptic to me in this context, but I did go with a temporary variable

tools/addrundata/rundata.go
greg-dennis: Yes, that does look like the norm, thanks.

tools/addrundata/rundata.go
```diff
@@ -103,22 +102,12 @@ func parseInit(sc *bufio.Scanner) (*mpb.Metadata, error) {
...
-	}{
-		UUID:        md.Uuid,
-		PlanID:      md.PlanId,
-		Description: md.Description,
-	})
+	bytes, err := prototext.Marshal(md)
+	if err != nil {
+		return err
+	}
+	_, err = w.Write(bytes)
```
liulk, line 111: Write the proto-file and proto-message also?

tools/addrundata/rundata.go
greg-dennis: Done. Had this in a subsequent PR but moved in here.

liulk (dismissed):

tools/addrundata/rundata.go
greg-dennis: Done

PR #1330, opened 2023-03-24
Replace rundata_test.go files with metadata.textproto files

greg-dennis:


greg-dennis:
I discovered I need to fix validate_paths.go first: https://github.com/openconfig/featureprofiles/pull/1342

greg-dennis:
All set now.

PR #1341, opened 2023-03-27
gNMI-1.9: Config hostname to prevent test failure when oc-related base config is empty

mojiiba:
The test assumes that some oc related base config is not empty that is not always true. Config the hostname to prevent the test failure when oc related config is empty.


prinikasn (commented):
/gcbrun

prinikasn (approved):

prinikasn (approved):

PR #1343, opened 2023-03-27
TE-13.1: add retry for gnoi after switchover

mojiiba:
Test misses to retry gnoi connection after switch over. This pull fixes the issue.


feature/experimental/gribi/ate_tests/route_addition_during_failover_test/route_addition_during_failover_test.go
```diff
@@ -134,6 +135,12 @@ func coreFileCheck(t *testing.T, dut *ondatra.DUTDevice, gnoiClient raw.GNOI, sy
 		Path: vendorCoreFilePath[dut.Vendor()],
 	}
 	validResponse, err := gnoiClient.File().Stat(context.Background(), in)
+	if err != nil {
+		if retry {
```
guoshiuan, line 139: So we only retry it once? Is it sufficient?

mojiiba: yes, it should recover with one retry.

prinikasn (commented):
/gcbrun

prinikasn (approved):

PR #1345, opened 2023-03-28
Refactor gnoi switchover and component reboot tests and add CheckInterfacesInBinding flag

mojiiba:
gnoi switchover and component reboot tests share redundant functions. This pull moves them to internal. Also, the tests check the interfaces status for all interfaces. However, this check can be flaky if reservation does not guarantee the connectivity of not-binded interfaces. For this I added CheckInterfacesInBinding to perfrom the check based on interfaces in binding file.


internal/helpers/helpers.go
```diff
@@ -0,0 +1,93 @@
...
+	for _, port := range upIntfs {
+		batch.AddPaths(gnmi.OC().Interface(port).OperStatus())
+		upInterfaces[port] = true
+	}
+	watch := gnmi.Watch(t, dut, batch.State(), timeout, func(val *ygnmi.Value[*oc.Root]) bool {
+		root, present := val.Val()
+		if !present {
+			return false
+		}
+		for _, port := range upIntfs {
```
guoshiuan, line 74: I think we can refactor the code to avoid the static analysis warning:

for _, port := range upIntfs {
  upInterfaces[port] = root.GetInterface(port).GetOperStatus() == oc.Interface_OperStatus_UP
  if !upInterfaces[port] {
    return false
  }  
}
return true

mojiiba: addressed the static check issue, slightly diffrent

internal/helpers/helpers.go
```diff
@@ -0,0 +1,93 @@
...
+// FetchOperStatusUPIntfs function uses telemetry to generate a list of all up interfaces.
+// When CheckInterfacesInBinding is set to true, all interfaces that are not defined in binding file are excluded.
+func FetchOperStatusUPIntfs(t *testing.T, dut *ondatra.DUTDevice) []string {
+	t.Helper()
+	intfsOperStatusUP := []string{}
+	intfs := gnmi.GetAll(t, dut, gnmi.OC().InterfaceAny().Name().State())
+	for _, intf := range intfs {
+		if *args.CheckInterfacesInBinding {
+			checkPort := false
+			for _, port := range dut.Ports() {
```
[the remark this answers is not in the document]
mojiiba: done

internal/helpers/helpers.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/openconfig/featureprofiles/internal/args"
+	"github.com/openconfig/ondatra"
+	"github.com/openconfig/ondatra/gnmi"
+	"github.com/openconfig/ondatra/gnmi/oc"
+	"github.com/openconfig/ygnmi/ygnmi"
+)
+
+// FetchOperStatusUPIntfs function uses telemetry to generate a list of all up interfaces.
+// When CheckInterfacesInBinding is set to true, all interfaces that are not defined in binding file are excluded.
+func FetchOperStatusUPIntfs(t *testing.T, dut *ondatra.DUTDevice) []string {
```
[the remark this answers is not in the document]
mojiiba: done

sachendras:
What happens if there are no interfaces in the binding file?

mojiiba:
> What happens if there are no interfaces in the binding file? Should there be a check to ensure there is atleast one interface in the binding file?

thanks for the comment, added check for when there is up interfaces, either because of binding or the connectivity of the device.

LimeHat:
> thanks for the comment, added check for when there is up interfaces, either because of binding or the connectivity of the device.

It isn't the same check as suggested by Sachendra though? With t.Log and only binding-specific ports we're potentially losing a test item.

E.g. in supervisor switchover scenario, if no ports are reserved, this step would be skipped
https://github.com/openconfig/featureprofiles/blob/40126f7080397dcc0a6d4fcf6f84fe0f3d63b16d/feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go#L55

internal/helpers/helpers.go
```diff
@@ -0,0 +1,93 @@
...
+		root, present := val.Val()
+		if !present {
+			return false
+		}
+		for _, port := range upIntfs {
+			if root.GetInterface(port).GetOperStatus() != oc.Interface_OperStatus_UP {
+				upInterfaces[port] = false
+				return false
+			}
+			upInterfaces[port] = true
```
LimeHat, line 80: Is L80 needed?
Seems that the same is done on L68

mojiiba:
> Additionally, with the new flag set to true by default, we're changing the existing behavior (slightly reducing the scope). I'd prefer to keep it as false to preserve the current behavior by default.

setting the flag to false means we prefer the test to be flaky by default for having more coverage that is implicit and dependent to the device base config.  I personally prefer explicit coverage and deterministic behavior rather than having a little more coverage.  I can still set it to false if you insist. @sachendras @guoshiuan @vosipchu  what do u suggest?

internal/helpers/helpers.go
mojiiba: yes, it is still needed since we may get more than one update for an interface. the assignment handle the case when we get down and then up event for an interface.

LimeHat:
> I could not fully follow your comment completely. if no port is up before switchover, we will lose the test item no matter we check from the binding or not (this is even the case in current code). 

We can have three different scenarios:
1) some interfaces are up and subset of those is a part of the binding for the test
2) some interfaces are up but they are not part of the binding for the test
3) no interfaces are up

I believe you're talking about scenario 3 which could happen in theory but quite unlikely in practice; I'm referring to scenario 2

> we should modify the test later to bring up specifics interface and then ensure they remain up after the switchover.

Sure, we can include it as a part of this PR I think? Seems like a simple change

> we prefer the test to be flaky by default for having more coverage that is implicit and dependent to the device base config. I personally prefer explicit coverage and deterministic behavior rather

Could you please elaborate? I think I'm missing the part why we presume flakiness.
E.g. things that could change is a) configuration of adjacent devices that will bring some of the links down, or b) physical topology change (NOC personnel disconnects a link?).
But given the fact that test execution time is relatively short and the test environment is usually semi-isolated or isolated and stable, these are somewhat unlikely?
But I don't have a strong opinion about the default value.

mojiiba:
> > I could not fully follow your comment completely. if no port is up before switchover, we will lose the test item no matter we check from the binding or not (this is even the case in current code).
> 
> We can have three different scenarios:
> 
>     1. some interfaces are up and subset of those is a part of the binding for the test
> 
>     2. some interfaces are up but they are not part of the binding for the test
> 
>     3. no interfaces are up
> 
> 
> I believe you're talking about scenario 3 which could happen in theory but quite unlikely in practice; I'm referring to scenario 2
> 
> > we should modify the test later to bring up specifics interface and then ensure they remain up after the switchover.
> 
> Sure, we can include it as a part of this PR I think? Seems like a simple change
> 
> > we prefer the test to be flaky by default for having more coverage that is implicit and dependent to the device base config. I personally prefer explicit coverage and deterministic behavior rather
> 
> Could you please elaborate? I think I'm missing the part why we presume flakiness. E.g. things that could change is a) configuration of adjacent devices that will bring some of the links down, or b) physical topology change (NOC personnel disconnects a link?). But given the fact that test execution time is relatively short and the test environment is usually semi-isolated or isolated and stable, these are somewhat unlikely? But I don't have a strong opinion about the default value.

For scenario 2, we can deactivate the flag  and can still have the same behavior as before. 

concerning the flakiness your examples are the good ones. There are also other cases which I can not share here.  Also, the test is not really short (reboot line card can take around 10 minutes). Further, tests should not rely on interfaces that are not specified in the binding.

mojiiba:
yes, I fully agree and i will do that in a separate pr.

mojiiba:
> > Further, tests should not rely on interfaces that are not specified in the binding.
> 
> Fair enough. What if we refactor the code to explicitly configure interfaces as you mentioned in the comment above (`for` loop over `dut.Ports()` perhaps) and enforce that we have at least 1 for the supervisor switchover?
> 
> Not sure if this is required for linecard reboots since it's not stated explicitly in the README.

yes, I fully agree and i will do that in a separate pr.

guoshiuan (dismissed):

prinikasn (commented):
/gcbrun

prinikasn (commented):
/gcbrun

internal/helpers/helpers.go
LimeHat: Understood, thx. Perhaps we can eliminate L68 then?

mojiiba: done

guoshiuan (approved):

PR #1346, opened 2023-03-28
rundata add metadata_test and comment about the cwd

liulk:


greg-dennis (commented):
Not sure metadata_test is necessary because it's effectively already tested here:
https://github.com/openconfig/featureprofiles/blob/main/internal/rundata/rundata_test.go#L147

And once the legacy rundata support is removed, I can simplify that test by using a metadata.textproto file in place of writing one dynamically. Do you still feel like this test is necessary and do you feel it would continue to be necessary when this additional simplification is made?

liulk:
Having an end to end test like this checks that `go test` actually chdir to the test package. If they ever change this behavior, I want to know about it through an end to end test.

On the other hand, there is still advantage for the unit test to write the metadata.textproto dynamically, so it can test more scenarios (e.g. graceful handling of parsing failure).

greg-dennis (approved):

PR #1356, opened 2023-03-29
TE-14.1 - Fixing nexthop address for instance vrf3 

cprabha:
Hi,

As per testplan , nexthop address for instance "vrf3" should not be >> tunnelDstIP   = "198.18.208.1" 

Thanks,
Prabha


greg-dennis (approved):
Approving for Go quality only

sachendras (approved):

PR #1358, opened 2023-03-30
Moving deviation flag P4RTMissingDelete to an accessor method

prinikasn:
Moving deviation flag P4RTMissingDelete to an accessor method. Tests which require deviation should call these methods instead of the deviation flags.

### DIFF
--- a/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
+++ b/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
@@ -278,7 +278,7 @@ func canWrite(t *testing.T, args *testArgs) (bool, error) {
 		}
 		return false, writeErr
 	}
-	if !*deviations.P4RTMissingDelete {
+	if !deviations.P4RTMissingDelete(ondatra.DUT(t, "dut")) {
 		if writeErr = writeTableEntry(args, t, pktIO, true); writeErr != nil {
 			t.Errorf("Error deleting table entry (highID %d, lowID %d): %v", args.highID, args.lowID, writeErr)
 		}
--- a/internal/deviations/deviations.go
+++ b/internal/deviations/deviations.go
@@ -69,12 +69,14 @@ import (
 	"github.com/openconfig/ondatra"
 )
 
-// P4RTMissingDelete adds deviation for device that does not support delete mode in P4RT write requests.
-func P4RTMissingDelete(dut *ondatra.DUTDevice) bool {
+// P4RTMissingDelete returns whether the device does not support delete mode in P4RT write requests.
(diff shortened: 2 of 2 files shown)


internal/deviations/deviations.go
```diff
@@ -65,8 +65,15 @@ package deviations
 
 import (
 	"flag"
+
+	"github.com/openconfig/ondatra"
 )
 
+// P4RTMissingDelete adds deviation for device that does not support delete mode in P4RT write requests.
```
greg-dennis, line 72: "adds deviation for device that" -> "returns whether the device"

internal/deviations/deviations.go
```diff
@@ -65,8 +65,15 @@ package deviations
...
+func P4RTMissingDelete(dut *ondatra.DUTDevice) bool {
```
greg-dennis, line 73: I guess let's name the "dut" parameter "_" for now to silence this static analysis

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -278,7 +278,7 @@ func canWrite(t *testing.T, args *testArgs) (bool, error) {
 		}
 		return false, writeErr
 	}
-	if !*deviations.P4RTMissingDelete {
+	if !deviations.P4RTMissingDelete(ondatra.DUT(t, "dut")) {
```
greg-dennis, line 281: Unfortunate that the dut pointer wasn't handy here, but I agree that this is preferable to threading the dut through the call chain, when our goal is to make the deviations go away entirely. In other migrations, let's be sure to reuse the existing dut variable where we can. (Nothing to change here, just FYI.)

greg-dennis (changes requested):

internal/deviations/deviations.go
```diff
@@ -65,8 +65,15 @@ package deviations
...
+	return *p4rtMissingDelete
+}
+
 // Vendor deviation flags.
```
greg-dennis, line 77: Let's also expand this comment with a new paragraph that says all new flags should not be exported (be lowercase) and accessed from tests through a public accessors like those above.

greg-dennis (changes requested):

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
prinikasn: Agree. I did check to see if I could reuse the dut here, but since it was not available went this route. 

Will reuse the existing dut variable whereever I can.

internal/deviations/deviations.go

greg-dennis (approved):

PR #1362, opened 2023-03-30
RT 1.7 adding deviation for last notification error

Wanmathy:
In TestDisconnect, added deviation for last notification error.


internal/deviations/deviations.go
```diff
@@ -217,6 +217,8 @@ var (
 	NetworkInstanceTableDeletionRequired = flag.Bool("deviation_network_instance_table_deletion_required", false,
 		"Set to true for device requiring explicit deletion of network-instance table, default is false")
 
+	MissingBgpLastNotificationErrorCode = flag.Bool("deviation_missing_bgp_last_notification_error_code", false, "Set to true to skip check for bgp/neighbors/neighbor/state/messages/received/last-notification-error-code leaf missing case")
```
prinikasn, line 220: Please use this new format for introducing new deviations: https://github.com/openconfig/featureprofiles/pull/1358.

We will be adding a check for this soon, all new flags should not be exported (define them in lowercase) and accessed from tests through a public accessors (ref in PR 1358)

prinikasn (changes requested):

internal/deviations/deviations.go
```diff
@@ -74,6 +73,11 @@ func P4RTMissingDelete(_ *ondatra.DUTDevice) bool {
 	return *p4rtMissingDelete
 }
 
+// MissingBgpLastNotificationErrorCode returns when the deviation leaf is missing in bgp.
```
greg-dennis, line 76: "when the deviation" -> "whether the last-notification-error-code"

internal/deviations/deviations.go
Wanmathy: Addressed this.

internal/deviations/deviations.go
Wanmathy: Addressed.

feature/bgp/tests/local_bgp_test/local_bgp_test.go
```diff
@@ -232,9 +232,16 @@ func TestDisconnect(t *testing.T) {
 	gnmi.Await(t, dut, nbrPath.SessionState().State(), time.Second*120, oc.Bgp_Neighbor_SessionState_ESTABLISHED)
 	gnmi.Replace(t, ate, ateConfPath.Bgp().Neighbor(dutIP).Enabled().Config(), false)
 	gnmi.Await(t, dut, nbrPath.SessionState().State(), time.Second*120, oc.Bgp_Neighbor_SessionState_ACTIVE)
-	code := gnmi.Get(t, dut, nbrPath.Messages().Received().LastNotificationErrorCode().State())
-	if code != oc.BgpTypes_BGP_ERROR_CODE_CEASE {
-		t.Errorf("On disconnect: expected error code %v, got %v", oc.BgpTypes_BGP_ERROR_CODE_CEASE, code)
+	code := gnmi.Lookup(t, dut, nbrPath.Messages().Received().LastNotificationErrorCode().State())
```
guoshiuan, line 235: If code is not present and deviation flag is not enable, should we emit an error?

Wanmathy: If code is not present, we handle the case using deviation flag where we display leaf not present, and we dont fail the case due to the deviation set.

If code is not present, and deviation not set, then the test would get failed due to error of the last notification error code leaf not present.

Wanmathy: Addressed.

guoshiuan (dismissed):

feature/bgp/tests/local_bgp_test/local_bgp_test.go
```diff
@@ -232,9 +232,18 @@ func TestDisconnect(t *testing.T) {
...
+	if code.IsPresent() {
+		value, _ := code.Val()
+		if value != oc.BgpTypes_BGP_ERROR_CODE_CEASE {
+			t.Errorf("On disconnect: expected error code %v, got %v", oc.BgpTypes_BGP_ERROR_CODE_CEASE, value)
+		}
+	} else {
+		if deviations.MissingBgpLastNotificationErrorCode(ondatra.DUT(t, "dut1")) {
```
prinikasn, line 242: I see we already have access to `dut` (from line 198). Please reuse that variable here instead of calling `ondatra.DUT...`

prinikasn (changes requested):

guoshiuan (approved):

prinikasn (approved):
LGTM from deviation code point of view.

PR #1365, opened 2023-03-30
Update README.md

xw-g:
* The code currently is checking for the gRIBI coredump only.  Update the README to reflect that.
* Add a TODO to check for any unexpected coredump, not just gRIBI.

### DIFF
--- a/feature/experimental/gribi/ate_tests/route_removal_during_failover_test/README.md
+++ b/feature/experimental/gribi/ate_tests/route_removal_during_failover_test/README.md
@@ -26,7 +26,7 @@ Validate gRIBI route flush during SSO
 
 *   Check for gRIBI coredumps in the DUT and validate that none are present post failover.
 
-    TODO: check for any coredumps.
+    TODO: check for any unexpected coredumps.
 
 *   Re-inject IPBlock1 in default VRF with NHGID: `1`.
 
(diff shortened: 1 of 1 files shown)


bstoll:
Shouldn't the statement be to check for any failures?  I don't think there is a way to look for a gribi coredump specifically - an implementation might have a gribi process, or it could be a part of something else and fail in a different way.  Checking GNOI healthz status before & after seems like the correct way for this type of validation.

xw-g:
> Shouldn't the statement be to check for any failures? I don't think there is a way to look for a gribi coredump specifically - an implementation might have a gribi process, or it could be a part of something else and fail in a different way. Checking GNOI healthz status before & after seems like the correct way for this type of validation.

I updated the description. PTAL.

bstoll (approved):
I misunderstood, I thought the checking for core dumps was unimplemented entirely.  SGTM.

rszarecki (approved):

self-maurya:
These changes are no longer required - https://github.com/openconfig/featureprofiles/blob/main/feature/gribi/otg_tests/route_removal_during_failover_test/README.md

All kind of core dumps are already covered as part of https://github.com/openconfig/featureprofiles/tree/main/internal/core internal lib which is enabled in troubleshooting mode.

PR #1373, opened 2023-03-31
TE-2.1 | Add a deviation for MAC override case

arvbaska1:
- Added a deviation for MAC override case
- With this deviation
     -  Configure another dummy IPv4 address on Egress Port2 & Port3
     -  Configure a Static ARP entry [Dummy IP + staticDstMAC]
     -  Change gRIBI programming to include above IP Address alongwith the staticDstMAC


internal/deviations/deviations.go
```diff
@@ -74,6 +74,10 @@ func P4RTMissingDelete(_ *ondatra.DUTDevice) bool {
 	return *p4rtMissingDelete
 }
 
+func GRIBIIpv4EntryMacOverride(_ *ondatra.DUTDevice) bool {
```
greg-dennis, line 77: Go initialism rules say "Ip"->"IP" and "Mac"->"MAC", so

GRIBIIPv4EntryMACOverride and gribiIPv4EntryMACOverride for the function and flag, respectively

arvbaska1: Updated them , thanks

feature/gribi/ate_tests/ipv4_entry_test/ipv4_entry_test.go
```diff
@@ -289,6 +301,28 @@ func TestIPv4Entry(t *testing.T) {
 
 			for _, tc := range cases {
 				t.Run(tc.desc, func(t *testing.T) {
+					//Creating a Static ARP entry for staticDstMAC
+					if deviations.GRIBIIPv4EntryMACOverride(ondatra.DUT(t, "dut")) && tc.desc == "Multiple next-hops with MAC override" {
```
guoshiuan, line 305: Instead of using string comparison, which is hard to maintain, can you create a new bool field in the cast struct? With that the testcase of "Multiple next-hops with MAC override" will be:

		{
			desc: "Multiple next-hops with MAC override",
			entries: []fluent.GRIBIEntry{ ... },
			wantGoodFlows: []*ondatra.Flow{ ... },
			wantOperationResults: []*client.OpResult{...},
                         gribiIPv4EntryMACOverride:  deviations.GRIBIIPv4EntryMACOverride(ondatra.DUT(t, "dut")),
		},

internal/deviations/deviations.go
```diff
@@ -224,4 +229,6 @@ var (
 		"Device skip isis restart-suppress check if value is true, Default value is false")
 
 	MacAddressMissing = flag.Bool("deviation_mac_address_missing", false, "Device does not support /system/mac-address/state.")
+
+	gribiIPv4EntryMACOverride = flag.Bool("deviation_gribi_ipv4_entry_mac_override", false, "Set to true for device not supporting programming a gribi flow with a next-hop entry of mac-address only, default is false")
```
xw-g, line 233: `gribiIMACOverrideWithStaticARP` I think is a better naming.

rszarecki (dismissed):

feature/gribi/ate_tests/ipv4_entry_test/ipv4_entry_test.go
arvbaska1: updated testcase as per this.

internal/deviations/deviations.go
arvbaska1: Changed the deviation name.

guoshiuan (approved):

PR #1375, opened 2023-03-31
P4RT-2.1 -Update deviation for UnsetElectionID with P4RTUnsetElectionIDUnsupported

rsindhujha:
P4RT 2.1 - A new deviation flag is introduced for Juniper (P4RTUnsetElectionIDUnsupported), where we ignore the error when stream disconnect happens for Unset election ID.


LimeHat:
@rsindhujha Would you please elaborate on this? Which error do you expect to see in such a scenario?

rsindhujha:
> @rsindhujha Would you please elaborate on this? Which error do you expect to see in such a scenario?

@LimeHat , UnsetElectionID sends code NotFound which is expected. Hence MasterArbitration Error needs to be ignored.So removed t.Errorf 

https://github.com/openconfig/featureprofiles/blob/6d91b187aff00107cf4da854b622eb30399ca60b/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go#L368

LimeHat:
>  UnsetElectionID sends code NotFound which is expected. 

I agree with the first part.
However, grpc status code is validated on lines L363-L366:
https://github.com/openconfig/featureprofiles/blob/6d91b187aff00107cf4da854b622eb30399ca60b/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go#L362-L370

I.e. my question is what error is expected together with a grpc NotFound status code when we retrieve it via `getRespCode(&test)` call on L363.

rsindhujha:
> > UnsetElectionID sends code NotFound which is expected.
> 
> I agree with the first part. However, grpc status code is validated on lines L363-L366:
> 
> https://github.com/openconfig/featureprofiles/blob/6d91b187aff00107cf4da854b622eb30399ca60b/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go#L362-L370
> 
> I.e. my question is what error is expected together with a grpc NotFound status code when we retrieve it via `getRespCode(&test)` call on L363.

I believe, except for status OK, rest are considered error codes by itself. There isn't any additional error here.

LimeHat:
Let me ask this another way.
As far as I see it, if you just return code 5 (`NOT_FOUND`) in `MasterArbitrationUpdate` message `status`, it will be captured in `resp` variable but not in `err`.

I'm trying to understand where does non-nil `err` comes from in this scenario.

rsindhujha:
> where does non-nil `err` comes from in this scenario.

When the grpc status code is other than 0, the function getRespCode, returns the grpc status code along with error itself. 
Hence the err itself is non-nil here.  

https://github.com/openconfig/featureprofiles/blob/b2929abe9cc0c9d694d93826319ab249ac2ee0ca/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go#L160

LimeHat:
> When the grpc status code is other than 0

Could you please be more specific here, in which field/message?
Thank you,

sgirishj:
> > When the grpc status code is other than 0
> 
> Could you please be more specific here, in which field/message? Thank you,

Line 363:
resp, err := getRespCode(&test)

The function getRespCode processes the ArbitrationResponse error and returns both the error code and the error itself back to the caller. Code snippet below:
			if err := p4rtutils.StreamTermErr(args.handle.StreamTermErr); err != nil {
				if err != nil {
					return int32(status.Code(err)), err
				}

Due to this, 'resp' contains the numeric error code and 'err' contains the error detail. Hence, 'err' wont be nil and that is expected. Previously, the script was erroring out if 'err' wasn't nil. I am changing it to a 'Logf' from an 'Errorf'. For all practical purposes we could delete the block that checks what 'err' contains, given that we have already checked 'resp'.

Hope that clarifies it. Thanks!

LimeHat:
> Hope that clarifies it. Thanks!

I'm sorry, but not really.. can you describe in P4RT terms what message with what fields set do you return that triggers this code path?

As I mentioned a few comments above, "if you just return code 5 (NOT_FOUND) in MasterArbitrationUpdate message status, it will be captured in resp variable but not in err."
On the other hand, if you, for instance, close the stream with the error code -> I'd expect would be captured just as you described.

Do you populate `error` message in `StreamMessageResponse` in this case?

rsindhujha:
> > Hope that clarifies it. Thanks!
> 
> I'm sorry, but not really.. can you describe in P4RT terms what message with what fields set do you return that triggers this code path?
> 
> As I mentioned a few comments above, "if you just return code 5 (NOT_FOUND) in MasterArbitrationUpdate message status, it will be captured in resp variable but not in err." On the other hand, if you, for instance, close the stream with the error code -> I'd expect would be captured just as you described.
> 
> Do you populate `error` message in `StreamMessageResponse` in this case?

Yes, you're right. The stream gets disconnected with unset Election ID, Hence err is not nil and status code comes through as NOT_FOUND.

LimeHat:
Right, but I don't think that should be the case?

@Ankur19 probably can confirm/clarify, but my read of the spec is that you are supposed to allow that connection and just return a status code 5 in masterArbitrationUpdate message, no errors.

I.e. [sections 5.3 && 5.4](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-updates) of the spec just talks about settings `status.code` to `google.rpc.NOT_FOUND`,

>Otherwise, the controller becomes a backup. If the controller was previously a primary (and downgraded itself), then an advisory message is sent to all controllers for this device_id and role. Otherwise, the advisory message is only sent to the controller that sent the initial MasterArbitrationUpdate. See the [following section](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-notification) for the format of the advisory message.

> Otherwise, if there is no primary currently, for all backup controllers, status is set to non-OK (with status.code set to google.rpc.NOT_FOUND).

and section 16.6 specifically says that arbitration errors [should not be reported via StreamError messages](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-stream-error-reporting)

> Note that client arbitration errors are never reported using the StreamError message. Invalid MasterArbitrationUpdate messages sent by the client cause the stream to be terminated and the appropriate error code to be returned to the client immediately. See section [5.3](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-updates).


And this is why, in my opinion, we should be hitting inly this line of getRespCode() fn, which actually extracts the status_code from arbitration update (and not the error code from some other place):
https://github.com/openconfig/featureprofiles/blob/2b2ce7e8ac64300b792ef107b5f00336831a9b52/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go#L167

Ankur19:
> Right, but I don't think that should be the case?
> 
> @Ankur19 probably can confirm/clarify, but my read of the spec is that you are supposed to allow that connection and just return a status code 5 in masterArbitrationUpdate message, no errors.
> 
> I.e. [sections 5.3 && 5.4](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-updates) of the spec just talk about settings `status.code` to `google.rpc.NOT_FOUND`,
> 
> > Otherwise, the controller becomes a backup. If the controller was previously a primary (and downgraded itself), then an advisory message is sent to all controllers for this device_id and role. Otherwise, the advisory message is only sent to the controller that sent the initial MasterArbitrationUpdate. See the [following section](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-notification) for the format of the advisory message.
> 
> > Otherwise, if there is no primary currently, for all backup controllers, status is set to non-OK (with status.code set to google.rpc.NOT_FOUND).
> 
> and section 16.6 specifically says that arbitration errors [should not be reported via StreamError messages](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-stream-error-reporting)
> 
> > Note that client arbitration errors are never reported using the StreamError message. Invalid MasterArbitrationUpdate messages sent by the client cause the stream to be terminated and the appropriate error code to be returned to the client immediately. See section [5.3](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-arbitration-updates).
> 
> And this is why, in my opinion, we should be hitting only this line of getRespCode() fn, which actually extracts the status_code from arbitration update (and not the error code from some other place):
> 
> https://github.com/openconfig/featureprofiles/blob/2b2ce7e8ac64300b792ef107b5f00336831a9b52/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go#L167

Hi! I think I see the issue here.

Currently `getRespCode()` is returning all error codes. Both `StreamTermErr` codes as well as `ArbitrationResponse.Status.Code` codes.

```
// Handle exception when GetCode is empty.
respCode := arbResp.Arb.GetStatus().GetCode()
```

These two errors are very different in the sense that `StreamTermErr` gets generated when stream gets `terminated` and `ArbitrationResponse.Status.Code` gets generated when stream is `not terminated`.

In our scenario the stream doesn't get terminated. So, instead of changing `t.Errorf` to `t.Logf`, here is what I propose:

* Split ``getRespCode()` into two functions: `getStreamTermErr()` and `getArbitrationResponseErr()`.
* Based on the scenario. Validate the correct `Error`. i.e. In our case: for `unset` electionID: ``getStreamTermErr()==null` && `getArbitrationResponseErr()==NOT_FOUND`

LimeHat:
Thanks Ankur. I agree it would be a nice enhancement to split this function into the two.

I don't think it would make a functional change for the "pass" behavior as it is written today, though -> it seems that the origin of this PR is a non-null StreamTermErr (which is not expected and correctly captured with a t.Error call)

Ankur19:
> Thanks Ankur. I agree it would be a nice enhancement to split this function into the two.
> 
> I don't think it would make a functional change for the "pass" behavior as it is written today, though -> it seems that the origin of this PR is a non-null StreamTermErr (which is not expected and correctly captured with a t.Error call)

Right. I think so too. For a correctly behaving device, `err` will be null.

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -365,9 +365,8 @@ func TestUnsetElectionid(t *testing.T) {
 				t.Fatalf("Incorrect status code received: want %d, got %d", test.wantStatus, resp)
 			}
 			if err != nil {
-				t.Errorf("Errors seen in ClientArbitration response: %v", err)
+				t.Logf("Errors seen when sending Master Arbitration as expected for unset ElectionID: %v", err)
```
greg-dennis, line 368: if these errors are expected, should we call t.Error if `err == nil`?

rsindhujha: hi @greg-dennis , this change is not going through. i will undo it.

rsindhujha:
> > Thanks Ankur. I agree it would be a nice enhancement to split this function into the two.
> > I don't think it would make a functional change for the "pass" behavior as it is written today, though -> it seems that the origin of this PR is a non-null StreamTermErr (which is not expected and correctly captured with a t.Error call)
> 
> Right. I think so too. For a correctly behaving device, `err` returned from `getRespCode()` should be null.

hi @Ankur19 ,

Thanks for the detailed reply.
Based on recent discussion, I would add deviation for Juniper for this test. Shall i modify the same git PR or open a new one?

internal/deviations/deviations.go
```diff
@@ -269,6 +274,8 @@ var (
 
 	p4rtMissingDelete = flag.Bool("deviation_p4rt_missing_delete", false, "Device does not support delete mode in P4RT write requests")
 
+	P4rtSkipStreamError = flag.Bool("deviation_p4rt_skip_streamerror", false, "Device closes stream channel for P4RT unset election ID")
```
Ankur19, line 277: Instead rename to `P4RTUnsetElectionIDUnsupported` ?

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -364,10 +364,9 @@ func TestUnsetElectionid(t *testing.T) {
 			if resp != test.wantStatus {
 				t.Fatalf("Incorrect status code received: want %d, got %d", test.wantStatus, resp)
 			}
-			if err != nil {
-				t.Errorf("Errors seen in ClientArbitration response: %v", err)
+			if err != nil && !deviations.P4RTSkipStreamError(ondatra.DUT(t, "dut")) {
```
Ankur19, line 367: Use `dut` object directly `if err != nil && !deviations.P4RTSkipStreamError(dut)`.

sachendras:
Can you please update the description to accommodate the recent change with deviation? Also, please add more context in to why the deviation.

sachendras (dismissed):

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
rsindhujha: Thanks Ankur. I modified it.

rszarecki (approved):

xw-g:
Sorry, I can not tell for sure what's the review status now? @Ankur19 @LimeHat any further concerns?

internal/deviations/deviations.go
```diff
@@ -74,6 +74,11 @@ func P4RTMissingDelete(_ *ondatra.DUTDevice) bool {
 	return *p4rtMissingDelete
 }
 
+// P4RTSkipStreamError returns whether Stream Disconnect for UnsetElectionID needs to be skipped.
```
Ankur19, line 77: Comment is throwing static analysis error.

Also can we unexport the `P4rtUnsetElectionIDUnsupported` flag. We don't want users to be able to use the method as well as the function.

internal/deviations/deviations.go
```diff
@@ -269,6 +274,8 @@ var (
 
 	p4rtMissingDelete = flag.Bool("deviation_p4rt_missing_delete", false, "Device does not support delete mode in P4RT write requests")
 
+	P4rtUnsetElectionIDUnsupported = flag.Bool("deviation_p4rt_unsetelectionid_unsupported", false, "Device closes stream channel for P4RT unset election ID")
```
prinikasn, line 277: Please unexport this variable i.e set variable name as `p4rtUnsetElectionIDUnsupported`

prinikasn (changes requested):

internal/deviations/deviations.go
rsindhujha: oh my bad, i modified the comment and then made the variable as p4rtUnsetElectionIDUnsupported

internal/deviations/deviations.go
rsindhujha: Thanks, this is taken care. Please check.

internal/deviations/deviations.go
```diff
@@ -74,6 +74,11 @@ func P4RTMissingDelete(_ *ondatra.DUTDevice) bool {
 	return *p4rtMissingDelete
 }
 
+// P4rtUnsetElectionIDUnsupported returns whether the device does not support unset election ID.
```
greg-dennis, line 77: please fix the static analysis error. the first word of the comment should be exactly "P4RTUnsetElectionIDUnsupported"

greg-dennis: And please fix "UseVendorNativeACLConfig" while here

greg-dennis (changes requested):

greg-dennis (approved):
Approving for Go quality only

thesrinath (approved):

PR #1381, opened 2023-04-03
Wrr traffic test changes

rszarecki:
This PR modifies WRR weight assigment to queues to values that most vendor should be able to accept:
* Highest weight value is 63 what prevent some vendor form scaling weight down and losing accuracy
* SUM of WRR rates is <100 (27) what prevent some vendor form rounding weight down and losing accuracy

Consequently expected loss-rate per class (PHB) has been also modified.

It this tests suite AF4 and NC1 are served as sptrict priority, hence weights are irrelevant
WRR weights for for AF3, AF2, AF1, BE0 and BE1 with weight 12, 8, 4, 1 and 2 respectively.


saarvind6481:
https://github.com/openconfig/featureprofiles/pull/1242  There is already PR with weights of 64, 16, 4, 1 and
    1. Probably change the highest weight to 63 and it helps to cover 1:4 for few queues

dplore:
FYI: Only issues representing the existence of a test should be part of the project lists.  I'm removing this closed item which is a change to a test from the two projects it's assigned to.

PR #1384, opened 2023-04-04
some of the checks should be skipped for Arista_7280 which is Not a m…

gforeal:
…odular system

Some of the checks in this test is mainly for the modular & chassis styled systems, so they should be skipped in the DUT model like Arista DCS-7280CR3K-32D4

These checks include:
"Fabric"
"Linecard"
"Supervisor"

For Arista DCS-7280CR3K-32D4, They are not in the returned telemetry states.

### DIFF
--- a/feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
+++ b/feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
@@ -451,7 +451,7 @@ func TestComponentParent(t *testing.T) {
 	for _, tc := range cases {
 		t.Run(tc.desc, func(t *testing.T) {
 
-			if len(compList[tc.desc]) == 0 && dut.Model() == "DCS-7280CR3K-32D4" {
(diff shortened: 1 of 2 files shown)


rszarecki:
I don't think we should use deviation for this purpose. Instead I would gate it by `dut.Model() == "model"`. 
Deviations are signals that sooner or later there would be fix needed. here is nothing to fix. Component just do not exist and nevwer will be existing.

gforeal:
actually that is exactly what the commit #1 - the initial commit about which conditioned on the dut.Model();
then I thought the popular deviation way maybe preferable by the reviewers in the community as it used does....
so I pushed in the commit #2 which is the current deviation one.
yes, I can just rollback to the original commit #1 if it satisfy the most!

thank for review & comment!

gforeal:
all reviewers, would you please update & continue the review - give your weight on which option you may still prefer after we have chatting talked the alternatives like the args....thanks!

Ankur19 (approved):

rszarecki (approved):

PR #1388, opened 2023-04-04
p4rt-2.1: fix usage of wantfail flag 

LimeHat:
This PR fixes the usage of wantFail flag for establishing p4rt stream channels; previously it was skipping error validation in some cases where it wasn't intended. 

---
<sup>


jjing1122:
filed internel bug b/277268421 for presubmit process

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -171,7 +170,7 @@ func getRespCode(args *testArgs) (int32, error) {
 }
 
 // getGDPParameter returns GDP related parameters for testPacketIn testcase.
-func getGDPParameter(t *testing.T) PacketIO {
+func getGDPParameter(_ *testing.T) PacketIO {
```
greg-dennis, line 173: Can you just remove the parameter entirely and while you're at it, drop the word "get" from the method name to conform to go style? The name "gdpParameter()" is fine. There appears to only be one caller.

LimeHat: I think we can remove this func altogether, moved it inline.

jjing1122 (approved):

PR #1390, opened 2023-04-05
RT-5.4: Update ATE config order to start LACP before IP

trathod1:
In this PR, updating ATE configuration to start LACP first for LAG and than configuring IP on LAG to have ARP/ND resolution done before traffic. This avoids momentary traffic loss.


feature/interface/aggregate/ate_tests/aggregate_forwarding_viable_test/aggregate_forwarding_viable_test.go
```diff
@@ -348,15 +348,16 @@ func (tc *testArgs) configureATE(t *testing.T) {
 	if is100gfr {
 		agg.Ethernet().FEC().WithEnabled(false)
 	}
+	tc.top.Push(t).StartProtocols(t)
```
xw-g, line 351: Any reason we start the protocol here, but then update the topology and restart again? 

Is this line really needed? Thanks.

LimeHat: @xw-g we observe that in some cases arp/nd adjacencies are not fully resolved when lacp and ip protocols are started together on ATE. This may cause a small packet loss on the first stream (since DUT has to query ATE for arp/nd info after receiving packets from a non-lag intf destined to ATE LAG intf).
This sequence allows to resolve this (first establish LACP, then add ipv4/v6 that will allow to resolve arp/nd info immediately.

Note we don't restart the protocols here, just add ipv4/v6 to the lag on top of the existing topology.

xw-g:
b/278081847 is the internal presubmit

PR #1391, opened 2023-04-05
Update wrr_traffic_test.go 

saarvind6481:
This PR is for changing weights for cisco in accordance with https://github.com/openconfig/featureprofiles/pull/1242/

### DIFF
--- a/feature/experimental/qos/ate_tests/wrr_traffic_test/wrr_traffic_test.go
+++ b/feature/experimental/qos/ate_tests/wrr_traffic_test/wrr_traffic_test.go
@@ -758,73 +758,71 @@ func TestWrrTraffic(t *testing.T) {
 		trafficDuration time.Duration
 	}
 
-	var cases []test
-
-	cases = []test{
+	cases := []test{
 		{
 			desc:            "Non-oversubscription AF3 and AF2 traffic",
 			trafficFlows:    nonOversubscribedTrafficFlows1,
-			trafficDuration: 20 * time.Second,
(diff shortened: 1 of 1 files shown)


feature/experimental/qos/ate_tests/wrr_traffic_test/wrr_traffic_test.go
```diff
@@ -1714,7 +1509,7 @@ func ConfigureCiscoQos(t *testing.T, dut *ondatra.DUTDevice) {
 		targetGrpoup string
 	}{{
 		desc:         "scheduler-policy-BE1",
-		sequence:     uint32(6),
+		sequence:     uint32(1),
```
guoshiuan, line 1512: Can we define human-friendly constants for these magic seq numbers: 0, 1, 4, 16?

saarvind6481: We can have , but general convention followed is for priority queues we use zero and one for non priority queues

guoshiuan (approved):

rszarecki (approved):

feature/experimental/qos/ate_tests/wrr_traffic_test/wrr_traffic_test.go
rszarecki: The OpenConfig MANDATES that sequence 0 is served 1st. And only when queue is empty (or PIR is achived) secheduler 1 is served. And so on.

So it is not convention. It's "standard".

PR #1399, opened 2023-04-06
Fix P4RT-5.2: Traceroute Packetout test

prakashbadri-arista:
The commit fixes the following issues in the Traceroute packetout test
- Modify the test to use router MAC for dst MAC in the packet. For the traceroute packet to be forwarded correctly, the dst MAC should be set to the router/system MAC instead of a random value
- Set the TTL to be 2 instead of 1 in the traceroute packet so that packet does not get resent to the CPU and instead gets forwarded correctly to the ATE.

Please note the following changes in the test only causes the ipv4 test to pass and not the ipv6 test with Arista. There is a pending fix in EOS that has been validated to fix the ipv6 traceroute packetout test. @


Ankur19 (approved):
LGTM!

Ankur19 (approved):

sachendras:
Hello @Ankur19 did we ensure that this PR will successfully run for all NOS? Google doesnt require to query the router mac address through gNMI.Get() before performing a traceroute. This seems to be a requirement for just one implementation. If so, why is this not a deviation? Since merging of this PR deviates from the original requirements shared w/ different implementations, this PR has now pushed a breaking change for other implementations. Hence I insist that this PR is rolled back and replaced w/ a deviation for the implementation that needs it.

sachendras:
@prakashbadri-arista, I request you to create a deviation instead for this change. This deviation then should be tracked via a Google vendor bug to track long term fix. @rszarecki for vis.

prakashbadri-arista:
@sachendras What is the exact regression here with the other vendor ? I would like to avoid introducing a deviation if we could find a test input that works for everyone. The test currently uses a random dmac for the traceroute packet. I was assuming using the router mac should not break another implementation.

PR #1414, opened 2023-04-12
removed delete mode for ipv4_entry_test

manan-patel:
For WBB Delete mode is not required so removing delete mode from tests


prinikasn (approved):
LGTM from test run prespective - also waiting on TL approvals

prinikasn (commented):
/gcbrun

PR #1239, opened 2023-03-09
Updated test step for ATE port-2 and DUT port-2 config with subinterface indices

mohanasm:
Updated test step for ATE port-2 and DUT port-2 config with subinterface indices. This is duplicate of PR#1086. Raised this PR as 1086 is having CLA issue.

PR #1417, opened 2023-04-12
Te 3.3 uprev

manan-patel:
PR is based on a request to enhance TE3.3 based on https://github.com/openconfig/featureprofiles/pull/1133 and https://github.com/openconfig/featureprofiles/pull/1239


feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
```diff
@@ -13,24 +13,22 @@ Configure ATE and DUT:
 
 *   Create a non-default VRF (VRF-1) that includes DUT port-1.
 
-*   On DUT port-2 and ATE port-2 create 18 L3 sub-interfaces each with a /30
-    subnet as below:
+*   For DUT port-2 interface, create a subinterface with Index 0 and IPv4 address 192.0.2.5, not in a VLAN.
```
sachendras, line 16: Why are we adding this change? Why cant this be starting from unit1 and vlanid 1?

Also, the PR is missing descriptions

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
```diff
@@ -86,28 +84,29 @@ WCMP width of 16 nexthops:
...
+    *   NH100: (31/32) * (1/241) = 0.402% traffic received by ATE port-2 VLAN 3
+
     *   for each VLAN ID in 3...18:
 
-        *   NH: (31/32) * (1/16) ~ 6.05% traffic received by ATE port-2 VLAN ID
+        *   NH: (31/32) * (16/241) = 6.432% traffic received by ATE port-2 VLAN ID
 
-    *   A deviation of 0.5% is allowed for each VLAN for now, since we only test
-        for 2 mins.
+    *   A deviation of 0.2% is allowed for each VLAN for now.
```
sachendras, line 109: Why is this updated to 0.2%? can we instead make this 1% to replicate the current test config?

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
vbhan-cisco: @sachendras ,
   We will add description, This PR is based on request to enhance this test based on https://github.com/openconfig/featureprofiles/pull/1133 . I see there were changes made in readme after PR 1133 was published. Can you and  @xw-g pls confirm  the readme to implement ?

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
sachendras: @xw-g, this is based on our conversation in b/270468786. Infact, imo there shouldnt be a deviation and we should just increase the tolerance to 1.5%. @xw-g, thoughts?

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
sachendras: @xw-g is there a need for the unit IDs to start from "0"? The updates were made in https://github.com/openconfig/featureprofiles/pull/1239 to ensure all implementations pass. 
As such, there is no business requirement to support VLANs. VLANs exist only for the tests to not carry complicated TE configs.

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
vosipchu: 0.2% came from Xiao (not sure about the math behind though). It's not a deviation  in the common sense we use here, more like just a deviation (tolerance as you called it) allowed to pass (as i read this)
Btw, why 1.5%, why not 15%? :)

sachendras: 0.2% came from Xiao (not sure about the math behind though)

- Well, the point I am trying to make is, even though it was originally defined in the readme as 0.5%, the code is looking for 1%. [Link to the main] (https://github.com/openconfig/featureprofiles/blob/main/feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go#L63). So my comment was, why not update the Readme to be 1% instead and not 0.2%. If this came from Xiao, is this in a bug, can you please point me there?

I am quoting the original bug discussion here for clarity and context. So by reducing this to 0.2%, we will add flakiness to the test results.

====
Hello Xiao

Need some clarification on TE-3.3 test. Currently the test uses only [256 flows](https://github.com/openconfig/featureprofiles/blob/main/feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go#L361). This being a hashing test, we either need to define more flows in the order of ~1000 or, we need to relax the pass thresholds which are [currently defined as 1%](https://github.com/openconfig/featureprofiles/blob/main/feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go#L63).

Why: Even with WCMP, the DUT relies on hashing to load share between different egress interfaces. When there are less number of flows and a very tight deviation to meet (1% in this case) between the different egress interfaces, there are high possibilities for the test results to be flaky. Please also note: The test is using only a single source address and 256 destination addresses. These are native IP flows with no payload. Hence, IXIA uses the payload protocol for IP headers to be constant as protocol 61. So there is not much of an entropy available to get proper hashing. Not sure what is the business requirement here. Any idea?

In the meantime, proposal is to:
Update test to generate ~1024 flows by changing destination subnet to /22.
====

xw-g:
@manan-patel can you help update the description, and also mention https://github.com/openconfig/featureprofiles/pull/1133? (feel free to close https://github.com/openconfig/featureprofiles/pull/1133). 

Also, can you help solve the merge conflict?

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
xw-g: Note that my updates of test plan(https://github.com/openconfig/featureprofiles/pull/1133) does not include this. 

@manan-patel I think might be because this PR base does not include #1239. Can you help rebase or merge from the main to get the latest head? Thanks.

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
```diff
@@ -13,24 +13,22 @@ Configure ATE and DUT:
...
 *   Create a non-default VRF (VRF-1) that includes DUT port-1.
 
-*   On DUT port-2 and ATE port-2 create 18 L3 sub-interfaces each with a /30
-    subnet as below:
+*   For DUT port-2 interface, create a subinterface with Index 0 and IPv4 address 192.0.2.5, not in a VLAN.
 
-    *   On DUT port-2, create subinterfaces with indices 1 to 18 mapped to VLAN
-        IDs 1 to 18 and corressponding IPv4 addresses 192.0.2.5, 192.0.2.9, ...,
-        192.0.2.73 respectively.
+*   For ATE port-2, create a subinterface with IPv4 address 192.0.2.6 and Default Gateway of 192.0.2.5, not in a VLAN.
```
xw-g, line 18: same comment as above about #1239

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
```diff
@@ -86,28 +84,29 @@ WCMP width of 16 nexthops:
...
+        *   NH#101 with weight:16, pointing to 192.0.2.22
 
         *   ...
 
-        *   NH#115 with weight:1, pointing to 192.0.2.79
+        *   NH#115 with weight:16, pointing to 192.0.2.79
 
 *   Validate with traffic:
 
     *   NH10: (1/32) * (3/8) = 1.171% traffic received by ATE port-2 VLAN 1
```
xw-g, line 99: my PR #1133 says `~`, but here is `=`, any reason why?

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
```diff
@@ -86,28 +84,29 @@ WCMP width of 16 nexthops:
...
 
     *   NH11: (1/32) * (5/8) = 1.953% traffic received by ATE port-2 VLAN 2
 
+    *   NH100: (31/32) * (1/241) = 0.402% traffic received by ATE port-2 VLAN 3
```
xw-g, line 103: same thing `~`

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
xw-g: * I'm ok with change the deviation to tolerance to avoid confusion.
* The 0.2% current is a arbitrary number. However, it does have some reasoning. (31/32) * (1/241) ~ 0.402% is the the lowest percentage we want to validate. So, we prefer the tolerance lowered that than. So, we start with 0.2% and we can revisit this number as we learn more from operation.
* @sachendras our b/270468786 discussion stands. I think the point is: if the DUT has challenges meeting the 0.2%, let's create a deviation flag or make the 0.2% value from an input arguments (this probably is better). If it's about that the test does not generate enough flows, then let's modify the code to generate more flows with enough hashing.  
* @sachendras @vosipchu for this test, I think we can start with generating more IP flows, say 254 x 254 = 64K (from more sources and more destinations). Later, as we have more clarification on hashing requirement, we can come back to add more parameters (e.g. IPinIP, etc). WDYT?

xw-g:
@manan-patel can you help ping me when the PR are updated? Thanks.

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
sachendras: "The 0.2% current is a arbitrary number. However, it does have some reasoning. (31/32) * (1/241) ~ 0.402% is the the lowest percentage we want to validate. So, we prefer the tolerance lowered that than. So, we start with 0.2% and we can revisit this number as we learn more from operation."

So based on the calculations and the weight definitions, we will be expecting 0.402% traffic on an interface at max sometimes. If so, looks like deviation_ucmp_traffic_tolerance w/ 1.5% tolerance of the total ingress traffic will pass the test even w/ 0% traffic on an interface, isnt it, or may be I am wrong here? :) Just a thought, making iterative progress towards the actual working number would be difficult. Anyhow, LGTM on the deviation.

- However, can we please update this PR for 254 x 254 = 64K flows and have the threshold set at 0.3% instead in this PR. It is still well within the requirement of ~ 0.402%. Additional PRs, just add toil to everyone :). @xw-g but I will leave it on u for 0.2% v/s 0.3%. But adding number of flows will definitely help in this PR.

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
manan-patel: done

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
manan-patel: Left it unchanged: 
for basic hierarchical weight, tolerance is 0.5
and hierarchical weight in boundary scenarios is 0.2

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
manan-patel: done

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
manan-patel: done

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
manan-patel: done

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
sachendras: @manan-patel can you please add 254 x 254 flows?

manan-patel:
@xw-g 

Addressed comments,

1: added src prefix entry, start, and end info for flow design
2: added 65k flows with (65k src - 65k dst mapping)
3: updated policy to match dscp bit (1)
4: updated flow to send traffic with dscp bit (1)

NOTE: 
case 1: testBasicHierarchicalWeight, matches hardcoded value 0.5 tolerance as suggested in README file, need to add deviation if it fails for any vendor (0.5 tolerance works for CISCO, it doesn't need a deviation)

case 2: testHierarchicalWeightBoundaryScenario, matches 0.2 tolerance for other vendors as suggested in README file and has a deviation of 1.5% which applies to CISCO

prinikasn (commented):
/gcbrun

feature/gribi/ate_tests/hierarchical_weight_resolution_test/hierarchical_weight_resolution_test.go
```diff
@@ -522,7 +530,7 @@ func testBasicHierarchicalWeight(ctx context.Context, t *testing.T, dut *ondatra
 	}
 	t.Run("testTraffic", func(t *testing.T) {
 		got := testTraffic(t, ate, top)
-		if diff := cmp.Diff(wantWeights, got, cmpopts.EquateApprox(0, deviation)); diff != "" {
+		if diff := cmp.Diff(wantWeights, got, cmpopts.EquateApprox(0, 0.5)); diff != "" {
```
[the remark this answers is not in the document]
manan-patel: Yes, we should set 0.2 for both cases now we have multiple flows.

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
```diff
@@ -11,7 +11,7 @@ Configure ATE and DUT:
 
 *   Connect ATE port-1 to DUT port-1. ATE port-2 to DUT port-2.
 
-*   Create a non-default VRF (VRF-1) that contains no interfaces.
+*   Create a non-default VRF (VRF-1) that includes DUT port-1.
```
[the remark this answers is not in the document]
xw-g: > Got it, addressed it in next PR

You already have a PR for it?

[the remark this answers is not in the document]
manan-patel: Its addressed in here

manan-patel:
@xw-g Correct, it would be one flow in the terms of outer destination IP, but 64K flows in the terms of inner source and destination IP

Since CLA since is not passing for some reason, I raised another PR with same changes https://github.com/openconfig/featureprofiles/pull/1444

xw-g:
> #1444

Thanks. 

* If #1417 and #1414 are the same, maybe we should just close this one for clarity? 
* nit: please help resolve conversations that you have already addressed. This helps reviews quickly follow up on the next round of review.

feature/gribi/ate_tests/hierarchical_weight_resolution_test/README.md
manan-patel: Added one flow in the terms of outer destination IP, but 64K (254x254) flows in the terms of inner source and destination IP.

manan-patel:
@xw-g, resolved all the comments. Closing the PR due to failing CLA check. 
Please review PR https://github.com/openconfig/featureprofiles/pull/1444

PR #1422, opened 2023-04-14
RT-2.1, RT-2.2 | Enable IS-IS wide metric-style by default without the deviation

arvbaska1:
Made below changes for the RT-2.1, RT-2.2 IS-IS testcases:

1) Enabled IS-IS wide metric-style by default without the deviation in the internal session.go .
Since the peer ATE is enabled with wide metric style, DUT needs to be enabled with wide metric style as well and it should not be with a deviation.

2) Removed IsisAfMetricStyleWideLevelRequired from deviations.go .


prinikasn (commented):
\gcbrun

prinikasn (approved):

PR #1428, opened 2023-04-15
P4RT: update P4RTNodesByPort to support breakout ports

kjahed:
This PR updates P4RTNodesByPort and the P4RT subtests of gNMI-1.10 to allow multiple ports to map to the same hardware-port. It also remove inferP4RTNodeName for CISCO.


guoshiuan:
b/278763487 has the internal test results.

guoshiuan (approved):

PR #1448, opened 2023-04-24
Create README.md for RT-5.6: Interface Loopback mode test

nagarushirao:
Test to verify Interface mode can be set to loopback mode and can be added as part of static LAG


sachendras (approved):

feature/interface/loopback/ate_tests/loopback_aggregate_test/README.md
```diff
@@ -0,0 +1,65 @@
...
+*   Add port-1 as part of Static LAG (lacp mode static(on))
+*   Validate that port-1 operational status is “UP”
+*   Validate that  LAG port status is “UP”
+
+### TestCase-2:
+
+*   Configure DUT port-1
+*   Configure “loopback mode” set to ”TERMINAL”
+*   Add port-1 as part of Static LAG (lacp mode static(on))
+*   Validate that port-1 operational status is “UP”
```
rszarecki, line 22: As per OpaneConfig: 
"
TERMINAL
A loopback which directs traffic received from an external source on the port back out the transmit side of the same port.
"
This loopback mode (in contrast fo FACILITY) do not define status on local port, hence I would remove this TC at all in order to avoid ambiguity.

feature/interface/loopback/ate_tests/loopback_aggregate_test/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Summary
+
+Ensure Interface mode can be set to loopback mode and can be added as part of static LAG.
+
+## Procedure
+
+### TestCase-1:
+
+*   Configure DUT port-1
```
rszarecki, line 11: At this point, We should ensure that we use port that is not connected to remote system, and therefore interface is in DOWN STATE.

nagarushirao: Updated test steps

feature/interface/loopback/ate_tests/loopback_aggregate_test/README.md
nagarushirao: for completeness of test, we need this test. moved to TBD as discussed offline with Rafal.

nagarushirao: also, removed TestCase-3 as its covered already in singleton tests

nagarushirao:
Rafal, PTAL

nagarushirao:
@dplore @rszarecki @sachendras @xw-g @vosipchu  PTAL

rszarecki (approved):

PR #1460, opened 2023-04-26
Fix ipv4_entry_test "Multiple next-hops with MAC override" testlet failure in EOS 

prakashbadri-arista:
In EOS, gnmi.Update to an interface primary address when there is already an ip address configured for that interface will fail.  
The following is the exact failure log:
"""
    ipv4_entry_test.go:311: Update(t) on dut(cmp304) /interfaces/interface[name=Ethernet3/5/1]: Update(t) at path origin:"openconfig" elem:{name:"interfaces"} elem:{name:"interface" key:{key:"name" value:"Ethernet3/5/1"}}: rpc error: code = InvalidArgument desc = error on request {
        prefix: {
          target: "cmp304"
        }
        update: {
          path: {
            origin: "openconfig"
            elem: {
              name: "interfaces"
            }
            elem: {
              name: "interface"
              key: {
                key: "name"
                value: "Ethernet3/5/1"
              }
            }
          }
          val: {
            json_ietf_val: "{\n  \"openconfig-interfaces:config\": {\n    \"description\": \"DUT Port 2\",\n    \"enabled\": true,\n    \"name\": \"Ethernet3/5/1\",\n    \"type\": \"iana-if-type:ethernetCsmacd\"\n  },\n  \"openconfig-interfaces:name\": \"Ethernet3/5/1\",\n  \"openconfig-interfaces:subinterfaces\": {\n    \"subinterface\": [\n      {\n        \"config\": {\n          \"index\": 0\n        },\n        \"index\": 0,\n        \"openconfig-if-ip:ipv4\": {\n          \"addresses\": {\n            \"address\": [\n              {\n                \"config\": {\n                  \"ip\": \"192.0.2.21\",\n                  \"prefix-length\": 30\n                },\n                \"ip\": \"192.0.2.21\"\n              }\n            ]\n          },\n          \"config\": {\n            \"enabled\": true\n          }\n        }\n      }\n    ]\n  }\n}"
          }
        }
        }: failed to apply: one and only one primary IPv4 address can be specified

"""

The ipaddress that was being added was meant to be a secondary ip address prefix meant to cover the gribi nexthop. In EOS, we cannot configure secondary IP address through OC and as a result adding the GRIBIMACOverrideStaticARPStaticRoute deviation to ipv4_entry_test so that a static route to the gribi nh aft entries as well the arp entry is created to make nexthops resolvable.


feature/gribi/ate_tests/ipv4_entry_test/ipv4_entry_test.go
```diff
@@ -304,8 +304,8 @@ func TestIPv4Entry(t *testing.T) {
 						d := gnmi.OC()
 						p2 := dut.Port(t, "port2")
 						p3 := dut.Port(t, "port3")
-						gnmi.Update(t, dut, d.Interface(p2.Name()).Config(), dutPort2DummyIP.NewOCInterface(p2.Name()))
-						gnmi.Update(t, dut, d.Interface(p3.Name()).Config(), dutPort3DummyIP.NewOCInterface(p3.Name()))
+						gnmi.Replace(t, dut, d.Interface(p2.Name()).Config(), dutPort2DummyIP.NewOCInterface(p2.Name()))
+						gnmi.Replace(t, dut, d.Interface(p3.Name()).Config(), dutPort3DummyIP.NewOCInterface(p3.Name()))
```
Ankur19, line 308: Hi @prakashbadri-arista This is not correct. With this change, for Arista we are basically updating the primary IP of the Interface itself. 

This would make the test wrong.

The idea behind using gNMI.Update() here is so that a `secondary` ip address is added to the interface, which can then be used in gRIBI next hop.

Instead we should be adding a Magic IP and interfaces in 
`/network-instances/network-instance/protocols/protocol/static-routes/static/next-hops/next-hop/interface-ref`

prakashbadri-arista: Hi Ankur,
I have updated the review accordingly. Please take a look.

Ankur19 (approved):

PR #1461, opened 2023-04-27
bumping up gosnappi version

bkreddy143:
OTG added support for EgressTracking as part of new release and need to bump the gosnappi version to use that.

This code is a Contribution to the OpenConfig Feature Profiles project ("Work") made under the Google Software Grant and Corporate Contributor License Agreement ("CLA") and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia's intellectual property are granted for any other purpose. This code is provided on an as is basis without any warranties of any kind.


jasdeep-hundal:
We'll test with the latest Gosnappi and update the Ondatra version simultaneously.

PR #1462, opened 2023-04-27
RT-1.5 BGP prefixlimit |  Fixes for overlimit case & remove deviation

arvbaska1:
Corrected below issues For Overlimit case & removed the deviation:

- The trafficDuration variable value was calculated inside the if condition tc.name == "OverLimit", but that cannot be used with sendtraffic function outside it, so added it inside the if condition.

- The duration for which traffic should be sent in the overlimit case should be
Restart time (trigger after max pfx limit is reached) - Elapsed time

- Removed BGPPrefixOverlimit from the deviations.go


prinikasn (commented):
/gcbrun

prinikasn (approved):

PR #1463, opened 2023-04-27
Changes to RT-1.4 & remove deviation

arvbaska1:
Made below changes to remove deviation

- Added Time duration unit for ate.Actions().NewBGPGracefulRestart() Action which was missing.

-For VerifyBGPNOTEstablished BGP state verification, changed the check != Established instead of check = Active/Connect
 BGP goes to IDLE with restart at ATE/apply Deny ACL, and in the BGP FSM from the IDLE state it can go to Connect or Active , & from Connect to Active or IDLE.

- Removed deviation BGPStateActiveACLDeny from deviations.go.


guoshiuan (approved):
b/280083311 has the internal test results.

PR #1466, opened 2023-04-27
gNMI-1.15 skip for other vendors

liulk:


dplore (commented):
Please add a TODO (perhaps in the in the readme?) to address adding in coverage for other dut.Vendor.

dplore (approved):

PR #1467, opened 2023-04-28
Add PR workflow action for requiring OTG changes

jasdeep-hundal:


.github/workflows/pull_request.yml
```diff
@@ -166,3 +166,17 @@ jobs:
       run: go install honnef.co/go/tools/cmd/staticcheck@latest
     - name: Run staticcheck
       run: GOGC=30 staticcheck ./...
+
+  otg_changes:
+    permissions: write-all
```
bstoll, line 171: Were these two lines necessary (write, write-all)? I wasn't seeing where this action uses any special permissions.

bstoll (dismissed):

.github/workflows/pull_request.yml
jasdeep-hundal: Thanks for the catch. They are not needed. (Was testing to see if we could modify the PR reviewer set, missed removing these.)

bstoll (approved):

PR #1484, opened 2023-05-02
initial experimental_management_cli Feature Profile

bensons:


dplore (commented):
I think we should make telnet separate, as telnet is an anti-requirement in any modern network?

IMO, should name the folders:
feature/system/ssh/
feature/system/telnet/

While a bit more granular that I would want at first glance, the need to keep telnet separate (as a non-requirement/anti-requirement) I think needs this.  

This follows precedent for system features and aligns to OC path structure which is the intent for other folders holding feature.proto.  You can think of the folders as:

feature/<feature-name/oc-toplevel-model-name>/<subfeature-name/oc-major-container-name>

dplore:
Closing this stale request

PR #1487, opened 2023-05-02
Fix P4RT-2.1: P4RT Election test failure for EOS

prakashbadri-arista:
In EOS, P4RT-2.1: P4RT Election test was failing for the following reasons
1 - EOS allows 0 election id as valid for primary arbitration messages. A deviation has been added to the test to expect different status code in TestUnsetElectionid. Also the test has been updated to allow SetForwardingPipelineConfig from a primary with election id of 0

2 - EOS always sends ALREADY_EXISTS in status code to secondary clients in
    the response to primaryAribitrationMessage when the election id
    is lesser than the current/previous primary election id. According
    to P4RT specification, ALREADY_EXISTS status is set when primary is
    present and NOT_FOUND is set when there is no primary. I am adding a
    deviation to expect ALREADY_EXISTS instead of NOT_FOUND status until
    EOS behavior is updated.

The test needs to be run with the following deviations with EOS:
-deviation_interface_enabled
-deviation_p4rt_missing_delete
-deviation_p4rt_zero_election_id_allowed
-deviation_bkup_arbitration_resp_code

### DIFF
--- a/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
+++ b/feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
@@ -339,6 +339,15 @@ func TestUnsetElectionid(t *testing.T) {
 			wantStatus: int32(codes.NotFound),
 		},
 	}
+	if *deviations.P4rtZeroElectionIdAllowed {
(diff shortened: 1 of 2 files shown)


prakashbadri-arista:
closing this one and opened a new PR based on the recent changes

PR #1458, opened 2023-04-26
initial security/aaa/local Feature Profile only

bensons:

PR #1500, opened 2023-05-02
add admin-username config to system feature

dplore:


feature/system/feature.textproto
```diff
@@ -111,6 +111,9 @@ config_path {
 telemetry_path {
     path: "/system/aaa/authentication/admin-user/state/admin-password-hashed"
 }
+config_path {
+    path: "/system/aaa/authentication/admin-user/config/admin-username"
```
bensons, line 115: Interestingly, this path doesn't exist. I think that admin-username is RO because it's not expected to be configurable, e.g. it's "root" or "admin" or something like that.

In order to configure an admin user with a unique username, I think the expectation is to create an regular (named) user account and give it an admin role using something like /system/aaa/authentication/users/user/config/role = SYSTEM_ROLE_ADMIN

(I've submitted [a PR](https://github.com/openconfig/featureprofiles/pull/1458) to create experimental_security_aaa_local Feature Profile with the relevant paths for regular user creation.)

bensons (changes requested):

PR #1501, opened 2023-05-03
Fix P4RT-2.1: P4RT Election test failure for EOS

prakashbadri-arista:
In EOS, P4RT-2.1: P4RT Election test was failing for the following reasons
1 - EOS allows 0 election id as valid for being elected as a primary
    client . A deviation (P4rtUnsetElectionIDPrimaryAllowed) has been added to the test to expect different status code
  in TestUnsetElectionid. Also the test has been updated to allow
  SetForwardingPipelineConfig from a primary with election id of 0

  2 - EOS always sends ALREADY_EXISTS in status code to secondary
    clients in the response to primaryAribitrationMessage when the election id
    is lesser than the current/previous primary election id. According
    to P4RT specification, ALREADY_EXISTS status is set when primary is
    present and NOT_FOUND is set when there is no primary. I am adding a
    deviation to expect ALREADY_EXISTS instead of NOT_FOUND status until
    EOS behavior is updated.

    The test needs to be run with the following deviations with EOS:
    -deviation_interface_enabled
    -deviation_p4rt_missing_delete
    --deviation_p4rt_unsetelectionid_primary_allowed
    -deviation_bkup_arbitration_resp_code


internal/deviations/deviations.go
```diff
@@ -452,6 +452,10 @@ var (
 
 	p4rtUnsetElectionIDUnsupported = flag.Bool("deviation_p4rt_unsetelectionid_unsupported", false, "Device does not support unset Election ID")
 
+	P4rtUnsetElectionIDPrimaryAllowed = flag.Bool("deviation_p4rt_unsetelectionid_primary_allowed", false, "Device allows unset Election ID to be primary")
```
prinikasn, line 455: Please use the new style guide to  add deviations i.e by using accessor methods and making this parameter private (lowercase)

prinikasn (changes requested):

internal/deviations/deviations.go
prinikasn: Refer to https://github.com/openconfig/featureprofiles/pull/1397/files for example

feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
```diff
@@ -670,6 +696,14 @@ func TestArbitrationUpdate(t *testing.T) {
 		// After updating electionID, statusCode also changes
 		// to secondary without a primary
 		test.wantStatus = 5
+		if *deviations.P4rtBackupArbitrationResponseCode {
+			// Change the expected status code to ALREADY_EXISTS for deviant implementations
+			// that send ALREADY_EXISTS instead of NOT_FOUND to secondary clients when there
+			// is no primary
+			test.wantStatus = 6
```
Ankur19, line 703: nit: can you change the hardcoded numbers to something like: `int32(codes.ALREADY_EXISTS)` and `int32(codes.NOT_FOUND)`

rszarecki (approved):

rszarecki (approved):

rszarecki (approved):

rszarecki (approved):

rszarecki (approved):

prinikasn (approved):
LGTM for deviation style guide

PR #1502, opened 2023-05-03
Add acl defined-sets profile

dplore:
This adds ACL paths needed to build a more complete set of ACL's than previously covered.


robshakir (commented):
Does this need the actual `/defined-sets` paths included -- or is that covered by a different feature profile?

It seems like they are likely coupled based on the fact that you probably can't test `defined-sets` without something referencing them, but I also see why they might be a separate feature given that they might be referenced from different contexts.

bensons:
> Does this need the actual `/defined-sets` paths included -- or is that covered by a different feature profile?
> 
> It seems like they are likely coupled based on the fact that you probably can't test `defined-sets` without something referencing them, but I also see why they might be a separate feature given that they might be referenced from different contexts.

@dplore -- Bumping this comment from @robshakir for attention.

Looking just now, I don't see defined-sets used anywhere in the ACL model. But it's possible that I'm missing something?

dplore:
> > Does this need the actual `/defined-sets` paths included -- or is that covered by a different feature profile?
> > It seems like they are likely coupled based on the fact that you probably can't test `defined-sets` without something referencing them, but I also see why they might be a separate feature given that they might be referenced from different contexts.
> 
> @dplore -- Bumping this comment from @robshakir for attention.
> 
> Looking just now, I don't see defined-sets used anywhere in the ACL model. But it's possible that I'm missing something?

@robshakir @bensons the scope for this PR changed to no longer include reference to defined-sets.  I have updated the title and description for this PR to reflect this.  

Since we haven't scoped test development to cover [defined-sets](https://github.com/openconfig/public/blob/master/release/models/defined-sets/openconfig-defined-sets.yang) yet I have opened https://github.com/openconfig/featureprofiles/issues/1883 to track.

bensons (approved):

PR #1505, opened 2023-05-03
Check that the testbed is specified in the metadata

greg-dennis:
PiperOrigin-RevId: 528000536

### DIFF
--- a/tools/addrundata/case.go
+++ b/tools/addrundata/case.go
@@ -67,7 +67,7 @@ func (tc *testcase) check() []error {
 	var errs []error
 
 	if tc.existing == nil {
-		errs = append(errs, errors.New("existing rundata is missing"))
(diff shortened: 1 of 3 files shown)


bstoll:
Can we update featureprofiles/tools/addrundata/README.md with the testbed requirements?  Previously the addrundata was completely automatic, but now it defaults to assigning a TESTBED_DUT_ATE_2LINKS.  I think this is OK but we should document that this default isn't always correct.

The CONTRIBUTING.md guide also references rundata_test.go that should be metadata.textproto now  (sort of unrelated to this...):
https://github.com/openconfig/featureprofiles/blob/bf1f9df03755b914f1f33351faed0fcfd6b49e47/CONTRIBUTING.md?plain=1#L135-L136

bstoll (approved):

PR #1511, opened 2023-05-03
gNMI-1.4: Telemetry Inventory Test - runtime improvements

sgirishj:
This Pull Req is bringing in run time improvements to gNMI-1.4 (Telemetry Inventory Test) as detailed below:
- Function 'findComponentsListByType' does a gNMI Subscribe for 'All Components', orders the components by type and returns the names of the components (ordered by type)
- Testcases TestHardwarecards and TestSwitchChip obtain the component names from the above, pass the names on to the function 'ValidateComponentState' which subsequently runs multiple gNMI Subscribe operations for verifying the data under each component
- Depending on the number of components of a given type present on the DUT, the number of gNMI operations could end up being quite high leading to increased script run time

Optimization proposed in this Pull Req:
- Modify the function 'findComponentsListByType' to return the ordered list of components with all information intact and as received from the 'gNMI Subscribe for All Components' call (instead of returning just the component names ordered by type)
- Modify test cases TestHardwareCards and TestSwitchChip to use the data as is from the returned Components data structure; pass the Components data set as is to the function 'ValidateComponentState' and verify the necessary fields

Verification steps involving fetching parent nodes or walking up the Component tree from a given node will still continue to execute the necessary gNMI Subscribe operations as needed.


greg-dennis (approved):
Approving for Go quality only

sachendras (approved):

PR #1522, opened 2023-05-04
Move IsisSingleTopologyRequired and ISISMultiTopologyUnsupported to accessor methods

prinikasn:
Move IsisSingleTopologyRequired and ISISMultiTopologyUnsupported to accessor methods

### DIFF
--- a/internal/deviations/deviations.go
+++ b/internal/deviations/deviations.go
@@ -69,6 +69,13 @@ import (
 	"github.com/openconfig/ondatra"
 )
 
+// AggregateAtomicUpdate returns if device requires that aggregate Port-Channel and its members be defined in a single gNMI Update transaction at /interfaces.
+// Otherwise lag-type will be dropped, and no member can be added to the aggregate.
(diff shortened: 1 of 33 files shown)


greg-dennis (approved):

internal/deviations/deviations.go
```diff
@@ -156,6 +156,16 @@ func ISISLevelAuthenticationNotRequired(_ *ondatra.DUTDevice) bool {
 	return *isisLevelAuthenticationNotRequired
 }
 
+// IsisSingleTopologyRequired sets isis af ipv6 single topology on the device if value is true.
+func IsisSingleTopologyRequired(_ *ondatra.DUTDevice) bool {
```
greg-dennis, line 160: I know this wasn't the case with the flag, but Go style says this needs to start with ISIS

greg-dennis (changes requested):

internal/deviations/deviations.go

PR #1536, opened 2023-05-05
TE-3.3: Updating load balance tolerance

mohanasm:
Observing some flakiness in loadbalancing in TE-3.3 testcase testHierarchicalWeightBoundaryScenario. Added args 'arg_hierarchical_weight_boundary_resolution_tolerance' to specify the tolerance


greg-dennis:
Why is this an arg and not a deviation?

sachendras:
> Why is this an arg and not a deviation?

sachendras:
> 

This was based on an internal discussion with xiaowa. https://screenshot.googleplex.com/46yyuZU2v8F7R6s has the production requirement. Arg will allow vendors to pass what works for them as long as it is within the requirements. Not a deviation because, 0.2% is not a production requirement.

greg-dennis:
This looks too messy to me. Are you sure we need an arg _and_ a deviation? And it seems to me the test should be testing what our production requirement is; if it isn't 0.2, then we shouldn't be testing 0.2 by default.

sachendras:
> 
agreed, we should update the test. There is nothing we can do as it ends up being hashing. The test already uses 64k flows.

@xw-g for viz

xw-g:
To clarify (and I totally agree with @greg-dennis) , that we should try to avoid using deviation and args together for the same thing.

I also think that there are use cases where we probably want to use args, instead of deviations. Scaling test is one example, different expectations might exists on different platforms/generations, and we probably want to adjust the test expectations accordingly.

Now come to this specific case in this PR, it's not yet confirmed if 0.2% is totally an unreasonable requirement to vendors, and it will take some time to clarify. So, I suggested maybe args here gives us more flexibility. But thinking it again just now, it also catches me that using args does not enforce the tracking (to follow up on the 0.2% value).

So, probably deviation is a good choice for now (from tracking's perspective). @greg-dennis thoughts?

rszarecki (approved):

xw-g:
Actually, since there is already a deviation (`HierarchicalWeightResolutionTolerance`), I think we can just close this PR, and use that deviation.

sachendras:
After offline discussions with @xw-g, I am closing this PR. We will use deviation HierarchicalWeightResolutionTolerance with desired value. Thanks

PR #1550, opened 2023-05-10
TE -13.1 /13.2 - Remove deviation

cprabha:
Hi,

This PR is created for the below changes.

1.) Removed deviation flag from TE-13.1 and TE13.2
2.) Added logf instead of fatalf during concurrent switchover/gribi route add. Since error is expected sometimes if switchover is triggered before all routes are added.

Thanks,
Prabha


sachendras (dismissed):

greg-dennis (approved):
Approving for Go quality only

jasdeep-hundal (approved):

PR #1551, opened 2023-05-10
Add disk and cpu to telemetry

Ankur19:
We are adding Hardware components CPU and STORAGE to the telemetry test, which are required but were missing in the original test.

The paths we are checking include: MfgName, PartNo, SerialNo.

While it is not required for STORAGE and CPU components to have their own PartNo and SerialNo, we do expect that, if we look at their parent component (or parent of parent component), we would eventually reach the CHASSIS component which must have a PartNo and SerialNo.


feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -426,228 +443,234 @@ func TestTempSensor(t *testing.T) {
...
+				t.Logf("Component %s PartNo: %v", cName, partNo)
+				if partNo == nil {
+					switch card.GetType() {
+					case componentType["Fan"]:
+						fallthrough
+					case componentType["Storage"]:
+						fallthrough
+					case componentType["Cpu"]:
+						parent := card.GetParent()
+						ppn, ok := gnmi.Lookup(t, dut, gnmi.OC().Component(parent).PartNo().State()).Val()
```
aaronmillisor, line 550: In the description of the PR it says:
"While it is not required for STORAGE and CPU components to have their own PartNo and SerialNo, we do expect that, if we look at their parent component (or parent of parent component), we would eventually reach the CHASSIS component which must have a PartNo and SerialNo."

However in the implementation it seems to be making assumptions as to which parts do or don't contain a part number. I think this should be refactored to detect if the serial/part-number exists, and if not search all parents until an ancestor is found that contains these fields. 

This means that perhaps you end up with the linecard, or perhaps the chassis, but you are also most likely to find the component which you would need to RMA.

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -426,228 +443,234 @@ func TestTempSensor(t *testing.T) {
...
+				t.Logf("Component %s SerialNo: %v", cName, serialNo)
+				if serialNo == nil {
+					switch card.GetType() {
+					case componentType["Fan"]:
+						fallthrough
+					case componentType["Storage"]:
+						fallthrough
+					case componentType["Cpu"]:
+						parent := card.GetParent()
+						psn, ok := gnmi.Lookup(t, dut, gnmi.OC().Component(parent).SerialNo().State()).Val()
```
aaronmillisor, line 572: Same comment as above. Also, unless I'm missing something, the PR also seems to be expecting that the immediate parent does contain a serial number/part number, no? What if the parent is a motherboard that is not included in the inventory? The same solution should suffice here where you may end up searching until you find the part number (i.e. maybe to the linecard, without an assumption as to the specific vendor hierarchy)

aaronmillisor (changes requested):
A few comments on assumptions, PTAL and let me know if I'm misunderstanding the logic.

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
AnandJyrm: We should stop at first parent which is removable for serial number also.

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
Ankur19: Right, currently this code doesn't check for the first FRU. e.g a CPU's direct parent may not be a FRU, a grandparent however might be. But the current code will fail in such a scenario since it will pass only when the direct parent is a FRU.

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
Ankur19: Done!

feature/platform/ate_tests/telemetry_inventory_test/telemetry_inventory_test.go
Ankur19: Done!

rszarecki (approved):

PR #1558, opened 2023-05-11
NTP server: place the new server into the same network-instance as these already exist in the device if any

gforeal:
Some vendors restricts the new server to be in the same network-instance as these already exist in the device

Currently the feature/system/ntp/tests/system_ntp_test/server_test is failing with on those vendors since the mgmt_vrf been enabled;

When mgmt vrf is enabled, the new adding NTP server(s) has to be in the same network-instance as these already configured in the device which are in the management vrf, regardless of IPv4 or IPv6 server address;
If no network-instance been specified, DUT will reject the Replace call.


bstoll:
This change is obsoleted by #1579.

PR #1560, opened 2023-05-11
Add a utility for computing the root path to the featureprofiles repo.

greg-dennis:
PiperOrigin-RevId: 531237118

### DIFF
--- a/internal/pathutil/pathutil.go
+++ b/internal/pathutil/pathutil.go
@@ -1,0 +1,41 @@
+// Package pathutil provides utilities for finding test-relative paths at runtime.
+package pathutil
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+)
+
+var (
+	mu       sync.Mutex
+	rootPath string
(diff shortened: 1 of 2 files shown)


internal/pathutil/pathutil.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+var (
+	mu       sync.Mutex
+	rootPath string
+)
+
+// RootPath returns an absolute path to the root "featureprofiles" directory.
+func RootPath() (string, error) {
+	mu.Lock()
```
Ankur19, line 18: Is RootPath() expected to be called by multiple go routines?

greg-dennis: I don't have any such anticipated use cases, but since it was cached I figured "why not" ... I can remove if you prefer

Ankur19: Its should be fine.

Ankur19 (approved):

PR #1564, opened 2023-05-11
Add deviation proto to metadata.proto

prinikasn:
PiperOrigin-RevId: 531281333


proto/metadata.proto
```diff
@@ -36,4 +38,238 @@ message Metadata {
   }
   // Testbed on which the test is intended to run.
   Testbed testbed = 4;
+
+  message PlatformExceptions {
+    message Platform {
```
greg-dennis, line 43: proto messages don't have to be nested just because they're single use. It's sometimes recommended to not nest them unless they are conceptually bound to the parent. In this case, it sort of feels right to me to move Platform and Deviations out to the top-level, because they are concepts that can stand independently. But to keep PlatformExceptions nested, since it is sort of used as a kind of glorified map field. Make sense to you?

greg-dennis (changes requested):
Please also run `make proto/metadata_go_proto/metadata.pb.go` in the root git directory to generate the Go code for it. See here:
http://google3/third_party/openconfig/featureprofiles/Makefile;l=28;rcl=525312988

proto/metadata.proto

PR #1569, opened 2023-05-12
Move MissingValueForDefaults to accessor

prinikasn:
PiperOrigin-RevId: 529717954


prinikasn:
OTG failure tracked in b/281542380

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
```diff
@@ -43,17 +43,17 @@ func TestMain(m *testing.M) {
 // deviation is set, in which case it uses check.EqualOrNil to allow the device
 // to return a nil value. This should only be used when `val` is the default
 // for this particular query.
-func EqualToDefault[T any](query ygnmi.SingletonQuery[T], val T) check.Validator {
-	if *deviations.MissingValueForDefaults {
+func EqualToDefault[T any](query ygnmi.SingletonQuery[T], val T, dut *ondatra.DUTDevice) check.Validator {
```
greg-dennis, line 46: Sorry that this creates more work for you, but I think these helpers would be better to not pass the full DUT in if we only need the value of a deviation. Could you change EqualToDefault and CheckPresence instead accept a boolean called missingDefaults and pass that in when called?

greg-dennis (changes requested):

feature/experimental/isis/ate_tests/base_adjacencies_test/base_adjacencies_test.go
prinikasn: done

greg-dennis (approved):

PR #1577, opened 2023-05-15
Add validation of software version

Ankur19:
Add validation of system/state/software-version


kjahed:
@Ankur19 Cisco does not support /system/state/software-version yet. Can we guard that request with
deviation_sw_version_unsupported ?

Ankur19:
> @Ankur19 Cisco does not support /system/state/software-version yet. Can we guard that request with deviation_sw_version_unsupported ?

Done!

PR #1578, opened 2023-05-15
Move GetSubcomponentPath to accessor method.

prinikasn:
PiperOrigin-RevId: 529713463
Move GetSubcomponentPath to accessor method.

### DIFF
--- a/internal/components/components.go
+++ b/internal/components/components.go
@@ -91,8 +91,8 @@ func FindMatchingStrings(components []string, r *regexp.Regexp) []string {
 }
 
 // GetSubcomponentPath creates a gNMI path based on the componnent name.
-func GetSubcomponentPath(name string) *tpb.Path {
-	if *deviations.GNOISubcomponentPath {
+func GetSubcomponentPath(name string, dut *ondatra.DUTDevice) *tpb.Path {
+	if deviations.GNOISubcomponentPath(dut) {
 		return &tpb.Path{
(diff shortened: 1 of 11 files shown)


internal/components/components.go
```diff
@@ -91,8 +91,8 @@ func FindMatchingStrings(components []string, r *regexp.Regexp) []string {
 }
 
 // GetSubcomponentPath creates a gNMI path based on the componnent name.
```
greg-dennis, line 93: can you fix the spelling of "component" here

internal/components/components.go
```diff
@@ -91,8 +91,8 @@ func FindMatchingStrings(components []string, r *regexp.Regexp) []string {
...
-func GetSubcomponentPath(name string) *tpb.Path {
-	if *deviations.GNOISubcomponentPath {
+func GetSubcomponentPath(name string, dut *ondatra.DUTDevice) *tpb.Path {
```
greg-dennis, line 94: Could you change the second argument to a boolean named "useNameOnly" and add a line to the godoc for this function that says something like "If useNameOnly is true, returns a path to the specified name instead of a full subcomponent path."

greg-dennis (changes requested):

internal/components/components.go
prinikasn: done. And moved the deviation. GNOISubcomponentPath call in the test file.

greg-dennis (approved):

PR #1631, opened 2023-05-15
Fp test power down up

Ankur19:
New test to check power-admin-state leaf


bstoll:
This isn't required, but it would be nice if the test could log the amount of time required for each transition from ACTIVE>DISABLED and DISABLED>ACTIVE, or some kind of log line with a timestamp after each await.  This test could take a while to run on a fully populated chassis and it might help for some future troubleshooting.

Ankur19:
> This isn't required, but it would be nice if the test could log the amount of time required for each transition from ACTIVE>DISABLED and DISABLED>ACTIVE, or some kind of log line with a timestamp after each await. This test could take a while to run on a fully populated chassis and it might help for some future troubleshooting.

Done!

rszarecki (approved):

PR #1638, opened 2023-05-16
RT-3.2: added check for ARP resolution from OTG

bkreddy143:
Added check for ARP resolution before traffic start 

This code is a Contribution to the OpenConfig Feature Profiles project ("Work") made under the Google Software Grant and Corporate Contributor License Agreement ("CLA") and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia's intellectual property are granted for any other purpose. This code is provided on an as is basis without any warranties of any kind


feature/experimental/policy/policy_vrf_selection/otg_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
```diff
@@ -332,6 +344,7 @@ func testTrafficFlows(t *testing.T, args *testArgs, expectPass bool, flows ...go
 	args.ate.OTG().StartProtocols(t)
 
 	t.Logf("*** Starting traffic ...")
+	waitOTGARPEntry(t)
```
jasdeep-hundal, line 347: Can use the existing function in otgutils instead. Example at :https://github.com/openconfig/featureprofiles/blob/bca2afa2a567c170b9b25e92acfa5e16c222cfd1/feature/experimental/backup_nhg/otg_tests/backup_nhg_test/backup_nhg_test.go#L251

jasdeep-hundal (changes requested):

feature/experimental/policy/policy_vrf_selection/otg_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
jasdeep-hundal: I would also move this right after StartProtocols/before the start traffic log entry, the logging will be more clear that way.

bkreddy143: updated with existing fucntion.

jasdeep-hundal (approved):

PR #1649, opened 2023-05-22
Add traceroute deviation to metadata textproto file

prinikasn:
PiperOrigin-RevId: 533597321
This PR does the following:

Adds the deviations deviation_traceroute_l4_protocol_udp and deviation_traceroute_fragmentation to the metadata textproto for the test that calls them
Removes these deviations from the vars sections.

### DIFF
--- a/internal/deviations/deviations.go
+++ b/internal/deviations/deviations.go
@@ -75,9 +75,11 @@ import (
 // lookupDutDeviations returns the deviations for the specified dut. If no duts match `nil` is returned.
 func lookupDUTDeviations(dut *ondatra.DUTDevice) *mpb.Metadata_Deviations {
 	for _, platformExceptions := range metadata.Get().PlatformExceptions {
-		vendor := platformExceptions.GetPlatform().Vendor.String()
+		if dut.Device.Vendor().String() != platformExceptions.GetPlatform().Vendor.String() {
+			continue
+		}
 		for _, hardwareModel := range platformExceptions.GetPlatform().HardwareModel {
-			if vendor == dut.Device.Vendor().String() && hardwareModel == dut.Device.Model() {
+			if dut.Device.Model() == hardwareModel {
 				return platformExceptions.GetDeviations()
 			}
 		}
@@ -87,9 +89,6 @@ func lookupDUTDeviations(dut *ondatra.DUTDevice) *mpb.Metadata_Deviations {
 }
 
 func logErrorIfFlagSet(name string) {
-	if !flag.Parsed() {
-		flag.Parse()
-	}
 	flag.Visit(func(f *flag.Flag) {
 		if f.Name == name {
 			log.Errorf("Value for %v is set using metadata.textproto. Flag value will be ignored!", name)


greg-dennis (changes requested):

internal/deviations/deviations.go
```diff
@@ -66,9 +66,23 @@ package deviations
...
 	"flag"
 
+	log "github.com/golang/glog"
+	"github.com/openconfig/featureprofiles/internal/metadata"
+	mpb "github.com/openconfig/featureprofiles/proto/metadata_go_proto"
 	"github.com/openconfig/ondatra"
 )
 
+// resolveDutDeviations returns the deviations for the specified dut. If no duts match `nil` is returned.
+func resolveDutDeviations(dut *ondatra.DUTDevice) *mpb.Metadata_Deviations {
```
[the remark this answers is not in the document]

feature/gnoi/system/tests/traceroute_test/metadata.textproto
```diff
@@ -5,3 +5,23 @@ uuid: "5384dcdb-c427-4b35-94d3-d5f571960634"
...
+platform_exceptions: {
+  platform: {
+    vendor: CISCO
+    hardware_model: 'CISCO-8202-32FH-M'
+  }
+  deviations: {
+    traceroute_fragmentation: true
+    traceroute_l4_protocol_udp: true
+  }
+}
```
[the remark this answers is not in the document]
prinikasn: done

internal/deviations/deviations.go
```diff
@@ -66,9 +66,23 @@ package deviations
...
 	"flag"
 
+	log "github.com/golang/glog"
+	"github.com/openconfig/featureprofiles/internal/metadata"
+	mpb "github.com/openconfig/featureprofiles/proto/metadata_go_proto"
 	"github.com/openconfig/ondatra"
 )
 
+// resolveDutDeviations returns the deviations for the specified dut. If no duts match `nil` is returned.
+func resolveDUTDeviations(dut *ondatra.DUTDevice) *mpb.Metadata_Deviations {
```
greg-dennis, line 76: Sorry, the verb "resolve" is bothering me a bit here, because I'm not sure there's much resolution going on. I think "lookup" may be more appropriate?

proto/metadata.proto
```diff
@@ -50,6 +50,12 @@ message Metadata {
     // Device does not support interface/ipv4/enabled,
     // so suppress configuring this leaf.
     bool ipv4_missing_enabled = 1;
+    // Device does not support fragmentation bit for traceroute.
+    bool traceroute_fragmentation = 2;
+    // Device only support UDP as l4 protocol for traceroute.
+    // Use this field to set default l4 protocol as UDP and skip the tests that
```
greg-dennis, line 56: How about we delete this line and the next. I think simple, declarative specifications are easier to maintain, and I don't think we should commit ourselves to consistently documenting what is skipped by enabling a deviation.

greg-dennis (changes requested):

internal/deviations/deviations.go

proto/metadata.proto

greg-dennis (dismissed):

greg-dennis (changes requested):
As discussed offline, let's instead log an error if the accessor is called and its flag is set.

internal/deviations/deviations.go
```diff
@@ -66,9 +66,34 @@ package deviations
...
+		}
+	}
+	log.Warningf("No platform exceptions for dut %v configured in platform exceptions metadata %v", dut, metadata.Get().PlatformExceptions)
+	return nil
+}
+
+// isFlagSetByUser returns if the specified flag value is set.
+func isFlagSetByUser(name string) bool {
+	found := false
+	flag.VisitAll(func(f *flag.Flag) {
```
greg-dennis, line 89: I think you need Visit, not VisitAll? Visit only visits flags that are set by the user

internal/deviations/deviations.go
```diff
@@ -66,9 +66,34 @@ package deviations
...
+			return platformExceptions.GetDeviations()
+		}
+	}
+	log.Warningf("No platform exceptions for dut %v configured in platform exceptions metadata %v", dut, metadata.Get().PlatformExceptions)
+	return nil
+}
+
+// isFlagSetByUser returns if the specified flag value is set.
+func isFlagSetByUser(name string) bool {
+	found := false
```
greg-dennis, line 88: delete this boolean. instead "return true" in place of "found = true" and "return false" in place of "return found"

internal/deviations/deviations.go
```diff
@@ -285,13 +310,23 @@ func MissingValueForDefaults(_ *ondatra.DUTDevice) bool {
 }
 
 // TraceRouteL4ProtocolUDP returns if device only support UDP as l4 protocol for traceroute.
-func TraceRouteL4ProtocolUDP(_ *ondatra.DUTDevice) bool {
-	return *traceRouteL4ProtocolUDP
+// Default value is false.
+func TraceRouteL4ProtocolUDP(dut *ondatra.DUTDevice) bool {
+	deviationFlag := "deviation_traceroute_l4_protocol_udp"
+	if isFlagSetByUser(deviationFlag) {
+		log.Errorf("Value for %v is set using metadata.textproto. Flag value will be ignored!!!", deviationFlag)
```
greg-dennis, line 317: one exclamation point is sufficient :)

internal/deviations/deviations.go
```diff
@@ -285,13 +310,23 @@ func MissingValueForDefaults(_ *ondatra.DUTDevice) bool {
...
+	}
+	return lookupDUTDeviations(dut).GetTracerouteL4ProtocolUdp()
```
greg-dennis, line 319: what do you think about passing in the name of the flag to "lookupDUTDeviations" so that the check doesn't need to be repeated in each accessor?

greg-dennis (changes requested):

internal/deviations/deviations.go
prinikasn: Updated this function to directly log the error instead of returning a bool.

internal/deviations/deviations.go
prinikasn: lookupDUTDeviations will also be called by new deviations which would be added before we complete this migration. So, I thought we could keep this function seperate. 

I did move these lines to log the error to a seperate function `logErrorForFlag`. This function would only be called for those deviations which have a flag value and are migrated to use the metadata.textproto value. 

Let me know if that sounds ok.

greg-dennis: Sounds good.

internal/deviations/deviations.go
```diff
@@ -66,9 +66,32 @@ package deviations
...
+		if platformExceptions.GetPlatform().Vendor.String() == dut.Device.Vendor().String() && platformExceptions.GetPlatform().HardwareModel == dut.Device.Model() {
+			return platformExceptions.GetDeviations()
+		}
+	}
+	log.Warningf("No platform exceptions for dut %v configured in platform exceptions metadata %v", dut, metadata.Get().PlatformExceptions)
+	return nil
+}
+
+// logErrorForFlag logs an error if specified flag is set.
+func logErrorForFlag(name string) {
```
greg-dennis, line 87: maybe a more descriptive name would be "logErrorIfFlagSet" and then I think you could delete the godoc comment, because the name would be self-explanatory.

feature/gnoi/system/tests/traceroute_test/metadata.textproto
```diff
@@ -5,3 +5,44 @@ uuid: "5384dcdb-c427-4b35-94d3-d5f571960634"
...
+  platform: {
+    vendor: CISCO
+    hardware_model: 'xrd'
+  }
+  deviations: {
+    traceroute_fragmentation: true
+    traceroute_l4_protocol_udp: true
+  }
+}
+
```
greg-dennis, line 48: looks like one extra stray line at the end here

greg-dennis (changes requested):

internal/deviations/deviations.go

feature/gnoi/system/tests/traceroute_test/metadata.textproto
prinikasn: done

internal/deviations/deviations.go

internal/deviations/deviations.go
```diff
@@ -66,9 +66,37 @@ package deviations
...
+	"github.com/openconfig/featureprofiles/internal/metadata"
+	mpb "github.com/openconfig/featureprofiles/proto/metadata_go_proto"
 	"github.com/openconfig/ondatra"
 )
 
+// lookupDutDeviations returns the deviations for the specified dut. If no duts match `nil` is returned.
+func lookupDUTDeviations(dut *ondatra.DUTDevice) *mpb.Metadata_Deviations {
+	for _, platformExceptions := range metadata.Get().PlatformExceptions {
+		vendor := platformExceptions.GetPlatform().Vendor.String()
+		for _, hardwareModel := range platformExceptions.GetPlatform().HardwareModel {
```
greg-dennis, line 79: Check the vendor first, because then you can avoid the whole loop

internal/deviations/deviations.go
```diff
@@ -66,9 +66,37 @@ package deviations
...
+				return platformExceptions.GetDeviations()
+			}
+		}
+	}
+	log.Warningf("No platform exceptions for dut platform %v or model %v configured in platform exceptions metadata %v", dut.Device.Vendor().String(), dut.Device.Model(), metadata.Get().PlatformExceptions)
+	return nil
+}
+
+func logErrorIfFlagSet(name string) {
+	if !flag.Parsed() {
```
greg-dennis, line 90: I think you can assume flags have been parsed by this point and remove this if statement. There is no way to reach this point without flag.Parse being called.

greg-dennis (changes requested):

internal/deviations/deviations.go
prinikasn: done

internal/deviations/deviations.go

greg-dennis (approved):

Issue #1655, opened 2023-05-25
New tests to verify behaviour of forwarding-viable in case ALL members of LAG are not viable (FALSE)


rszarecki:
We need to verify that in case all members of LAG are not vialble, this LAG is still able to accept incoming traffic and IP routed (ISIS, iBGP) traffic is rerouted to alternative path.

rszarecki:
DRAFT pull request has procedure description
https://github.com/openconfig/featureprofiles/pull/1657

dplore:
Removing from the project "Test Dev Group 2" as forwarding viable is not a feature requirement for this project

PR #1656, opened 2023-05-25
Rszarecki/issue1655

rszarecki:
New test procedure description case when all LAG memeber are set with forwarding-viable=FALSE.
ref: issue1655


dplore:
Removing this from project "Test Dev Group 2" as the forwarding viable feature is not a requirement for that project.

dplore:
Removed from Test Dev Group 1 because this item is closed (abandoned) and it is a duplicate of https://github.com/openconfig/featureprofiles/issues/1655

PR #1658, opened 2023-05-25
Make hardware model in deviations proto a repeated field

prinikasn:


proto/metadata.proto
```diff
@@ -43,7 +43,7 @@ message Metadata {
     // Vendor of the device.
     ondatra.Device.Vendor vendor = 1;
     // Hardware model of the device.
```
greg-dennis, line 45: update the comment, too. just "model" -> "models"?

greg-dennis (changes requested):

proto/metadata.proto

greg-dennis (approved):

PR #1668, opened 2023-05-26
move `deviation_hierarchical_weight_resolution_tolerance` to metadata textproto

prinikasn:


internal/deviations/deviations.go
```diff
@@ -277,8 +277,13 @@ func SwVersionUnsupported(_ *ondatra.DUTDevice) bool {
 }
 
 // HierarchicalWeightResolutionTolerance returns the allowed tolerance for BGP traffic flow while comparing for pass or fail conditions.
-func HierarchicalWeightResolutionTolerance(_ *ondatra.DUTDevice) float64 {
-	return *hierarchicalWeightResolutionTolerance
+// Default minimum value is 0.2. Anything less than 0.2 will be set to 0.2.
+func HierarchicalWeightResolutionTolerance(dut *ondatra.DUTDevice) float64 {
+	logErrorIfFlagSet("deviation_hierarchical_weight_resolution_tolerance")
```
greg-dennis, line 282: I think we need to avoid referencing deviation fields directly, because that will cause a nil deference if there are no deviations for the DUT. We should also only call `lookupDUTDeviations(dut)` once in any accessor. Maybe:

```
logErrorIfFlagSet("deviation_hierarchical_weight_resolution_tolerance")
hwrt := lookupDUTDeviations(dut).GetHierarchicalWeightResolutionTolerance()
if minHWRT := 0.2; hwrt < minHWRT {
	return minHWRT
}
return hwrt
```

proto/metadata.proto
```diff
@@ -56,6 +56,9 @@ message Metadata {
     bool traceroute_l4_protocol_udp = 3;
     // Device does not support bgp/neighbors/neighbor/afi-safis/afi-safi/state/prefixes/received-pre-policy.
     bool prepolicy_received_routes = 4;
+    // Expected ucmp traffic tolerance. Minimum value is 0.2, anything lesser
```
greg-dennis, line 59: lesser -> less

greg-dennis (changes requested):

greg-dennis (approved):

PR #1670, opened 2023-05-28
Update qos_policy_config_test.go

saarvind6481:
cisco-dp-1.2-uprev


saarvind6481:
will open a new PR as the directory has changed

PR #1672, opened 2023-05-28
Update qos_basic_test.go

saarvind6481:
cisco-dp-1.14-uprev


feature/qos/ate_tests/qos_basic_test/qos_basic_test.go
```diff
@@ -397,23 +397,15 @@ func TestBasicConfigWithTraffic(t *testing.T) {
...
-					"dutQosDroppedOctetsAfterTraffic",
-				}
-			} else {
-				counterNames = []string{
-
-					"ateOutPkts", "ateInPkts", "dutQosPktsBeforeTraffic", "dutQosOctetsBeforeTraffic",
-					"dutQosPktsAfterTraffic", "dutQosOctetsAfterTraffic", "dutQosDroppedPktsBeforeTraffic",
-					"dutQosDroppedPktsAfterTraffic",
-				}
+			counterNames = []string{
```
prinikasn, line 401: Fix static check failure - `should merge variable declaration with assignment on next line `

feature/qos/ate_tests/qos_basic_test/qos_basic_test.go
```diff
@@ -952,147 +944,194 @@ func ConfigureCiscoQos(t *testing.T, dut *ondatra.DUTDevice) {
...
-		}
-		fwdgroups := q.GetOrCreateForwardingGroup(tc.targetGroup)
-		fwdgroups.Name = ygot.String(tc.targetGroup)
-		fwdgroups.OutputQueue = ygot.String(tc.targetGroup)
-		gnmi.Replace(t, dut, gnmi.OC().Qos().Config(), q)
+			term.SetId(tc.termID)
+			action := term.GetOrCreateActions()
+			action.SetTargetGroup(tc.targetGroup)
+			condition := term.GetOrCreateConditions()
+			if tc.name == "dscp_based_classifier_ipv4" {
```
prinikasn, line 1125: check for type instead of name? same previous PRs

prinikasn (changes requested):
Please make the applicable changes to tests under otg_tests/folder too

feature/qos/ate_tests/qos_basic_test/qos_basic_test.go
saarvind6481: Fixed it

feature/qos/ate_tests/qos_basic_test/qos_basic_test.go
saarvind6481: The original tests function which ConfigureQoS had similar way to differentiate between v4 and v6 conditions was with comparing with string.
condition := term.GetOrCreateConditions()
if tc.name == "dscp_based_classifier_ipv4" {
condition.GetOrCreateIpv4().SetDscpSet(tc.dscpSet)
} else if tc.name == "dscp_based_classifier_ipv6" {
condition.GetOrCreateIpv6().SetDscpSet(tc.dscpSet)
}

But since in the previous releases , we did not support different policy-map name for V4 or v6 under same , I used type to modify the condition. Since in this release , we support the same, changed the test back how the original test was written.
Please let me know if we still have to change the test by using the type to derive the condition , I can do the same

saarvind6481: fixed it

prinikasn (approved):

guoshiuan:
b/286408229 has the internal test results.

PR #1673, opened 2023-05-28
Update one_sp_queue_traffic_test.go

saarvind6481:


feature/qos/ate_tests/one_sp_queue_traffic_test/one_sp_queue_traffic_test.go
```diff
@@ -1299,14 +1345,11 @@ func ConfigureCiscoQos(t *testing.T, dut *ondatra.DUTDevice) {
 		action := term.GetOrCreateActions()
 		action.SetTargetGroup(tc.targetGroup)
 		condition := term.GetOrCreateConditions()
-		if tc.classType == oc.Qos_Classifier_Type_IPV4 {
+		if tc.name == "dscp_based_classifier_ipv4" {
```
prinikasn, line 1348: Why are we comparing tc name instead of tc type?

saarvind6481: The original tests function which ConfigureQoS had similar way to differentiate between v4 and v6 conditions was with comparing with string.
condition := term.GetOrCreateConditions()
if tc.name == "dscp_based_classifier_ipv4" {
condition.GetOrCreateIpv4().SetDscpSet(tc.dscpSet)
} else if tc.name == "dscp_based_classifier_ipv6" {
condition.GetOrCreateIpv6().SetDscpSet(tc.dscpSet)
}

But since in the previous releases , we did not support different policy-map name for V4 or v6 under same , I used type to modify the condition. Since in this release , we support the same, changed the test back how the original test was written.
Please let me know if we still have to change the test by using the type to derive the condition , I can do the same

guoshiuan (approved):

PR #1676, opened 2023-05-29
Cisco dp 1.10 uprev

saarvind6481:


internal/deviations/deviations.go
```diff
@@ -575,8 +570,6 @@ var (
 
 	bgpMD5RequiresReset = flag.Bool("deviation_bgp_md5_requires_reset", false, "Device requires a BGP session reset to utilize a new MD5 key")
 
-	qosDroppedOctets = flag.Bool("deviation_qos_dropped_octets", false, "Set to true to skip checking QOS Dropped octets stats for interface")
```
prinikasn, line 578: This deviation is currently used by other vendors too, so I dont think we can remove them.

feature/qos/ate_tests/mixed_sp_wrr_traffic_test/mixed_sp_wrr_traffic_test.go
```diff
@@ -554,9 +548,9 @@ func TestMixedSPWrrTraffic(t *testing.T) {
 				counters["dutQosPktsBeforeTraffic"][data.queue] = gnmi.Get(t, dut, gnmi.OC().Qos().Interface(dp3.Name()).Output().Queue(data.queue).TransmitPkts().State())
 				counters["dutQosOctetsBeforeTraffic"][data.queue] = gnmi.Get(t, dut, gnmi.OC().Qos().Interface(dp3.Name()).Output().Queue(data.queue).TransmitOctets().State())
 				counters["dutQosDroppedPktsBeforeTraffic"][data.queue] = gnmi.Get(t, dut, gnmi.OC().Qos().Interface(dp3.Name()).Output().Queue(data.queue).DroppedPkts().State())
-				if !deviations.QOSDroppedOctets(dut) {
```
prinikasn, line 557: This deviation is currently used by other vendors too, so we can remove them yet.

internal/deviations/deviations.go
saarvind6481: This code is in cisco specific config function will not affect other vendors.I have added deviations in deviations folder

feature/qos/ate_tests/mixed_sp_wrr_traffic_test/mixed_sp_wrr_traffic_test.go
saarvind6481: Added back the deviations

prinikasn (approved):
/gcbrun

prinikasn (approved):
/gcbrun

PR #1677, opened 2023-05-29
Update bursty_traffic_test.go

saarvind6481:
cisco-dp-1.11-uprev
Below are  resaons that has  brought about changes in PR from Cisco 


prinikasn (changes requested):
please also update the otg_tests/ files for this test.

feature/qos/ate_tests/bursty_traffic_test/bursty_traffic_test.go
```diff
@@ -400,9 +392,9 @@ func TestBurstyTraffic(t *testing.T) {
 				counters["dutQosOctetsBeforeTraffic"][data.queue] = gnmi.Get(t, dut, gnmi.OC().Qos().Interface(dp3.Name()).Output().Queue(data.queue).TransmitOctets().State())
 				counters["dutQosDroppedPktsBeforeTraffic"][data.queue] = gnmi.Get(t, dut, gnmi.OC().Qos().Interface(dp3.Name()).Output().Queue(data.queue).DroppedPkts().State())
 				//counters["dutQosDroppedOctetsBeforeTraffic"][data.queue] = gnmi.Get(t, dut, gnmi.OC().Qos().Interface(dp3.Name()).Output().Queue(data.queue).DroppedOctets().State())
-				if !deviations.QOSDroppedOctets(dut) {
```
prinikasn, line 403: This deviation is currently used by other vendors too, so we can remove them yet.

saarvind6481: This is in cisco specific config code, where we are not using the deviations, as we support Dropped Octets now
So this should not affect other vendors.

saarvind6481: Added back the deviations

saarvind6481: Also added the OTG tests

prinikasn (approved):
/gcbrun

PR #1678, opened 2023-05-29
deviations for power admin down up

manan-patel:


internal/deviations/deviations.go
```diff
@@ -475,6 +475,16 @@ func InterfaceRefConfigUnsupported(_ *ondatra.DUTDevice) bool {
 	return *interfaceRefConfigUnsupported
 }
 
+// PlatformPowerEnableWaitRequired returns timer value to wait while platform is powered on
```
prinikasn, line 478: As discussed offline please follow the new guidelines for adding these new deviations.

Here is the style guide - https://github.com/openconfig/featureprofiles/blob/main/internal/deviations/README.md
And here is a sample PR - https://github.com/openconfig/featureprofiles/pull/1663/files

internal/deviations/deviations.go
```diff
@@ -645,4 +655,8 @@ var (
 	routePolicyUnderAFIUnsupported = flag.Bool("deviation_route_policy_under_afi_unsupported", false, "Set true for device that does not support route-policy under AFI/SAFI, default is false")
 
 	interfaceRefConfigUnsupported = flag.Bool("deviation_interface_ref_config_unsupported", false, "Device does not support interface-ref configuration when applying features to interface")
+
+	platformPowerEnableWaitRequired = flag.Int("deviation_platform_power_enable_wait_required", 0, "Set timer in minutes to wait for shut platform to power up")
```
prinikasn, line 659: With the new method of adding deviations, you no longer have to add these flag values here. So these lines can be removed.

internal/deviations/deviations.go
```diff
@@ -658,4 +677,8 @@ var (
 	interfaceRefConfigUnsupported = flag.Bool("deviation_interface_ref_config_unsupported", false, "Device does not support interface-ref configuration when applying features to interface")
 
 	storageComponentUnsupported = flag.Bool("deviation_storage_component_unsupported", false, "Set to true for device that does not support telemetry path /components/component/storage")
+
+	_ = flag.Bool("deviation_platform_power_down_up_skip", false, "Set flag to skip as Device does not support power-admin-state leaf")
+
+	_ = flag.Int64("deviation_platform_power_enable_wait", 0, "Set timer in minutes to wait for shut platform to power up")
```
prinikasn, line 683: These lines can be removed, since these are new deviations, they do not have to be introduced as flags.

internal/deviations/deviations.go
```diff
@@ -486,6 +486,25 @@ func StorageComponentUnsupported(_ *ondatra.DUTDevice) bool {
...
+	if min_ppwewait < ppewait {
+		return min_ppwewait
+	}
+	return ppewait
+}
+
+// PlatformPowerDownUpSkip returns whether the device should skip the Platform Power Down Up case.
+// Default value is false.
+func PlatformPowerDownUpSkip(dut *ondatra.DUTDevice) bool {
+	logErrorIfFlagSet("deviation_platform_power_down_up_skip")
```
prinikasn, line 504: the `logErrorIfFlagSet` line can be removed from both functions. Since this is a new deviation the flag value is not avialable and we do not expect anyone setting these deviation via flag

internal/deviations/deviations.go
```diff
@@ -486,6 +486,25 @@ func StorageComponentUnsupported(_ *ondatra.DUTDevice) bool {
 	return *storageComponentUnsupported
 }
 
+// PlatformPowerEnableWait returns timer value to wait while platform is powered on.
+// Default value is 0.
+func PlatformPowerEnableWait(dut *ondatra.DUTDevice) int64 {
+	logErrorIfFlagSet("deviation_platform_power_enable_wait")
+	ppewait := lookupDUTDeviations(dut).GetPlatformPowerEnableWait()
+	var min_ppwewait int64 = 5
```
prinikasn, line 494: Does this mean that we do not want this proto field to be less than 5 ? 
If yes, why? Is 0 an invalid value?

prinikasn (changes requested):

feature/platform/tests/power_admin_down_up_test/power_admin_down_up_test.go
```diff
@@ -60,6 +63,9 @@ func TestLinecardPowerAdmin(t *testing.T) {
 
 func TestControllerCardPowerAdmin(t *testing.T) {
 	dut := ondatra.DUT(t, "dut")
+	if deviations.PlatformPowerDownUpSkip(dut) {
```
LimeHat, line 66: Could you please create separate deviations for fabric and controller card tests?
We'd like to reuse just one of those

manan-patel: Sure

internal/deviations/deviations.go
manan-patel: Updated logic

internal/deviations/deviations.go
```diff
@@ -490,6 +490,29 @@ func StorageComponentUnsupported(_ *ondatra.DUTDevice) bool {
 	return *storageComponentUnsupported
 }
 
+// PlatformPowerEnableWait returns timer value to wait while platform is powered on.
+// Default value is 0.
+func PlatformPowerEnableWait(dut *ondatra.DUTDevice) int64 {
+	ppewait := lookupDUTDeviations(dut).GetPlatformPowerEnableWait()
+	var min_ppwewait int64 = 0
```
prinikasn, line 497: don't use underscores in Go names; var min_ppwewait should be minPpwewait

internal/deviations/deviations.go
```diff
@@ -490,6 +490,29 @@ func StorageComponentUnsupported(_ *ondatra.DUTDevice) bool {
...
+	if ppewait > min_ppwewait {
```
prinikasn, line 498: Curious, default value returned by `GetPlatformPowerEnableWait()` is 0.

If the default value of the proto is 0, then would it make sense to directly return `lookupDUTDeviations(dut).GetPlatformPowerEnableWait()` ?

prinikasn (changes requested):

feature/platform/tests/power_admin_down_up_test/metadata.textproto
```diff
@@ -5,3 +5,17 @@ uuid: "7e1f1570-a141-4758-8aa5-228c932e3bee"
 plan_id: "FP-1.1"
 description: "Power admin DOWN/UP Test"
 testbed: TESTBED_DUT_ATE_2LINKS
+platform_exceptions: {
+  platform: {
+    vendor: CISCO
+    hardware_model: "CISCO-8808"
```
prinikasn, line 11: hardware_model field is deprecated, and replaced by hardware_model_regex.

However, in this case, if this deviation applies to all CISCO models, you can remove the hardware_model.* field completly. 

So your platform_exceptions will look like this:

```
platform_exceptions: {
  platform: {
    vendor: CISCO
  }
  deviations: {
    platform_power_enable_wait: 5
    platform_power_down_up_fabric_skip: true
    platform_power_down_up_controller_card_skip: true
  }
}
```

feature/platform/tests/power_admin_down_up_test/power_admin_down_up_test.go
```diff
@@ -143,12 +146,13 @@ func powerDownUp(t *testing.T, dut *ondatra.DUTDevice, name string, cType oc.E_P
 	}
 	t.Logf("Component %s, power-admin-state after %f minutes: %v", name, time.Since(start).Minutes(), power)
 
-	oper, ok := gnmi.Await(t, dut, c.OperStatus().State(), timeout, oc.PlatformTypes_COMPONENT_OPER_STATUS_DISABLED).Val()
-	if !ok {
-		t.Errorf("Component %s oper-status, got: %v, want: %v", name, oper, oc.PlatformTypes_COMPONENT_OPER_STATUS_DISABLED)
+	if !deviations.ComponentPowerDownReturnsInactiveState(dut) {
```
LimeHat, line 149: Can you add an explicit check for the inactive state based on this deviation flag (instead of skipping the oper status check altogether)?

LimeHat (changes requested):

prinikasn (approved):

PR #1680, opened 2023-05-30
ntp deviation

manan-patel:


internal/deviations/deviations.go
```diff
@@ -475,6 +475,11 @@ func InterfaceRefConfigUnsupported(_ *ondatra.DUTDevice) bool {
 	return *interfaceRefConfigUnsupported
 }
 
+// SkipNtpNonDefaultVrf returns if we skip the ntp nondefault vrf case
+func SkipNtpNonDefaultVrf(_ *ondatra.DUTDevice) bool {
```
prinikasn, line 479: As discussed offline please follow the new guidelines for adding these new deviations.

Here is the style guide - https://github.com/openconfig/featureprofiles/blob/main/internal/deviations/README.md
And here is a sample PR - https://github.com/openconfig/featureprofiles/pull/1663/files

internal/deviations/deviations.go
```diff
@@ -645,4 +650,6 @@ var (
 	routePolicyUnderAFIUnsupported = flag.Bool("deviation_route_policy_under_afi_unsupported", false, "Set true for device that does not support route-policy under AFI/SAFI, default is false")
 
 	interfaceRefConfigUnsupported = flag.Bool("deviation_interface_ref_config_unsupported", false, "Device does not support interface-ref configuration when applying features to interface")
+
+	skipNtpNonDefaultVrf = flag.Bool("deviation_skip_ntp_non_default_vrf", false, "Set true to skip the non default vrf ntp case")
```
prinikasn, line 654: With the new method of adding deviations, you no longer have to add these flag values here. So these lines can be removed.

prinikasn (changes requested):

internal/deviations/deviations.go
```diff
@@ -481,6 +481,13 @@ func StorageComponentUnsupported(_ *ondatra.DUTDevice) bool {
 	return *storageComponentUnsupported
 }
 
+// NtpNonDefaultVrf returns true if the device does not support ntp nondefault vrf.
+// Default value is false.
+func NtpNonDefaultVrf(dut *ondatra.DUTDevice) bool {
+	logErrorIfFlagSet("deviation_ntp_non_default_vrf")
```
prinikasn, line 487: since this is a new deviation we can remove the `logErrorIfFlagSet("deviation_ntp_non_default_vrf")` line. No need to define the flag, and since the flag is not present, users will not be able to set it.

internal/deviations/deviations.go
```diff
@@ -652,4 +659,6 @@ var (
 	interfaceRefConfigUnsupported = flag.Bool("deviation_interface_ref_config_unsupported", false, "Device does not support interface-ref configuration when applying features to interface")
 
 	storageComponentUnsupported = flag.Bool("deviation_storage_component_unsupported", false, "Set to true for device that does not support telemetry path /components/component/storage")
+
```
prinikasn, line 662: this line `_ = flag.Bool("deviation_ntp_non_default_vrf", false, "Set to true if device does not support ntp nondefault vrf")` can be removed. Since all new deviations will be proto fields we no longer have to set them as flags.

proto/metadata.proto
```diff
@@ -59,6 +59,8 @@ message Metadata {
     // Expected ucmp traffic tolerance. Minimum value is 0.2, anything less
     // will be coerced to 0.2.
     double hierarchical_weight_resolution_tolerance = 5;
+    // Device does not support the ntp nondefault vrf case
```
prinikasn, line 62: nit. add a period at the end.

prinikasn (changes requested):

proto/metadata.proto
```diff
@@ -80,6 +80,8 @@ message Metadata {
     bool isis_explicit_level_authentication_config = 13;
     // Device skip isis restart-suppress check if value is true.
     bool isis_restart_suppress_unsupported = 14;
+    // Device does not support the ntp nondefault vrf case.
+    bool ntp_non_default_vrf = 15;
```
prinikasn, line 84: Change the deviation name to `ntp_non_default_vrf_unsupported` to make it a bit more clear?

prinikasn (approved):

guoshiuan (approved):
b/286406754 has the internal test result.

PR #1684, opened 2023-05-31
Add README for deviations

prinikasn:


internal/deviations/README.md
```diff
@@ -0,0 +1,88 @@
+## Guidelines to add deviations to FNT tests
+
+* Add the deviation to the `Deviation` message in the [proto/metadata.proto](https://github.com/openconfig/featureprofiles/blob/main/proto/metadata.proto) file.
+
+```
+message Deviations {
```
greg-dennis, line 6: *Deviations

internal/deviations/README.md
```diff
@@ -0,0 +1,88 @@
...
+go list -f 'protobuf-import/{{ .Path }}' -m github.com/openconfig/ondatra | xargs -L1 dirname | sort | uniq | xargs mkdir -p
+go list -f '{{ .Dir }} protobuf-import/{{ .Path }}' -m github.com/openconfig/ondatra | xargs -L1 -- ln -s
+protoc -I='protobuf-import' --proto_path=proto --go_out=./ --go_opt=Mmetadata.proto=proto/metadata_go_proto metadata.proto
+goimports -w proto/metadata_go_proto/metadata.pb.go
+```
+
+* Add the accessor function for this deviation to the [internal/deviations/deviations.go](https://github.com/openconfig/featureprofiles/blob/main/internal/deviations/deviations.go) file. This function will need to accept a parameter `dut` of type `*ondatra.DUTDevice` to lookup the deviation value for a specific dut. This accessor function must call `lookupDUTDeviations` and return the deviation value. Test code will use this function to access deviations.
+	* If default value of deviation is same as default value for proto field, accessor method can directly call the `Get*()` method for deviation field. For e.g. for `bool` fields, the default is `false`, and the accessor method will have one line `return lookupDUTDeviations(dut).GetTracerouteFragmentation()`.
```
greg-dennis, line 34: Add some missing "the" words:
If *the* default value of the deviation is the same as *the* default value for *the* proto field, *the* accessor method...

internal/deviations/README.md
```diff
@@ -0,0 +1,88 @@
...
+	}
+	```
+
+	* If the default value of deviation is not the same as the default value of the proto field, the accessor method can add a check and return the required default value.
+
+	```
+	// HierarchicalWeightResolutionTolerance returns the allowed tolerance for BGP traffic flow while comparing for pass or fail conditions.
+	// Default minimum value is 0.2. Anything less than 0.2 will be set to 0.2.
+	func HierarchicalWeightResolutionTolerance(dut *ondatra.DUTDevice) float64 {
+    hwrt := lookupDUTDeviations(dut).GetHierarchicalWeightResolutionTolerance()
```
greg-dennis, line 50: The body of the function needs to be indented. Look at the preview to see how it's rendered:
https://github.com/openconfig/featureprofiles/blob/2fbd835c1654be786744edce1f9a4254b99ca397/internal/deviations/README.md

internal/deviations/README.md
greg-dennis, line 34: ".e.g." means "for example" so "For e.g." means "For for example" :)

How about changing the second sentence to:

"For example, the boolean `traceroute_fragmentation` deviation, which has a default value of false, will have an accessor method with the single line `return lookupDUTDeviations(dut).GetTracerouteFragmentation()`."

internal/deviations/README.md
```diff
@@ -0,0 +1,88 @@
...
+```
+if deviations.TraceRouteFragmentation(dut) {
+...
+}
+```
+
+* Example PR - https://github.com/openconfig/featureprofiles/pull/1649
+
+## Notes
+* If you run into issues with the `make proto/metadata_go_proto/metadata.pb.go` you may need to check if `protoc` module is installed in your environment. Also depending on your Go version you may also need to update your PATH and GOPATH.
```
greg-dennis, line 87: ... if *the* `protoc` module ...

greg-dennis (changes requested):

greg-dennis (approved):

internal/deviations/README.md
prinikasn: done

internal/deviations/README.md
prinikasn: done

internal/deviations/README.md
prinikasn: done

internal/deviations/README.md
prinikasn: Done

prinikasn:
Addressed the comments and made the indentation changes to the code snippets.
Also added couple changes: 
 * Expanded the example for the non default deviation section
 * Added PR1668 to the example PRs

greg-dennis (approved):

PR #1689, opened 2023-05-31
DP-1.8: Two strict priority queue traffic test - Updating vendor specific queue numbers and scheduler weights

charantejag504:
DP-1.8: Two strict priority queue traffic test - Updating vendor specific queue numbers and scheduler weights
1.   Update Queue number changes
2.  Update Scheduler weight as there is a weight limit of 100

### DIFF
--- a/feature/qos/ate_tests/two_sp_queue_traffic_test/two_sp_queue_traffic_test.go
+++ b/feature/qos/ate_tests/two_sp_queue_traffic_test/two_sp_queue_traffic_test.go
@@ -156,13 +156,13 @@ func TestTwoSPQueueTraffic(t *testing.T) {
 			"BE1": "g_BE1",
 		},
 		ondatra.JUNIPER: {
-			"NC1": "3",
-			"AF4": "2",
-			"AF3": "5",
-			"AF2": "1",
-			"AF1": "4",
+			"NC1": "7",
+			"AF4": "6",
+			"AF3": "4",
+			"AF2": "3",
+			"AF1": "2",
 			"BE1": "0",
-			"BE0": "6",
+			"BE0": "1",
 		},
 		ondatra.NOKIA: {
 			"NC1": "7",
@@ -1214,13 +1214,13 @@ func ConfigureQoS(t *testing.T, dut *ondatra.DUTDevice) {
 
 	if dut.Vendor() == ondatra.JUNIPER {
 		qos = qosVals{
-			be0: "6",
+			be0: "1",
 			be1: "0",
-			af1: "4",
-			af2: "1",
-			af3: "5",
-			af4: "2",
-			nc1: "3",
+			af1: "2",
+			af2: "3",
+			af3: "4",
+			af4: "6",
+			nc1: "7",
 		}
 	}
 
@@ -1502,7 +1502,7 @@ func ConfigureQoS(t *testing.T, dut *ondatra.DUTDevice) {
 		priority:    oc.Scheduler_Priority_STRICT,
 		inputID:     "AF4",
 		inputType:   oc.Input_InputType_QUEUE,
-		weight:      uint64(100),
+		weight:      uint64(99),
 		queueName:   qos.af4,
 		targetGroup: "target-group-AF4",
 	}, {
@@ -1511,7 +1511,7 @@ func ConfigureQoS(t *testing.T, dut *ondatra.DUTDevice) {
 		priority:    oc.Scheduler_Priority_STRICT,
 		inputID:     "NC1",
 		inputType:   oc.Input_InputType_QUEUE,
-		weight:      uint64(200),
+		weight:      uint64(100),
 		queueName:   qos.nc1,
 		targetGroup: "target-group-NC1",
 	}}


greg-dennis (approved):
Approving for Go quality only

greg-dennis (dismissed):
Re-approving for Go quality

greg-dennis (approved):
Re-approving for Go quality

sachendras (dismissed):

greg-dennis (changes requested):
The queue number changes can be reverted, because this is handled by the entity-naming library now. To update, please submit a PR that updates this function: https://github.com/openconfig/entity-naming/blob/main/internal/juniper/juniper.go#L114

Then this PR can include the go.mod and go.sum changes to bump the entity-naming library version.

charantejag504:
Hi Greg,

Please update this file with D35 queue names.
https://github.com/openconfig/entity-naming/blob/main/internal/juniper/juniper.go#L114

i.e revert this pull request https://github.com/openconfig/entity-naming/pull/20/files

Thanks,
Charan

greg-dennis (approved):

feature/qos/ate_tests/bursty_traffic_test/bursty_traffic_test.go
```diff
@@ -704,9 +704,10 @@ func ConfigureQoS(t *testing.T, dut *ondatra.DUTDevice) {
 	for _, tc := range classifierIntfs {
 		i := q.GetOrCreateInterface(tc.intf)
 		i.SetInterfaceId(tc.intf)
-		if deviations.ExplicitInterfaceRefDefinition(dut) {
-			i.GetOrCreateInterfaceRef().Interface = ygot.String(tc.intf)
-			i.GetOrCreateInterfaceRef().Subinterface = ygot.Uint32(0)
+		i.GetOrCreateInterfaceRef().Interface = ygot.String(tc.intf)
+		i.GetOrCreateInterfaceRef().Subinterface = ygot.Uint32(0)
+		if deviations.InterfaceRefConfigUnsupported(dut) || deviations.IntfRefConfigUnsupported(dut) {
```
mojiiba, line 709: to my understanding  deviations are for interfaceref/sub-interface and interfaceref/interface. Right? Can we rename the deviations to reflect what they are for? InterfaceRef and IntfRef both have the same meaning. Also, we need to handle each deviation separately since it is possible the device requires interfaceref/interface to be set, but does not support interfaceref/sub-interface or sub-interface 0.

feature/qos/ate_tests/qos_basic_test/qos_basic_test.go
```diff
@@ -1517,10 +1518,8 @@ func ConfigureJuniperQos(t *testing.T, dut *ondatra.DUTDevice) {
 	for _, tc := range classifierIntfs {
 		i := q.GetOrCreateInterface(tc.intf)
 		i.SetInterfaceId(tc.intf)
-		if deviations.ExplicitInterfaceRefDefinition(dut) {
-			i.GetOrCreateInterfaceRef().Interface = ygot.String(tc.intf)
-			i.GetOrCreateInterfaceRef().Subinterface = ygot.Uint32(0)
-		}
+		i.GetOrCreateInterfaceRef().Interface = ygot.String(tc.intf)
```
mojiiba, line 1521: what this test is not consistent with other tests in using deviations InterfaceRefConfigUnsupported(dut)  and deviations.IntfRefConfigUnsupported

feature/qos/ate_tests/qos_basic_test/qos_basic_test.go
```diff
@@ -1628,34 +1627,32 @@ func ConfigureJuniperQos(t *testing.T, dut *ondatra.DUTDevice) {
...
 	}}
 
 	t.Logf("qos output interface config: %v", schedulerIntfs)
 	for _, tc := range schedulerIntfs {
 		i := q.GetOrCreateInterface(dp3.Name())
 		i.SetInterfaceId(dp3.Name())
-		if deviations.ExplicitInterfaceRefDefinition(dut) {
-			i.GetOrCreateInterfaceRef().Interface = ygot.String(dp3.Name())
-		}
+		i.GetOrCreateInterfaceRef().Interface = ygot.String(dp3.Name())
```
mojiiba, line 1655: same as above

feature/qos/ate_tests/qos_basic_test/qos_basic_test.go
charantejag504: This is under ConfigureJuniperQos, it is specific to Juniper and the mentioned deviation is not applicable.

feature/qos/ate_tests/bursty_traffic_test/bursty_traffic_test.go
charantejag504: it is consistent with the current deviation used.
reference: https://github.com/openconfig/featureprofiles/pull/1768
changing name of deviations should not be under the scope of our pull.

feature/qos/ate_tests/qos_basic_test/qos_basic_test.go
charantejag504: This is under ConfigureJuniperQos, it is specific to Juniper and the mentioned deviation is not applicable

feature/qos/ate_tests/bursty_traffic_test/bursty_traffic_test.go
mojiiba: you are changing all the test in one pull, so that would be great if you also improve the misleading names and handle the deviations properly.
Also, concerning the the scope and the guideline, we open one pull per test,  can u please follow the guideline. Also, please do not resolve the comment if you have not resolved.  

@vosipchu FYI

mojiiba (changes requested):
I noticed my comment is resolved without addressing it. Please improve the misleading names for deviations and handle them separately.

greg-dennis (changes requested):
I'm going to investigate this week whether I can unblock our import of this test internally.

greg-dennis:
@mojiiba, there's a bit of a complicated reason why there are two deviations with the same meaning. It's a rare case where we need to sometimes pull the deviation value from a flag and sometimes from the metadata proto. I agree they should converged to one, but then we also need to document how we handle similar cases in the future. @prinikasn and I were discussing it right before she went OOO. If we can hold our nose on that one for another week, I can make sure it is top priority for her when she returns.

greg-dennis:
Sorry, my comment about unblocking merging this internally should have been on another PR. This one looks good to go.

greg-dennis (approved):

PR #1692, opened 2023-06-01
TE-14.1/TE-3.3/TE-3.1: Refactored the deviation for interface-ref config

mohanasm:
Refactored TE-14.1, TE-3.1 and TE-3.3 with deviation InterfaceRefConfigUnsupported


greg-dennis (changes requested):
PR title is misleading, because it does not remove the deviation entirely; it only replaces some uses of it.

mohanasm:
> PR title is misleading, because it does not remove the deviation entirely; it only replaces some uses of it.

Meant that it is removed for TE-14.1/TE-3.3 and added the testcase number in the title to make it specific. To have better clarity, updated the title and description. Please check

greg-dennis (dismissed):

sachendras (dismissed):

feature/gribi/ate_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go
```diff
@@ -179,7 +179,7 @@ func applyForwardingPolicy(t *testing.T, ingressPort string) {
 	pfPath := gnmi.OC().NetworkInstance(deviations.DefaultNetworkInstance(dut)).PolicyForwarding().Interface(ingressPort)
 	pfCfg := d.GetOrCreateNetworkInstance(deviations.DefaultNetworkInstance(dut)).GetOrCreatePolicyForwarding().GetOrCreateInterface(ingressPort)
 	pfCfg.ApplyVrfSelectionPolicy = ygot.String(policyName)
-	if deviations.ExplicitInterfaceRefDefinition(dut) {
+	if !deviations.InterfaceRefConfigUnsupported(dut) {
```
xw-g, line 182: I think we should try to avoid double negative to improve readability. 

Is it better to write something like the following?

```
pfCfg.GetOrCreateInterfaceRef().Interface = ygot.String(ingressPort)
pfCfg.GetOrCreateInterfaceRef().Subinterface = ygot.Uint32(0)
if deviations.InterfaceRefConfigUnsupported(dut) {
   // unset the interface ref
}
```

mohanasm: Updated to unset interfaceref with deviation InterfaceRefConfigUnsupported as per your suggestion.

xw-g:
> PR title is misleading, because it does not remove the deviation entirely; it only replaces some uses of it.

+1, @mohanasm you can simply say that this PR refactors/reword the deviation naming.

rszarecki (commented):
LGTM

PR #1696, opened 2023-06-01
update isis feature.textproto

dplore:
* Fixed large number of path errors due to model changes 
* Fixed paths with typo in interfaces/interface
* added folder for link-state-database as there are quite a lot of lsdb telemetry paths and seemed good to separate for a separate test
* I did not include two paths which were thought to be required but are no longer part of the OC model set:
`/network-instances/network-instance/protocols/protocol/isis/interfaces/interface/levels/level/adjacencies/adjacency/state/local-extended-system-id`
`/network-instances/network-instance/protocols/protocol/isis/interfaces/interface/levels/level/adjacencies/adjacency/state/remaining-hold-time`


dplore:
@bensons after many hours fixing paths, this is ready for review

feature/isis/link-state-database/feature.textproto
```diff
@@ -0,0 +1,156 @@
+id {
+  name: "isis_link-state-database"
```
robshakir, line 2:
```suggestion
  name: "isis_lsdb"
```

robshakir (approved):

bensons (approved):

feature/isis/link-state-database/feature.textproto
bensons: It seems to me that we should keep the name of the Feature Profile consistent with the directory structure, when possible... I think there are some tools that make this assumption.

dplore: concise and readable suggestion, but the OC path spells this out, so I wanted to align with that for consistency

PR #1703, opened 2023-06-05
Add max recv size call option so that gNOI responses can have up to 40MB data per recv call.

Ankur19:
MaxCallRecvMsgSize has a default value of 4MB. However, devices may send more than 4MB data per Recv in the gNOI.Healthz.Get response. 

An example of that is Healtz.Get for `Chassis` component path here: https://github.com/openconfig/featureprofiles/blob/main/feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go#L98

This change would allow setting the MaxRecvMsgSize for all gRPC connections or for individial g* APIs


topologies/proto/binding.proto
```diff
@@ -110,6 +110,9 @@ message Options {
 
   // gRPC request timeout (second)
   int32 timeout = 7;
+
+  // gRPC dial option to set the maximum recv message size.
```
liulk, line 114:
```suggestion
  // gRPC dial option to set the maximum recv message size in bytes.
```
https://github.com/grpc/grpc-go/blob/v1.55.0/rpc_util.go#L342

topologies/atedut.binding
```diff
@@ -27,6 +27,11 @@ duts {
     target: "dut-proxy-hostname:6030"
   }
 
+  # Options specific to gNOI
+  gnoi {
+    max_recv_msg_size: 1048576
```
liulk, line 32: The unit is in bytes, so this sets the size to 1MB, not 1GB.

1GB is excessive. If we are sending more than 4MB in one response, we probably should explore other alternatives such as streaming gRPC.

Ankur19: Done! Updated to 40MB which should suffice.

liulk (approved):

PR #1705, opened 2023-06-05
TE-13.1 - Setting parent intf as L3 using the existing InterfaceEnabled deviation

sambhavjain-204:
Fixing the test to use the existing deviation `InterfaceEnabled` to configure parent interface as L3.


feature/experimental/gribi/ate_tests/route_addition_during_failover_test/route_addition_during_failover_test.go
```diff
@@ -295,6 +295,12 @@ func generateSubIntfPair(t *testing.T, dut *ondatra.DUTDevice, dutPort *ondatra.
 		configureATE(t, top, atePort, name, vlanID, dutIPv4, ateIPv4+"/30")
 		nextHops = append(nextHops, ateIPv4)
 	}
+	if deviations.InterfaceEnabled(dut) {
+		i := d.GetOrCreateInterface(dutPort.Name())
+		i.Type = oc.IETFInterfaces_InterfaceType_ethernetCsmacd
+		s := i.GetOrCreateSubinterface(0).GetOrCreateIpv4()
```
LimeHat, line 301: Could you elaborate on this change? It seems that you're creating a new subinterface 0 (which is not used in any capacity in this test) under port-2. I don't think this matches the description or established pattern of use of `InterfaceEnabled` deviation.

prakashbadri-arista: The problem here is that the subinterfaces are not routed ( ipv4/ipv6 not enabled ) as expected due to the parent interface not being a routed port.  The reason for this is a deviation in defaults between EOS interface configuration ( ipv4/ipv6 enabled is false ) and open config defaults for an interface where ipv4(6) enabled path is True.  The "InterfaceEnabled" deviation is used typically for handling this difference in behavior. This particular change enables routing in the parent interface explicitly when this deviation is set.

LimeHat: > as expected due to the parent interface not being a routed port.

Is there an OC definition of this? L301 creates a new _subinterface_ (with the index 0)

If an implementation requires interface 0 to exist in order to use L3 configuration on a non-zero subinterface, that should be a new deviation IMO.
`InterfaceEnabled` can and should be used to enable interfaces and subinterfaces address families on existing interfaces/subinterfaces that are part of the test.

sambhavjain-204: I've added a new deviation for this and pushed the changes.

Ankur19 (approved):

PR #1722, opened 2023-06-06
Add additional KNE topologies and add aggregate link testbeds

alexmasi:
This change adds two new testbeds:

- ate_dut with 4 agg links
- ate_dut with 9 agg links

Additionally topologies for kne are added to fit all testbeds with an additional topo that fits all testbeds

Changing the enums will come in a follow up PR


topologies/atedut_12.binding
```diff
@@ -1,149 +0,0 @@
-# proto-file: github.com/openconfig/featureprofiles/blob/main/topologies/proto/binding.proto
```
greg-dennis, line 1: Why was this deleted? There is still an atedut_12.testbed.

greg-dennis (dismissed):
One question below. Also, I'd like @bstoll to sign off on the cloudbuild changes.

topologies/atedut_12.binding
alexmasi: Ah I also meant to delete the .testbed

In general are any of these shapes being used? I assumed not because of there not being an enum in metadata for 12

greg-dennis: Aha, there may be some vendors who are using the 12 ports in a one-off way to "scale" test beyond the 9 link default. Can we leave them around until we make the announcement to stop providing the testbed flag? It's also part of the README file, which may need to be updated in this PR:
https://github.com/openconfig/featureprofiles/blob/main/README.md?plain=1#L95

alexmasi:
> One question below. Also, I'd like @bstoll to sign off on the cloudbuild changes.

@bstoll I considered just deleting the auto presubmit `test.sh` as it doesn't do anything of value unless the binding itself gets changed. The new ci-trigger essentially can replace it by running tests that are actually modified. Maybe its better to keep it but turn off the auto run so it requires the gcbrun comment to manually trigger if its a PR that the presub would add a valuable signal. Either way let me know what you think and if the changes look good to you

README.md
```diff
@@ -26,68 +26,126 @@ Tests below are implemented using the
...
+1. Create the topology:
+
+```
+kne create topologies/kne/arista/ceos/topology.textproto
+```
+
+2. Run a sample test:
+
+```
+go test ./feature/system/tests/... -topology $PWD/topologies/kne/arista/ceos/topology.textproto -vendor_creds ARISTA/admin/admin
```
bstoll, line 46: I think this should be `-kne-topology` not `-topology`.

bstoll:
My goal was to remove `/cloudbuild/cloudbuild.yml`, `/cloudbuild/test.sh`, and the general CloudBuild-GitHub integration once ci-trigger was a bit more proven reliable (ie #1727, still finding some minor problems), since ci-trigger does the same thing but in a more granular way.

Is the goal to have KNE do matching such that a single `topology.textproto` can cover all of these testbeds?  Or is there going to be anything that uses the `topology.textproto` testbed?

bstoll (approved):

alexmasi:
> Is the goal to have KNE do matching such that a single topology.textproto can cover all of these testbeds?

Yes, the topology.textproto can match any of the testbeds. The drawback is sometimes 2 DUTs are created when the test only needs 1 DUT and this can unnecessarily increase resource usage. However it simplifies things greatly and can be very convenient when you want to bring up a single topo and run multiple tests against it in sequence using resets instead of a full tear down.

So with those tradeoffs outlined it seems like the CI-trigger could make good use of the topology.textproto (if it runs the tests in sequence) to save some time by avoiding a full tear down.

bstoll (approved):

PR #1724, opened 2023-06-06
Update metadata.textproto

sachendras:
Update metadata.proto for the test to include newly added deviation in pull/1653


bstoll:
It looks like #1699 included this change already.

PR #1731, opened 2023-06-08
Add enum for agg testbeds and update metadata

alexmasi:


alexmasi:
@greg-dennis I'm not sure how to determine which of the other tests (outside of the aggregate dir) require a LAG testbed. We had previously discussed that some of the 4 link testbeds might need the lag testbed:

https://github.com/search?q=repo%3Aopenconfig%2Ffeatureprofiles+path%3A*%2Fmetadata.textproto+TESTBED_DUT_ATE&type=code

bstoll:
Not sure if I missed any, but I think these are the ones using aggregates:

```
gnmi/ate_tests/telemetry_port_speed_test
interface/aggregate/ate_tests/aggregate_forwarding_viable_test
interface/aggregate/ate_tests/aggregate_test
interface/aggregate/ate_tests/balancing_test
interface/aggregate/otg_tests/aggregate_forwarding_viable_test
interface/aggregate/otg_tests/aggregate_test
interface/aggregate/otg_tests/balancing_test
system/gnmi/set/tests/gnmi_set_test
```

greg-dennis:
Yeah, you probably want to look at which ate tests are using the `WithLAG` call

alexmasi:
Thanks for the tips:

- `feature/system/gnmi/set/tests/gnmi_set_test` has no mention of LAG and is already passing on KNE with the current testbed so I will leave it as is
- `gnmi/ate_tests/telemetry_port_speed_test` has the WithLAG call but:
    - there are only 2 ports in the testbed (the lag grouping requires that the first port is not included in the LAG so this is a noop)
    - this is only an ATE test (not OTG) so it should not make a difference

This PR should cover the rest of the LAG cases, I've removed 4_lag enum and testbed because it isn't currently needed

proto/metadata.proto
```diff
@@ -35,6 +35,7 @@ message Metadata {
     TESTBED_DUT_ATE_2LINKS = 3;
     TESTBED_DUT_ATE_4LINKS = 4;
     TESTBED_DUT_ATE_9LINKS = 5;
```
greg-dennis, line 37: Is there technically still a need for the 9LINKS or do all the 9LINK testbeds involve lags now?

greg-dennis:
What does this mean?

>   the lag grouping requires that the first port is not included in the LAG so this is a noop

proto/metadata.proto
alexmasi: all of the 9 links got migrated to lags, so the enum is not currently in use. would you prefer I leave it as is or remove it?

alexmasi:
> What does this mean?
> 
> > the lag grouping requires that the first port is not included in the LAG so this is a noop

All of the LAG testbeds follow the pattern of port1 is not part of the LAG and ports 2-`n` for `n` links make up the lag. So for the 2 link testbed I believe its a noop.

This follows the description of the agg tests: https://github.com/openconfig/featureprofiles/tree/main/feature/interface/aggregate/otg_tests/aggregate_test#procedure

proto/metadata.proto
greg-dennis: Um, I think remove. The fewer things we are maintaining, the better.

alexmasi: done

PR #1733, opened 2023-06-08
Fix kne topologies for complete shape

alexmasi:
otg1 + otg2 doesn't work, instead combine the ports on one otg


alexmasi:
@octpetre

bstoll (approved):

PR #1740, opened 2023-06-12
Te 3.1 deviation

manan-patel:
TE-3.1 deviation, adding dummy ipv4 address for nexthop programming using macaddress
set "gribiMACOverrideWithStaticARP" to true


feature/gribi/ate_tests/base_hierarchical_route_installation_test/base_hierarchical_route_installation_test.go
```diff
@@ -84,6 +84,19 @@ var (
 		IPv4:    "192.0.2.6",
 		IPv4Len: ipv4PrefixLen,
 	}
+
+	dutPort2DummyIP = attrs.Attributes{
+		Desc:    "dutPort2",
+		IPv4:    "192.0.2.21",
+		IPv4Len: 30,
```
guoshiuan, line 91: why does the dut/ate port pair have different  mask length?

guoshiuan (approved):
b/286947203 has the internal test results.

PR #1749, opened 2023-06-13
Update to read vendor and hardware model from binding

prinikasn:


topologies/proto/binding.proto
```diff
@@ -84,6 +86,12 @@ message Device {
 
   // Dial options for OTG Ixia-C (HW Only).
   Options otg = 18;
+
+  // Vendor of the device.
+  ondatra.Device.Vendor vendor = 19;
+
+  // Hardware models of the device.
+  string hardware_model = 20;
```
greg-dennis, line 94: Please add software_version as well

topologies/binding/binding.go
```diff
@@ -322,6 +322,18 @@ func dims(td *opb.Device, bd *bindpb.Device) (*binding.Dims, error) {
 	if err != nil {
 		return nil, err
 	}
+	if bd.GetVendor().String() != "" && bd.GetHardwareModel() != "" {
```
greg-dennis, line 325: These dimensions should be able to override the testbed dimensions individually. Like:

```
dims := &binding.Dims{
	Name:            bd.Name,
	Vendor:          bd.GetVendor(),
	HardwareModel:   bd.GetHardwareModel(),
	SoftwareVersion: bd.GetSoftwareVersion(),
	Ports:           portmap,
}

// Populate empty binding dimensions with testbed dimensions. 
// TODO(partnercorp bug link): Remove once all vendors are using binding dimensions exclusively.
if tdVendor := td.GetVendor(); tdVendor != opb.VENDOR_UNSPECIFIED {
	if dims.Vendor != opb.VENDOR_UNSPECIFIED && dims.Vendor != tdVendor {
		return nil, fmt.Errorf("binding vendor %v and testbed %v vendor do not match", dims.Vendor, tdVendor)
	} 
	dims.Vendor = tdVendor
}
// similarly for hardware model and software version

return dims
```

greg-dennis (changes requested):

topologies/proto/binding.proto
prinikasn: done

topologies/binding/binding.go
prinikasn: Done. 

I was also looking through the code base and noticed most TODO's had ldap instead of bug id. 
So followed that style and added my ldap and filed an internal bug b/287292956. Let me know if the bug needs to be moved to another component and added here.

greg-dennis (approved):

PR #1751, opened 2023-06-13
CI-Trigger: Use git merge-base for generating diffs

bstoll:
CI Trigger compares the diff between the "base" and "head" commit hashes received from the GitHub API in a pull request.  The "base" commit hash received turns out to be misleading; is not the common ancestor of the merging branch and the head commit.  This is hinted at in the docs in github/docs#431 that the base value will not change as a pull request progresses.

This change implements the `git merge-base` behavior in ci-trigger instead of relying on GitHub API to provide the correct value.

This should fix the behavior where extra tests are listed as modified in a PR when the PR is not changing them.


tools/ci-trigger/git.go
```diff
@@ -81,3 +90,35 @@ func modifiedFiles(repo *git.Repository, head, base string) ([]string, error) {
...
+		commits = append(commits, commit)
+	}
+
+	res, err := commits[0].MergeBase(commits[1])
+	if err != nil {
+		return "", fmt.Errorf("could not traverse the repository history: %w", err)
+	}
+
+	if len(res) < 1 {
+		return "", fmt.Errorf("no common git ancestor found for hash '%s'", head)
```
greg-dennis, line 120: nit '%s' can be replaced with %q

tools/ci-trigger/git_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if _, err := rm.Write([]byte(contents)); err != nil {
+		return err
+	}
+
+	if _, err := wt.Add(fileName); err != nil {
+		return err
+	}
+
+	if _, err := wt.Commit("test commit message", &git.CommitOptions{
```
greg-dennis, line 35: can save a couple lines with:

```
_, err = wt.Commit("test commit message", &git.CommitOptions{
	All:       true,
	Author:    &author,
	Committer: &author,
}
return err
```

greg-dennis (dismissed):
A few nits that aren't too important

greg-dennis (approved):

PR #1752, opened 2023-06-14
Move remaning shared deviations to metadata.textproto

prinikasn:
PiperOrigin-RevId: 540098730

Move the following deviations to metadata:
 - subinterface_packet_counters_missing
 - connect_retry
 - gribi_mac_override_with_static_arp
 - components_software_module_unsupported
 - route_policy_under_afi_unsupported


proto/metadata.proto
```diff
@@ -171,6 +171,18 @@ message Metadata {
     bool explicit_interface_in_default_vrf = 42;
     // Set to true to skip checking QOS Dropped octets stats for interface.
     bool qos_dropped_octets = 43;
+    // Device is missing subinterface packet counters for IPv4/IPv6.
+    bool subinterface_packet_counters_missing = 44;
+    // Connect-retry is not supported
+    // /bgp/neighbors/neighbor/timers/config/connect-retry.
+    bool connect_retry = 45;
+    // Set to true for device not supporting programming a gribi flow with a
```
greg-dennis, line 179: In a separate PR, can you remove all the "Set to true" phrases in the proto? For example, this should just be "Device does not support programming ..."

greg-dennis (dismissed):

proto/metadata.proto
prinikasn: Will work on that on here: https://github.com/openconfig/featureprofiles/pull/1764

greg-dennis (approved):

PR #1754, opened 2023-06-14
Update prefixlimit.feature.textproto

sachendras:
Add missing OC paths in the textproto based on the feature requests in the various bugs from the catalog.


dplore (changes requested):
These paths are present in https://github.com/openconfig/featureprofiles/blob/add9e8dddde998961a0b8e56b6bcf04626cb1aa9/feature/bgp/feature.textproto#L248.  

They should be moved here as you have indicated. Can you also remove them from https://github.com/openconfig/featureprofiles/blob/add9e8dddde998961a0b8e56b6bcf04626cb1aa9/feature/bgp/feature.textproto#L248?

self-maurya:
This change is no longer required as we decided to deprecate the feature.textproto - https://github.com/openconfig/featureprofiles/pull/2793

PR #1756, opened 2023-06-14
BGP route-reflector-at-scale-test

sachendras:
Proposed test addresses requirements in the internal bug b/287237486

Validates:
- BGP RR capability where IBGP peering address is learnt over IS-IS
- OC leaves that capture BGP transport session attributes like local-address, peer-type
- Ensure supported-capabilites leaf functionality


feature/bgp/routereflector/route-reflector-test
```diff
@@ -0,0 +1,45 @@
...
+  - Connect ATE Port1 to DUT port1 (EBGP peering)
+  - Connect ATE Port2 to DUT port2 (For IS-IS adjacency and IBGP peer reachaility)
+  - Connect ATE Port3 to DUT port3 (For IS-IS adjacency and IBGP peer reachaility)
+
+Procedure:
+  - Establish IS-IS adjacency between ATE Port2 <-> DUT Port2, ATE Port3 <-> DUT Port3. 
+  - Establish BGP sessions as follows between ATE and DUT . 
+    - ATE port 2 and ATE port3 are emulating RR clients peered with the DUT acting as the RR server. DUT's loopback address should be used for the IBGP peering and 
+      "transport/config/local-address" OC path should be used on DUT to configure BGP transport address for IBGP peering. ATE addresses used for the IBGP peering (different from ATE Port1 and ATE Port2 addreses) and DUT 
+      loopback addresses should be reachable via IS-IS. Each of the IBGP peering should advertise 5M IPv4 and 1M IPv6 prefixes.
```
bensons, line 16: It would be helpful to give additional guidance about these routes, e.g. how many of them are the same prefix vs unique prefixes from each RRC? Basic RR reflection would cause each RRC to see all prefixes, so the validation could look for reflected unique (i.e. otherwise unknown to the RR) prefixes on each RRC.

feature/bgp/routereflector/route-reflector-test
```diff
@@ -0,0 +1,45 @@
...
+    - The DUT Port1 has eBGP peering with ATE Port 1 and is receiving 10M IPv4 and 3M IPv6 prefixes. Please note: The FIB is expected to be 1.5M v4 routes and 600K v6 routes.
```
bensons, line 17: Related to my comment above: How do we expect the RIB->FIB math to work out here? If there are 5M unique prefixes each from 2 RRCs, how did we reduce them to 1.5M FIB entries etc? For that matter, is validation of the FIB something that is needed in the scope of the RR test?

feature/bgp/routereflector/route-reflector-test
```diff
@@ -0,0 +1,45 @@
+Summary: 
+ - BGP route reflector capabilities check
+ - Ensure functionality of different OC paths for "supported-capabilities", "BGP peer-type", "BGP Neighbor details" and "BGP transport session parameters"
+
+Topology:
+ATE (Port1) <-EBGP-> (Port1) DUT (Port2) <-IS-IS/IBGP-> (Port2) ATE, DUT (Port3) <-IS-IS/IBGP-> (Port3) ATE
```
bensons, line 6: Not review feedback, really, but rather an observation: it would be useful to have variations of this test that don't have dependencies on IS-IS etc. Regardless I think this test is a realistic case, so this test makes sense to me.

feature/bgp/routereflector/route-reflector-test
```diff
@@ -0,0 +1,45 @@
+Summary:
```
dplore, line 1: Please add title and test ID assignment
Please add a github issue and place it in https://github.com/orgs/openconfig/projects/5 to track these test requirements and implementation.

feature/bgp/routereflector/route-reflector-test
sachendras: Well, I thought it is implicit here that 1.5M v4 and 600k v6 are unique, not clear? Didnt mention policies here. These are prefixes learnt from a single EBGP peer simulating an external peer w/ internet scale routes. 

Anyways, I added more text to clarify

feature/bgp/routereflector/route-reflector-test
sachendras: Added more text to clarify

feature/bgp/routereflector/route-reflector-test
sachendras: This is more to capture our use case. IMO, we can add more at a later date

feature/bgp/routereflector/route-reflector-test
sachendras: Done

feature/bgp/routereflector/README.md
```diff
@@ -0,0 +1,44 @@
...
+*   Connect ATE Port3 to DUT port3 (For IS-IS adjacency and IBGP peer reachaility)
+
+## Procedure
+*   Establish IS-IS adjacency between ATE Port2 <-> DUT Port2, ATE Port3 <-> DUT Port3.
+*   Establish BGP sessions as follows between ATE and DUT.
+  *  ATE port 2 and ATE port3 are emulating RR clients peered with the DUT acting as the RR server. DUT's loopback address should be used  for the IBGP peering and "transport/config/local-address" OC path should be used on DUT to configure BGP transport address for IBGP peering. ATE addresses used for the IBGP peering (different from ATE Port1 and ATE Port2 addreses) and DUT loopback addresses should be reachable via IS-IS. Each of the IBGP peering should advertise 5M IPv4 and 2M IPv6 prefixes. Among these, 2M v4 and 1M v6 prefixes are unique. Among the 2M v4 prefixes as well, 1.5M are non RFC1918 or RFC6598 addresses and represent Internet prefixes.
+  *   The DUT Port1 has eBGP peering with ATE Port 1 and is receiving 10M IPv4 and 3M IPv6 routes. These simulate Internet routes. Out of these, only 1.5M v4 and 600k IPv6 prefixes are unique and will be programmed to the FIB.
```
dplore, line 16: Which direction are the prefixes being advertised?  (I am guessing from DUT to ATE?)

Are the 1.5M 'Internet' prefixes the same 1.5M prefixes advertised from DUT to ATE the same 1.5M 'Internet' prefixes learned from the iBGP RRC's?

feature/bgp/routereflector/README.md
```diff
@@ -0,0 +1,44 @@
...
+*   The DUT should automatically determine the BGP transport source address based on the nearest interface. Hence, the OC path "transport/config/local-address" shouldnt be used.
+*   Validate session state on ATE ports and DUT using telemetry.
+  *   The check should also include accurately receiving values for the path "transport/state/local-address" for RRCs as well as for the EBGP peering.
+  *   Validate accuracy of the peer-type leaf (neighbor/config/peer-type) for EBGP and IBGP peering.
+*   Validate session state and capabilities received on DUT using telemetry.
```
dplore, line 21: Which BGP capabilities are expected to be present?

feature/bgp/routereflector/route-reflector-test
sachendras: This is to address b/299390379 which is RR @ scale. So this test should cover for that requirement. I have updated the test naming to also reflect the same.

feature/bgp/routereflector/README.md
sachendras: I added more context and details ptal. Mapping w/ a real world scenario

feature/bgp/routereflector/README.md
sachendras: Nothing in specific. I am guessing Juniper can just pick random communities. Idea is to map w/ a real world scenario for a given BGP update message.

feature/bgp/routereflector/route-reflector-test
dplore: @sachendras, I think you still  need to specify if the two iBGP sessions should contain the same 5+1M prefixes or if they are expected to be unique?  I am guessing the two sessions are expected to contain the exact same 5+1M prefixes? If yes, can you state that?

feature/bgp/routereflector/route-reflector-test
dplore: For the FIB, of the  1.5M v4 routes and 600K v6 routes, how many of those are from eBGP vs. iBGP?   Maybe best to explain how many unique prefixes there are for each iBGP and the one eBGP session.

feature/bgp/routereflector/README.md
dplore: ok, this is now clear to me for the eBGP prefixes: "Out of these, only 1.5M v4 and 600k IPv6 prefixes are unique and will be programmed to the FIB"

feature/bgp/routereflector/README.md
dplore: You need to be specific about the pass/fail criteria.  Since we don't care which capabilities, we only care if there is a non-zero number of capabilities,  that is ok, but you should state that.

feature/bgp/routereflector/route-reflector-test
dplore: Ok, there is a good explanation for the eBGP session unique prefixes.  But need the same for the iBGP sessions

dplore (commented):
Please be more specific on the iBGP routes and the pass fail criteria for the bgp capabilities.  Then I think this is ready

feature/bgp/routereflector/route-reflector-test
sachendras: I thought I clarified: Here's what I had for local prefixes


- Among the 2M IPv4 prefixes as well, 1.5M are non RFC1918 or RFC6598 addresses and represent Internet prefixes. (adding new for clarity) These prefixes should be common between the 2xDUTs with different path-attributes for protocol next-hop, AS-Path and community
- Balance 500k IPv4 and 400k IPv6 prefixes represent internal subnets and should include some prefixes that are unique to each of the ATEs. (adding new for more clarity) Remaining prefixes in the mix need to be common between the 2xATEs and should have identical path attributes except for the protocol next-hops.

feature/bgp/routereflector/route-reflector-test
sachendras: Added more clarity. Hopefully that helps.

feature/bgp/routereflector/README.md
sachendras: updated and added expectations for eBGP and IBGP peering as well

feature/bgp/routereflector/route-reflector-test
dplore: Got it, I think I reviewed an earlier commit that didn't include this text.  I can understand it now, thanks!

dplore (approved):

PR #1770, opened 2023-06-17
restore mac+interface-ref addNH() code in gribi.go helper

LimeHat:
In #1740, code for gribi addNH() helper was modified to remove interface-ref from intf+MAC NH case statement, that doesn't seem to be correct
restoring the previous version

---
<sup>
</sup>

### DIFF
--- a/internal/gribi/gribi.go
+++ b/internal/gribi/gribi.go
@@ -229,7 +229,7 @@ func (c *Client) AddNH(t testing.TB, nhIndex uint64, address, instance string, e
 				nh = nh.WithSubinterfaceRef(opt.Interface, opt.SubInterface).
 					WithMacAddress(opt.Mac)
 			} else {
-				nh = nh.WithIPAddress(opt.Dest).WithMacAddress(opt.Mac)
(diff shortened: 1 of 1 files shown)


xw-g:
LGTM.

@guoshiuan, @manan-patel, I agree that #1740 incorrectly introduced code changes outside of the deviation. This PR is reverting back the specific line. 

@guoshiuan I don't think it requires presubmit. (because it reverts a previously approved behavior)

@guoshiuan if you agree, can you help approve and merge this PR and then notify the author @manan-patel to follow up with another PR? Thanks.

guoshiuan (approved):

manan-patel:
Adding the dummy address under the same interface to handle mac address programming

 description dutPort2
 ipv4 address 192.0.2.21 255.255.255.252 (dummy entry)
 ipv4 address 192.0.2.5 255.255.255.252 secondary

programming is done successfully with interface reference, however all the traffic will be lost.
HundredGigE0/0/0/1, NH mac address: 00:1a:11:00:0a:bc

having ip address along with mac address solves this problem
192.0.2.22, NH mac address: 00:1a:11:00:0a:bc

I'll open a separate PR to handle it.

PR #1762, opened 2023-06-15
Fix base_hierarchical_route_installation_test

Ankur19:
Some devices only support FIB_ACK when NextHop and NextHopGroup are added together.

The old script were adding NH and NHG as two standalone transactions. The change updates the test to add NH and NHG as one single transaction.

Changes in `gribi.go` also adds the option to add destination IP when NH add type is `MACwithInterface`

PR #1775, opened 2023-06-21
TE-3.1_fix : gribi programming support with mac and ipaddress combination

manan-patel:
Adding gribi NH programming support for mac and ip address combination, which fixes TE-3.1 TC

Dummy address is needed for gribi NH programming with mac address, with dummy address present using interfaceref or subinterfaceref with mac for NH programming will lead to incomplete forwarding and traffic will end up dropping.

Having nh dummy address plus mac fixes this problem.


internal/gribi/gribi.go
```diff
@@ -225,11 +225,15 @@ func (c *Client) AddNH(t testing.TB, nhIndex uint64, address, instance string, e
 		}
 	case "MACwithInterface":
 		for _, opt := range opts {
-			if opt.SubInterface != 0 {
-				nh = nh.WithSubinterfaceRef(opt.Interface, opt.SubInterface).
-					WithMacAddress(opt.Mac)
-			} else {
+			if opt.Dest != "" {
 				nh = nh.WithIPAddress(opt.Dest).WithMacAddress(opt.Mac)
```
LimeHat, line 229: This statement will not use interface-ref at all though.
Is this the intention?

If you want just to supplement interface-ref with an ip. addr, I think Ankur resolves this in a more appropriate way in #1762

manan-patel: Correct, I was checking if we have Destination ip provided in which case we use mac + ip
else we further check if subinterface is there and use subinterface + Mac else Interface+Mac

LimeHat: In that case (if interface-ref cannot be supported / Ankur's PR is not going to solve it), I'd propose to do this as a separate `case` statement in the helper.

Overloading `MACwithInterface` case with a scenario where, in fact, interface-reference is not used or present is very confusing/misleading.

manan-patel: Okay, let me add another case

LimeHat: thanks 👍

guoshiuan:
b/288899973 has the internal test results.

guoshiuan (approved):

PR #1776, opened 2023-06-22
AFI-SAFI oc_defaults

sachendras:
Summary:
  - When operating in "openconfig mode", NOS (network operating system) defaults should match what OC defines as the defaults i.e,
    - For BGP, there are no defaults for AFI-SAFI at the neighbor and peer-group levels. However at the global level the default is "false"
  - This test currently only verifies the defaults for ipv4-unicast and ipv6-unicast families. However, this test can be extended further to cover for other AFI-SAFIs as well in future.
  - The test will check for default implementations under the neighbor, peer-group and global-level hierarchies and also test for inheritance rules as was specified in pull/774 and pull/815.


dplore:
Hi @sachendras, can you double check the operational requirements?  I think the only operational requirements we expect are:
- Explicitly enable AF IPv4 only at global, peer-group and neighbor levels.  AF IPv6 should remain disabled (OC default)
- Explicitly enable AF IPv4 and IPv6  at global, peer-group and neighbor levels
 
Although the OC model has AF IPv4 disabled by default, we do not have an operational requirement for this (and most/all implementations also do not support this).

rszarecki (approved):

feature/bgp/afi_safi/oc_defaults/README.md
```diff
@@ -0,0 +1,68 @@
...
+  /network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/state/session-state
+  /network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/state/supported-capabilities
+  /network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/state/peer-type
+  /network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/state/peer-as
+  /network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/state/supported-capabilities
+
+  /network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/state/peer-type
+  /network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/state/peer-as
+  /network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/state/local-as
+  /network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/state/peer-group
```
dplore, line 67: Should create a feature.textproto for these paths, but it can be in a future PR.

dplore (approved):

PR #1778, opened 2023-06-22
TE-3.3: Increased tolerance to avoid flakiness observed

mohanasm:
Increased tolerance to 0.4 to avoid flakiness observed.


sachendras (approved):

feature/gribi/ate_tests/hierarchical_weight_resolution_test/metadata.textproto
sachendras, line 1: Did we try with 0.35?

arulkumarsekar: per the conversation on group chat, this is getting merged with .4 due to the difference between ptx10008 & ptx10001

PR #1788, opened 2023-06-26
Add hardware and software regex to metadata proto

prinikasn:
PiperOrigin-RevId: 543528687


internal/deviations/deviations.go
```diff
@@ -72,20 +74,60 @@ import (
 	"github.com/openconfig/ondatra"
 )
 
+func matchRegex(regex string, deviceInfo string) bool {
+	match, err := regexp.MatchString(regex, deviceInfo)
+	if err != nil {
+		log.Errorf("Error with regex match %v", err)
+	} else if match {
```
greg-dennis, line 81: drop the "else if" and just have an un-guarded "return match"

internal/deviations/deviations.go
```diff
@@ -72,20 +74,60 @@ import (
...
+		return true
+	}
+	return false
+}
+
 // lookupDutDeviations returns the deviations for the specified dut.
```
greg-dennis, line 87: this comment is not styled correctly "Dut" should be "DUT", but it doesn't seem to add anything beyond what the method name already communicates, so I'd just drop it entirely.

proto/metadata.proto
```diff
@@ -44,6 +44,11 @@ message Metadata {
     ondatra.Device.Vendor vendor = 1;
     // Hardware models of the device.
     repeated string hardware_model = 2;
+    // Regex for hardware model of the device.
```
greg-dennis, line 47: Add a comment on the new fields that leaving them empty matches any {hardware model, software version}.

internal/deviations/deviations.go
```diff
@@ -72,20 +74,60 @@ import (
...
+		if hardwareModelRegex != "" {
+			// Return an error if both hardware_model_regex and repeated hardware_model fields are set.
+			if hardwareModelCount > 0 {
+				log.Fatalf("Cannot set repeated hardware model field: %v and \n Hardware model regex: %v", platformExceptions.GetPlatform().GetHardwareModel(), platformExceptions.GetPlatform().GetHardwareModelRegex())
+			}
+			// If hardware_model_regex is set and does not match, continue
+			if !matchRegex(hardwareModelRegex, dut.Device.Model()) {
+				continue
+			}
+			if softwareVersionRegex != "" {
```
greg-dennis, line 108: I think it's easier to reason about if the software version is just matched independently of the hardware model

internal/deviations/deviations.go
```diff
@@ -72,20 +74,60 @@ import (
...
-			if dut.Device.Model() == hardwareModel {
+
+		hardwareModelRegex := platformExceptions.GetPlatform().GetHardwareModelRegex()
+		softwareVersionRegex := platformExceptions.GetPlatform().GetSoftwareVersionRegex()
+		hardwareModelCount := len(platformExceptions.GetPlatform().GetHardwareModel())
+
+		if hardwareModelRegex != "" {
+			// Return an error if both hardware_model_regex and repeated hardware_model fields are set.
+			if hardwareModelCount > 0 {
+				log.Fatalf("Cannot set repeated hardware model field: %v and \n Hardware model regex: %v", platformExceptions.GetPlatform().GetHardwareModel(), platformExceptions.GetPlatform().GetHardwareModelRegex())
```
greg-dennis, line 102: Suggested changes here:

- "Exitf" instead of "Fatalf" for a cleaner error message
- drop the word "repeated"
- "hardware model" -> "hardware_model"
- "Hardware model regex" -> "hardware_model_regex"

internal/deviations/deviations.go
```diff
@@ -72,20 +74,60 @@ import (
...
+	if err != nil {
+		log.Errorf("Error with regex match %v", err)
+	} else if match {
+		return true
+	}
+	return false
+}
+
 // lookupDutDeviations returns the deviations for the specified dut.
 func lookupDUTDeviations(dut *ondatra.DUTDevice) *mpb.Metadata_Deviations {
```
greg-dennis, line 88: we need this to fail if multiple platforms match the dut

greg-dennis (changes requested):

internal/deviations/deviations.go
prinikasn: done

internal/deviations/deviations.go
prinikasn: done.

proto/metadata.proto
prinikasn: done

internal/deviations/deviations.go

internal/deviations/deviations.go

internal/deviations/deviations.go

internal/deviations/deviations.go
```diff
@@ -64,6 +64,8 @@
 package deviations
 
 import (
+	"regexp"
+
```
greg-dennis, line 68: We don't need the blank line and can move flag below regexp, right?

```
"flag"
"regexp"
```

proto/metadata.proto
greg-dennis: Instead, I would include a unique line per comment, i.e.

```
    // Regex for hardware model of the device.
    // The empty string will match any hardware model.
    string hardware_model_regex = 3;
    // Regex for software version of the device.
    // The empty string will match any software version.
    string software_version_regex = 4;
```

internal/deviations/deviations.go
```diff
@@ -72,20 +74,73 @@ import (
...
-// lookupDutDeviations returns the deviations for the specified dut.
+func matchRegex(regex string, deviceInfo string) bool {
+	match, err := regexp.MatchString(regex, deviceInfo)
+	if err != nil {
+		log.Errorf("Error with regex match %v", err)
+	}
+	return match
+}
+
 func lookupDUTDeviations(dut *ondatra.DUTDevice) *mpb.Metadata_Deviations {
```
greg-dennis, line 85: Let's break this into two methods: `lookupDeviations` which returns `(*mpb.Metadata_Deviations, error)` and `mustLookupDeviations` which returns `*mpb.Metadata_Deviations` and calls `log.Exit` if `lookupDeviations` returns an error.

greg-dennis (changes requested):

proto/metadata.proto
prinikasn: Done

internal/deviations/deviations.go
prinikasn: Add a `lookupDeviations` which returns (*mpb.Metadata_PlatformExceptions, error) and instead of `mustLookupDeviations` updated the `lookupDUTDeviations ` to call `log.Exit` if lookupDeviations returns an error.

Let me know if this looks ok.

internal/deviations/deviations.go
```diff
@@ -65,27 +65,91 @@ package deviations
...
+		return fmt.Errorf("vendor should be specified in textproto %v", platformExceptions)
+	}
+	if platformExceptions.GetPlatform().GetHardwareModelRegex() != "" && len(platformExceptions.GetPlatform().GetHardwareModel()) > 0 {
+		return fmt.Errorf("vendor should be specified in textproto %v", platformExceptions)
+	}
+	return nil
+}
+
+func lookupDeviations(dut *ondatra.DUTDevice) (*mpb.Metadata_PlatformExceptions, error) {
+	matchedPlatformException := &mpb.Metadata_PlatformExceptions{}
```
greg-dennis, line 96: I'd prefer if this were initially nil and you use non-nil in place of isPlatformExceptionsMatched

internal/deviations/deviations.go
```diff
@@ -65,27 +65,91 @@ package deviations
...
+				continue
+			}
+		}
+
+		// TODO(prinikasn): Refactor after hardware_model field is removed.
+		if len(platformExceptions.GetPlatform().GetHardwareModel()) == 0 {
+			if isPlatformExceptionsMatched {
+				return nil, fmt.Errorf("cannot have more than one match within platform_exceptions fields %v and %v", matchedPlatformException, platformExceptions)
+			}
+			matchedPlatformException = platformExceptions
```
greg-dennis, line 125: It would be more consistent with the prior checks if you continue when there is a non-match, and then you only set this at the end of the loop (if there were no prior continues)

internal/deviations/deviations.go
```diff
@@ -65,27 +65,91 @@ package deviations
...
+	return matchedPlatformException, nil
+}
+
+func lookupDUTDeviations(dut *ondatra.DUTDevice) *mpb.Metadata_Deviations {
+	platformExceptions, err := lookupDeviations(dut)
+	if err != nil {
+		log.Exitf("Error looking up deviations: %v", err)
+	}
+	if platformExceptions.GetPlatform().Vendor.String() == "" {
+		log.Errorf("Did not match any platform_exception %v, returning default values", metadata.Get().GetPlatformExceptions())
```
greg-dennis, line 149: This should only be logged at an Info level. There is no error if there are no deviations -- that's what we ideally want :)

greg-dennis (changes requested):

internal/deviations/deviations.go
```diff
@@ -65,27 +65,91 @@ package deviations
...
 	mpb "github.com/openconfig/featureprofiles/proto/metadata_go_proto"
 	"github.com/openconfig/ondatra"
 )
 
-// lookupDutDeviations returns the deviations for the specified dut.
-func lookupDUTDeviations(dut *ondatra.DUTDevice) *mpb.Metadata_Deviations {
+func matchRegex(regex string, deviceInfo string) bool {
+	match, err := regexp.MatchString(regex, deviceInfo)
+	if err != nil {
+		log.Errorf("Error with regex match %v", err)
```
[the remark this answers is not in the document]
prinikasn: Agree, if the regex is invalid it should be a hard error. Logged with error with Exitf instead of Errorf here so that this function returns a bool and keeps the if condition below simple.

internal/deviations/deviations.go
```diff
@@ -65,27 +65,91 @@ package deviations
...
+	}
+	return match
+}
+
+func isPlatformExceptionsValid(platformExceptions *mpb.Metadata_PlatformExceptions) error {
```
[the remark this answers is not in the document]

internal/deviations/deviations.go

internal/deviations/deviations.go
```diff
@@ -65,27 +65,91 @@ package deviations
...
+		if err := isPlatformExceptionsValid(platformExceptions); err != nil {
+			return nil, err
+		}
 		if dut.Device.Vendor().String() != platformExceptions.GetPlatform().Vendor.String() {
 			continue
 		}
+		// Exit test if both hardware_model_regex and hardware_model fields are set.
+		if hardwareModelRegex := platformExceptions.GetPlatform().GetHardwareModelRegex(); hardwareModelRegex != "" {
+			// If hardware_model_regex is set and does not match, continue
+			if !matchRegex(hardwareModelRegex, dut.Device.Model()) {
```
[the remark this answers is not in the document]

internal/deviations/deviations.go

internal/deviations/deviations.go
prinikasn: Agree. done

internal/deviations/deviations.go
greg-dennis: Unfortunately, it makes the difference between the two functions unclear, and the Go style rule is to prefix these functions with "must": https://google.github.io/styleguide/go/decisions.html#must-functions

Was it avoided, because it requires updating lots of callers?

internal/deviations/deviations.go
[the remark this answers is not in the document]
greg-dennis: Ug, Go code should really handle and propagate errors, leaving the "must" behavior at the top-level.
https://google.github.io/styleguide/go/decisions.html#must-functions

I know you moved the matchRegex calls into the if-condition at my suggestion, but at the time I didn't connect that suggestion to my suggestion here that this should result in a hard error. Idiomatic Go would just prefer that you propagate the error here and nest the matchRegex call. In fact, you can just get rid of matchRegex at his point and inline the regexp.MatchString call directly.

greg-dennis (changes requested):

internal/deviations/deviations.go
prinikasn: thank you for pointing to this, I did not know about the `must` prefix. I updated the function names accordingly.

Yes, I did not want to change the name of the `lookupDUTDeviations` function, because it is called from every deivaiton setter and is referenced in the README file. I felt changing this name might also create confusion for PR authors who have existing PRs adding new deviations.

internal/deviations/deviations.go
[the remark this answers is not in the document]
prinikasn: done.

internal/deviations/deviations.go
```diff
@@ -65,27 +65,77 @@ package deviations
...
+func matchRegex(regex string, deviceInfo string) bool {
+	match, err := regexp.MatchString(regex, deviceInfo)
+	if err != nil {
+		log.Exitf("Error with regex match %v", err)
+	}
+	return match
+}
+
+func lookupDeviations(dut *ondatra.DUTDevice) (*mpb.Metadata_PlatformExceptions, error) {
+	var matchedPlatformException *mpb.Metadata_PlatformExceptions = nil
```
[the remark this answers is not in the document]

internal/deviations/deviations.go
```diff
@@ -65,27 +65,77 @@ package deviations
...
+		// TODO(prinikasn): Remove after hardware_model field is removed.
+		matchedHwRepeated := false
 		for _, hardwareModel := range platformExceptions.GetPlatform().HardwareModel {
 			if dut.Device.Model() == hardwareModel {
-				return platformExceptions.GetDeviations()
+				matchedHwRepeated = true
 			}
 		}
+
+		if len(platformExceptions.GetPlatform().GetHardwareModel()) != 0 && !matchedHwRepeated {
```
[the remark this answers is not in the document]

greg-dennis (approved):

PR #1792, opened 2023-06-27
cisco-leaf-ref-changes

saarvind6481:
We have added interface-ref and leaf validation in this release. Modifying cisco code accordingly


guoshiuan (approved):
b/289541211 has the internal test results.

PR #1797, opened 2023-06-27
Update qos_basic_test.go

saarvind6481:


guoshiuan (approved):
b/289541912 has the internal test results.

PR #1800, opened 2023-06-27
Update wrr_traffic_test.go

saarvind6481:


guoshiuan (approved):
b/289542974 has the internal test results.

PR #1802, opened 2023-06-27
[TE-8.1] Change the gnoikill to validate the process is killed and restarted successfully

mojiiba:
This pull adds new flag GNXIUseSameProcess to account for devices that use same process to serve all gNXI services. It also change the to validate if the kill is applied correctly.


LimeHat:
I find this semantics confusing; usually gNXI means gNOI, gNMI, gNSI (hence the X part); however the example in the flag description implies that you mean all grpc-based services (gribi+p4rt+gNXI) instead? If so, would you consider renaming it?

Also not sure why the default value of the flag is true.

internal/args/args.go
```diff
@@ -37,4 +37,5 @@ var (
 	FanNamePattern           = flag.String("arg_fan_name_pattern", "", "This name pattern is used to filter out Fan components.")
 	FabricChipNamePattern    = flag.String("arg_fabricChip_name_pattern", "", "This name pattern is used to filter out FabricChip components.")
 	CheckInterfacesInBinding = flag.Bool("arg_check_interfaces_in_binding", true, "GNOI tests perform interface status validation based on all interfaces. This can cause flakiness in testing environments where only connectivity of interfaces in binding is guaranteed.")
+	GNXIUseSameProcess       = flag.Bool("arg_gnxi_use_same_process", true, "the device uses the same OS process to serve gnxi services. Use this flag to change the expected behavior correspondingly. E.g., killing gribi requires killing the related process that leads to killing other services (gnmi, gnoi, p4rt).")
```
greg-dennis, line 40: Please don't create new flags if we can avoid it. Let's try to use the metadata proto deviation values exclusively.

PR #1803, opened 2023-06-27
Support for self-deletion of CI VM instances.

bstoll:
In some cases, it is possible for Cloud Build to terminate a job immediately without cleanup.  This can leave VM instances orphaned because Cloud Build is the only task that can clean up an instance.  With this change, VMs will delete themselves after 2 hours.  This is greater than the 2700s (45min) limit currently set, so it should only be possible to trigger this event if something has gone wrong.

### DIFF
--- a/cloudbuild/cleanup.sh
+++ b/cloudbuild/cleanup.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 sleep 2h
-echo "Execution timeout reached; deleting self"
-NAME=$(curl -X GET http://metadata.google.internal/computeMetadata/v1/instance/name -H 'Metadata-Flavor: Google')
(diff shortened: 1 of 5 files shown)


cloudbuild/cleanup.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+sleep 2h
+echo "Execution timeout reached; deleting self"
+NAME=$(curl -X GET http://metadata.google.internal/computeMetadata/v1/instance/name -H 'Metadata-Flavor: Google')
```
greg-dennis, line 4: These should probably be quoted, because Google bash style says to always quote command substitutions

greg-dennis (approved):

cloudbuild/cleanup.sh
bstoll: I updated the quoting.

greg-dennis (approved):

PR #1812, opened 2023-06-29
Fix for protocol_dscp_rules_for_vrf_selection_test failure in EOS

prakashbadri-arista:
The test was failing for the following reasons
- The subinterface was not a routed port which is not supported in EOS when the parent is routed
- OC configuration for applying vrf selection policy in the ingress interface was incorrect and erroring out. After fixing the above 2, the test is passing fine


feature/experimental/policy/policy_vrf_selection/otg_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
```diff
@@ -203,6 +203,10 @@ func configNetworkInstance(t *testing.T, dut *ondatra.DUTDevice, vrfname string,
 	} else {
 		si.GetOrCreateVlan().GetOrCreateMatch().GetOrCreateSingleTagged().VlanId = ygot.Uint16(vlanID)
 	}
+	s4 := si.GetOrCreateIpv4()
```
Ankur19, line 206: Do we need this? Since we already do this in `getSubInterface()`

Ankur19 (approved):
LGTM!

PR #1814, opened 2023-06-29
Remove gNOI-5.2 Traceroute do_not_fragment case

arvbaska1:
Removing the "Check traceroute with IPv4 DF bit" case in gNOI-5.2 traceroute_test, as this behavior is not required.
& the deviation for this.


guoshiuan (approved):
b/289542964 has the internal test results.

PR #1816, opened 2023-06-29
Remove sw version deviation for cisco

kjahed:
Removes software version deviation for cisco in gNMI-1.10 and gNOI-4.1.


feature/gnoi/os/tests/osinstall/metadata.textproto
```diff
@@ -15,16 +15,4 @@ platform_exceptions: {
...
-    hardware_model: "CISCO-8808"
-    hardware_model: "CISCO-8202-32FH-M"
-    hardware_model: "8201"
-    hardware_model: "xrd"
-  }
-  deviations: {
-    sw_version_unsupported: true
-  }
-}
+}
```
greg-dennis, line 18: restore the new line at end of file

greg-dennis (changes requested):

feature/gnoi/os/tests/osinstall/metadata.textproto
kjahed: done

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -499,14 +499,21 @@ func TestSoftwareVersion(t *testing.T) {
...
 			// Arista_7280 OC component EOS has parent type Chassis
-			if dut.Model() == "DCS-7280CR3K-32D4" {
+			case "DCS-7280CR3K-32D4":
 				if got == chassisType {
 					t.Logf("Got a valid parent %v with a type %v for the component %v", v, got, os)
 				} else {
 					t.Errorf("Got a parent %v with a type %v for the component %v, want %v", v, got, os, chassisType)
 				}
-			} else {
+			case "CISCO-8808":
```
prinikasn, line 510: Is there a better way we can do this instead of a switch case with `dut.Model`? 

Deviations can now be enabled for a specific vendor+hardware model so perhaps we could add a deviation here instead of a switch case with `dut.Model()` ?

kjahed: Thanks @prinikasn. Do you have any example of a metadata.proto that targets specific hardware model for additional deviations? We have a set of deviations that apply to all models and one additional deviation that only applies to CISCO-8808. Having two 'platform' entries with the same vendor gives
```
Error looking up deviations: cannot have more than one match within platform_exceptions fields platform:{vendor:CISCO}
```

feature/gnmi/ate_tests/telemetry_basic_check_test/metadata.textproto
```diff
@@ -8,12 +8,26 @@ testbed: TESTBED_DUT_ATE_2LINKS
 platform_exceptions: {
   platform: {
     vendor: CISCO
+    hardware_model: "CISCO-8202-32FH-M"
```
prinikasn, line 11: hardware_model has been deprecated. 

You can use hardware_model_regex here. 

I think for this section you could use hardware_model_regex: "820.*|xrd" and in the next section you could use hardware_model_regex: "CISCO-8808" so that they are mutually exclusive.

kjahed: Thanks I was wondering why it's throwing an error all the sudden. Fixed and pushed.

guoshiuan (dismissed):
b/290127695 has the results for ATE and OTG tests.

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -499,19 +499,17 @@ func TestSoftwareVersion(t *testing.T) {
...
-				} else {
-					t.Errorf("Got a parent %v with a type %v for the component %v, want %v", v, got, os, chassisType)
-				}
-			} else {
-				if got == supervisorType {
+			switch got {
+			case supervisorType:
+				t.Logf("Got a valid parent %v with a type %v for the component %v", v, got, os)
+			case linecardType, chassisType:
+				if deviations.OSComponentParentIsChassis(dut) || deviations.OSComponentParentIsLinecard(dut) {
```
greg-dennis, line 506: This check isn't as strong as it could be. Note that if the expected parent type is CHASSIS but the actual type is LINECARD, this incorrectly succeeds. I suggest refactoring this to calculate the expected parent type first, i.e.

```
want := supervisorType
switch {
case deviations.OSComponentParentIsChassis(dut):
  want = chassisType
case deviations.OSComponentParentIsLinecard(dut):
  want = linecardType
}

if got != want {
  t. Errorf(...)
}
```

greg-dennis (changes requested):
Also, please sync forward and resolving the conflicts

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
kjahed: Thanks Greg. Pushed the changes.

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -499,19 +500,20 @@ func TestSoftwareVersion(t *testing.T) {
...
-					t.Logf("Got a valid parent %v with a type %v for the component %v", v, got, os)
-				} else {
-					t.Errorf("Got a parent %v with a type %v for the component %v, want %v", v, got, os, chassisType)
-				}
+			want := []oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT{supervisorType}
+			if deviations.OSComponentParentIsChassis(dut) {
+				want = []oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT{chassisType}
+			}
+			if deviations.OSComponentParentIsLinecard(dut) {
+				want = []oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT{supervisorType, linecardType}
```
greg-dennis, line 508: I don't get it. If the `deviations.OSComponentParentIsLinecard` is true, why are we permitting the parent to be of a supervisor type?

greg-dennis (changes requested):

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
kjahed: For CISCO 8808 some OS components have linecard as parent while others have supervisor. I updated the deviation name to reflect that fact.

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -499,19 +500,20 @@ func TestSoftwareVersion(t *testing.T) {
...
-				}
+			want := []oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT{supervisorType}
+			if deviations.OSComponentParentIsChassis(dut) {
+				want = []oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT{chassisType}
+			}
+			if deviations.OSComponentParentIsSupervisorOrLinecard(dut) {
+				want = []oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT{supervisorType, linecardType}
+			}
+
+			if slices.IndexFunc(want, func(w oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT) bool {
```
greg-dennis, line 511: I think you can use slices.Contains(want, got)

greg-dennis (changes requested):

feature/gnmi/ate_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
kjahed: slices.Contains didn't work due a type inference error since got is a union type. Changing want to the union type does not work either because the type does not implement comparable.

greg-dennis: Got it, thanks

proto/metadata.proto
```diff
@@ -279,6 +279,10 @@ message Metadata {
     bool isis_interface_afi_unsupported = 85;
     // Devices don't support modify table entry operation in P4 Runtime.
     bool p4rt_modify_table_entry_unsupported = 86;
+    // Parent of OS component is of type SUPERVISOR or LINECARD
```
greg-dennis, line 282: each of these comments should end with a period

feature/gnmi/otg_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
```diff
@@ -502,19 +503,20 @@ func TestSoftwareVersion(t *testing.T) {
...
+			if deviations.OSComponentParentIsChassis(dut) {
+				want = []oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT{chassisType}
+			}
+			if deviations.OSComponentParentIsSupervisorOrLinecard(dut) {
+				want = []oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT{supervisorType, linecardType}
+			}
+
+			if slices.IndexFunc(want, func(w oc.E_PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT) bool {
+				return w == got
+			}) != -1 {
```
greg-dennis, line 516: It's easier to read conditions with a positive polarity, i.e. I would change this to check "== -1" and swap the Errorf and Logf statements.

greg-dennis (changes requested):
Just a couple nits ...

proto/metadata.proto
kjahed: thanks, fixed

feature/gnmi/otg_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
kjahed: Got it. Fixed.

greg-dennis (dismissed):
Approving for Go quality

guoshiuan (approved):

PR #1825, opened 2023-06-30
update lldp feature.textproto

dplore:
add 2 missing paths


dplore:
This is a duplicate of #1696

PR #1464, opened 2023-04-27
initial version of security_aaa_tacacs Feature Profile

bensons:

Issue #280, opened 2022-06-28
AAA feature profile


bstoll:
As mentioned in #266, we need to flesh out user authentication/management feature profiles.  It is probably too complicated to bundle all authentication setup into the base system feature profile.

We need a "system_aaa" feature profile (or something along these lines) to support:
* General AAA management (TACACS, etc)
* Local user management (Move the existing paths from "system" to "system_aaa" feature profile)

We will also need test coverage for this, but that can be handled in another issue.

PR #1826, opened 2023-06-30
Add system/aaa feature.textproto

dplore:


dplore:
@bensons would you consider this instead of #1464 ?

bstoll (approved):

bstoll:
Looks like this also resolves an old issue #280.

PR #1831, opened 2023-07-05
Gnmi 1.7 readme for npu asic counters leaves

Wanmathy:


feature/gnmi/ate_tests/telemetry_npu_asic_counters_test/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+*   Packet Counters - Host-Block
+
+    *   Subscribe to the following, and capture the values of the 4 leaves(1),
+        * /packet/host-interface-block/in-pkts
+        * /packet/host-interface-block/out-pkts
+        * /packet/host-interface-block/in-bytes (TBD)
+        * /packet/host-interface-block/in-bytes (TBD)
+    *   Define variance = 0.99
+    *   Send ping packets to ATEPort2IP with count 1000. Wait for 20 secs.
```
rszarecki, line 44: Why you think this traffic would enter host-interface-block? My understanding is that ATEPort2IP is in FIB and have resolved ARP/ND, hence traffic would be forwarded.
Did you mean DUTPort2IP ?
Or you want to set TTL=1?

feature/gnmi/ate_tests/telemetry_npu_asic_counters_test/README.md
```diff
@@ -0,0 +1,176 @@
...
+    *   drop/interface-block/state/oversubscription (TBD)
+
+*   Drop Counters - Queueing Block
+    *   Connect a gRIBI client to the DUT, make it become leader, flush and inject the following,
+
+        NHG#1 --> NH#1 {next-hop: ATEPort2IP}
+    *   Subscribe to /drop/queueing-block/oversubscription, and capture the leaf value(1).
+    *   Send traffic, 
+        * Flow 1 from ATE port 1 to ATE port 2 with line rate 75%.
+        * Flow 2 from ATE port 3 to ATE port 2 at 75% linerate.
```
rszarecki, line 123: This should not work.
The OpenConfig states `Number of packets dropped due to oversubscription of the integrated-circuit subsystem block.` Not interface oversubscription.
Depending how we define "Queueng Block" but I would say that **block** capacity would be rather >> 100GE !
One possible scenario is oversubscription of HBM I/O (2Tbps in most ASICS). But it may be more complicated due to On-Chip/Off-Chip buffers and "slicing" of chip. 

I think right test would be to generate ingress traffic ~ 50% of chip capacity on single chip and direct it to single egress interfaces. Base on my understanding of HBM I/O this should overload it and cause drops.

rszarecki (changes requested):

feature/gnmi/ate_tests/telemetry_npu_asic_counters_test/README.md
Wanmathy: Icmp packets are sent here, with remote ip <atePort2IP> which will enter the host interface block cpu counters. Have seen regular icmp packets hitting the cpu counters without setting ttl as 1 or any other value.

rszarecki: Maybe this is something implemetation specific (but I doubt) but in general:
For following topology:
`HOST1--ROUTER1-HOST2`
Wnen HOST1 sendts ICMP ECHO Request with destination IP equal to address of HOST2, There is no reason ROUTER1's ASIC to expose this packet on ASIC's host interface (conencting ASIC to routers control plane).
This is because destination IP is NOT owned by ROUTER1, TTL are not expired, no router-alerts are set. It is just transit traffic as any other.
There is one possibility when ASIC host-path is involved - when ARP request must be generated. When the MAC address of HOST2  interface is unknown/timed-out.

You can simply fix test by setting ping destination to be **DUT**Port2IP, intead of **ATE**Port2IP.

rszarecki (changes requested):

feature/gnmi/ate_tests/telemetry_npu_asic_counters_test/README.md
Wanmathy: Sure, let me confirm on the behavior internally and get back on this.

feature/gnmi/ate_tests/telemetry_npu_asic_counters_test/README.md
saarvind6481: The oversubscription is per VOQ, which is per egress interface.
We can verify per VOQ traffic getting evicted to HBM and verify it.
That is my understanding,  I can check and get back.

feature/gnmi/telemetry_npu_asic_counters_test/tests/README.md
```diff
@@ -0,0 +1,176 @@
...
+    *   Verify in-bytes and out-bytes (TBD)
+
+*   Packet Countes - Queueing Block
+
+    *   Subscribe to the following, and capture the values of the 4 leaves(1),
+        * /packet/queueing-block/in-pkts(TBD)
+        * /packet/queueing-block/in-bytes(TBD)
+        * /packet/queueing-block/out-pkts,
+        * /packet/queueing-block/out-bytes
+    *   Define variance = 0.98.
```
rszarecki, line 58: why 98% ? why not 99% as previous? why not 99.9%

feature/gnmi/telemetry_npu_asic_counters_test/tests/README.md
```diff
@@ -0,0 +1,176 @@
...
+    * Stop traffic.
+
+*   Packet Counters - Host-Block
+
+    *   Subscribe to the following, and capture the values of the 4 leaves(1),
+        * /packet/host-interface-block/in-pkts
+        * /packet/host-interface-block/out-pkts
+        * /packet/host-interface-block/in-bytes (TBD)
+        * /packet/host-interface-block/in-bytes (TBD)
+    *   Define variance = 0.99
```
rszarecki, line 43: why 99%

feature/gnmi/telemetry_npu_asic_counters_test/tests/README.md
```diff
@@ -0,0 +1,176 @@
...
+    *   Start traffic, verify traffic received on atePort2.
+    *   Shutdown ateport2. Wait for 30 secs.
+    *   Verify /drop/lookup-block/no-route counter populated (2).
+        * Step 2-Step 1(no-route), calculation 30 * pps.       
+    *   Unshut ateport2, verify traffic recovers on atePort2.
+
+    *   Configure static Null0 route for outerDst_IP1.
+    *   After 30 secs, Verify /drop/lookup-block/no-nexthop leaf populated(3).
+        * Step 3- Step 1(no-nexthop), calcuation 30 * pps.
+    *   Unconfigure static route, and verify traffic recovers on atePort2, stop traffic.
```
rszarecki, line 89: I'm not sure if Null0 is "no-nexthop". Actually Null0 is a next-hop.
The better test would be:
- remove NH#1 by gRIBI w/o removing NGH#1.

feature/gnmi/telemetry_npu_asic_counters_test/tests/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+    *   Send traffic with custom checksum in ipv4 header.
```
rszarecki, line 91: Why custom checksum?

feature/gnmi/telemetry_npu_asic_counters_test/tests/README.md
```diff
@@ -0,0 +1,176 @@
...
+    *   After 30 secs, Verify /drop/lookup-block/invalid-packet leaf populated(4).
+        * Step 4 - Step 1(invalid-packet), calculation 30*pps.
+    *   Stop traffic. 
+
+    *   /drop/lookup-block/forwarding-policy <TBD>
+
+    *   Send traffic to Outer_Dst_IP1, and verify traffic received on atePort2.
+    *   Remove ipv4 address on dutport1.
+    *   After 30 secs, Verify /drop/lookup-block/incorrect-software-state populated(5).
```
rszarecki, line 100: Actuallay, how DUT can judge that lack of IP address is incorrect? It follows controller intent, so IT IS CORRECT STATE fron DUT point of view.
To exercise this counter we would need to homehow introduce illegal register combination on ASIC. IMHO this is impossible to test in black-box acenario, outside of vendor engineering lab.

feature/gnmi/ate_tests/telemetry_npu_asic_counters_test/README.md
saarvind6481: Even if a single interface is oversubscribed, we have to evict the packets to HBM and we will get VOQ drops on the ingress, since the oq scheduler on the egress will not provide grants.
The egress interface can be any where on the system on different NP, slice , different line card, but  the buffering will happen on ingress NP.

feature/gnmi/telemetry_npu_asic_counters_test/tests/README.md
```diff
@@ -0,0 +1,176 @@
...
+*   OuterDstIP_1, OuterSrcIP_1, OuterSrcIP_2: IPinIP outer IP addresses.
+*   InnerDstIP_1, InnerSrcIP_1: IPinIP inner IP addresses.
+*   ATEPort2IP: testbed assigned interface IP to ATE port 2.
+*   ATEPort3IP: testbed assigned interface IP to ATE port 3.
+*   Connect a gRIBI client to the DUT, make it become leader and inject the following,
+
+    NHG#1 --> NH#1 {next-hop: ATEPort2IP, network-instance:DEFAULT}
+    OuterDstIP_1/32  --> NHG#1.
+
+*   Send IPinIP traffic to OuterDstIP_1. Validate that ATE port-2 receives the IPinIP traffic.
```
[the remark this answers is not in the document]
Wanmathy: Yes, modified the statement now.

feature/gnmi/telemetry_npu_asic_counters_test/tests/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+*   Packet Counters - Host-Block
+
+    *   Subscribe to the following, and capture the values of the 4 leaves(1),
+        * /packet/host-interface-block/in-pkts
+        * /packet/host-interface-block/out-pkts
+        * /packet/host-interface-block/in-bytes (TBD)
+        * /packet/host-interface-block/in-bytes (TBD)
+    *   Define variance = 0.99
+    *   Send ping packets to ATEPort2IP with count 1000. Wait for 20 secs.
```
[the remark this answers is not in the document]
Wanmathy: Have updated the step to "Send ICMP ECHO REQUEST form ATEPort1IP to DUTPort2IP".

feature/gnmi/telemetry_npu_asic_counters_test/tests/README.md
Wanmathy: There could be other control/slowpath packets and so kept variance as 99%

feature/gnmi/telemetry_npu_asic_counters_test/tests/README.md
Wanmathy: Have verified this case, where /drop/lookup-block/incorrect-software-state gets populated with the mentioned trigger. Sending ipv4 traffic, with ip disabled on the interface causes this counter to populate due to incorrect state during lookup.

feature/gnmi/telemetry_npu_asic_counters_test/tests/README.md
Wanmathy: Had written the calculation part for out-pkts as an exact match for expected packet count. The expectation for variance is a redundant item here, have modified the variance value to 0.999.

feature/gnmi/telemetry_npu_asic_counters_test/tests/README.md
saarvind6481: No-nexhop can also simulated with glean adjacency.

dplore:
This collection of paths is no longer a requirement or priority for featureprofiles development at this time.  The focus should instead be on the set of paths added in https://github.com/openconfig/public/pull/715.

To minimize confusion, this PR should be closed and a new PR for a  README will need to be created for a test covering https://github.com/openconfig/public/pull/715.

PR #1849, opened 2023-07-07
Run Go and Protobuf checks on a daily schedule

wenovus:


dplore (approved):
Note that results of cron initiated CI checks can be viewed in the 'Actions' tab of the github web UI.

PR #1857, opened 2023-07-10
DP-1.14: Updating queue names for queue names for Juniper

charantejag504:
DP-1.14: Updating queue names for queue names for Juniper

For Juniper function - ConfigureJuniperQos, Updating queue names  in DP-1.14 according to recent changes in https://github.com/openconfig/entity-naming/pull/22


greg-dennis (approved):
Approving for Go quality

PR #1860, opened 2023-07-10
Initial commit for TC RT-1_20

### LINKED ISSUE #1590: RT-1.20: BGP 2-byte and 4-byte ASN support with policy

s-nandu:
Initial commit for TC - RT-1.20: BGP 2-Byte and 4-Byte ASN support with policy

This will resolve #1590


feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
```diff
@@ -0,0 +1,445 @@
...
+	peer := bgpDut1.AddPeer().WithPeerAddress(ateParams.peerIP).WithLocalASN(ateParams.localAS)
+
+	if connectionType == connInternal {
+		peer.WithTypeInternal()
+	} else {
+		peer.WithTypeExternal()
+	}
+
+	if ateParams.isV4 {
+		network := iDut1.AddNetwork("bgpNeti1")
```
greg-dennis, line 372: to be more concise, define the three networks before the if-statement and set them within the if-statements, .ie.

```
network1 := iDut1.AddNetwork("bgpNeti1")
network2 := iDut1.AddNetwork("bgpNeti2")
network3 := iDut1.AddNetwork("bgpNeti3")

if ateParams.isV4 {
  network1.IPv4().WithAddress(prefixes[0]).WithCount(1)
  ...
} else [
  network1.IPv6().WithAddress(prefixes[0]).WithCount(1)
  ...
}
```

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
```diff
@@ -0,0 +1,445 @@
...
+		network.IPv4().WithAddress(prefixes[0]).WithCount(1)
+		network.BGP().WithNextHopAddress(ateSrc.IPv4).AddASPathSegment(55000, 4400, 3300)
+
+		network = iDut1.AddNetwork("bgpNeti2")
+		network.IPv4().WithAddress(prefixes[1]).WithCount(1)
+		network.BGP().WithNextHopAddress(ateSrc.IPv4).AddASPathSegment(55000, 7700)
+		network.BGP().WithNextHopAddress(ateSrc.IPv4).Communities().WithPrivateCommunities("200:1")
```
greg-dennis, line 379: no need to set NextHopAddress here; it is already set on the previous line

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
```diff
@@ -0,0 +1,445 @@
...
+		network.BGP().WithNextHopAddress(ateSrc.IPv4)
+	} else {
+		network := iDut1.AddNetwork("bgpNeti1")
+		network.IPv6().WithAddress(prefixes[0]).WithCount(1)
+		network.BGP().WithNextHopAddress(ateSrc.IPv6).AddASPathSegment(55000, 4400, 3300)
+
+		network = iDut1.AddNetwork("bgpNeti2")
+		network.IPv6().WithAddress(prefixes[1]).WithCount(1)
+		network.BGP().WithNextHopAddress(ateSrc.IPv6).AddASPathSegment(55000, 7700)
+		network.BGP().WithNextHopAddress(ateSrc.IPv6).Communities().WithPrivateCommunities("200:1")
```
greg-dennis, line 392: no need to set NextHopAddress here; it is already set on the previous line

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
```diff
@@ -0,0 +1,445 @@
...
+
+	pg.GetOrCreateApplyPolicy().ImportPolicy = []string{policyName}
+
+	return ni_proto
+}
+
+func createBgpNeighbor(nbr *bgpNbr, dut *ondatra.DUTDevice) *oc.NetworkInstance_Protocol {
+	d := &oc.Root{}
+	ni1 := d.GetOrCreateNetworkInstance(deviations.DefaultNetworkInstance(dut))
+	ni_proto := ni1.GetOrCreateProtocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP")
```
greg-dennis, line 418: no underscores in go variable names, only camelCase

greg-dennis (changes requested):

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
```diff
@@ -0,0 +1,445 @@
...
+	rejectCommunity     = "REJECT-COMMUNITY"
+	rejectAspath        = "REJECT-AS-PATH"
+	aclStatement1       = "10"
+	prefixSubnetRangeV4 = "24..32"
+	prefixSubnetRangeV6 = "64..128"
+)
+
+var prefixV4 = []string{"30.0.2.0/24", "40.0.2.0/24", "50.0.2.0/24"}
+var prefixV6 = []string{"3001::/64", "4001::/64", "5001::/64"}
+var community_Set = []string{"200:1"}
```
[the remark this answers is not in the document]
s-nandu: Addressed

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
s-nandu: Addressed with latest commit

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
s-nandu: Addressed with latest commit

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
s-nandu: Addressed with latest commit

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
s-nandu: Addressed with latest commit

greg-dennis (dismissed):
Approving for Go quality

sachendras (approved):

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
```diff
@@ -0,0 +1,449 @@
...
+	return niProto
+}
+
+func createBgpNeighbor(nbr *bgpNbr, dut *ondatra.DUTDevice) *oc.NetworkInstance_Protocol {
+	d := &oc.Root{}
+	ni1 := d.GetOrCreateNetworkInstance(deviations.DefaultNetworkInstance(dut))
+	niProto := ni1.GetOrCreateProtocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP")
+	bgp := niProto.GetOrCreateBgp()
+
+	global := bgp.GetOrCreateGlobal()
```
arvbaska1, line 425: Could you enable Global IPv4/IPv6 AFI as well here ?

global.GetOrCreateAfiSafi(oc.BgpTypes_AFI_SAFI_TYPE_IPV4_UNICAST).Enabled = ygot.Bool(true)
global.GetOrCreateAfiSafi(oc.BgpTypes_AFI_SAFI_TYPE_IPV6_UNICAST).Enabled = ygot.Bool(true)

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
```diff
@@ -0,0 +1,449 @@
...
+	global.As = ygot.Uint32(nbr.localAS)
```
arvbaska1, line 426: The BGP local-AS has to be different from the Global BGP AS# here.
Test needs to be modified to use different local-AS & global-AS numbers.

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
```diff
@@ -0,0 +1,449 @@
...
+func applyBgpPolicy(policyName string, dut *ondatra.DUTDevice) *oc.NetworkInstance_Protocol {
+	d := &oc.Root{}
+	ni1 := d.GetOrCreateNetworkInstance(deviations.DefaultNetworkInstance(dut))
+	niProto := ni1.GetOrCreateProtocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP")
+	bgp := niProto.GetOrCreateBgp()
+
+	pg := bgp.GetOrCreatePeerGroup("ATE")
+	pg.PeerGroupName = ygot.String("ATE")
+
+	pg.GetOrCreateApplyPolicy().ImportPolicy = []string{policyName}
```
arvbaska1, line 414: BGP RPL should be under Peer-group AFI, could you use RoutePolicyUnderAFIUnsupported to add RPL under PG directly ?

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
sachendras: I would recommend Cisco to address this comment in a separate Pull. Because of the following,

As per the [yang definition](https://github.com/openconfig/public/blob/85a6f6ef5216fad43187ea60d0fe6722b855b293/release/models/bgp/openconfig-bgp-common-multiprotocol.yang#L198C1-L208), the default is "disabled". Even though there are discussions for making global/ipv4-unicast and global/ipv6-unicast enabled as a default in pull/815, the Pull is still under discussions. Given the current situation, we may need further discussions on this topic to decide if this request needs to be addressed by vendor specific deviation or we just consider sending global/afi-safi==enabled for ipv4-unicast and ipv6-unicast.

We have future tests in the pipeline already https://github.com/openconfig/featureprofiles/pull/1776 to check specifically on this requirement for compliance w/ the yang definitions w/ expectations that the non-compliant implementations would have a deviation.

For viz: @dplore for further clarifications as he was actively involved on this topic.

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
sachendras: @s-nandu to update the code as requested.

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
sachendras: @s-nandu to address the code change as requested.

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
arvbaska1: Yes by default Global AFI/SAFI is Disabled/False for cisco, so we need to set it to true to enable them at bgp global.
Same is done in all of the other BGP tests.

dplore: Global enable should be ok.  However, we should be able to enable at the neighbor level and leave the global enable as-is.  Is that supported by Cisco?

sachendras: Well if I understand the ask here clearly, it is that, even though the test is enabling the afi/safi for IPv4 and IPv6 under the neighbor hierarchy, ask is to enable it under the global AFI/SAFI as well? This is irrespective of the fact that the yang states that the default == false https://github.com/openconfig/public/blob/master/release/models/bgp/openconfig-bgp-common-multiprotocol.yang#L217 at the global level. I know we (Google) internally decided that while pushing configs we will ensure that the global/ipv4-unicast is updated to enabled == true always. However, this arrangement doesn't exist for global/ipv6-unicast. Seems some implementations require that global/ipv6-unicast should also be == enabled and just having neighbor/ipv6-unicast == enabled wont work? This is an agreement we need to make internally.

sachendras: So post internal discussions, lets enable IPv4-unicast and IPv6-unicast at the global level as well while we are still working on deciding what should be the long term state here. 

@s-nandu, can you please accommodate the change?

dplore: I'd prefer to see the test only enable IPv6 at the neighbor level.

Cisco may submit a PR to add a deviation to this test which also enables IPv6 at the global level.

sachendras: @dplore based on our offline conversation, seems we had an agreement to not explicitly enable both v4-unicast and v6-unicast at the global level and all implementations that require these to be explicitly enabled even though these address-families are already enabled either at the neighbor or at the peer-group level need to have 2 different deviations, one for enabling v4 at the global level and another for enabling v6 at the global level. Is your understanding the same as mine?

aaronmillisor: > 

@dplore re: ipv6 enabled at the neighbor, but disabled globally, what is the goal of this test? IOS XR in particular forbids configuration of an afi on a neighbor if that afi is not enabled globally. It seems that this would be a misconfiguration, no?

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
s-nandu: Updated with latest commit

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
s-nandu: Updated with latest commit

feature/experimental/bgp/ate_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
sachendras: Well, as per the yang definition https://github.com/openconfig/public/blob/master/release/models/bgp/openconfig-bgp-common-multiprotocol.yang#L217, default for Global AFI/SAFI is enabled=false.

Intent is that, if this boolean is set at the neighbor/peer-group level, that should be the config. If not defined at the peer-group/neighbor level then the setting in the global level should take effect, which btw has a default of AFI/SAFI disabled.

sachendras (approved):

PR #1861, opened 2023-07-10
Clarify bootz-3.1 wanted behavior.

xw-g:
Bootz should check for OV.

It would be useful for Bootz to also support an "insecure" mode where OV check is skipped (e.g. in some lab environment). That will require an update to https://github.com/openconfig/bootz first, and then update this test case later to cover the "insecure" mode.


xw-g:
Close this for now in favor of https://github.com/openconfig/featureprofiles/pull/1864

PR #1865, opened 2023-07-11
test staticcheck

wenovus:


wenovus:
So it's confirmed that an error annotation by itself does not cause the CI check to fail.

PR #1867, opened 2023-07-12
RT-5.2 TestPlan: extending test to include SDN Port ID

rszarecki:


xw-g (approved):

xw-g (commented):
I think we should create an issue, and link it to the TODO?

xw-g (changes requested):
I think we should create an issue, and link it to the TODO?

PR #1869, opened 2023-07-12
Factor out a function for setting a QoS input classifier

greg-dennis:


internal/qoscfg/qoscfg.go
```diff
@@ -0,0 +1,40 @@
...
+	t.Helper()
+	intf := qos.GetOrCreateInterface(intfID)
+	intf.GetOrCreateInterfaceRef().SetInterface(intfID)
+	if dut.Vendor() != ondatra.CISCO {
+		intf.GetOrCreateInterfaceRef().SetSubinterface(0)
+	}
+	if deviations.InterfaceRefConfigUnsupported(dut) || deviations.IntfRefConfigUnsupported(dut) {
+		intf.InterfaceRef = nil
+	}
+	intf.GetOrCreateInput().GetOrCreateClassifier(classType).SetName(className)
```
bstoll, line 38: Should there be something like `intf.GetOrCreateInput().GetOrCreateClassifier(classType).SetType(classType)` here too?  I don't think creating the classifier will implicitly initialize the type field.

bstoll (commented):
I still need to verify the results of the test after this change (or was this already done?), but this change generally SGTM.

internal/qoscfg/qoscfg.go
greg-dennis: It does implicitly set the classType. The class type is the key to the list, and the list key is implicitly set when `GetOrCreate*` methods are called. I will be removing all the redundant key-setting that happens throughout the qos tests.

greg-dennis:
> I still need to verify the results of the test after this change (or was this already done?), but this change generally SGTM.

Thank you! So I ran the qos_basic_test across all vendors, and it passed on Arista, Cisco, and Juniper, and it failed on Nokia in the same way the Nokia test is failing today.

What is the easiest way to say "run all the qos tests with this change"?

bstoll (approved):

PR #1880, opened 2023-07-13
Move deviation to metadata textproto

prinikasn:
PR contains the following changes:
 - move 7 deviations from flag to metadata textproto
 - Remove hardware_model from metadata proto where ever file is modified
 - Move private deviations to byexceptions.go file


prinikasn (commented):
/fptest virtual

greg-dennis (approved):

PR #1881, opened 2023-07-13
Fix P4Runtime performance test in EOS

prakashbadri-arista:
The test was failing intermittently in EOS because the ATE ingress
counters check for verifying the expected packetOut counters was done
right after injecting the traceroute/LLDP/GDP packets with no delay.

There is no guarantee that all the packetOut requests have been
guaranteed by that time. Added a 30s delay, to account for any trailing
inflight packets.


ram-mac:
These changes are already there in the test. So we can close this PR.

Issue #1734, opened 2023-06-09
gNOI-5.3 update to fetch artifact


dplore:
[gNOI-5.3 test](https://github.com/openconfig/featureprofiles/blob/main/feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go) has a remaining TODO item.  The test must be updated to fetch an artifact indicated by the healthz response.  

That is:

* The test must verify the [healthz get response](ComponentStatus) produces a reresponse for component type CHASSIS
* Invoke the [health check rpc](https://github.com/openconfig/gnoi/blob/6884aef91ca81a8009eb349bcaeb9dbabe80f333/healthz/healthz.proto#L57) for component type CHASSIS to trigger generation of  [Artifact ID(s)](https://github.com/openconfig/gnoi/blob/891b1d10e452f3fc4f4138b5ec80456860b88337/healthz/healthz.proto#LL112C7-L112C7) equivalent to 'show tech support.   
*  Invoke  [ArtifactRequest](https://github.com/openconfig/gnoi/blob/891b1d10e452f3fc4f4138b5ec80456860b88337/healthz/healthz.proto#L150) to transfer the requested artifact ID.  

Pass criteria is 
* getResponse contains valid health data
* Check RPC returns valid CheckResponse
* ArtifactResponse contains data.  Fail is if the ArtifactReponse cannot be successfully retrieved or contains zero data.

sgirishj:
Pull Req https://github.com/openconfig/featureprofiles/pull/1890 opened to resolve this

PR #1890, opened 2023-07-14
gNOI-5.3: Adding chassis component artifact fetch test

sgirishj:
Adding Artifact Fetch test for Chassis component. This PR resolves https://github.com/openconfig/featureprofiles/issues/1734


feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go
```diff
@@ -117,3 +117,52 @@ func findProcessByName(ctx context.Context, t *testing.T, dut *ondatra.DUTDevice
...
+				},
+				{
+					Name: "component",
+					Key:  componentName,
+				},
+			},
+		},
+	}
+	t.Logf("Executing Healthz Check RPC for component %v", componentName["name"])
+	validResponse2, err2 := gnoiClient.Healthz().Check(context.Background(), chkReq)
```
LimeHat, line 140: Nit: it is idiomatic in golang to reuse the same `err` variable, i.e. no need to create err2, ..., errX

Also you may want to shorten/use more meaningful names for response vars, e.g. `chkRes` pairs nicely with `chkReq`, `artRes` with `artReq`, etc

feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go
```diff
@@ -117,3 +117,52 @@ func findProcessByName(ctx context.Context, t *testing.T, dut *ondatra.DUTDevice
...
+	t.Logf("Artifacts received for component %v: %v", componentName["name"], artifacts)
+	// Fetch artifact details by executing ArtifactRequest and passing the artifact ID along.
+	for _, artifact := range artifacts {
+		artId := artifact.GetId()
+		t.Logf("Executing ArtifactRequest for artifact ID %v", artId)
+		artReq := &hpb.ArtifactRequest{
+			Id: artId,
+		}
+		// Verify that a valid response is received.
+		validResponse3, err3 := gnoiClient.Healthz().Artifact(context.Background(), artReq)
```
LimeHat, line 158: same as above

feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go
```diff
@@ -117,3 +117,52 @@ func findProcessByName(ctx context.Context, t *testing.T, dut *ondatra.DUTDevice
...
+		if err3 != nil {
+			t.Fatalf("Unexpected error on executing Healthz Artifact RPC: %v", err3)
+		}
+		h1, err31 := validResponse3.Header()
```
LimeHat, line 162: same as above

LimeHat (changes requested):

greg-dennis (approved):
Approving for Go quality only

feature/gnoi/system/tests/copying_debug_files_test/README.md
```diff
@@ -6,10 +6,16 @@ Validate that the debug files can be copied out of the DUT.
 
 ## Procedure
 
-*   Issue gnoi System.KillProcessRequest to the DUT to crash a software process. 
-*   Issue gnoi System.Healthz Get RPC to chassis.
-*   TODO: Validate that the device returns the vendor relevant information for
-    debugging via gnoi Healthz.Artifact
+*   Test #1 - Software Process Health Check
+    *   Issue gnoi System.KillProcessRequest to the DUT to crash a software process.
```
sachendras, line 10: I think this test should be rewritten completely

There should only be one function "TestCopyingDebugFiles". Please don't add the new function "TestChassisComponentArtifacts". Idea is to crash a process to create core dumps. Then use healthz.artifacts RPC to collect Core dumps + "Show tech-support or RSI" eqvt.

Also, in place of "SIGNAL_KILL" operation, probably "SIGNAL_ABRT" operation should be used to "terminate the process immediately and dump a core file." as per 
 https://github.com/openconfig/gnoi/blob/main/system/system.proto#L324-L330

sachendras: So based on the offline discussion, the concern here is that the healthz.attributes RPC that checks for different components wont be able to get crash dump for a failed service like the BGP process or RPD process because there doesnt exist corresponding component in OC.

I feel the intent of this test is to ensure we are able to,
1. Collect core-dumps when they happen
2. "healthz artifacts" RPC works for different components.

In order for "1", we can use the gNOI.file.Get operation to collect the core dump. But we need to know the file name of the core dump file. This needs us to use scrape through the show tech output to find the core dump file. Challenge is that the Show tech output will be different for different vendors. I will keep this comment open and lets see if @dplore and @marcushines have some suggestions.

dplore: So far we only have a requirement  healthz check and artifact for component CHASSIS.  ref:  #1734

dplore: It is true we expect the CHASSIS artifact to be similar to `show tech support` or other vendor equivalent system wide debug data.  Parsing of the artifact is not required for this test, only confirmation that it is sent and is non-zero bytes.      (ref #1734 )

The use case for this is the healthz artifact for CHASSIS  will be passed to the device developer for root cause analysis.  The device operator is not expected to parse or otherwise directly read this data for this use case.

sachendras: Thanks Darren

Since in "Test #1", killing the process is unrelated to what the subtest is eventually accomplishing with gNOI.Healthz.Getrequest and  gNOI.Healthz.Getresponse on component chassis, it is best to maintain Test#1 given that it provides coverage for gNOI.Healthz.Getrequest and  gNOI.Healthz.Getresponse RPC operations and ensures that the implementation is error free in the NOS being tested. Hence I see value in Test#1. In future as healthz becomes more mature, killing the BGP process and then checking for healthz on the respective component would make most sense. For the timebeing we can maintain test#1 as is.

Test#2 on the other hand (introduced by this pull) with gNOI.healthz.checkrequest and gnoi.healthz.CheckResponse and testing for corresponding Artifacts is accurate.

sachendras (approved):

PR #1896, opened 2023-07-17
Use latest version of Ondatra and enable atomic-aware ygnmi API

wenovus:
Per ondatra@v0.2.0 `telemetry-atomic` nodes no longer provide path access to descendants: https://github.com/openconfig/ondatra/releases/tag/v0.2.0


wenovus:
I've checked the KNE test results for both and I don't believe any of the failures are a direct cause of my change. There is one test case that failed at where I modified the code:
```
    --- FAIL: TestRouteRemovalNonDefaultVRFFlush/flushNonZeroReference (36.34s)
    route_removal_non_default_vrf_test.go:289: ipv4-entry/state/prefix does not contain entry, expected: 198.51.100.0/24
```

Even if my change triggered this failure, it would be a gNMI compliance issue and not an issue in this PR.

bstoll (approved):

PR #1898, opened 2023-07-17
Move interface_ref_config_unsupported to byexceptions file.

prinikasn:
This PR has the following changes:
 - Replaces `deviation_intf_ref_config_unsupported` with `deviation_interface_ref_config_unsupported` .
 - Add proto field for each private deviation.
 - Add logic for private deviations to check flag value if present and if not use metadata proto value.


Ankur19:
Hi @prinikasn `interface_ref_config_unsupported` and `intf_ref_config_unsupported` are the same deviation. Why not just have one?

greg-dennis (approved):

Ankur19 (approved):
LGTM

PR #1903, opened 2023-07-18
Sec-3.1 native user

MarcCharlebois:
1. Allow configuration of users/aaa through native model
2. Add port 57400 for gNMI connection
3. add ssh.Password() auth method
4. Deviation for configuring users/aaa (deviation_set_native_user)


internal/deviations/deviations.go
```diff
@@ -572,6 +572,12 @@ func QOSOctets(dut *ondatra.DUTDevice) bool {
 	return lookupDUTDeviations(dut).GetQosOctets()
 }
 
+// SetNativeUser creates a user and assigns role/rbac to that user via native model.
+func SetNativeUser(dut *ondatra.DUTDevice) bool {
+	logErrorIfFlagSet("deviation_set_native_user")
```
prinikasn, line 577: Calling this function `logErrorIfFlagSet` is not required here since this is a new deviation and there is no flag called`deviation_set_native_user`.

This accesor can simply be:

```
func SetNativeUser(dut *ondatra.DUTDevice) bool {
	return lookupDUTDeviations(dut).GetSetNativeUser()
}
```

MarcCharlebois: Thanks @prinikasn , fixed in latest commit

cfernanz (approved):

PR #1905, opened 2023-07-19
Remove hardware model field from metadata proto

prinikasn:
PiperOrigin-RevId: 549169993

### DIFF
--- a/feature/experimental/p4rt/tests/p4rt_election/metadata.textproto
+++ b/feature/experimental/p4rt/tests/p4rt_election/metadata.textproto
@@ -5,12 +5,3 @@ uuid: "73bf3d88-e5c1-4d30-9fae-f40b8363044f"
 plan_id: "P4RT-2.1"
 description: "P4RT Election"
 testbed: TESTBED_DUT_ATE_2LINKS
-platform_exceptions: {
-  platform: {
-    vendor: NOKIA
-    hardware_model: "7250-ixr10e"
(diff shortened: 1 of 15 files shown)


feature/experimental/p4rt/tests/p4rt_election/metadata.textproto
```diff
@@ -5,12 +5,3 @@ uuid: "73bf3d88-e5c1-4d30-9fae-f40b8363044f"
 plan_id: "P4RT-2.1"
 description: "P4RT Election"
 testbed: TESTBED_DUT_ATE_2LINKS
-platform_exceptions: {
-  platform: {
-    vendor: NOKIA
-    hardware_model: "7250-ixr10e"
```
greg-dennis, line 11: Were you intending to remove the whole platform_exceptions or just the hardware_model fields?

greg-dennis (changes requested):

feature/experimental/p4rt/tests/p4rt_election/metadata.textproto
prinikasn: This PR is supposed to only remove hardware_models. 

But in this case, I realiazed there are no deviations listed. The only deviation Nokia had for this test was removed ~2 days ago, but this section was not cleaned up. Hence I removed the whole platform_exceptions section here.

greg-dennis (approved):

PR #1909, opened 2023-07-19
Fix BGP policy config to only configure each policy once.

wenovus:


arulkumarsekar (approved):
test passed on Juniper..

PR #1912, opened 2023-07-19
Fixing KNE textproto for Nokia

MarcCharlebois:
Changed interface key nomenclature, removed public image link for tagged srlinux:latest

### DIFF
--- a/topologies/kne/nokia/srlinux/dut.textproto
+++ b/topologies/kne/nokia/srlinux/dut.textproto
@@ -4,7 +4,7 @@ nodes: {
     vendor: NOKIA
     model: "ixr10"
     config: {
-        image: "ghcr.io/nokia/srlinux:latest"
+        image: "srlinux:latest"
         file: "config.cfg"
         cert: {
             self_signed: {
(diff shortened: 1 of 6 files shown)


topologies/kne/nokia/srlinux/dut.textproto
```diff
@@ -4,7 +4,7 @@ nodes: {
     vendor: NOKIA
     model: "ixr10"
     config: {
-        image: "ghcr.io/nokia/srlinux:latest"
```
alexmasi, line 7: Any reason for switching this value? It was originally targeting the public image by default to make it as easy as possible for users to run tests (without having to load in a custom image into their k8 cluster)

The interface name changes look good to me

MarcCharlebois: The main purpose for FP is testing OC compliance. OC is only available for specific chassis types that requires a license, which means that in *most* cases people would have a custom image loaded.

If someone wants to run the public image in a k8s cluster, I would suggest just the textproto from https://github.com/openconfig/kne/tree/main/examples/nokia/srlinux-services as it's more of a generic use case.

alexmasi: Okay thats fine with me, can you also update https://github.com/openconfig/featureprofiles/blob/131970e41bb2a5359e8e943f43c32ead408a05d0/cloudbuild/virtual.sh#L114 for the CI trigger?

MarcCharlebois: No problem, I can package it in this PR, just to confirm that it looks right

Change from: 
` sed -i "s/ghcr.io\/nokia\/srlinux:latest/us-west1-docker.pkg.dev\/gep-kne\/nokia\/srlinux:ga/g" /tmp/kne/"${kne_topology}" `

to: 
`sed -i "s/srlinux:latest/us-west1-docker.pkg.dev\/gep-kne\/nokia\/srlinux:ga/g" /tmp/kne/"${kne_topology}"`

alexmasi (approved):

PR #1913, opened 2023-07-19
P4RT-1.1 , update DUT links from 2 to 4

rsindhujha:
Changed metadata.textproto from TESTBED_DUT_ATE_2LINKS to TESTBED_DUT_ATE_4LINKS, to have a better chance of different nodes getting selected during run.


sachendras (approved):

LimeHat:
@rsindhujha you're modifying p4rt-2.1 instead of p4rt-1.1 (which is stated in the subject). Is this intended?

rsindhujha:
> @rsindhujha you're modifying p4rt-2.1 instead of p4rt-1.1 (which is stated in the subject). Is this intended?

Thanks for pointing out @LimeHat. My bad, modified the wrong file. Fixed it now.

greg-dennis (approved):
Approving for Go quality (there is no Go code here)

greg-dennis (changes requested):
Actually, hold on, I'm not sure the point of this PR. The testbed should specify what the test requires, and this test only request 2 links, right?

rsindhujha:
> Actually, hold on, I'm not sure the point of this PR. The testbed should specify what the test requires, and this test only request 2 links, right?

On Juniper ptx10001, with 2 port topology there are  chances that both ports are picked from the same FAP. To help ports getting picked from multiple FAPs increasing it to 4 port topology helps.

sachendras:
We are fixing this internally hence the pull is not required anymore, closing!

PR #1931, opened 2023-07-22
fixing the otg_test cases my_station_mac_test.go

abhijit-dhar:
The test case "feature/experimental/interface/my_station_mac/otg_tests/my_station_mac_test/my_station_mac_test.go" was failing with cisco vxr dut.

I also saw that the code required to support the deviation "--deviation_mac_address_missing", has been deleted. I brought that back, simply because - if this deviation is not there, the test will fail much earlier from the code segment "gnmi.Get(t, dut, gnmi.OC().System().MacAddress().RoutingMac().State()); strings.ToUpper(got)".

I am aware that after a certain time the deviations will be accepted from the metadata file. So I have inserted the deviation in the metadata file as well.


self-maurya:
@abhijit-dhar Could you please resolve the merge conflicts.

self-maurya (approved):

PR #1936, opened 2023-07-24
Update metadata.textproto for drained_configuration_convergence

Ankur19:
Added these deviations after modifications to the internal/setup.go files that apply to both full_configuration_replace and drained_configuration_convergence.


feature/experimental/system/gnmi/benchmarking/ate_tests/drained_configuration_convergence_time/metadata.textproto
```diff
@@ -10,6 +10,7 @@ platform_exceptions: {
     vendor: CISCO
   }
   deviations: {
+    isis_instance_enabled_not_required: true
```
arvbaska1, line 13: Not specific to this PR change, but the IS-IS ‘instance’ leaf has been deprecated now from the OC ISIS model https://github.com/openconfig/public/blob/master/release/models/isis/openconfig-isis.yang#L243.
Should we remove the usage of this from the FP tests as well ?

Ankur19: Sounds good! Could you submit a PR to change the name of the deviation to `isis_instance_enabled_required` and enable it for the other two vendors?

arvbaska1: Sure i will open a PR, thanks.

gforeal (approved):

PR #1938, opened 2023-07-24
OTG TE-3.7 - Updating flows with multiple ports as destinations

### LINKED ISSUE #582: OTG: TE-3.7: Base Hierarchical NHG Update

octpetre:
Resolves #582 
- adding multiple ports as destination for flow
- using flow stats
- updated lossPct to float32


jasdeep-hundal (changes requested):
@octpetre : Can we update to full matching functionality with the ATE test now that we have egress tracking available?

ANISH-GOTTAPU:
> @octpetre : Can we update to full matching functionality with the ATE test now that we have egress tracking available?

@jasdeep-hundal : I have aligned the test with ATE and included egress tracking. Please note that we are using 12-bit tracking bits instead of the 15-bit ones used in ATE to ensure seamless operation of the test against both software and hardware.

jasdeep-hundal (approved):

PR #1361, opened 2023-03-30
RT-5.4 : Adding ping before traffic test to resolve adjacency

agvarghe-cisco:
- Adding ping before traffic test to resolve adjacency

PR #1942, opened 2023-07-25
Clarify reason for separate LACP/IP ATE push in RT-5.4 ATE test

jasdeep-hundal:


LimeHat:
Hi @jasdeep-hundal, this change was needed to avoid intermittent packet loss due to unresolved ARP/ND. A conversation about this was held in #1390 and also #1361 

Any reason to remove this?

jasdeep-hundal:
@LimeHat : I was unaware of that reason. Changed this patch to adding that documentation instead.

LimeHat:
Lgtm, thank you for the change!

guoshiuan (approved):

PR #1946, opened 2023-07-26
deviations for drained_configuration

Ankur19:
* New deviations added for L1 configuration in ISIS and OSPF configuration in BGP.

* For BGP MED, apply ACCEPT_ROUTE policy and during verification only check routes advertised by ATE port1.


feature/experimental/system/gnmi/benchmarking/ate_tests/drained_configuration_convergence_time/drained_configuration_convergence_time_bgp_test.go
```diff
@@ -73,6 +74,10 @@ func setMED(t *testing.T, dut *ondatra.DUTDevice, d *oc.Root) {
 	actions5 := stmt.GetOrCreateActions()
 	setMedBGP := actions5.GetOrCreateBgpActions()
 	setMedBGP.SetMed = oc.UnionUint32(bgpMED)
+	actions5.PolicyResult = oc.RoutingPolicy_PolicyResultType_ACCEPT_ROUTE
```
cprabha, line 77: Please add condition here to accept only bgp routes.  Otherwise it will accept all routes in RIB and causes issue.

stc := stmt.GetOrCreateConditions()
stc.InstallProtocolEq = oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP

We have added similar way in https://github.com/openconfig/featureprofiles/blob/main/feature/experimental/bgp/ate_tests/bgp_remove_private_as/bgp_remove_private_as_test.go

cprabha: Same change is needed in setASPath function. Thanks

Ankur19: Using this is currently a source of contention between different vendors since there are two ways to do the same thing(https://github.com/openconfig/public/issues/931).

Anyways, I am now filtering only on the BGP routes advertised by Port 1.

cprabha: @Ankur19  gotMED is filtered based on prefixes but we still import all routes into BGP.  "isConverged" is called but exits after max iterations are completed.  We never hit pass criteria.  But scripts works as gotMED is having entries equal to routecount. Is it possible to add one more check to see if gotMED equal to routeCount  ?

feature/experimental/system/gnmi/benchmarking/ate_tests/drained_configuration_convergence_time/drained_configuration_convergence_time_bgp_test.go
```diff
@@ -246,10 +251,26 @@ func verifyBGPSetMED(t *testing.T, dut *ondatra.DUTDevice, ate *ondatra.ATEDevic
...
+			for _, route := range routes {
+				ip, _, _ := net.ParseCIDR(route.GetPrefix())
+				pref = append(pref, route.GetPrefix())
+				if ip.Mask(mask).Equal(masked) {
+					idx := route.GetAttrIndex()
+					if idx >= uint64(len(attrs)) {
+						t.Errorf("Invalid attr-index %d for prefix: %s", idx, route.GetPrefix())
+						continue
+					}
+					gotSetMED = append(gotSetMED, attrs[idx].GetMed())
```
cprabha, line 271: Please add length of gotSetMED should be equal to routeCount.  ?

Ankur19: This is already being done in cmp.Diff. The `wantSetMED` list is of length equal to routeCount. 

So if `gotSetMED` is not of the same length, the test would throw error.

rszarecki (approved):

PR #1962, opened 2023-07-31
Add retry support when updating GitHub comments.

bstoll:
We have seen the GitHub API occasionally return 5xx errors when updating comments.  This adds basic logic to retry updating comments if there is a failure.


tools/ci-trigger/pr_test.go
```diff
@@ -234,3 +236,72 @@ func TestPopulateTestDetail(t *testing.T) {
 		t.Errorf("populateModifiedTests(%v): -want,+got:\n%s", modifiedTests, diff)
 	}
 }
+
+func TestWithRetry(t *testing.T) {
+	var retryCount int
```
greg-dennis, line 241: I noticed the language alternating between "retry" and "attempt." I'm not sure if you had a particular reason why you one over another in the various in the code and test, but maybe something to consider. I don't have a strong opinion either way, although some seem to "feel" right for reasons I can't articulate yet.

greg-dennis (dismissed):
A couple optional nits

tools/ci-trigger/pr_test.go
bstoll: That one variable name did make it inconsistent.  Each individual retry is an attempt, but the overall operation is a retry.  Hopefully it makes more sense now.

greg-dennis (approved):

PR #1963, opened 2023-08-01
Do not overwrite grpc.dialoptions when they are set by the caller

mojiiba:
this pull ensure that grpc.dialoptions  are not overwritten if they are set by the caller. related pull: https://github.com/openconfig/ondatra/pull/85


topologies/binding/options.go
```diff
@@ -87,11 +87,11 @@ func (d *dialer) loadCertificates() (*x509.CertPool, tls.Certificate, error) {
 	}
 	trusBundle := x509.NewCertPool()
 	if !trusBundle.AppendCertsFromPEM(caCertBytes) {
-		return nil, tls.Certificate{}, fmt.Errorf("error in loading ca trust bundle")
+		return nil, tls.Certificate{}, err
```
greg-dennis, line 90: err here will be nil. was the intention to return no error here?

topologies/binding/options.go
```diff
@@ -87,11 +87,11 @@ func (d *dialer) loadCertificates() (*x509.CertPool, tls.Certificate, error) {
...
 	}
 	keyPair, err := tls.LoadX509KeyPair(d.CertFile, d.KeyFile)
 	if err != nil {
-		return nil, tls.Certificate{}, fmt.Errorf("could load the client keys")
+		return nil, tls.Certificate{}, err
```
greg-dennis, line 94: err here will be nil. was the intention to return no error here?

topologies/binding/options.go
```diff
@@ -101,43 +101,48 @@ func (d *dialer) loadCertificates() (*x509.CertPool, tls.Certificate, error) {
...
-		tc := insecure.NewCredentials()
-		opts = append(opts, grpc.WithTransportCredentials(tc))
-	case d.SkipVerify:
-		tc := credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})
-		opts = append(opts, grpc.WithTransportCredentials(tc))
-	case d.MutualTls:
-		trusBundle, keyPair, err := d.loadCertificates()
-		if err != nil {
-			return nil, err
+	if len(opts) == 0 {
```
greg-dennis, line 104: Rather than this, can you instead change it so that the provide opts are appended to the default opts? That should cause any transport credentials provided in the opts to overwrite the default credentials.

topologies/binding/options.go
```diff
@@ -101,43 +101,48 @@ func (d *dialer) loadCertificates() (*x509.CertPool, tls.Certificate, error) {
...
-	}
-	if d.MaxRecvMsgSize != 0 {
-		opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(int(d.MaxRecvMsgSize))))
 	}
 	if d.Timeout == 0 {
 		return grpc.DialContext(ctx, d.Target, opts...)
+	} else {
+		ctx, cancelFunc := context.WithTimeout(ctx, time.Duration(d.Timeout)*time.Second)
+		defer cancelFunc()
+		return grpc.DialContext(ctx, d.Target, opts...)
```
greg-dennis, line 144: let's rework this logic so that there is a single grpc.DialContext call:

```
if d.Timeout > 0 {
  var cancel context.CancelFunc
  ctx, cancel = context.WithTimeout(ctx, d.Timeout * time.Second)
  defer cancel()
}
return grpc.DialContext(ctx, d.Target, opts...)
```

greg-dennis (changes requested):

topologies/binding/options.go
bstoll: I think something like this might work to allow a test to override whatever the base config sets up:

```go
func (d *dialer) dialGRPC(ctx context.Context, overrideOpts ...grpc.DialOption) (*grpc.ClientConn, error) {
	var opts []grpc.DialOption
	switch {
	case d.Insecure:
		tc := insecure.NewCredentials()
		opts = append(opts, grpc.WithTransportCredentials(tc))
	case d.SkipVerify:
		tc := credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})
		opts = append(opts, grpc.WithTransportCredentials(tc))
	case d.MutualTls:
		trusBundle, keyPair, err := d.loadCertificates()
		if err != nil {
			return nil, err
		}
		tls := &tls.Config{
			Certificates: []tls.Certificate{keyPair},
			RootCAs:      trusBundle,
		}
		tlsConfig := credentials.NewTLS(tls)
		opts = append(opts, grpc.WithTransportCredentials(tlsConfig))
	}
	if d.Username != "" {
		c := &creds{d.Username, d.Password, !d.Insecure}
		opts = append(opts, grpc.WithPerRPCCredentials(c))
	}
	if d.MaxRecvMsgSize != 0 {
		opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(int(d.MaxRecvMsgSize))))
	}
	if d.Timeout != 0 {
		retryOpt := grpc_retry.WithPerRetryTimeout(time.Duration(d.Timeout) * time.Second)
		opts = append(opts,
			grpc.WithStreamInterceptor(grpc_retry.StreamClientInterceptor(retryOpt)),
			grpc.WithUnaryInterceptor(grpc_retry.UnaryClientInterceptor(retryOpt)),
		)
		var cancelFunc context.CancelFunc
		ctx, cancelFunc = context.WithTimeout(ctx, time.Duration(d.Timeout)*time.Second)
		defer cancelFunc()
	}
	opts = append(opts, overrideOpts...)
	return grpc.DialContext(ctx, d.Target, opts...)
}
```

topologies/binding/options.go
mojiiba: addressed

topologies/binding/options.go
mojiiba: here the error will not be nil.

topologies/binding/options.go
mojiiba: done

topologies/binding/options.go
mojiiba: done

topologies/binding/options.go
greg-dennis: Looks like this wasn't quite done right. In bstoll's example overrideOpts is the parameter name, but in what you have, it's declared as a local variable and left empty?

mojiiba: my bad, will fix that

greg-dennis (approved):

PR #1966, opened 2023-08-01
Remove IS-IS instance leaf use & update deviation

arvbaska1:
- In the latest OC IS-IS yang model the leaf instance has been deprecated 
- Updating tests to not use this leaf by default, and 
- use deviation isis_instance_enabled_required for vendors requiring this config


feature/experimental/isis/ate_tests/isis_change_lsp_lifetime_test/metadata.textproto
```diff
@@ -39,5 +37,6 @@ platform_exceptions: {
     missing_value_for_defaults: true
     interface_enabled: true
     default_network_instance: "default"
+    isis_instance_enabled_required: true
```
Ankur19, line 40: Why the flip? If isis.Global().Instance leaf is not supported, I would name the deviation as `isis_global_instance_name_unsupported`

arvbaska1: Its changes as per this discussion here https://github.com/openconfig/featureprofiles/pull/1936#discussion_r1272656562.
This leaf is deprecated , but some other vendors still require it i think, so have named this deviation such.

Ankur19: Aah. Thanks for reminding me! LGTM!

guoshiuan (approved):

PR #1967, opened 2023-08-01
Remove ISISprotocolEnabledNotRequired deviation

arvbaska1:
Removing ISISprotocolEnabledNotRequired deviation from the tests.


proto/metadata.proto
```diff
@@ -288,6 +286,7 @@ message Metadata {
     
     // Reserved field numbers and identifiers.
     reserved 84;
+    reserved 9;
```
prinikasn, line 289: This can be combined with above line.

```
reserved 84, 9;
```

arvbaska1: updated to combine them

prinikasn (approved):

PR #1968, opened 2023-08-01
gNMI-1.10 TestPlan: Fabric redundancy

rszarecki:


feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
```diff
@@ -0,0 +1,71 @@
...
+  *   /components/component/state/serial-no
+  *   /components/component/state/type
+
+  *   /components/component/state/location
+  *   /components/component/state/last-reboot-time
+* store list of present components of FABRIC_CARD type
+
+### test 2 redundancy
+* Power down all but one component
+* Run traffic between ATE por1 and port 2 for 16 millions of packets
```
LimeHat, line 36: Would be good to clarify traffic stream parameters (is 16M a total # of packets or 16 Mpps? also what's the packet size)

rszarecki: added. "on 100kpps rate".
Packet size is of less importance I think.

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
```diff
@@ -0,0 +1,72 @@
+# gNMI-1.14 Fabric redundnacy test
+
+## Summary
+- collect inventory data for each fabric card
+- Verify last restart time is updated
+- verify trassic could be forwarded with only one Fabric Card active.
```
vosipchu, line 6: trassic -> traffic
But for this statement - why do we need to verify that? Probability of all but one fabrics going done is miserable. At the same time in prod we won't ever get to this state. We will more likely drain a device in case of a FC failure to get it back to the full scale. 
So, what's the goal of 1xFC test ?

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Procedure
+### topology and basic setup
+*  Connect OTG port1 to DUT port1 and OTG port2 to DUT port2
+*  Configure IPv6 addresses on both links
+### test 1 Fabric inventory
+
+> NOTE: Theis test is practically identical to Fabric part of
+> https://github.com/openconfig/featureprofiles/tree/main/feature/platform/tests/telemetry_inventory_test;
+> Hence code could be reused and then removed form telemetry_inventory_test.
```
vosipchu, line 16: let's maybe not overcomplicate? Let's remove this note at all and just either keep this in the telemetry test or add here, or keep in both.

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+* collect folloing attributes for each component of FABRIC_CARD type and verify corectness (mostly non-empty string)
```
vosipchu, line 18: folloing ->following

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
```diff
@@ -0,0 +1,72 @@
...
+  *   /components/component/state/hardware-version
+  *   /components/component/state/id
+  *   /components/component/state/mfg-name
+  *   /components/component/state/name
+  *   /components/component/state/oper-status
+  *   /components/component/state/parent
+  *   /components/component/state/part-no
+  *   /components/component/state/serial-no
+  *   /components/component/state/type
+
```
vosipchu, line 29: why do we have an empty line?

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
```diff
@@ -0,0 +1,72 @@
...
+  *   /components/component/state/location
+  *   /components/component/state/last-reboot-time
+* store list of present components of FABRIC_CARD type
+
+### test 2 redundancy
```
vosipchu, line 34: as i mentioned above, i don't think we need this test case. Are we trying to test whether we have a DOA card? I would be more interested to test more what we will get in prod. I've observed with old generations of LC/FC pairs on routers of some vendor that even with little traffic load there were drops observed on fabrics. Testing N-1 redundancy will make more sense (for me). And we can increase the rate and the amount of packets.

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
```diff
@@ -0,0 +1,72 @@
...
+* Power down all but one component
+* Run traffic between ATE por1 and port 2 for 16 millions of packets on 100kpps rate and using 4000B packets.
+* verify loss-lessness (with 10E-6 tolerance); Since single fabric is not overloaded in any form
+  with above traffic pattern 0 losses is expected
+* Power up all components
+
+### test 3 last reboot time
+* Select one component from list of available
+* store last-reboot-time for this component as "No_want"
```
vosipchu, line 43: what do "want" and "no_want" mean? Maybe we can find better names?

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
```diff
@@ -0,0 +1,72 @@
...
+* Power down this component, wait 60 sec.
+* Power up this component
+* Wait
+* get last-reboot-timea and compare with "want"
```
vosipchu, line 47: last-reboot-timea -> last-reboot-time

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
```diff
@@ -0,0 +1,72 @@
...
+  * "No_want" must be smaller (earlier) then recently collectrd last-reboot-time
```
vosipchu, line 48: collectrd -> collected

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
rszarecki: PREVIOUS_REBOOT_TIME

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
rszarecki: Whoever will implement may remove it. After considering if he want take it out of telemetry test.

The telemetry test is one big atomic PASS/FAIL. It do not line-up with WBB Catalog features.
Also we do not know, maybe we will get in future device w/o fabric at all but with 2 x controller-card. or vice-a-versa?

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
rszarecki: If we have last fabric standing and able communicate through it, i proves a) device can operate with failed fabrics. and b) internal handling of packets do not requite at lest N fabric in operation.
Alternativly we can try with failed one fabric. but this will not prove b)

On drain not - it takes time to drain and somthimes there is some ftaffic on device anyway. so between failure of fabric and drein be fully enforced, we relay of redundancy.

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
vosipchu: We define the rules and need specify exactly what to be done. Less ambiguity for them, better results for us.

feature/platform/fabric/otg_tests/fabric_redundancy_test/README.md
```diff
@@ -0,0 +1,70 @@
+# gNMI-1.14 Fabric redundnacy test
+
+## Summary
+- collect inventory data for each fabric card
+- Verify last restart time is updated
+- verify traffic could be forwarded with only one Fabric Card active.
```
vosipchu, line 6: i still think we don't need this case with 1xFC

vosipchu (approved):

PR #1973, opened 2023-08-03
Remove the ATE version of the gNMI-1.10 and gNMI-1.11 tests

jasdeep-hundal:


greg-dennis:
Looks like this revealed a need to update the OTG Changes analysis to allow for deletion of ATE tests

greg-dennis (approved):

PR #1982, opened 2023-08-08
Fix backup_nhg_action_test for EOS

prakashbadri-arista:
The following changes were required to make the test work in EOS
- Configure vrf selection policy on the ingress interface under "BackupNHGRequiresVrfWithDecap" deviation
- Add NH aft and NHG aft in one modifyRequest message
- The test will also need the following configuration in startup configuration. """ vrf selection policy next-hop fallback decapsulation vrf ip hardware fib optimize vrf VRF-A VRF-B VRF-C prefix-length 32 """


feature/experimental/gribi/otg_tests/backup_nhg_action/backup_nhg_action_test.go
```diff
@@ -227,6 +230,19 @@ func TestBackupNHGAction(t *testing.T) {
...
+		ni := d.GetOrCreateNetworkInstance(vrfA)
+		pf := ni.GetOrCreatePolicyForwarding()
+		fp1 := pf.GetOrCreatePolicy(policyID)
+		fp1.SetType(oc.Policy_Type_VRF_SELECTION_POLICY)
+		fp1.GetOrCreateRule(1).GetOrCreateIpv4().Protocol = oc.UnionUint8(ipOverIPProtocol)
+		fp1.GetOrCreateRule(1).GetOrCreateAction().NetworkInstance = ygot.String(vrfA)
+		p1 := dut.Port(t, "port1")
+		intf := pf.GetOrCreateInterface(p1.Name())
+		intf.ApplyVrfSelectionPolicy = ygot.String(policyID)
+		gnmi.Replace(t, dut, gnmi.OC().NetworkInstance(vrfA).PolicyForwarding().Config(), pf)
```
Ankur19, line 244: HI @prakashbadri-arista  can we push this as part of the test:

```vrf selection policy next-hop fallback decapsulation vrf ip hardware fib optimize vrf VRF-A VRF-B VRF-C prefix-length 32
```

prakashbadri-arista: the problem it also requires an agent/process to be restarted if this configuration is not part of the startup config
SandL3Unicast agent needs to be restarted after this config is enabled

Ankur19 (approved):

PR #1984, opened 2023-08-08
Common CI workflow

wenovus:


wenovus:
Unfortunately it looks like static analysis runs out of disk space after getting the cache. Not sure what can be done here.

I'll just post a different PR to speed up `go vet` for now.

PR #1986, opened 2023-08-09
topology.textproto: Corrected the port numbering for the topologies for cPTX

stthummuri:
Corrected the port numbering for the topologies for cPTX


alexmasi (approved):
Thanks!

PR #1987, opened 2023-08-09
P4RT GDP requires vlan in EthernetHeader.

Ankur19:
Add VlanID to Ethernet frames in packetIN and packetOUT


feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetin_test/google_discovery_protocol_packetin_test.go
```diff
@@ -257,14 +279,19 @@ func TestMain(m *testing.M) {
...
 	}
-
-	s := i.GetOrCreateSubinterface(0)
+	var s *oc.Interface_Subinterface
+	if hasVlan && deviations.P4RTGdpRequiresDot1QSubinterface(dut) {
+		s = i.GetOrCreateSubinterface(1)
+		s.GetOrCreateVlan().GetOrCreateMatch().GetOrCreateSingleTagged().VlanId = &vlanID
+	} else {
+		s = i.GetOrCreateSubinterface(0)
+	}
```
sachendras, line 294: If the requirement for the test is to accept tagged frames for a specific VLAN, a sub-interface needs to configured with that VLAN-id. A sub-interface that is not configured for vlan-id, can receive only untagged frames. Therefore the code logic seems incorrect. The NOS implementations that need expliciti VLAN-id configuration are not the ones that need a deviation because their operation is accurate. It doesnt make sense for an implementation that is not configured for dot1q tagging to accept tagged frames. That implementation seems incorrect and should instead use a deviation.

Some relevance to this argument is also available in https://github.com/openconfig/public/blob/master/release/models/vlan/openconfig-vlan.yang#L264-L275

```
 grouping vlan-logical-config {
    description
      "VLAN related configuration that is part of subinterface
      (logical interface) configuration.  These are generally
      L3 VLANs with an id that is local or L2 VLANs attached with a wire
      transport service, such as a local cross-connect or an MPLS
      pseudo-wire.

      For ingress packets this VLAN configuration is used to match packets
      with an interface; for egress packets, the 'single-tagged' and
      'double-tagged' containers may be used to specify the VLAN identifiers
      that are added to packets as they leave the system.";
```

Ankur19: While this is true for a normal ethernet frame being processed by the device, it may not be completely true for this scenario. 

In this scenario, the P4 Info set on the device should define the data plane behavior. And as per the P4Info defined, any packet with the GDP Ether type should be forwarded by the ASIC to the P4RT controller.

I am not sure which implementation is correct here. Maybe the P4 spec would have some answers.

Ankur19: Another thing to note here is that, the bytes being sent can't actually be considered an Ethernet frame (even though it very much looks like one). It is a GDP (Google Discovery Protocol) packet and should never be processed like a ethernet frame.

singh-prem: I agree with @sachendras that we should have subinterfaces enabled on the ROUTER to accept the tagged frame.
For Ethertype = 0x6007. It doesnt seem to be a random payload. Some reference to this EtherType show it as:
 #define ETHERTYPE_DEC_SCA       0x6007 
 ref: https://www.wireshark.org/docs/wsar_html/etypes_8h_source.html
or https://datatracker.ietf.org/doc/html/rfc2896

Even if we assume it as random payload, a dot1q padded frame has an ethertype of 0x8100 and the original ethertype is pushed as second ethertype. Ref: https://en.wikipedia.org/wiki/EtherType

feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetout_test/google_discovery_protocol_packetout_test.go
rsindhujha, line 1: @Ankur19, for GDP Packetout scenario, could you please include the same deviation "p4rt_gdp_requires_dot1q_subinterface" for JUNIPER and update the logic in the go code func configInterfaceDUT ?

feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetin_test/google_discovery_protocol_packetin_test.go
```diff
@@ -475,11 +502,15 @@ func (gdp *GDPPacketIO) GetTrafficFlow(ate *ondatra.ATEDevice, frameSize uint32,
 	ethHeader := flow.Packet().Add().Ethernet()
 	ethHeader.Src().SetValue(*gdp.SrcMAC)
 	ethHeader.Dst().SetValue(*gdp.DstMAC)
-	ethHeader.EtherType().SetValue(int32(*gdp.EthernetType))
+	ethHeader.EtherType().SetValue(int32(0x8100))
+
+	vlan := flow.Packet().Add().Vlan()
+	vlan.Id().SetValue(int32(vlanID))
+	vlan.Tpid().SetValue(int32(*gdp.EthernetType))
```
LimeHat, line 509: I'm no OTG expert, so just a question about tpid and ethertype in this configuration context

In canonical terms, TPID is an indicator that dot1q frame follows and set as 0x8100; and then ethertype of the payload follows the vlan-id
I.e. pseudo-header looks like `<dst_mac> <src_mac> <tpid> <prio/cfi> <vlan_id> <ethertype> <payload>`

Are the tpid/ethertype values reversed here, or it is just an OTG syntax and 0x8100 will be first, and 0x6007 will be second in the actual frame?

feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetout_test/google_discovery_protocol_packetout_test.go
Ankur19: Will do! Thanks

feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetin_test/google_discovery_protocol_packetin_test.go
Ankur19: True. But when I actually print out the bytes of the packet I get:

```
<dst_mac> <src_mac> <0x8100> <prio/cfi> <vlan_id> <tpid> <payload>
```

Ankur19: It may be a standard syntax because, I had to do the same thing in gopacket layers

LimeHat: Okay, got it. Will assume it's just a syntax thing. Thx.

feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetin_test/google_discovery_protocol_packetin_test.go
sachendras: Well I am taking a full U turn now after getting more in to the depths of the expectations and talking to several folks internally. Let me first start w/ packet-out scenario as this is explicitly called out in the HLD

"The controller will inject packets that are to be forwarded by the device to support GDP through specifying the egress_port to which they should be forwarded. Ports are uniquely identified through the mechanism described in P4RT-4, and via the gRPC channel associated with the specified switching chip.
Additionally, the GDP packets may or may not have vlan-id set in them even if the interface is untagged. Device should be able to punt these packets up, and forward them out."

Having quoted that for the packet-out behavior from HLD, expectations for packet-in are also the same i.e.
The controller will program a set of rules for incoming GDP packets w/ ether-type 0x6007 for certain ingress interfaces w/ actions to either punt them to the controller or have them follow the FIB. These frames may carry vlan-id or not.  The device is expected to accept the packets even if the interface is untagged and follow the instructions programmed by the controller. 
Therefore, this is not a regular tagged packet that in traditional practice a NOS can accept only if the interface is configured for VLANs. Hence the code logic in this pull seems accurate and satisfies our production requirement. Any different behavior is a deviation.

Now, how to implement this in the NOS to cover for the latter i.e the packet-in expectations?

In the case of an ethernet frame with Dot1Q header, right after the "source mac address" comes the 2Byte TPID (Tag protocol identifier) of 0x8100. An NOS should be looking for the presence of this TPID. If one exists, ethertype of 0x6007 will appear after a 2Byte offset (offset to cover for VLAN ID etc). If after the source Mac the 2Byte TPID is missing then that is where the ethertype of 0x6007 can be found.

Hope this makes sense?

feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetin_test/google_discovery_protocol_packetin_test.go
```diff
@@ -257,14 +279,19 @@ func TestMain(m *testing.M) {
...
+func configInterfaceDUT(i *oc.Interface, a *attrs.Attributes, dut *ondatra.DUTDevice, hasVlan bool) *oc.Interface {
 	i.Description = ygot.String(a.Desc)
 	i.Type = oc.IETFInterfaces_InterfaceType_ethernetCsmacd
 	if deviations.InterfaceEnabled(dut) {
 		i.Enabled = ygot.Bool(true)
 	}
-
-	s := i.GetOrCreateSubinterface(0)
+	var s *oc.Interface_Subinterface
+	if hasVlan && deviations.P4RTGdpRequiresDot1QSubinterface(dut) {
```
sachendras, line 289: Please add 2xsubinterfaces to cover for production scenario. One subinterface to accept untagged frames and another subinterface to cover for GDP packets w/ dot1Q encapsulation. These subinterfaces should be applied only when a NOS satisfies deviation.

feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetin_test/google_discovery_protocol_packetin_test.go
```diff
@@ -111,32 +112,48 @@ type testArgs struct {
...
 		return err
 	}
 	return nil
 }
 
-// decodePacket decodes L2 header in the packet and returns source and destination MAC and ethernet type.
-func decodePacket(t *testing.T, packetData []byte) (string, string, layers.EthernetType) {
+// decodePacket decodes L2 header in the packet and returns source and destination MAC
+// vlanID, and ethernet type.
+func decodePacket(t *testing.T, packetData []byte) (string, string, uint16, layers.EthernetType) {
```
sachendras, line 131: As discussed, please add 3 types of flows for both Packet-in and Packet-out

=== Packet-in scenarios
* GDP with vlanID - punted to p4rt controller.
* GDP without vlanID - punted to p4rt controller.
* Non-GDP Ethernet without vlanID - using FIB

=== Packet-out scenarios
* Inject GDP with vlanID punted out a non trunk interface
* Inject GDP without vlanID punted out a non trunk interface
* Inject Non-GDP Ethernet without vlanID to follow the FIB

singh-prem: This testcase seems to evolve beyond original requirements. Just curious if non-GDP ethernet injection is a requirement of the solution?

feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetin_test/google_discovery_protocol_packetin_test.go
singh-prem: I think packet out scenario is fine. Its only the packet in behavior which deviates from standard practice imo. I will leave it to NOS developers for any comment.

feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetin_test/google_discovery_protocol_packetin_test.go
Ankur19: The intent is to emulate production traffic which will be non GDP and may be untagged ethernet.

gforeal (approved):

PR #2013, opened 2023-08-14
Add tags to classify which areas tests apply to.

thesrinath:
An empty tag means it applies to all areas.


proto/metadata.proto
```diff
@@ -298,4 +298,15 @@ message Metadata {
...
+
+  enum Tags {
+    AGGREGATION = 0;
+    DATACENTER_EDGE = 1;
+    EDGE = 2;
+    TRANSIT = 3;
+  }
+
+  // The `tags` used to identify the area(s) testcase applies to. An empty tag
+  // is the default implying it applies to all areas.
```
prinikasn, line 310: Also run `make proto/metadata_go_proto/metadata.pb.go` to generate the pb.go files

proto/metadata.proto
```diff
@@ -298,4 +298,15 @@ message Metadata {
   // The `platform` field for each `platform_exceptions` should be mutually
   // exclusive. Duplicate matches will result in a test failure.
   repeated PlatformExceptions platform_exceptions = 5;
+
+  enum Tags {
```
prinikasn, line 302: https://protobuf.dev/programming-guides/proto3/#default

```
For enums, the default value is the first defined enum value, which must be 0.
```

Based on the comment below, I suppose there should be "all = 0" for empty `tag` ?

proto/metadata.proto
```diff
@@ -298,4 +298,16 @@ message Metadata {
...
```
prinikasn, line 302: nits/questions: 

Instead of `DEFAULT` should we rename this to `ALL` so that it reflects what the default is meant to do.

Also instead of tags, should this be renamed to `TestArea` to be more clear, because technically all fields in the metadata proto are like tags?

marcushines: proto format should be

TAGS_UNSPECIFIED
TAGS_AGGREGATION
TAGS_DATA_CENTER_EDGE ...

thesrinath: Done

vosipchu (approved):

sachendras (approved):

PR #2019, opened 2023-08-15
Default label for issues

dplore:
Automatically set the 'enhancement' label for new github issues.   This means new issues are considered 'feature requests' instead of 'bugs' by default.  

The enhancement label can be manually removed to be treated like a bug.

### DIFF
--- a/.github/workflows/label-issue
+++ b/.github/workflows/label-issue
@@ -1,0 +1,21 @@
+name: Default enhancement label for issues
+on:
+  issues:
+    types:
+      - reopened
+      - opened
+jobs:
+  label_issues:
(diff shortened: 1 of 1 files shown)


sachendras (approved):

bstoll:
The file should end with `.yml` - at least to keep the same pattern as all the other workflows.

I don't think this is a good idea.  I don't like having a bot apply tags that could be wrong - it lessens the value of the tag.  How about instead we lock down issue creation and disable the ability to open blank issues?  Then, we can have issue templates type for bugs/feature requests/new test plans/general questions and tag them accordingly at creation time.  This would effectively force all new issues to have some sort of tag applied to them.

PR #2022, opened 2023-08-15
DP-1.3: Adding test support for Nokia

trathod1:
Following changes are made in this PR -
    1. Adding `testNokiaECNConfig` test function for Nokia
    2.  Having queue-id configurations as discussed in b/297188157


bensons:
Internal issue tracking validation: b/296290043

internal/qoscfg/qoscfg.go
```diff
@@ -32,6 +32,16 @@ func SetForwardingGroup(t *testing.T, dut *ondatra.DUTDevice, qos *oc.Qos, group
 	gnmi.Replace(t, dut, gnmi.OC().Qos().Config(), qos)
 }
 
+func SetForwardingGroupWithFabricPriority(t *testing.T, dut *ondatra.DUTDevice, qos *oc.Qos, groupName, queueName string, fabricPriority uint8) {
```
bensons, line 35: This func should have a comment documenting its purpose.
e.g.,
// SetForwardingGroupWithFabricPriority sets a forwarding group with fabric priority for a given queue in the specified qos config

(please correct as appropriate)

feature/qos/tests/qos_ecn_config_test/qos_ecn_config_test.go
```diff
@@ -63,6 +63,14 @@ var (
...
 		},
 	}
+	QoSNokiaEcnConfigTestcases = []Testcase{
+		{
+
+			name: "testNokiaECNConfig",
+			fn:   testNokiaECNConfig,
+		},
+	}
+	maxBurstSize = uint32(268435456)
```
bensons, line 73: Why is maxBurstSize defined at the package level, instead of in the testcase function along with other parameters?

bensons (dismissed):

internal/qoscfg/qoscfg.go
trathod1: Removed this function from the `qoscfg` library. Considering, it is not required after fabric-priority configurations are removed as discussed in b/297188157.

feature/qos/tests/qos_ecn_config_test/qos_ecn_config_test.go
trathod1: Moved this function specific variable from package level to specific function.

xw-g:
I don't think per vendor test function is what we want here. The following is our recommendation, from @vosipchu, @rszarecki, @sachendras, and I: 

Nokia will do the following via #2022:
1. Remove the function: testCiscoECNConfig, testJuniperECNConfig and testNokiaECNConfig.
2. Make necessary update to the vendor agnostic function `QoSEcnConfigTestcases`. This will be reviewed by @rszarecki .
3. This #2022  will not fix errors/failures for other vendors. Other vendors will need to file follow-up PRs (also to be reviewed by @rszarecki to update the vendor agnostic functions, (e.g. deviations if needed). to make them pass.

rszarecki (changes requested):
Please remove any Vendor specific code. The fundamental property of open featureprofiles test is its vendor independence.
The 'switch dut.Vendor()' in Line 102 should gone.
If vendor needs some specific operation in test it MUST be exposed via deviation and gated by respective IF condition.

trathod1:
Updated script with following changes -
 1. Removed all vendor specific functions
 2. Added 2 new deviations -
    - `qos_queue_requires_id`
    - `qos_buffer_allocation_config_required`
 3. Vendor agnostic test functions share configurations, so adding `q` variable as argument to `testECNConfig` and `testQoSOutputIntfConfig`
 4. Moved usage of deviation `state_path_unsupported` and `drop_weight_leaves_unsupported` from vendor specific function to vendor agnostic function

rszarecki (dismissed):

rszarecki (approved):

PR #2031, opened 2023-08-16
CloudBuild Virtual Tests: Start cleanup timer service

bstoll:
The `systemctl enable cleanup.timer` makes the cleanup execute if there is a reboot, but the 'start' command is also needed to start the cleanup timer for the running system.  We have not had many failures lately so this problem did not stand out.


alexmasi (approved):
Note that we have a more resilient solution in the works for KNE presubmit, will enable here as well in the near future

PR #2032, opened 2023-08-16
Add Tags to Edge testcases

thesrinath:


feature/experimental/p4rt/tests/metadata_validation_test/metadata.textproto
```diff
@@ -29,3 +29,4 @@ platform_exceptions: {
     p4rt_modify_table_entry_unsupported: true
   }
 }
+tags: TAGS_EDGE
```
sachendras, line 32: This is not anCX

sachendras: We have feature/experimental/bgp/ate_tests/bgp_remove_private_as/metadata.textproto for RT-1.11: BGP remove private AS

thesrinath: It is tagged as such in your list. Can you update that and I will revert it here?

vosipchu (approved):

PR #2037, opened 2023-08-17
add a new test cases for partial down scenario

xw-g:


feature/gribi/ate_tests/ipv4_entry_test/ipv4_entry_test.go
```diff
@@ -325,6 +325,44 @@ func TestIPv4Entry(t *testing.T) {
 					AsResult(),
 			},
 		},
+		{
+			desc:     "One of the referenced next-hop interfaces is down",
+			downPort: ate.Port(t, "port2"),
```
jasdeep-hundal, line 330: Will need a change like https://github.com/openconfig/featureprofiles/commit/0ef32a2080a63b9b899ae395ca0aaf85e7086969#diff-446a8de58a33efff05bd4b6a8f500e980eb8aac9e163e869a2b62fc483054cc8R301-R306 here as well.

(AFAICT, should merge cleanly with the ipv4_entry_test changes that went in with that patch.)

jasdeep-hundal (changes requested):

PR #2041, opened 2023-08-18
Align TE-3.6 ATE/OTG tests

jasdeep-hundal:


feature/gribi/otg_tests/ack_in_presence_other_routes/metadata.textproto
```diff
@@ -18,7 +18,6 @@ platform_exceptions: {
     vendor: NOKIA
   }
   deviations: {
-    explicit_gribi_under_network_instance: true
```
LimeHat, line 21: @jasdeep-hundal I think it's better to keep it.
we don't necessarily require it when only default NI is used (hence it is was not added to all ATE cases), for consistency it's good to keep it here.

thanks!

prinikasn: IMO, If a deviation is not present in the test, it should also be removed from the metadata file. 

We moved deviations to a per-test metadata file to make it easier to identify which deviations are used in each test.

LimeHat: @prinikasn what do you mean by "not present in the test"? It's right there on L157-L159 (we should keep those lines as well, of course)

LimeHat: I'd say yes.

prinikasn: @LimeHat Ah, ok. 

Since the comment was on the metadata.textproto file and the PR removed the deviation from the test, I thought you meant keep the line in metadata.textproto only. 

Thank you for clarifying.

cfernanz (approved):

PR #2053, opened 2023-08-23
Make ATE port operations deviation settable via a flag

jasdeep-hundal:


octpetre (approved):
Flag works fine in kne

octpetre:
@jasdeep-hundal could you please merge this into main so we can run kne tests which require this

PR #2054, opened 2023-08-23
Fix traffic flow in GDP test. Test was passing without traffic.

Ankur19:
Test was configuring an invalid flow in OTG. That was causing the traffic test to pass without actually passing any traffic.


LimeHat:
Thanks @Ankur19, lgtm

gforeal (approved):

PR #2057, opened 2023-08-24
Update interface_loopback_aggregate_test to apply deviation

Ankur19:
* LoopbackMode can only be applied to member ports for it to work on an Aggregate interface


Ankur19:
@prakashbadri-arista please take a look! Thanks.

gforeal (approved):

PR #2060, opened 2023-08-24
Update ci-trigger archive creation

bstoll:
This PR updates ci-trigger code archive creation behavior.  This is necessary to support offline compilation of Go tests.

* Generate/reuse a single code archive for all vendors instead of per-vendor.
* Vendor go deps into code archive to support offline compilation.
* Process events asynchronously from the GitHub API HTTP triggering event.
* Write PR and SHA to code archive metadata.
* Use go1.21 for ci-trigger container.
* Enable stderr logging in ci-trigger container.
* Add badge_pubsub arg to optionally disable processing badge pubsub events.
* Fix a typo in metadata restoration that could break the "raw log" comment link.

### DIFF
--- a/tools/ci-trigger/pr.go
+++ b/tools/ci-trigger/pr.go
@@ -18,8 +18,10 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"io/fs"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -27,6 +29,7 @@ import (
 
 	"cloud.google.com/go/storage"
 	"github.com/google/go-github/v50/github"
+	"github.com/google/uuid"
 	"google.golang.org/api/cloudbuild/v1"
 
 	"github.com/go-git/go-git/v5"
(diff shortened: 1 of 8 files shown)


tools/ci-trigger/pr.go
```diff
@@ -78,30 +81,62 @@ func (d *deviceType) String() string {
 	return d.Vendor.String() + " " + d.HardwareModel
 }
 
+// createArchive uploads the compressed repository to Object Store and returns the path to the object.
+func (p *pullRequest) createArchive(ctx context.Context, storClient *storage.Client) (string, error) {
```
alexmasi, line 85: are these archives getting cleaned up? Not sure if this is a regression from the current impl

alexmasi (approved):

tools/ci-trigger/pr.go
bstoll: They currently are not cleaned up, but my intent was to set an expiration policy on the gcp bucket to clean them up after 90d or so.

It is an improvement because we are going from one per vendor to only a single file, although bundling the go vendor deps is canceling out a lot of those savings.

PR #2062, opened 2023-08-25
OTG: Align the OTG test TE11.3 with ATE

ANISH-GOTTAPU:
This PR addresses below comments:

1)Needs egress tracking 
2)Needs ATE port shutdown
3)Needs a fix to the 'vip2' constant

Note:
Please note that we are using 12-bit tracking bits instead of the 15-bit ones used in ATE to ensure seamless operation of the test against both software and hardware.


feature/experimental/gribi/otg_tests/backup_nhg_action/backup_nhg_action_test.go
```diff
@@ -467,6 +496,11 @@ func createFlow(t *testing.T, ate *ondatra.ATEDevice, top gosnappi.Config, name
 	innerIPHeader := flow.Packet().Add().Ipv4()
 	innerIPHeader.Src().SetValue(innerSrcIP1)
 	innerIPHeader.Dst().Increment().SetStart(innerDstIP1).SetCount(1)
+	flow.EgressPacket().Add().Ethernet()
+	ipTracking := flow.EgressPacket().Add().Ipv4()
+	ipDst := ipTracking.Dst().MetricTags().Add()
+	ipDst.SetName("ipDstTracking").SetOffset(0).SetLength(12)
+
```
jasdeep-hundal, line 503: The second procedure does require checking for source IP in one case, can we add that for this case here?

https://github.com/openconfig/featureprofiles/tree/main/feature/experimental/gribi/otg_tests/backup_nhg_action#procedure

jasdeep-hundal (changes requested):

feature/experimental/gribi/otg_tests/backup_nhg_action/backup_nhg_action_test.go
ANISH-GOTTAPU: @jasdeep-hundal : Custom offset is not supported yet, so I have validated src ip and dst ip separately. Please check if that works.

jasdeep-hundal: Validating src/dst separately is preferred because it's more explicit (and it's represented easily in the disaggregated OTG flow stats as well.)

jasdeep-hundal (approved):

PR #2073, opened 2023-08-29
ci-trigger: Add support for launching physical tests

bstoll:
This PR adds frontend/UI support for launching physical hardware tests from PR comments.  Note that the "all" keyword does not include physical devices for now - I want physical tests to be explicitly launched while ironing out any problems.


tools/ci-trigger/pr.go
```diff
@@ -151,6 +158,41 @@ func (p *pullRequest) createBuild(ctx context.Context, buildClient *cloudbuild.S
...
+				for _, v := range physicalDevice.Tests {
+					if v.Status != "pending authorization" {
+						continue physicalDeviceLoop
+					}
+				}
+				jobID, err := uuid.NewRandom()
+				if err != nil {
+					return fmt.Errorf("uuid.NewRandom device %q: %w", physicalDevice.Type.String(), err)
+				}
+				p.Physical[i].ArchivePath = objPath
```
greg-dennis, line 176: isn't p.Physical[i] the same as physicalDevice here and below?

tools/ci-trigger/pr.go
```diff
@@ -151,6 +158,41 @@ func (p *pullRequest) createBuild(ctx context.Context, buildClient *cloudbuild.S
 				}
 			}
 		}
+	physicalDeviceLoop:
+		for i, physicalDevice := range p.Physical {
+			if physicalDevice.Type == d {
+				if len(physicalDevice.Tests) == 0 {
+					continue
+				}
+				for _, v := range physicalDevice.Tests {
```
greg-dennis, line 167: nit "t"->"v"?

tools/ci-trigger/pr.go
```diff
@@ -151,6 +158,41 @@ func (p *pullRequest) createBuild(ctx context.Context, buildClient *cloudbuild.S
...
+				if err != nil {
+					return fmt.Errorf("json.Marshal device %q: %w", physicalDevice.Type.String(), err)
+				}
+				result := pubsubTopic.Publish(ctx, &pubsub.Message{Data: jsonMsg})
+				id, err := result.Get(ctx)
+				if err != nil {
+					return fmt.Errorf("pubsubTopic.Publish device %q: %w", physicalDevice.Type.String(), err)
+				}
+				glog.Infof("Sent Physical Test Job for PR%d at commit %q for device %q via %s", p.ID, p.HeadSHA, physicalDevice.Type.String(), id)
+				for j := range physicalDevice.Tests {
```
greg-dennis, line 191: iterator over the values instead, i.e.

for _, t := range physicalDevice.Tests {
  t.Status = "setup"
}

greg-dennis (changes requested):

tools/ci-trigger/pr.go
bstoll: Done

tools/ci-trigger/pr.go
bstoll: Done

tools/ci-trigger/pr.go
bstoll: It is but there were no pointers being used so modifying physicalDevice directly wouldn't work.  I updated this and think it is easier to follow now.

greg-dennis (approved):

PR #2115, opened 2023-09-12
Migrate off remaining deprecated raw api uses

greg-dennis:


feature/gnoi/factory_reset/tests/factory_reset_test/factory_reset_test.go
```diff
@@ -121,7 +121,7 @@ func deviceBootStatus(t *testing.T, dut *ondatra.DUTDevice) {
 // performs factory reset
 func factoryReset(t *testing.T, dut *ondatra.DUTDevice, devicePaths []string) {
 	createFiles(t, dut, devicePaths)
-	gnoiClient := dut.RawAPIs().GNOI().New(t)
+	gnoiClient := dut.RawAPIs().GNOI(t)
```
bstoll, line 124: factoryReset() can be called twice in a row so I am not sure about reusing a cached connection here.

bstoll (approved):

PR #2135, opened 2023-09-15
ni_proto -> niProto

greg-dennis:


robshakir (approved):
Thanks for cleaning this up :-)

PR #2150, opened 2023-09-19
gNMI-1.11 add check for monitoring traffic counters per 30 seconds

cfernanz:
Enhancing gNMI-1.11 [Telemetry: Interface Packet Counters - Feature Profiles] 
* to monitor interface in and out packet counters per 30 seconds


feature/gnmi/otg_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
```diff
@@ -173,6 +175,50 @@ func TestInterfaceCounters(t *testing.T) {
 	}
 }
 
+func validateInAndOutPktsPerSecond(t *testing.T, dut *ondatra.DUTDevice, i1, i2 *interfaces.InterfacePath) bool {
+	if deviations.InterfaceCountersFromContainer(dut) {
+		time.Sleep(10 * time.Second)
+		return true
+	}
+	inSamples := gnmi.Collect(t, dut.GNMIOpts().WithYGNMIOpts(ygnmi.WithSubscriptionMode(gpb.SubscriptionMode_SAMPLE), ygnmi.WithSampleInterval(1*time.Second)), i1.Counters().InUnicastPkts().State(), 10*time.Second)
+	outSamples := gnmi.Collect(t, dut.GNMIOpts().WithYGNMIOpts(ygnmi.WithSubscriptionMode(gpb.SubscriptionMode_SAMPLE), ygnmi.WithSampleInterval(1*time.Second)), i2.Counters().OutUnicastPkts().State(), 10*time.Second)
```
dplore, line 184: Sample interval should be 30 seconds as this is the operational use case and pass/fail criteria.  

THere was some offline discussion of possibly a future version of this test could be a performance/scale test where we attempt to find the limit / fastest sample rate.  But that is considered out of scope for this functional test at this time.

feature/gnmi/otg_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
```diff
@@ -173,6 +175,51 @@ func TestInterfaceCounters(t *testing.T) {
...
+		return pktCounterOK
+	}
+
+	for i := 1; i < len(inPkts); i++ {
+		inValOld, _ := inPkts[i-1].Val()
+		outValOld, _ := outPkts[i-1].Val()
+		inValLatest, _ := inPkts[i].Val()
+		outValLatest, _ := outPkts[i].Val()
+		t.Logf("Incoming Packets: %d, Outgoing Packets: %d", inValLatest, outValLatest)
+		if inValLatest-inValOld != outValLatest-outValOld {
```
dplore, line 214: I think we also want to check that each value pair has increased.   Bug we want to detect is if the old and latest are the same value.

dplore: for example, if inValOld is 100 and inValLatest is 100, and outValLatest and outValOld is also 100,  yet we are sending traffic, then the test should fail.

cfernanz: Done

dplore (approved):

PR #2158, opened 2023-09-20
Create RT-1.21 README BGP MSS and PMTUD

dplore:
BGP TCP MSS and PMTUD


sachendras:
@dplore: I actually wrote the test in Juniper's https://github.com/openconfig/featureprofiles/pull/1910 already and is automated now

PR #2164, opened 2023-09-20
Update RT-5.1:  README.md to include sub test case for link flaps

nraghavendran:
/interfaces/interface/state/counters/carrier-transitions is the OC path that captures link flaps. adding a sub-testcase under singleton interface to perform link flaps, and verify that the state transitions of the port are captured on the oc path above.


feature/interface/singleton/otg_tests/singleton_test/README.md
```diff
@@ -15,6 +15,7 @@ a new testbed configuration with the desired port types.
     the octets of IPv4.
   * Ensure: ARP discovers static MAC address specified when port is
         configured with static MAC.
+### TC1 - singleton interface verification:
```
dplore, line 18: Suggestion follow style from [test requirements template](https://github.com/openconfig/featureprofiles/blob/main/doc/test-requirements-template.md):

```suggestion
### Subtest 1 - singleton interface verification:
```

feature/interface/singleton/otg_tests/singleton_test/README.md
```diff
@@ -30,6 +31,14 @@ a new testbed configuration with the desired port types.
 
 [^1]: The MTU specified above refers to the L3 MTU, which is the payload portion
     of an Ethernet frame.
+### TC2 - link flaps:
```
dplore, line 34:
```suggestion
### Subtest 2 - link flaps:
```

feature/interface/singleton/otg_tests/singleton_test/README.md
```diff
@@ -30,6 +31,14 @@ a new testbed configuration with the desired port types.
...
+* Bring down the physical layer of ATE port-1, and bring it back up.
+    Repeat this a few times (minimum 2)
```
dplore, line 36: if there any verification needed inbetween up/down transition?  Just wait for ack from ATE and repeat?  Is the ATE call blocking or asynchronous?  Just want to make sure the ATE actually brings down the physical layer.  

2 times seems like a small number?  Is there an operational issue/target we can reference?

nraghavendran: I will add a verification step on the DUT/ATE for every state transition. this would be a blocking call. I have left the transition value to a min. value of 2, and we could increase this, if we identify a target reference later.

sachendras (approved):

PR #2178, opened 2023-09-22
[4/5] gRIBI MPLS compliance test for popping N labels from stack.

robshakir:
```
* (M) feature/gribi/mplsutil:
  - Add compliance test for popping N labels from an incoming packet
    programmed by gRIBI.
 * (M) feature/gribi/otg_tests/mpls_compliance:
  - Add test with no traffic verification for validating gRIBI
    programming API.
```

### DIFF
--- a/feature/gribi/mplsutil/compliance.go
+++ b/feature/gribi/mplsutil/compliance.go
@@ -336,3 +336,73 @@ func PopNLabels(t *testing.T, c *fluent.GRIBIClient, defaultNIName string, popLa
 		})
 	}
 }
+
+// PopOnePushN implements a test whereby one (the top) label is popped, and N labels as specified
+// by pushLabels are pushed to the stack for an input MPLS packet. Two LFIB entries (100 and 200)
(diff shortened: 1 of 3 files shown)


feature/gribi/mplsutil/compliance.go
```diff
@@ -270,3 +270,69 @@ func PopTopLabel(t *testing.T, c *fluent.GRIBIClient, defaultNIName string, traf
...
+			c.Modify().AddEntry(t,
+				fluent.NextHopGroupEntry().
+					WithNetworkInstance(defaultNIName).
+					WithID(1).
+					AddNextHop(1, 1))
+
+			c.Modify().AddEntry(t,
+				fluent.LabelEntry().
+					WithLabel(100).
+					WithPoppedLabelStack(popLabels...).
```
alshabib, line 301: In the single label case the pop'ing was done in the next hop entry, why is it done here in this case?

alshabib (approved):

PR #2181, opened 2023-09-25
feat: mtu-1.3: large ip packet transmission

cm-nokia:
* test description outlined four input/output ports, but this will run with one or more ports just fine for the bundle flavor
* if using a kne setup will need to make sure to have two lag groups configured! (obvious in hindsight but easy to forget :p) 
* tested on both kne/virtual and real hardware
* my first fp test so be gentle :D

### DIFF
--- a/feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/large_ip_packet_transmission_test.go
+++ b/feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/large_ip_packet_transmission_test.go
@@ -1,0 +1,723 @@
+// Copyright 2023 Nokia
+//
+// This code is a Contribution to OpenConfig Feature Profiles project ("Work")
+// made under the Google Software Grant and Corporate Contributor License
+// Agreement ("CLA") and governed by the Apache License 2.0. No other rights
+// or licenses in or to any of Nokia's intellectual property are granted for
+// any other purpose.  This code is provided on an "as is" basis without
+// any warranties of any kind.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package large_ip_packet_transmission_test
+
+import (
+	"fmt"
+	"sort"
+	"testing"
+	"time"
+
+	otgtelemetry "github.com/openconfig/ondatra/gnmi/otg"
+	"github.com/openconfig/ygnmi/ygnmi"
+
+	"github.com/open-traffic-generator/snappi/gosnappi"
+	"github.com/openconfig/featureprofiles/internal/attrs"
+	"github.com/openconfig/featureprofiles/internal/deviations"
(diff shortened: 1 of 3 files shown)


feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/large_ip_packet_transmission_test.go
```diff
@@ -0,0 +1,723 @@
...
+
+	portNames := make([]string, len(allAtePorts))
+	for idx, port := range allAtePorts {
+		portNames[idx] = port.ID()
+	}
+
+	// note that it seems max in otg containers is 9000 so bundle tests > 1500 bytes will fail,
+	// for whatever reason individual ports work just fine > 1500 bytes though! also, physical gear
+	// seems to work just fine as well, so we'll set this to the max we can for kne tests.
+	otgConfig.Layer1().Add().SetName("layerOne").SetPortNames(portNames).SetMtu(9000)
```
jasdeep-hundal, line 585: OTG `Layer1` objects default the port speed to 10 gpbs, which is causing issues for this test. Can we explicitly set the speed for the ports here based on the information in the reservation? (via something like `SetSpeed(Layer1Speed.SPEED_100_GBPS)`)

jasdeep-hundal (changes requested):

feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/large_ip_packet_transmission_test.go
cm-nokia: @jasdeep-hundal copy -- updated to set based on zeroith port speed in the testbed ports. perhaps not the absolute best but looks like a reasonable approach, lemme know if you wanna see any tweaks!

rszarecki (approved):

LimeHat:
@self-maurya we don't have permissions to do /gcbruns, has to be done by Google.

cm-nokia:
`we don't have permissions to do /gcbruns, has to be done by Google.` I was just too excited, just tryin to be part of the team 😁 🥳

self-maurya:
As discussed with @Ankur19 , gcbrun is not required
going ahead with merge

sachendras:
Nokia. why was this pull not sent to review of all implementations? I will revert pulls in future if they fail because you didnt open the pull for comments from other vendors.

LimeHat:
@sachendras the PR was open on github for a month, anyone is welcome to comment.
is there a special procedure to "sent to review of all implementations"? If so, we're not aware about it, and we're not getting any special invitations to review either, unless someone from G pings us directly which doesn't happen in most of the cases. For ancx cases we received exactly 0 such requests, specifically

PR #2183, opened 2023-09-25
Update gNOI OSInstall test

sachendras:
Update the existing test "feature/gnoi/os/tests/osinstall" to include configuration push failure scenario


feature/gnoi/os/tests/osinstall/README.md
```diff
@@ -2,78 +2,96 @@
...
+   5. Verify that the supervisor image has moved to the new image:
+       
+      a. Verify that the supervisor has a valid image by issuing gnoi.os.Verify rpc.
+         * Expect a VerifyResponse with the version field set to the version specified in messages above eventually.
+         * Verify the standby supervisor version.
+         * Expect that the VerifyResponse.verify_standby has the same version in messages above.
+
+               
+* Subtest-2 : Configuration push verification post OS change and check health of the software-module that allows configuration.
+  1. Check the health of the software-module component that allows configuration of the router and verify if it is healthy using the leaf /components/component[**process that handles configuration of the DUT**]/healthz/status/
```
dplore, line 60: IMO, the goal should be to add a, '[smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software))' to confirm a successful software upgrade occurred.  

Instead of introspection into implementation specific process names, we can accomplish the same goal by pushing a configuration and confirming it is accepted and applied using gNMI.   Coverage of gNOI healthz rpc's should be in other, healthz specific tests IMO.

In other words, remove step 1.    

Step 3 can be changed to verify configuration is the same as what is pushed with gNMI.Get

Remove step 4.

marcushines: i think we should try to use healthz for this test - why wouldn't wouldn't this be the entity library providing a clear "lookup" for this component

feature/gnoi/os/tests/osinstall/README.md
```diff
@@ -2,78 +2,96 @@
...
+  2. Do a gNMI.GET() RPC to extract the current configuration on the DUT as backup.
+     
+  3. Push test configuration to the router using gNMI.Set() RPC with "replace operation" and reverify the status of the leaf /components/component[process that handles configuration of the DUT]/healthz/status/.
+     
+     a. If unhealthy, run HealthZ.Get() and HealthZ.Artifact() RPCs on the subject component to fetch artifacts corresponding to the event.
+        [TODO: Below step needs to be discussed to inderstand how to recover the DUT. May just need to depend on the Test infrastructure]
+        * Push the backup configuration fetched in Subtest-2 bullet#2 above to the DUT to recover the DUT.
+        * Mark the test as a failure due to issues with the OS upgrade and exit the test.
+          
+  4. Kill the process that manages device configuration using the "gNOI.KillProcessRequest_SIGNAL_ABRT" and restart flag set to False. This will terminate the process and will also dump a Core file, while maintaing the process in a down state. Verify if the leaf /components/component/healthz/state/ transitioned to unhealthy state.
```
marcushines, line 75: define the actual proto message you want here rather than just the "text"

sachendras: Updated to reflect the actual proto message.

feature/gnoi/os/tests/osinstall/README.md
sachendras: @dplore 
- Remove step 1:
  - I added this step because it is possible that a config already present in the router may cause config daemon to crash after software upgrade because the new software no longer supports a particular option in the config. So before we go to step2, imo verifying healthz right after OS upgrade is important.
 - Remove step 4:
   - Added step4 to test `config daemon failure reporting` here because it seemed more relevant to the OS install and config compatibility check. If the installed OS is compatible w/ the pushed configuration thats good, but then is the artifacts collection part of this daemon is functioning post upgrade? If not, then the OS is buggy and the OS install test failed.

feature/gnoi/os/tests/osinstall/README.md
```diff
@@ -2,78 +2,96 @@
...
-    primary supervisor.
-
-    *   Set the version field of the ActivateRequest message to be the same as
-        the version specified in the TransferRequest message above.
-    *   Set the no_reboot flag to true.
-    *   Set the standby_supervisor to TRUE this time.
+## Procedure
 
-*   Reboot the switch:
+* Subtest-1 : OS Install on the primary and secondary controller cards.
```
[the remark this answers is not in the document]
sachendras: Subtest-1, 2 so on is the process being followed in other tests too. Also, the test numbering is a little confusing right now. Check for example https://source.corp.google.com/piper///depot/google3/net/management/wbb/catalog/testregistry.textproto;l=40-51. BMP-2.1, 2.2, 2.3, 2.4 ... are separate tests rather than being subtests.

feature/gnoi/os/tests/osinstall/README.md
```diff
@@ -2,78 +2,96 @@
...
+      
+   5. Verify that the supervisor image has moved to the new image:
+       
+      a. Verify that the supervisor has a valid image by issuing gnoi.os.Verify rpc.
+         * Expect a VerifyResponse with the version field set to the version specified in messages above eventually.
+         * Verify the standby supervisor version.
+         * Expect that the VerifyResponse.verify_standby has the same version in messages above.
+
+               
+* Subtest-2 : Configuration push verification post OS change and check health of the software-module that allows configuration.
```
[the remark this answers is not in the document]
sachendras: Same comment as above

feature/gnoi/os/tests/osinstall/README.md
LimeHat: Hi @sachendras 
Re
> I added this step because it is possible that a config already present in the router may cause config daemon to crash after software upgrade because the new software no longer supports a particular option in the config. 

isn't the general assumption that such configuration should be either
a) automatically transformed during the upgrade, or
b) rejected by the node (instead of causing a crash)?
I believe both behaviors are widely implemented.

Indeed a daemon may crash during a reload if there's a SW bug; however, I believe that this logic should be extended to other daemons as well if the objective to ensure overall post-upgrade health of the node. Why limit to a single daemon?
I.e. pre-upgrade the controller should gather health state of all SW components (that report it / must report it), and compare it with the health state post-upgrade.

> If the installed OS is compatible w/ the pushed configuration thats good, but then is the artifacts collection part of this daemon is functioning post upgrade? If not, then the OS is buggy and the OS install test failed.

But this is not specific to an upgrade procedure per se. Was if the artifacts collection part of this daemon wasn't working pre-upgrade either? Wouldn't a better place for a such procedure be in a healthZ specific test as suggested by Darren?

sachendras:
```
isn't the general assumption that such configuration should be either
a) automatically transformed during the upgrade, or
b) rejected by the node (instead of causing a crash)?
I believe both behaviors are widely implemented.
```

Sure, but IDK what is the behavior of all implementations. For now, I dont want to enforce a requirement through this test on that particular use case. We will see in future if that needs to change. Either ways, net result for subtest-1 is a failure.

```
Indeed a daemon may crash during a reload if there's a SW bug; however, I believe that this logic should be extended to other daemons as well if the objective to ensure overall post-upgrade health of the node. Why limit to a single daemon?
I.e. pre-upgrade the controller should gather health state of all SW components (that report it / must report it), and compare it with the health state post-upgrade.
```
The subtest says "Configuration push verification post OS change and check health of the software-module that allows configuration". This test here is trying to address a particular DVT found bug whereby post upgrade an implementation had its config daemon crashing. However, it had the base config loaded and working so streaming and regular forwarding was working, just that we couldnt push new config to the DUT. I would like to keep this test focused only to the daemon that manages configuration and addresses the DVT found bug. I dont want to muddy this test by looking for unrelated daemons. I think heathz check would be more relevant for out-of-context checks for all software-modules like you said.

```
But this is not specific to an upgrade procedure per se. What if the artifacts collection part of this daemon wasn't working pre-upgrade either? Wouldn't a better place for such procedure be in a healthZ specific test as suggested by Darren?
```
Sure, we can move this negative test to HealthZ instead.

feature/gnoi/os/tests/osinstall/README.md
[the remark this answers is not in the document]
marcushines: subtests should be 

BMP-2.1.1 or such not just subtest-1 that doesn't hold any meaning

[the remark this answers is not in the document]
sachendras: You mean gNOI-4.1.1 and gNOI-4.1.2? I updated so, ptal

feature/gnoi/os/tests/osinstall/README.md
```diff
@@ -2,81 +2,89 @@
...
+   5. Verify that the supervisor image has moved to the new image:
+       
+      a. Verify that the supervisor has a valid image by issuing gnoi.os.Verify rpc.
+         * Expect a VerifyResponse with the version field set to the version specified in messages above eventually.
+         * Verify the standby supervisor version.
+         * Expect that the VerifyResponse.verify_standby has the same version in messages above.
+
+               
+* gNOI-4.1.2: Configuration push verification post OS change and check health of the software-module that allows configuration.
+  1. Check the health of the software-module component that allows configuration of the router and verify if it is **HEALTHY** using the leaf /components/component[**process that handles configuration of the DUT**]/healthz/state/status/
```
dplore, line 60: What should happen if gnmi get for healthz path is not supported  or the below gnoi healthz calls are not supported 
 but the DUT is actually otherwise healthy?  Should gNOI-4.1.2 pass or fail?

For example: : gNOI-4.1.1 succeeds, and gNOI-4.1.2 steps 2 and 3 can succeed, except for the calls to healthz?)

Seems a likely scenario that a DUT could fully support OS install, but not support healthz.  IMO, we should be able to tell which part is failing

sachendras: `What should happen if gnmi get for healthz path is not supported or the below gnoi healthz calls are not supported
but the DUT is actually otherwise healthy? Should gNOI-4.1.2 pass or fail?`
**Response:** This situation should be handled in the similar fashion as we always do for unsupported paths i.e.
 a. Add a deviation
 b. Raise a Google bug (w/ assignee as the implementation that do not support the feature) to track long-term fix for the deviation
 c. Once the fix is available, remove the deviation.

Example: If I was to write a test for BGP, I would write it such that
1. BGP configuration is Pushed
2. Traffic is being forwarded as expected
3. Check all the relevant state paths for BGP to ensure the implementation supports those paths and the results are accurate to the context for each subtest involved.

I wouldnt just do steps 1 and 2 and write a separate test for all state paths and cover "3" there. Imo it is not right  to call a test pass/fail just because bullet#2 is pass/fail and in production we need the state paths to be accurately streamed as well matching the context of the situation.
Similarly in this test i.e. gNOI-4.1, HealthZ check for the process in context seems important. 
As for unsupported paths, we have the respective implementations raise a deviation like we always do and follow the a-c steps above. I have 2xNotes in the readme to cover for implementations that may have a different behavior. The Deviation should have pretty much similar impact as before for the implementations not supporting HealthZ for Software-module other than the fact that this readme also asks to push some configuration. Sorry for the long response but hopefully it helps explain the thought behind the test and the process to handle it when the support doesnt exist?

`Seems a likely scenario that a DUT could fully support OS install, but not support healthz. IMO, we should be able to tell which part is failing`
**Response:** I see the following chain of events for an implementation that do not support HealthZ
1. They will pass/fail gNOI-4.1.1 the same as before.
2. For gNOI-4.1.2 the deviation will skip healthZ check and the implementation will push a configuration. I see 2 offshoots of this operation. Either,
    a. The implementation allows unsupported configuration to be pushed and its daemon keeps crashing. I am hoping the gNMI.Get() RPC call would fail in which case the test will fail.
    b.  The implementation wouldnt allow gNMI.Set() RPC for the unsupported conifguration in which case the test is a failure.
If the configuration push succeeds and the gNMI.Get() operation is also successful for the same configuration then the test is a pass. Hopefully this make sense?

dplore: osinstall test should ensure the os install completed successfully.  ie: the new os in installed and running.  Leave it to the  other tests to verify all the other features.    os install affects every process and every path on the entire system.  To fully know it's successful one should run all the tests in the entire featureprofiles repository, which includes healthz feature checks.    

This update does include gNMI set and get which will identify the bug that was experienced.  gNMI set and get are well known, widely supported features that are a baseline requirement for nearly all featureprofiles tests.  Healthz is not.  Let the the healthz tests confirm the healthz feature is supported.     Using the bgp analogy, do we know if BGP works after os_install?  We don't know, so we should check it.   Should that check be done in this test?  No, do it in the BGP tests.

Using deviations should be avoided.  Adding healthz here creates the possibility for additional deviations to be required which are not directly related or required for the feature being tested.   

That said, all the above does not mean including healthz breaks this test.  This test is now a os_install AND healthz test.   While I don't agree this is the minimum, this test can still work.  So I will approve.

feature/gnoi/os/tests/osinstall/README.md
```diff
@@ -60,24 +62,29 @@
...
-        * Mark the test as a failure due to issues with the OS upgrade and exit the test.
+        * Rollback to the previous OS version by following the steps in 3, 4 and 5 above from gNOI-4.1.1 to recover the DUT from the faulty state.
+        * Mark the test as a failure due to issues with the OS upgrade process and exit the test.
           
   2. Do a gNMI.GET() RPC to extract the current configuration on the DUT as backup.
      
-  3. Push test configuration to the router using gNMI.Set() RPC with "replace operation" and reverify the status of the leaf /components/component[process that handles configuration of the DUT]/healthz/state/status/.
+  3. Push test configuration to the router using gNMI.Set() RPC with "replace operation" and reverify the status of the leaf /components/component[process that handles configuration of the DUT]/healthz/state/status/. 
      
      a. If **UNHEALTHY**, run HealthZ.Get() and HealthZ.Artifact() RPCs on the subject component to fetch artifacts corresponding to the event.
```
dplore, line 72:
```suggestion
     a. If gNMI.Set fails, run HealthZ.Get() and HealthZ.Artifact() RPCs on the configuration component to fetch artifacts corresponding to the event.
```

feature/gnoi/os/tests/osinstall/README.md
```diff
@@ -60,24 +62,29 @@
...
         [TODO: Below step needs to be discussed to inderstand how to recover the DUT. May just need to depend on the Test infrastructure]
-        * Push the backup configuration fetched in Subtest-2 bullet#2 above to the DUT to recover the DUT.
+        * Push the backup configuration fetched in gNOI-4.1.2 bullet#ii above to the DUT to recover the DUT.
         * Mark the test as a failure due to issues with the OS upgrade and exit the test.
+     
+     b. If the configuration push is successful, make a gNMI.Get() RPC call and compare the configuration received with the originally pushed configuration and check if the configuration is a match. Test is a failure if either the gNMI.Get() operation fails or the configuration do not match with the one that was pushed.
+
+     **Please Note:** In some implementations it is possible that the gNMI.Set operation itself would fail because the configuration being pushed to the box is not supported by the new OS. The test would be a failure in such situations.
```
dplore, line 79: I think this text is not needed if you apply my suggestion for 3.a.

feature/gnoi/os/tests/osinstall/README.md
```diff
@@ -60,24 +62,29 @@
...
+  3. Push test configuration to the router using gNMI.Set() RPC with "replace operation" and reverify the status of the leaf /components/component[process that handles configuration of the DUT]/healthz/state/status/. 
      
      a. If **UNHEALTHY**, run HealthZ.Get() and HealthZ.Artifact() RPCs on the subject component to fetch artifacts corresponding to the event.
         [TODO: Below step needs to be discussed to inderstand how to recover the DUT. May just need to depend on the Test infrastructure]
-        * Push the backup configuration fetched in Subtest-2 bullet#2 above to the DUT to recover the DUT.
+        * Push the backup configuration fetched in gNOI-4.1.2 bullet#ii above to the DUT to recover the DUT.
         * Mark the test as a failure due to issues with the OS upgrade and exit the test.
+     
+     b. If the configuration push is successful, make a gNMI.Get() RPC call and compare the configuration received with the originally pushed configuration and check if the configuration is a match. Test is a failure if either the gNMI.Get() operation fails or the configuration do not match with the one that was pushed.
```
dplore, line 77: If the Get fails or the config doesn't match, do you want to rollback the os install as per the other failures?

dplore (approved):

feature/gnoi/os/tests/osinstall/README.md
sachendras: Thinking this through, the test will probably end up flaky if the daemon is crashing continuously as the reported health will depend on its state in time. I have changed the write-up to only update and read config. This seems a better approach and the test would fail if either config push fails or the GET operation fails.

However the business requirement is also to ensure that the health of the config daemon can be streamed and supported by all implementations. Hence I also have https://github.com/openconfig/featureprofiles/pull/2240 to test the paths in a more stable environment.

feature/gnoi/os/tests/osinstall/README.md
sachendras: updated the test

feature/gnoi/os/tests/osinstall/README.md
sachendras: Isnt that the case either ways, PASS/FAIL? The test should restore the DUT back to its original OS, no?

feature/gnoi/os/tests/osinstall/README.md
sachendras: I updated the test to just include config set/get operations.

dplore (approved):

PR #2185, opened 2023-09-26
Using correct global AS variable for bgp_2byte_4byte_asn_test

s-nandu:
Current test fails while verifying Global AS an localAS variable is used for configuration
Using globalAS solves the issue


Ankur19 (approved):

Ankur19:
Merging this as this is required for test

PR #2186, opened 2023-09-26
TE-11.2:Updated script to use flow metrics for traffic verification

mohanasm:
TE-11.2 uses port metrics to verify the traffic flow and port counters can include control packets like ARP., This leads to incorrect packet count resulting in flakiness of test result. Updated the script to use flow metrics for traffic verification similar to other OTG scripts.


feature/gribi/otg_tests/backup_nhg_multiple_nh_test/backup_nhg_multiple_nh_test.go
```diff
@@ -333,6 +333,7 @@ func (a *testArgs) testIPv4BackUpSwitch(t *testing.T) {
 	// create flow
 	dstMac := gnmi.Get(t, a.ate.OTG(), gnmi.OTG().Interface(atePort1.Name+".Eth").Ipv4Neighbor(dutPort1.IPv4).LinkLayerAddress().State())
 	BaseFlow := a.createFlow(t, "BaseFlow", dstMac)
+        time.Sleep(5 * time.Second)
```
mojiiba, line 336: could you please elaborate what is flaky here and how does this sleep help? We do not add sleep in fp tests  and it is better we add checking for preconditions if is required.

mohanasm: Script used port counters for verifying traffic flow and port counters can account for control packets too (like arp). This lead to flakiness. Updated the script to use flow counters like in other OTG scripts.

sachendras:
Just for posterity

The test is being corrected as the current version of it is looking at the IXIA Port stats for sent and received packets as pass/fail criteria. This make the test flaky as the IXIA sent packets may also include ARP etc. Changes in this Pull is to update checking port stats to checking flow stats for accuracy.

sachendras (approved):

greg-dennis (approved):
Approving for Go quality

sachendras (approved):

PR #2187, opened 2023-09-26
Skip firmware-version telemetry check for non 400G transceivers

Ankur19:
firmware-version leaf is not required for all transceiver types. It is required for 400G transceivers.

### DIFF
--- a/feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
+++ b/feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
@@ -17,6 +17,7 @@ package telemetry_inventory_test
 import (
 	"fmt"
 	"regexp"
+	"strings"
 	"testing"
 
 	"github.com/openconfig/featureprofiles/internal/args"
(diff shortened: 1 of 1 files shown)


feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -651,8 +661,18 @@ func ValidateComponentState(t *testing.T, dut *ondatra.DUTDevice, cards []*oc.Co
 			if p.fwVerValidation {
 				fwVer := card.GetFirmwareVersion()
 				t.Logf("Component %s FirmwareVersion: %s", cName, fwVer)
+
+				isTransceiver := card.GetType() == componentType["Transceiver"]
+				is400G := false
+				if isTransceiver {
+					is400G = strings.Contains(card.GetTransceiver().GetEthernetPmd().String(), "ETH_400GBASE")
```
dplore, line 668:
```suggestion
					is400G = strings.Contains(card.GetTransceiver().GetEthernetPmd().String(), "ETH_400GBASE_ZR")
```

dplore (changes requested):

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
dplore: https://github.com/openconfig/public/blob/5d38d8531ef9c5b998262207eb6dbdae8968f9fe/release/models/optical-transport/openconfig-transport-types.yang#L1112C12-L1115

dplore: I think the requirement is for 400ZR only.  Can you confirm Xulin?

dplore: After checking with our SME for optical transceivers, we expect all 400G transceivers to expose a firmware version.

dplore (approved):

PR #2190, opened 2023-09-27
Query

xw-g:


self-maurya:
comments from @nflath:
NHG#102 --> [NH#104 {decap-encap, src:OuterSrcIP_2, dst:OuterDstIP_FAILURE, network-instance: VRF-C}, backupNHG: NHG#103]

IE vendor didn't support the flow you suggested, so we actually put a fake destination IP that won't resolve and the backup is decap

nflath (approved):
LGT

PR #2199, opened 2023-09-30
Add subtest for P4RT interface id

dplore:


dplore:
@bstoll regarding the build failure here, do I need a separate PR to first add the package I am moving? "github.com/openconfig/featureprofiles/internal/p4rt/p4rtutils"

bstoll:
> @bstoll regarding the build failure here, do I need a separate PR to first add the package I am moving? "github.com/openconfig/featureprofiles/internal/p4rt/p4rtutils"

The import is using "internal/p4rt/p4rtutils" but the file was moved to "internal/p4rtutils".  If you update this line to be `"github.com/openconfig/featureprofiles/internal/p4rtutils"` this should fix the issue with the p4rt_daemon_failure_test you see.

https://github.com/openconfig/featureprofiles/blob/ff18bb6c129b8352966f791ef2a9f67f076ca1ea/feature/p4rt/otg_tests/p4rt_daemon_failure_test/p4rt_daemon_failure.go#L28

You'll also need to update the import path in these files to reflect the move from "feature/experimental/p4rt/internal/p4rtutils" to "internal/p4rtutils":

```
feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetin_test/google_discovery_protocol_packetin_test.go
feature/experimental/p4rt/otg_tests/traceroute_packetin_test/traceroute_packetin_test.go
feature/experimental/p4rt/otg_tests/traceroute_packetin_test/packetin_test.go
feature/experimental/p4rt/otg_tests/traceroute_packetout_test/traceroute_packetout_test.go
feature/experimental/p4rt/otg_tests/lldp_packetin_test/lldp_packetin_test.go
feature/experimental/p4rt/otg_tests/lldp_packetout_test/lldp_packetout_test.go
feature/experimental/p4rt/otg_tests/performance_test/performance_test.go
feature/experimental/p4rt/otg_tests/google_discovery_protocol_packetout_test/google_discovery_protocol_packetout_test.go
feature/experimental/p4rt/otg_tests/base_p4rt/base_p4rt_test.go
feature/experimental/p4rt/tests/p4rt_election/p4rt_election_test.go
feature/experimental/p4rt/tests/metadata_validation_test/metadata_validation_test.go
```

Another alternative would be to keep p4rtutils in old location and do two PRs since moving the library out of experimental is independent of the test changes you are doing.

internal/attrs/attrs.go
```diff
@@ -42,6 +42,7 @@ type Attributes struct {
 	IPv4Len uint8  // Prefix length for IPv4.
 	IPv6Len uint8  // Prefix length for IPv6.
 	MTU     uint16
+	ID      uint32 // p4rt interface id
```
bstoll, line 45: I think I would prefer named something like P4ID or P4InterfaceID.  ID is so generic it opens the field up to misuse.

dplore: It's a good suggestion, but if we change it here, it makes it different from the OC path.  Maybe I should add a comment with the full path name?

This ID maps directly to the OC path [/interfaces/interface/config/id](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-interfaces.html#interfaces-interface-config-id)

dplore:
> > @bstoll regarding the build failure here, do I need a separate PR to first add the package I am moving? "github.com/openconfig/featureprofiles/internal/p4rt/p4rtutils"
> 
> The import is using "internal/p4rt/p4rtutils" but the file was moved to "internal/p4rtutils". If you update this line to be `"github.com/openconfig/featureprofiles/internal/p4rtutils"` this should fix the issue with the p4rt_daemon_failure_test you see.

Ah thanks, I will fix that.

> Another alternative would be to keep p4rtutils in old location and do two PRs since moving the library out of experimental is independent of the test changes you are doing.

go doesn't allow using a package containing a path `internal` outside of the tree it is defined in, so it has to move.

feature/p4rt/otg_tests/p4rt_daemon_failure_test/p4rt_daemon_failure.go
```diff
@@ -0,0 +1,378 @@
...
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/open-traffic-generator/snappi/gosnappi"
+	"github.com/openconfig/featureprofiles/internal/attrs"
+	"github.com/openconfig/featureprofiles/internal/deviations"
+	"github.com/openconfig/featureprofiles/internal/fptest"
+	"github.com/openconfig/featureprofiles/internal/gribi"
+	"github.com/openconfig/featureprofiles/internal/p4rt/p4rtutils"
```
Ankur19, line 28: Moving a test from experimental to feature directory needs to be a separate PR.

We would need to update all tests using p4rtutils, if we want to move it to the /internal folder in root.

dplore: I will update all all tests using p4rtutils if you think it's not too bad to include in this PR.

internal/attrs/attrs.go
dplore: Added comment with the full path to disambiguate

Ankur19 (dismissed):

Ankur19:
Does not build due to OTG Api update

feature/p4rt/otg_tests/p4rt_daemon_failure_test/p4rt_daemon_failure.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+	if deviations.ExplicitInterfaceInDefaultVRF(dut) {
+		fptest.AssignToNetworkInstance(t, dut, p1.Name(), deviations.DefaultNetworkInstance(dut), 0)
+		fptest.AssignToNetworkInstance(t, dut, p2.Name(), deviations.DefaultNetworkInstance(dut), 0)
+	}
+}
+
+func configureATE(t *testing.T, ate *ondatra.ATEDevice) gosnappi.Config {
+	otg := ate.OTG()
+	top := otg.NewConfig(t)
```
Ankur19, line 161: `top := gosnappi.NewConfig()`

dplore: Thanks!

Ankur19 (approved):
LGTM!

PR #2200, opened 2023-10-02
TUN-1.6: Tunnel End Point Resize README.md

sachendras:
Rewrite TUN-1.6: Tunnel End Point Resize - Interface Based GRE Tunnel


feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
```diff
@@ -0,0 +1,90 @@
...
+        H[DUT:Port5\lag2] --> I[Port5\lag2:OTG];
+        J[DUT:Port6] --> K[Port6:OTG];
+```
+
+
+
+## SUB-Test:1 Encap Test
+*   The test should configure 32 GRE tunnel interfaces each with different /32 and /128 tunnel destination prefixes of the same /27 and /64 prefix range. Each of these tunnel interfaces should use the loopback address as the source-address for their respective address-families.
+ *   Configure an IPv4 and an IPv6 address for each of the 32xtunnel interfaces.
+ *   Add 32 different v4 and v6 static routes for the same /24 IPv4 and /64 IPv6 destination prefixes. These static routes should point at the 32 different tunnel interface addresses by their respective address families as next-hop.
```
sugrimov, line 23: If we are talking about Cisco's implementation they are the same static route pointing to NH tunnel interface. eg
  10.255.255.255/32 tunnel-ip1200
  10.255.255.255/32 tunnel-ip1201
  10.255.255.255/32 tunnel-ip1202
  10.255.255.255/32 tunnel-ip1203
  ...
  
  Also why are you mentioning /24 here?

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+## SUB-Test:2 Encap Test
+
+## Verification steps for Subtest-1
+*   Verify the next hop counters for packet being diverted or sent for encapsulation
+        *   Verify the tunnel interfaces counters to confirm traffic encapsulation
+        *   Verify the tunnel interfaces for optimal load balancing of traffic/flow. Acceptable deviation is <=3%
+
+## Verification steps for Subtest-2
+ *   After decapsulation, traffic should be load balanced/hashed to all available L3 ECMP which untimately resolves over a combination of LAG and Singleton interfaces. Acceptable deviation is <=3%.
```
sugrimov, line 48: Decap doesn't apply here.

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
```diff
@@ -0,0 +1,90 @@
...
+ *   Verify the tunnel interfaces counters to confirm the traffic decapsulation
+ *   Validate system for:
+        * Health-1.1
+        * No feature related error or drop counters incrementing
+
+## Verification steps for Subtest-2    
+*   After decapsulation, traffic should be load balanced/hashed to all available L3 ECMP or LAG or combination of both features
```
sugrimov, line 55: Decap doesn't apply here.

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
sachendras: There are 2 types of static routes here per address family

1. /24 and /64 static routes for the inner payload pointing at the tunnel interfaces
2. /27 and /64 static routes for the tunnel destination addresses.

The mention here is for "1".

Btw, I made some changes to the test. PTAL

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
sachendras: Right... doesnt makes sense now that I look at it. Removed in the updated version of the test

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
dplore, line 1: Please use the test requirements template format: https://github.com/openconfig/featureprofiles/blob/main/doc/test-requirements-template.md

Note, break up the test into subtests where some config and it's associated 'verification' are grouped together.  Be more explicit about the steps by making a numbered list of steps for each subtest (step1, step2, step3).  This will make it easier to translate your intention into code.

Subtest1  
 * Configure DUT
 * Configure  OTG 
 * Verify tunnel state on DUT and OTG

Subtest 2 - Send the traffic and verify counters.  Specify which counters (in OC) to be verified and the pass / fail criteria for each.

Subtest 3 add tunnel interfaces, verify counters.   Specify which counters (in OC) to be verified and the pass / fail criteria for each.

Subtest 4 remove tunnel interfaces, verify coutners.   Specify which counters (in OC) to be verified and the pass / fail criteria for each.


For health validation, include a link to the relevant test to be executed for clarity.

dplore (changes requested):

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
```diff
@@ -0,0 +1,97 @@
...
+        * Health-1.1. checks.
+        
+
+    
+## Config Parameter coverage
+
+TODO: OpenConfig definition required for Tunnel protocol under interfaces/interfaces/interface/tunnel/ as GRE, IP-IP, GUE etc. 
+ *   interfaces/interfaces/interface/tunnel/config/dest/
+ *   interfaces/interfaces/interface/tunnel/config/src/
+ *   interfaces/interfaces/interface/tunnel/ipv4/addresses/address/config/ip
```
dplore, line 72: Should this also cover?
/interfaces/interface/tunnel/config/ttl 
/interfaces/interface/tunnel/config/gre-key

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
```diff
@@ -0,0 +1,97 @@
...
+5. Add 128 different v4 and v6 static routes (64 for each address-family) for the 2 sets of /24 IPv4 and /64 IPv6 payload prefixes in bullet#4 above. The static routes should be such that, the first set of 1x /24 IPv4 and 1x /64 IPv6 prefixes should point at the 32x tunnel interfaces configured w/ IPv4 tunnel destination and the second set of 1x /24 IPv4 and 1x /64 IPv6 prefixes should point at the 32x tunnel interfaces w/ IPv6 tunnel destinations.
+
+### Verification:
+ *   Ensure all tunnel interfaces are up
+ *   Ensure all LAG and singleton interfaces are up
+ *   Run Health-1.1. checks
+
+## Subtest-2: Gather baseline stats by passing traffic
+ *   Start with the configuration in Subtest-1 above.
+ *   The test should send 2000 IPv4 and 2000 1Pv6 flows from the ATE:Port1 to the DUT:Port1. These flows should equally be split between the 2x /24 IPv4 and 2x /64 IPv6 destination prefixes mentioned in Subtest-1 bullet#4 above. The flows should use variations in the following,
```
dplore, line 34:
```suggestion
 *   The test should send 2000 IPv4 and 2000 1Pv6 flows from the ATE:Port1 to the "tunnel payload destination prefixes". These flows should equally be split between the 2x /24 IPv4 and 2x /64 IPv6 destination prefixes mentioned in Subtest-1 bullet#4 above. The flows should use variations in the following,
```

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
```diff
@@ -0,0 +1,97 @@
...
+        * Destination IP adresses.
+        * Destination Port
+        * Payload protocol
+ *   Incoming traffic on DUT:PORT1 should be load shared across all 64 Tunnel interfaces for encapsulation and then routed out the combination of lag and singleton interfaces, i.e. lag1, lag2 and Port6.
+ *   Follow all the "verifications" steps in [*] below
+
+## Subtest-3: Reduce number of tunnel interfaces 
+ *   Start with the state in Subtest-2 above with active traffic flow.
+ *   Reduce the number of Tunnel interfaces, e.g. from 32 to 16 for each of the tunnel destination group by address-family:
+        * If the static routes are used to forward traffic to tunnel, please disable or delete the static routes in this test to simulate the reduction in available paths.
```
dplore, line 46: Shouldn't the tunnel endpoints be removed by deleting ` interfaces/interfaces/interface/tunnel/config/dest/ ` entries?  If there are unreachable tunnel endpoints (due to no route), is the DUT behavior well defined?     Is the intent to test for this condition?  Or just the removal of endpoints in the config?

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
```diff
@@ -0,0 +1,97 @@
...
+ *   Incoming traffic on DUT:PORT1 should now be load shared across all remaining 16 Tunnel interfaces per tunnel-destination by address-family.
+ *   Follow all the verifications steps in [*] below
+ *   Inflight reduction in number of tunnel interfaces shouuldn't cause traffic drop.
```
dplore, line 49: Minor nit: Is this redundant to the Verification steps below?

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Subtest-4: Restore the number of tunnel interfaces 
+ *   Start from the condition in Subtest-3 above with inflight traffic from OTG:Port1.
+ *   Increase number of Tunnel interfaces, e.g. From 16 to 32 per tunnel-destination by address-family.
+ *   Incoming traffic on DUT:PORT1 should now be load shared across all 32 Tunnel interfaces per tunnel-destination by address-family. Follow all the verifications steps in [*] below
+ *   Inflight increase in the number of tunnel interfaces and related static route shouldn’t cause traffic drops.
```
dplore, line 55: Minor nit: Is this redundant to the Verification steps below?

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
sachendras: Well, we will have multiple static routes for the same destination prefixes to be encaped, mapped to several next hops which basically will be the tunnel IP addresses. If you remove some static routes, the remaining static routes should still be functional. Net effect is that we will have reduced number of tunnel interfaces. 
Other approach is as you said, remove the tunnel destination addresses from the tunnel interfaces. I am guessing that will take down the tunnel interfaces and hopefully will make the corresponding static routes passive and therefore not useable. But this approach has several nuances
1. We need to test if the chain of things I mentioned above really happens so
2. I dont know how different NOS will take the change. Possible that a NOS wouldn't let us commit a config because the tunnel interface has missing tunnel destination.

Tl;dr: Earlier approach seems more predictable.

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
sachendras: Updated, thanks

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
sachendras: - /interfaces/interface/tunnel/config/gre-key
   The goal of gre-key was to tie flows to the same tunnel interface in both directions. That is not the requirement in Google. 
- /interfaces/interface/tunnel/config/ttl
We dont set it. Not a requirement either. By default TTL is copied from inner header

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
dplore: Ah, I see.  I agree that removing only a portion of the static routes has a simple, expected behavior.    It's confusing to call this 'removal of tunnel interfaces'.  This approach is 'removal of static routes', causing some  the tunnel interfaces to not be used.  

This IMO is testing if adding/removing static routes causes packet loss.  This is not testing if adding and removing tunnels causes packet loss.   

Changing the static routes is an ok test and I don't think we have coverage of this in our current tests that use static routes (which seem to be [present in gribi tests, while the static route feature has none](https://github.com/search?q=repo%3Aopenconfig%2Ffeatureprofiles%20static%20route&type=code)).   But I think this portion of the test is not really exercising tunnels, it's static routing that is getting tested and would apply equally to tunnel interfaces and physical interfaces.

Do you mind updating the description to say "removal of static routes, causing some  the tunnel interfaces to not be used?"

I think we need a subtest in this readme which covers actually changing the tunnel configuration, not just static routes.    Here's why:  I think what we want to happen operationally is configuration generation will create a new test of tunnels and this will be applied using a config replace at /interfaces in one step.   There won't be a multi-step configuration change where routing changes (static or otherwise) is applied to 'drain' or 'isolate' a device and then make the configuration change.    So we need to test this way.     Example high level logic:
* Configure tunnel interface and routes set 1 in single gnmi.Replace
* Start traffic
* Configure tunnel interface and routes set 2 in single gnmi.Replace
* Check for packet loss

In production, are we using static routes?  Or dynamic routing?  Maybe the routes for this subtest should be isis+bgp?

dplore (changes requested):
Last comment to address is  regarding static routes and removing tunnel interfaces.

feature/experimental/tunnel/otg_tests/tunnel_interface_based_resize_test/README.md
sachendras: `Do you mind updating the description to say "removal of static routes, causing some the tunnel interfaces to not be used?"`
- Done

`I think we need a subtest in this readme which covers actually changing the tunnel configuration, not just static routes. `
- Done

`In production, are we using static routes? Or dynamic routing? Maybe the routes for this subtest should be isis+bgp?`
- Static routes. We will learn about the destination prefixes for which the traffic has to be encaped over IBGP from the WAN ports. But then these static routes are in place to change the local router's behavior to punt the traffic to the tunnel interface instead using static routes. These routes will be locally significant for these routers.

dplore (approved):
LGTM!  Thanks for all the changes.  Especially subtest 5.  This is a good example of chaining a list of subtests which start basic and add additional actions and criteria which should help pinpoint where a failure is.

PR #2208, opened 2023-10-04
Add support of OC slot-id path

arulkumarsekar:


dplore (dismissed):
LGTM for test intent.

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -528,7 +530,12 @@ func ValidateComponentState(t *testing.T, dut *ondatra.DUTDevice, cards []*oc.Co
 					t.Errorf("Component %s Description: got empty string, want non-empty string", cName)
 				}
 			}
-
+			if p.slotIDValidation && card.GetType() == oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_LINECARD {
```
greg-dennis, line 533: Get rid of slowIDValidation and just do the validation whenever the card type is oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_LINECARD? It looks like that wouldn't change the behavior for any test.

If you really think the validation ought not to be automatic, then I would get rid of the "card.GetType() == oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_LINECARD" condition. That would at least prevent someone from adding extraneous `slotIDValidation: true` fields.

greg-dennis (changes requested):

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
arulkumarsekar: @greg-dennis updated. can you check now pls?

greg-dennis (approved):

PR #2213, opened 2023-10-05
RT2.7-Adding interface passive state deviation

divyarema:
Added deviation to skip interface passive state check


internal/deviations/deviations.go
```diff
@@ -625,3 +625,9 @@ func ISISLspMetadataLeafsUnsupported(dut *ondatra.DUTDevice) bool {
 func QOSQueueRequiresID(dut *ondatra.DUTDevice) bool {
 	return lookupDUTDeviations(dut).GetQosQueueRequiresId()
 }
+
+// IsisInterfacePassiveStateUnsupported returns true for devices that do not
+// support interface passive state.
+func IsisInterfacePassiveStateUnsupported(dut *ondatra.DUTDevice) bool {
```
greg-dennis, line 631: "Isis" -> "ISIS"

greg-dennis (changes requested):

internal/deviations/deviations.go
divyarema: modified in latest commit

mohanasm:
Closing as this change is no longer required.

PR #2219, opened 2023-10-05
RT-1.5 / refactor: exception for explicit prefix limit *received*

cm-nokia:
- Adds deviation "BGPExplicitPrefixLimitReceived" for devices that set received prefix limits explicitly under prefix-limit-received rather than under "prefix-limit"
- fixed conditional check that was comparing addresses rather than value in the "verifyPrefixLimitTelemetry" sub test


LimeHat:
b/297188147 to track the deviation

rohit-rp (approved):

PR #2229, opened 2023-10-05
Add a new scenario in TE-3.1

xw-g:


xw-g:
deprecated by https://github.com/openconfig/featureprofiles/pull/2231

PR #2233, opened 2023-10-07
RT-5.2: Cleanup LACP and LAG configs to avoid potential conflicts in config

sgirishj:
Adding support to cleanup LACP and LAG configs before applying testcase config. This is to avoid having potentially conflicting configs between the LAG and LACP hierarchies for a given LAG interface; for e.g. LAG mode being set to STATIC for the bundle under oc-interfaces hierarchy and having the interface bundle enabled for LACP under oc-lacp hierarchy.


feature/interface/aggregate/otg_tests/aggregate_test/aggregate_test.go
```diff
@@ -227,6 +226,18 @@ func (tc *testCase) configureDUT(t *testing.T) {
...
+	resetBatch := &gnmi.SetBatch{}
+	gnmi.BatchDelete(resetBatch, aggPath.Config())
+	gnmi.BatchDelete(resetBatch, lacpPath.Config())
+	resetBatch.Set(t, tc.dut)
+
+	// Apply LACP config only when LAG mode is LACP
+	if tc.lagType == lagTypeLACP {
+		gnmi.Replace(t, tc.dut, lacpPath.Config(), lacp)
+	}
+
```
dplore, line 240: Earlier code always does a replace on both lacpPath.Config and aggPath.Config. even if tc.lagType =  lagTypeSTATIC.  Doing a gnmi.Replace should force an overwrite of any existing configuration.  It should not be necessary to delete configuration.    It seems if we are needing to delete configuration within these trees there is a bug?

dplore (changes requested):

feature/interface/aggregate/otg_tests/aggregate_test/aggregate_test.go
sgirishj: When lagType = lagTypeSTATIC, the interface should not be enabled under LACP hierarchy since LACP should not be run in that case. Previously, the script was enabling the interface under LACP for both lagTypeSTATIC and lagTypeLACP. The code change in this PR essentially is to achieve the below: start with a clean slate under both Interface_AGG and LACP hierarchies (by deleting the configs through gnmi.BatchDelete). Enable the interface under LACP only for lagType = lagTypeLACP.
Hope that clarifies.

Ankur19: Isn't below block of code doing the same thing?

```
if deviations.AggregateAtomicUpdate(tc.dut) {
		tc.clearAggregate(t)
		tc.setupAggregateAtomically(t)
	}
```

Enabling `aggregate_atomic_update` deviation should suffice

sgirishj: Just enabling the deviation wont help here since:
- clearAggregate() does not clear LACP configs
- the issue being addressed is not technically atomic vs non-atomic related. The issue at hand is that when LAG type is set to STATIC for an interface, the interface should be removed from under LACP.

Thanks!

Ankur19: But the issue that I see here is that there shouldn't be any existing config.

```
aggID := netutil.NextAggregateInterface(t, dut)
```

Wouldn't this util always choose a non existing AggregateID?

Ankur19: Anyways, for Arista we would need to move the below section such that it is called after the `delete`.

```
if deviations.AggregateAtomicUpdate(tc.dut) {
		tc.clearAggregate(t)
		tc.setupAggregateAtomically(t)
	}
```

sgirishj: > But the issue that I see here is that there shouldn't be any existing config.
> 
> ```
> aggID := netutil.NextAggregateInterface(t, dut)
> ```
> 
> Wouldn't this util always choose a non existing AggregateID?

Yes, but after choosing a fresh aggID, we test lagTypeLACP and then lagTypeSTATIC on the same aggID.

sgirishj: Made the change set simpler with the latest commit. LACP config will be applied only when lagType is LACP. Also the config delete is not needed in the middle if we test STATIC mode followed by LACP mode. Kindly review. Thanks!

sachendras:
Why this change now? Please provide more context.

sachendras:
> Why this change now? Please provide more context.

Okay, seems the new code version in your implementation has added a commit check that will look for such discrepancies and fail?

sachendras (approved):

sgirishj:
> > Why this change now? Please provide more context.
> 
> Okay, seems the new code version in your implementation has added a commit check that will look for such discrepancies and fail?

Thats right.

> I will approve, but please make sure to close this comment before merging.

Thanks!

dplore (approved):

Ankur19 (approved):

PR #2235, opened 2023-10-09
Add standard system client test

marcushines:


feature/system/tests/system_base_test/README.md
```diff
@@ -1,3 +1,41 @@
-# OC-1.1: System Configuration
+# System testing
 
-TODO(bstoll): fill in test plan from code already written.
+## Summary
+
+Ensures the device support basic system requirements for a WBB device.
```
robshakir, line 5: `s/WBB/device supporting OpenConfig APIs/` or g* APIs.

feature/system/tests/system_base_test/README.md
```diff
@@ -1,3 +1,41 @@
...
+## service-1
+
+### Procedure
+
+Each test will require the DUT configured with a basic service configuration that
+should be provided as part of the basic configuration.  This setup should also include
+any security setup for connecting to the services.
+
+The default setup should expect a CA signed certifate and trust bundle which can be
+used for MTLS
```
robshakir, line 18:
```suggestion
used for mTLS.
```

feature/system/tests/system_base_test/g_protocol_test.go
```diff
@@ -0,0 +1,112 @@
...
+func TestGNMIClient(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	addr := fmt.Sprintf("%s:9339", dut.Name())
+	t.Logf("gNMI standard port test: %q", addr)
+	conn, err := grpc.Dial(addr, grpc.WithBlock())
+	if err != nil {
+		t.Fatalf("grpc.Dial failed to: %q", addr)
+	}
+	c := gpb.NewGNMIClient(conn)
+	_, err = c.Get(context.Background(), &gpb.GetRequest{})
```
robshakir, line 44: Suggest to do `if _, err := c.Get(...); err != nil`.

feature/system/tests/system_base_test/g_protocol_test.go
```diff
@@ -0,0 +1,112 @@
...
+	spb "github.com/openconfig/gnoi/system"
+	authzpb "github.com/openconfig/gnsi/authz"
+	gribipb "github.com/openconfig/gribi/v1/proto/service"
+	p4rtpb "github.com/p4lang/p4runtime/go/p4/v1"
+)
+
+// TestGNMIClient validates that the DUT listens on standard gNMI Port.
+func TestGNMIClient(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	addr := fmt.Sprintf("%s:9339", dut.Name())
```
robshakir, line 37: Is using `dut.Name()` as a valid DNS name something that we can rely on? I think in KNE environments we don't have anything that would create such a resolver entry. Should we rather use `dut.RawAPIs` or a call that gets the IP of the assigned device here? (Although I expect that `RawAPIs` then introduces new layers that are being dialed through, but I don't remember fully.)

feature/system/tests/system_base_test/g_protocol_test.go
```diff
@@ -0,0 +1,112 @@
...
+func TestGNOIClient(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	addr := fmt.Sprintf("%s:9339", dut.Name())
+	t.Logf("gNOI standard port test: %q", addr)
+	conn, err := grpc.Dial(addr, grpc.WithBlock())
+	if err != nil {
+		t.Fatalf("grpc.Dial failed to: %q", addr)
+	}
+	c := spb.NewSystemClient(conn)
+	_, err = c.Time(context.Background(), &spb.TimeRequest{})
```
robshakir, line 60: We should ensure that the featureprofile here requires us to have `gNOI.Time` implemented. I don't recall whether it's been spelled out as an explicit requirement for any specific test thus far.

feature/system/tests/system_base_test/g_protocol_test.go
```diff
@@ -0,0 +1,112 @@
...
+func TestGRIBIClient(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	addr := fmt.Sprintf("%s:9340", dut.Name())
+	t.Logf("gRIBI standard port test: %q", addr)
+	conn, err := grpc.Dial(addr, grpc.WithBlock())
+	if err != nil {
+		t.Fatalf("grpc.Dial failed to: %q", addr)
+	}
+	c := gribipb.NewGRIBIClient(conn)
+	_, err = c.Get(context.Background(), &gribipb.GetRequest{})
```
robshakir, line 92: Per the spec ([here](https://github.com/openconfig/gribi/blob/master/doc/specification.md#421-getrequest-message)), an empty `GetRequest` isn't valid -- so I'd expect that this would return an error.

It's probably best to send:

```proto
all: {}
aft: ALL
```

which should be valid and doesn't rely on any knowledge of which AFTs or NIs exist on the DUT.

feature/system/tests/system_base_test/g_protocol_test.go
marcushines: it is not required :P

robshakir: Confused! What happens if the server returns `Unimplemented` here -- this test fails at that point, right?

feature/system/tests/system_base_test/g_protocol_test.go
marcushines: this actually succeeds?

feature/system/tests/system_base_test/g_protocol_test.go
marcushines: i would prefer to leave it as very likely we would really want to verify a behavior from the returned client / resp at some point but i just didn't need it now

PR #2237, opened 2023-10-10
Te3dot7 drain

bkreddy143:
1. Removed skip that enables test for all vendors 
2. corrected interface-id in delete interface from vrf 
3. restoring speed setting as replace is removing speed.
4. moved configureDUTDrain() call ahead of OTG protocols start. 


nsadhasivam (changes requested):
Can this PR be updated with the latest script version?

feature/gribi/otg_tests/base_hierarchical_nhg_update/base_hierarchical_nhg_update_test.go
```diff
@@ -865,14 +856,13 @@ func deleteinterfaceconfig(t *testing.T, dut *ondatra.DUTDevice) {
...
 	if deviations.ExplicitInterfaceInDefaultVRF(dut) {
 		ni := deviations.DefaultNetworkInstance(dut)
-		gnmi.Delete(t, dut, d.NetworkInstance(ni).Interface(p2.Name()+".").Subinterface().Config())
-		gnmi.Delete(t, dut, d.NetworkInstance(ni).Interface(p3.Name()+".").Subinterface().Config())
-		gnmi.Delete(t, dut, d.NetworkInstance(ni).Interface(p4.Name()+".").Subinterface().Config())
-	} else {
-		gnmi.Delete(t, dut, d.Interface(p2.Name()).Subinterface(0).Config())
-		gnmi.Delete(t, dut, d.Interface(p3.Name()).Subinterface(0).Config())
-		gnmi.Delete(t, dut, d.Interface(p4.Name()).Subinterface(0).Config())
+		gnmi.Delete(t, dut, d.NetworkInstance(ni).Interface(p2.Name()+"."+"0").Config())
```
nsadhasivam, line 859: Please use Subinterface() instead of hardcoding the value as 0

Please do not club gnmi.Delete with deviations.ExplicitInterfaceInDefaultVRF.
If explicit gnmi.Delete is needed and it works differently for Nokia when compared to other vendors.
Have it captured under a new deviation.

https://github.com/openconfig/featureprofiles/blob/main/internal/deviations/README.md

nsadhasivam (changes requested):

feature/gribi/otg_tests/base_hierarchical_nhg_update/base_hierarchical_nhg_update_test.go
nsadhasivam: @dplore can you please confirm if having gnmi.Delete along with deviations.ExplicitInterfaceInDefaultVRF accepted?

Regarding "Subinterface() instead of hardcoding the value as 0" , it is not actually subinterface is used here . it is interface id that was used while adding subitnerface into DEFAULT vrf i.e " interace_name.subinterface_id " https://github.com/openconfig/featureprofiles/blob/main/internal/fptest/networkinstance.go#L51 . Regarding usage of "0" , it is not hard coded value the script is designed to use subitnerface "0" at the initial interface config so entire script followed that same.
               
                   Accepted.

dplore: The preferred style is not using gnmi.Delete in general.  Instead we should use gnmi.Replace to over-write configuration.  It seems the configDUTDrain does this at least partially at:
https://github.com/openconfig/featureprofiles/pull/2237/files/d6f3b60b44a682505aef859bfa971961237b1e2a#diff-8acaeb616dcb7318148a8e7c202fb4ffc7e4c7aaf98c5a59a9ed34d410583dfdR877

Why does nokia need to delete the interface before doing a replace?

feature/gribi/otg_tests/base_hierarchical_nhg_update/base_hierarchical_nhg_update_test.go
```diff
@@ -737,14 +727,15 @@ func testImplementDrain(ctx context.Context, t *testing.T, args *testArgs) {
 	p2Flow := createFlow(t, p2FlowName, args.top, true, false, &atePort2)
 	p3Flow := createFlow(t, p3FlowName, args.top, true, false, &atePort3)
 	p4Flow := createFlow(t, p4FlowName, args.top, true, false, &atePort4)
+	deleteinterfaceconfig(t, args.dut)
```
gggsmith, line 730: Curious on why is the interface cleanup config is needed only here, but not for the other tests in `TestBaseHierarchicalNHGUpdate`?

bkreddy143: @gggsmith , following fucntioncall configDUTDrain() is trying to add the interfaces into LAG bundle without deleting subinterfaces and its references from network-instance . We need to delete subinterfaces to use port(interface) as LAG member port.

PR #2245, opened 2023-10-11
RT-1.12: BGP always compare MED  :  Adding Import policy

cprabha:
Added allow policy after configuring MED to accept the bgp route instead of using default implicit policy


Ankur19:
Hi @cprabha This is not required. I already had a PR out for explicitly applying the policy: https://github.com/openconfig/featureprofiles/pull/2238

cprabha:
Closing due to https://github.com/openconfig/featureprofiles/pull/2238

PR #2260, opened 2023-10-16
Update subtest headings to match template

dplore:
Update subtest headings to match template


dplore:
closing as this is obsolete

PR #2272, opened 2023-10-20
BGP Default Policies README.md

sachendras:
## Summary

Following expectation for default-policies at the peer-group and neighbor levels for BGP
*   Default import and export policies are expected to be functional for both IBGP and EBGP peers when policy definition in the policy chain is not satisfied.
*   For IBGP and EBGP peers, when no policy is attached, actions defined in the default-policy (when one exists) should apply.
*   For IBGP peers when no policy is attached including the default-policy, default should be import and export all BGP routes.
*   For EBGP peers when no policy is attached includign the default-policy, default should be to disallow import and export of all BGP routes.
*   In all the above cases, BGP default-policy is applicable only to the BGP learnt routes or the routes that are redistributed in to BGP from other protocols. Import/Export of routes from other protocols should be governed by their respective policies and should not either be exported or imported based on the actions of the BGP default-policy.

### DIFF
--- a/feature/bgp/policybase/otg_tests/default_policies_test/README.md
+++ b/feature/bgp/policybase/otg_tests/default_policies_test/README.md
@@ -5,9 +5,9 @@
 Following expectation for default-policies at the peer-group and neighbor levels
 *   Default import and export policies are expected to be functional for both IBGP and EBGP peers when policy definition in the policy chain is not satisfied.
 *   For IBGP and EBGP peers, when no policy is attached, actions defined in the default-policy (when one exists) should apply.
-*   For IBGP peers when no policy is attached and a default-polciy doesnot exist, default should be import and export all BGP routes.
-*   For EBGP peers when no policy is attached and a default-polciy doesnot exist, default should be to disallow import and export of all BGP routes.
-*   In all the above cases, BGP default-policy is applicable only to BGP learnt routes or the routes that are redistributed in to BGP. Routes from other protocols should be governed by their respective policies and should not either be exported or imported based on the actions of the BGP default-policy.
+*   For IBGP peers when no policy is attached including the default-policy, default should be import and export all BGP routes.
+*   For EBGP peers when no policy is attached includign the default-policy, default should be to disallow import and export of all BGP routes.
+*   In all the above cases, BGP default-policy is applicable only to the BGP learnt routes or the routes that are redistributed in to BGP from other protocols. Import/Export of routes from other protocols should be governed by their respective policies and should not either be exported or imported based on the actions of the BGP default-policy.
   
 ## Topology
 ```mermaid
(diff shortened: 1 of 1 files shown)


feature/bgp/policybase/otg_tests/default_policies_test/README.md
```diff
@@ -0,0 +1,135 @@
+# RT-7: BGP default policies
+
+## Summary
+
+Following expectation for default-policies at the peer-group and neighbor levels
+*   Default import and export policies are expected to be functional for both IBGP and EBGP peers when policy definition in the policy chain is not satisfied.
+*   For IBGP and EBGP peers, when no policy is attached, actions defined in the default-policy (when one exists) should apply.
```
dplore, line 7: Let's assert that for eBGP we require [RFC 8212](https://www.rfc-editor.org/rfc/rfc8212.html) compliant behavior where all routes are rejected for import and export if NO policy is applied to the (neighbor, peer, afi/safi).

Let's assert that for iBGP we require all routes are accepted for import and export if NO policy is applied to the (neighbor, peer, afi/safi).

Add TODO for OC clarification to confirm the eBGP behavior and reference https://github.com/openconfig/public/issues/981
Add TODO for OC clarification to confirm the iBGP behavior and reference https://github.com/openconfig/public/issues/981

feature/bgp/policybase/otg_tests/default_policies_test/README.md
```diff
@@ -0,0 +1,135 @@
...
+*   For IBGP peers when no policy is attached including the default-policy, default should be import and export all BGP routes.
+*   For EBGP peers when no policy is attached includign the default-policy, default should be to disallow import and export of all BGP routes.
+*   In all the above cases, BGP default-policy is applicable only to the BGP learnt routes or the routes that are redistributed in to BGP from other protocols. Import/Export of routes from other protocols should be governed by their respective policies and should not either be exported or imported based on the actions of the BGP default-policy.
```
dplore, line 10: Let's assert that for eBGP we require [RFC 8212](https://www.rfc-editor.org/rfc/rfc8212.html) compliant behavior where all routes are rejected for import and export if NO policy is applied to the (neighbor, peer, afi/safi).

Let's assert that for iBGP we require all routes are accepted for import and export if NO policy is applied to the (neighbor, peer, afi/safi).

Add TODO for OC clarification to confirm the iBGP behavior and reference https://github.com/openconfig/public/issues/981

```suggestion
*   For eBGP peers, when no policy is attached all routes should be rejected for import and export per [RFC 8212](https://www.rfc-editor.org/rfc/rfc8212.html).
    *.    TODO: OC clarification to confirm the eBGP behavior, https://github.com/openconfig/public/issues/981


*   For IBGP peers when no policy is attached including the default-policy, default should be import and export all BGP routes.
    *    TODO:  OC clarification on iBGP behavior, https://github.com/openconfig/public/issues/981

```

dplore (approved):

PR #2284, opened 2023-10-23
DP-1.3 - Adding required interface ref config

charantejag504:
DP-1.3 - Adding required interface ref config

https://github.com/openconfig/featureprofiles/pull/2022 removed all vendor specific functions (ciscoEcnConfig, juniperEcnConfig etc). Interface-ref config was part of those vendor specific functions. When vendor specific functions were removed, interface-ref config was not added in the common function. I am adding it back as it is a mandatory config for /qos/interfaces/


sachendras:
Please add more details to the `Description` to provide more context. You add **what** you are doing which I think anyone can tell from the code. Important is **why** you are doing, so that it will help w/ review.

sachendras (dismissed):

sachendras:
> Please add more details to the `Description` to provide more context. You add **what** you are doing which I think anyone can tell from the code. Important is **why** you are doing, so that it will help w/ review.

Next time please also make sure to include the original PR where the breaking config was added in this case https://github.com/openconfig/featureprofiles/pull/2022.. Providing more context helps w/ the review process.

arulkumarsekar (approved):

PR #2288, opened 2023-10-24
Add test for ZR RX input and TX Output power streaming

ahsaanyousaf:


feature/platform/transceiver/zr_input_output_power_test/README.MD
```diff
@@ -0,0 +1,112 @@
+# gNMI-1.20: Telemetry: 400ZR RX input and TX output power telemetry values streaming.
```
dplore, line 1: Let's assign the next test id.  You should also update testregistry.textproto by [adding a new entry right after this one](https://github.com/openconfig/featureprofiles/blob/main/testregistry.textproto#L874-L879).

```suggestion
# TRANSCEIVER-4: Telemetry: 400ZR RX input and TX output power telemetry values streaming. 
```

feature/platform/transceiver/zr_input_output_power_test/README.MD
```diff
@@ -0,0 +1,112 @@
...
+    stream any invalid string values like "nil" or "-inf" until valid values
+    are available for streaming.
+
+*   RX Input and TX output power values must always be of type decimal64.
+    When link interfaces are in down state 0 must be reported as a valid
+    value.
+
+**Note:** For min, max, and avg values, 10 second sampling is preferred. If 
+          10 seconds is not supported, the sampling interval used must be
+          communicated.
```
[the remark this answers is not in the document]
dplore: The use of 'deviation' to assert differences in the DUT behavior versus requirements is the standard method.  You can see the [list of all deviations](https://github.com/openconfig/featureprofiles/blob/main/internal/deviations/deviations.go).  As test code is implemented, you can see the [deviations applied to each test](https://github.com/openconfig/featureprofiles/blob/main/feature/platform/tests/telemetry_inventory_test/metadata.textproto#L8-L25) .

dplore (changes requested):

feature/platform/transceiver/zr_input_output_power_test/README.MD
```diff
@@ -0,0 +1,112 @@
...
+As per CMIS ZR modules report two types of RX input power and one TX output
+power.
+* RX Signal Power
+  * Reports the actual signal power after filtering out any extra noise.
+  * Is mapped to /component/optical-channel/ full path shown below
+* RX Total Power
+  * Reports RX Signal Power plus noise without any filtering.
+  * Is mapped to /component/transceiver/physical-channel full path shown below
+* TX Output Power
+  * This is the total TX output power
```
ejbrever, line 17: Is there such thing as signal power on the TX or just the total TX output power? I wasn't sure if CMIS or the hardware supported both.

ahsaanyousaf: There is no Signal TX power defined in CMIS. This is primarily because on RX side we have a filter, power measure after the filter is Signal power and before the filter is total power. No such filter on TX side.

ejbrever: Thanks for confirming.

ejbrever (approved):

dplore (approved):

PR #2298, opened 2023-10-26
P4RT-1.2: fix bug in subscribeOnChangeInterfaceName

kjahed:
The watcher in subscribeOnChangeInterfaceName never returns a value as dutPort1.Name is not set. This PR:
1- Changes the watcher condition to use dut.Port("port1").Name() instead.
2- Subscribe to /interface/id after p4rt daemon kill rather than before. This is necessary for systems that use the same process for p4rt and gnmi.


Ankur19:
HI @kjahed For the second change `Subscribe to /interface/id after p4rt daemon kill rather than before. This is necessary for systems that use the same process for p4rt and gnmi.` : we may need to skip the test altogether for such systems.

The intent of the test is to check that the system doesn't update the `/interface/id` leaf to `null` with an `on_change` gnmi notification during the p4rt process shutdown. 
This scenario is not applicable when gnmi itself shuts down when p4rt is down.

LimeHat:
> we may need to skip the test altogether for such systems.
> 
> The intent of the test is to check that the system doesn't update the `/interface/id` leaf to `null` with an `on_change` gnmi notification during the p4rt process shutdown.

@Ankur19 was this meant to be only about the telemetry part, or the whole test (including dataplane validation during daemon restart)? IMO only telemetry should be skipped.

Ankur19:
> > we may need to skip the test altogether for such systems.
> > The intent of the test is to check that the system doesn't update the `/interface/id` leaf to `null` with an `on_change` gnmi notification during the p4rt process shutdown.
> 
> @Ankur19 was this meant to be only about the telemetry part, or the whole test (including dataplane validation during daemon restart)? IMO only telemetry should be skipped.

Right @LimeHat ! I meant to say we need to skip the Test for the value in /interface/id.

@kjahed I meant the test for change in /interface/id

kjahed:
> > > we may need to skip the test altogether for such systems.
> > > The intent of the test is to check that the system doesn't update the `/interface/id` leaf to `null` with an `on_change` gnmi notification during the p4rt process shutdown.
> > 
> > 
> > @Ankur19 was this meant to be only about the telemetry part, or the whole test (including dataplane validation during daemon restart)? IMO only telemetry should be skipped.
> 
> Right @LimeHat ! I meant to say we need to skip the Test for the value in /interface/id.
> 
> @kjahed I meant the test for change in /interface/id

Got it. Updated accordingly.

feature/p4rt/otg_tests/p4rt_daemon_failure_test/p4rt_daemon_failure_test.go
```diff
@@ -359,12 +358,15 @@ func TestP4RTDaemonFailure(t *testing.T) {
...
-	// Verify interfaceID did not change since the last time we read it.
-	changedID, notOk := watchID.Await(t)
-	if notOk {
-		t.Errorf("FAIL: DUT changed /interfaces/interface/state/id  during p4rt process restart.  want:%q got:%q", dutPort1.ID, changedID.String())
+	// Skip check for CISCO devices that use the same process for P4RT & gNMI.
+	if dut.Vendor() != ondatra.CISCO {
+		// Verify interfaceID did not change since the last time we read it.
+		changedID, notOk := watchID.Await(t)
+		if notOk {
+			t.Errorf("FAIL: DUT changed /interfaces/interface/state/id  during p4rt process restart.  want:%q got:%q", dutPort1.ID, changedID.String())
```
robshakir, line 366:
```suggestion
			t.Errorf("DUT changed /interfaces/interface/state/id during p4rt process restart.  want: %q got: %q", dutPort1.ID, changedID.String())
```

No need to put `FAIL` at the start of the string, since Go will do this itself.

feature/p4rt/otg_tests/p4rt_daemon_failure_test/p4rt_daemon_failure_test.go
```diff
@@ -359,12 +358,15 @@ func TestP4RTDaemonFailure(t *testing.T) {
 	t.Logf("Stop traffic")
 	ate.OTG().StopTraffic(t)
 
-	// Verify interfaceID did not change since the last time we read it.
-	changedID, notOk := watchID.Await(t)
-	if notOk {
-		t.Errorf("FAIL: DUT changed /interfaces/interface/state/id  during p4rt process restart.  want:%q got:%q", dutPort1.ID, changedID.String())
+	// Skip check for CISCO devices that use the same process for P4RT & gNMI.
+	if dut.Vendor() != ondatra.CISCO {
```
robshakir, line 362: Can you expand on the change that you made here? `subscribeOnChangeInterfaceName` is checking that you get an `ON_CHANGE` notification for `/interfaces/interface[name=eth0]/state/name` ahead of the test -- this seems fine (as the comment says we're checking a well-supported path).

However, now for Cisco you are never validating that `/interfaces/interface[name=...]/state/id` doesn't change across a process restart -- which is the key thing in this test. It's not clear to me why running on the same port invalidates the need to check that the ID doesn't change.

If the concern is that the subscription dies during the test because gNMI restarted, then one could implement this differently -- and still validate this coverage -- by blocking on gNMI coming back (calling watchName or something similar with a longer timeout -- failing if it doesn't return within X seconds) and then checking the ID didn't change afterwards.

Either way, removing the check completely for Cisco seems a regression in coverage for something that's important.

robshakir (changes requested):

feature/p4rt/otg_tests/p4rt_daemon_failure_test/p4rt_daemon_failure_test.go
kjahed: Thanks @robshakir. For `subscribeOnChangeInterfaceName`, the predicate for the Watch call was comparing the received value against dutPort1.Name which is not set in the test. Changed the function to take the actual port name.

For the interface ID, based on what I understood from @Ankur19 comment's above, the intent of the test is to validate that interface is not nullified **during** process restart. Obviously, I can get the interface id before and after the process restart and make sure they match, but that does not guarantee that the interface id has not changed at all, which is what the test is currently doing with an ON_CHANGE subscription. If gNMI is restarted with P4RT, that subscription stream is lost.

robshakir: My recommendation would be to ensure that it does not change -- but I agree that it also should not change during.

@greg-dennis -- is the expectation that there is no reconnection of underlying streams should the gNMI connection be lost during a test? It seems like something that should maybe be supported to say "persistently reconnect until I get a value for path /a/b/c". WDYT?

greg-dennis: Is the gnmi connection being restarted a practical problem that is happening or a theoretical problem?

The ygnmi library uses the gnmi client it is given, so doing that would be a matter of configuring the gnmi client, to automatically reconnect. If it is a problem occurring today,  could you try the `grpc.WaitForReady(true)` call option to see if it goes away?

kjahed: Thanks Rob and Greg. grpc.WaitForReady(true) did not help. Await() after P4RT/gNMI server process restart still returns EOF. Also tried adding grpc_retry.WithMax(5) but it seems retry is not supported for streams. 

```
p4rt_daemon_failure_test.go:364: Await(t) on &{SF-D %!s(func(context.Context) (gnmi.GNMIClient, error)=0x8810c60) map[] %!s(bool=true) <nil> [%!s(ygnmi.Option=0x98003a0)]} at /interfaces/interface[name=TenGigE0/0/0/4/0]/state/id: error receiving gNMI response: rpc error: code = Unavailable desc = error after request {
        subscribe:  {
          prefix:  {
            target:  "SF-D"
          }
          subscription:  {
            path:  {
              origin:  "openconfig"
              elem:  {
                name:  "interfaces"
              }
              elem:  {
                name:  "interface"
                key:  {
                  key:  "name"
                  value:  "TenGigE0/0/0/4/0"
                }
              }
              elem:  {
                name:  "state"
              }
              elem:  {
                name:  "id"
              }
            }
            mode:  ON_CHANGE
          }
          encoding:  PROTO
        }
        }: error reading from server: EOF
```

prinikasn, line 362: Should this be a deviation instead of a check `if dut.Vendor() != ondatra.CISCO` ?

robshakir: @greg-dennis I think the issue here is that if you get an error a `ygnmi.Watcher` exits because it gets EOF, so even though the `grpc.ClientConn` that you provided reconnects, nothing recreates the `gnmi.Subscribe` on top of that connection until you call something that creates the watcher.

I was wondering whether `gnmi.OC().Await()` in ONDATRA should cause the watcher to be recreated following any loss of the connection.

kjahed: I don't think that's a deviation.

greg-dennis: @robshakir, I don't know whether it makes sense to add in `Await` logic. Maybe, but it complicates the semantics and code and perhaps should be left to higher-level retrying logic the wraps the Await call. If it is included in Await directly, it would probably need to be guarded by an option, at least to start. I'd like to get @DanG100's and @wenovus's opinion on it. Regardless, would be resolved separately from this PR.

robshakir: Sure -- OK. Since this is also gated on a specific vendor, I'm OK with this.

robshakir: Can you explain why this isn't a deviation? I tend to agree with Prinika here.

aaronmillisor: In Cisco's current implementation all gRPC services are managed by the same process (emsd), and so killing any process related to grpc will kill all hosted services. 

Is it the goal of feature profiles to mandate a specific implementation (i.e. separate daemons for each service)?

prinikasn (approved):

feature/p4rt/otg_tests/p4rt_daemon_failure_test/p4rt_daemon_failure_test.go
robshakir: Apologies, I thought I had replied here.

The intention is not to imply any requirements here. I see this more as just "where do we store differences between implementations" -- I'd suggest that generally doing this through deviations is cleaner as it leads to a single place to store these differences, and avoids the need for there to be `if` conditions with many vendors in them through the code.

I'll approve, but would suggest to the wider team here that for the sake of maintainability we keep these in a single place.

robshakir (approved):

PR #2303, opened 2023-10-27
Bgp asn policy update

s-nandu:
Adding accept term to policy


sachendras:
Lettle more context would help w/ the review process

sachendras (approved):

PR #2314, opened 2023-10-27
Removing InterfaceRefConfigUnsupported deviation for Policy-forwarding

arvbaska1:
Made below changes 

- Removing InterfaceRefConfigUnsupported deviation for RT-3.1,RT-3.2,TE-14.1,TE-3.1,TE-3.3 tests.
- Have not removed the InterfaceRefConfigUnsupported from deviations, since its used in other tests.


LimeHat:
What's the purpose of interface renaming?

arvbaska1:
> What's the purpose of interface renaming?

As per https://github.com/openconfig/public/blob/master/release/models/interfaces/openconfig-interfaces.yang#L177, it recommends the interface-id should be a combination of Interface name + subinterface index. So its changed to follow this format here.

arvbaska1:
> That is true, however we also have [openconfig/public/pull/802](https://github.com/openconfig/public/pull/802) that clarified how interface-ref is supposed to be used. If interface.subinterface format is the requirement of an implementation, IMO this should be a deviation based on the above.
> 
> If it is not a requirement and you intend to just remove the deviation - then all nice-to-haves can be a separate PR

I have added interface.subinterface with a deviation

guoshiuan (approved):

PR #2126, opened 2023-09-14
P4RT-5.2: add subtests for all metadata combinations

kjahed:
Adds subtests for all combinations of metadata. Combination are generated from:

egress_port={not_set, "TBD BY SWITCH" (i.e., invalid), "2100" (i.e., valid)}
X
submit_to_ingress={not_set, 0x0, 0x1} 
X
padding={not_set, "0"}

The test validates that traffic was sent out the correct port.

PR #2316, opened 2023-10-30
Modify for P4RT-5.2

ArpitJoshi18:
Modifying P4RT-5.2 Traceroute Packet-out to select 2 ports out of 4 ports which belongs in the same P4RT Node name. 
Traffic is also sent and expected on the same selected ports.


ArpitJoshi18:
> Can you elaborate as to why ports have to be on the same node?

@LimeHat Please find discussion in comments from this PR  https://github.com/openconfig/featureprofiles/pull/2126/#issuecomment-1737878950

LimeHat:
Are you referring to the last comment from the PR?

> We have observed "The DUT port1 and port2 needs to be in the same device/node for the packetout traffic with submit_to_ingress set to 0x0 or NIL and egress_port_id set to '2101' for traffic to be forwarded, else traffic will be dropped as port2 belongs to a difference device/node for which the current controller has no client session".

If so, it's still uncelar why that should be the case. Port-2 is a regular ip interface destination, where does the p4rt session dependency comes from (is there a reference to the spec)?

Ankur19:
> Are you referring to the last comment from the PR?
> 
> > We have observed "The DUT port1 and port2 needs to be in the same device/node for the packetout traffic with submit_to_ingress set to 0x0 or NIL and egress_port_id set to '2101' for traffic to be forwarded, else traffic will be dropped as port2 belongs to a difference device/node for which the current controller has no client session".
> 
> If so, it's still unclear why that should be the case. Port-2 is a regular ip interface destination (with an assigned p4rt id), where does the p4rt session dependency comes from (is there a reference to the spec)?

Hi @LimeHat Consider a scenario where dut:port1 belongs to ASIC1 and dut:port2 belongs to ASIC2. 

Now from ASIC1, we can send packetOuts only to ports mapped to ASIC1 for egress_port_id scenario right?

LimeHat:
Ok, so it's not about the presence of the session as indicated in the quoted comment above?

> Now from ASIC1, we can send packetOuts only to ports mapped to ASIC1 for egress_port_id scenario right?

That's not my interpretation of the spec, but I see how a case can be made for this. 
Perhaps it would be good to update the README first to define the expected behavior:
- packetOut from a controller should be able to reach any sdn port
- packetOut from a controller should be able to reach only ports that belong to the same ASIC
- both behaviors are acceptable

Ankur19:
> Ok, so it's not about the presence of the session as indicated in the quoted comment above? More about the permission management.
> 
> > Now from ASIC1, we can send packetOuts only to ports mapped to ASIC1 for egress_port_id scenario right?
> 
> That's not my interpretation of the spec, but I see how a case can be made for this.
> 
> Perhaps it would be good to update the README first to define the expected behavior:
> 
> * packetOut from a controller should be able to reach any sdn port
> * packetOut from a controller should be able to reach only ports that belong to the same ASIC (for which the controller is a primary)
> * both behaviors are acceptable
> 
> And after we get this clarification the test can be updated to validate the behavior with both positive and negative (if opt#2 is required) scenarios.

It is related to the session as well. A leader client would be attached to a single `device_id` right ((device_id, role, election_id) being unique)? I think the question is: should this client be able to send packets out through a port mapped to another device_id.

For `submit_to_ingress` scenario this is definitely true because you are sending the packet out using an ip address. THe confusion is about the `egress_port` scenario where we are trying to send a packet out through a specific port (that may not be controlled by the current device_id)

LimeHat:
@Ankur19 agree, that's pretty much what i meant / that's another way to phrase it.

If it's challenging to clarify this requirement right away, I'd suggest to modify the PR in a much simpler way that will allow both behaviors for the time being: just swap roles of port 1 and port 2 (right now all `submit_to_ingress=1` traffic is expected to arrive on port-1, and `submit_to_ingress=0` on port-2, while the client is connected to the port-1 ASIC). E.g. client can be connected to port-2 instead.

ArpitJoshi18:
> It is related to the session as well. A leader client would be attached to a single `device_id` right ((device_id, role, election_id) being unique)? I think the question is: should this client be able to send packets out through a port mapped to another device_id.
> 
> For `submit_to_ingress` scenario this is definitely true because you are sending the packet out using an ip address. THe confusion is about the `egress_port` scenario where we are trying to send a packet out through a specific port (that may not be controlled by the current device_id)

Thanks @Ankur19,

Script performs port-selection, and ensures that the selected two ports belongs to the same ASIC, for which P4RT session is established. This helps the packet_out to be the successful when 'submit_to_ingress' is not set, where 'egress_port' will point to the port which belongs in the same ASIC, for which P4 controller has mastership. 

On the other hand. when 'submit_to_ingress' is set, we know that packet_out will be purely based on the destination IP address ('egress_port' METADATA  will be silently ignored) .

LimeHat:
@ArpitJoshi18 will the above suggestion work for you then?

One-liner with a change at L145 if i'm not mistaken:
https://github.com/openconfig/featureprofiles/blob/1d26646337e0fdecf7d0da8bb43cb4e98d230427/feature/experimental/p4rt/otg_tests/traceroute_packetout_test/traceroute_packetout_test.go#L145
Replace "port1" with "port2" here.

ArpitJoshi18:
> https://github.com/openconfig/featureprofiles/blob/1d26646337e0fdecf7d0da8bb43cb4e98d230427/feature/experimental/p4rt/otg_tests/traceroute_packetout_test/traceroute_packetout_test.go#L145
> 
> 
> Replace "port1" with "port2" here.

Yes @LimeHat , it works. I will edit it accordingly.

LimeHat:
thanks for the change @ArpitJoshi18 , lgtm

sezhang2 (approved):
It is a minor port change. We need to verify that the change does not break existing test before merge.

PR #2342, opened 2023-11-03
check interface forwarding-viable state

self-maurya:


feature/interface/aggregate/otg_tests/aggregate_forwarding_viable_test/aggregate_forwarding_viable_test.go
```diff
@@ -511,10 +511,21 @@ func (tc *testArgs) verifyCounterDiff(t *testing.T, before, after []*oc.Interfac
...
-		gnmi.Update(t, tc.dut, gnmi.OC().Interface(tc.dutPorts[1].Name()).ForwardingViable().Config(), forwardingViable)
+		gnmi.Update(t, tc.dut, gnmi.OC().Interface(pName).ForwardingViable().Config(), forwardingViable)
+	}
+
+	v := gnmi.Lookup(t, tc.dut, gnmi.OC().Interface(pName).ForwardingViable().State())
+	got, present := v.Val()
+	t.Logf("First port %s forwarding-viable: got %v, present %v, want %v", pName, got, present, forwardingViable)
+	if present && got != forwardingViable {
+		t.Errorf("First port %s forwarding-viable: got %t, want %t", pName, got, forwardingViable)
+	} else if !present && !deviations.MissingValueForDefaults(tc.dut) {
```
LimeHat, line 527: If the deviation is `true`, this will skip the check for both default (true) and non-default (false) values of `forwarding-viable`, e.g. if the device did not return `false` when expected to do so.
Perhaps an extra condition (`&& want == true`) should be added. Or a new deviation if the state flag is not supported at all.

self-maurya: Right, if the value is not present and not a deviations then it's an error as value is not defined (already covered)

Another case would be, value not present and it's a deviation and want false, it should also result in an error as with deviation we assume default value to be true which is not equal to want value

LimeHat: Thanks, looks good now

arulkumarsekar (approved):

PR #2348, opened 2023-11-06
gnmi_subscriptionlist_test

sachendras:
This is to test for gNMI `Subscription` to multiple paths with different `SubscriptionMode` in a single `SubscriptionRequest` message using the `Subscriptionlist` field. Goal here is to,
  * Ensure that the NOS supports "Subscriptionlist" field with multiple `Subscription` messages and also supports the desired `Subscriptionmode` per path in each `Subscription` message.
  * The tests also check if the DUT is responding back everytime with a `SubscriptionResponse` message that has the `sync_response` field set to `true`


feature/gnmi/gnmi_subscriptionlist_test/README.md
```diff
@@ -0,0 +1,51 @@
...
+    * /interfaces/interface/state/counters/out-discards
+    * /interfaces/interface/state/counters/in-errors
+    * /interfaces/interface/state/counters/out-errors
+    * /interfaces/interface/state/counters/in-fcs-errors
+    * /qos/interfaces/interface/output/queues/queue/state/transmit-pkts
+    * /qos/interfaces/interface/output/queues/queue/state/transmit-octets
+    * /qos/interfaces/interface/output/queues/queue/state/dropped-pkts
+    * /components/component/integrated-circuit/backplane-facing-capacity/state/available-pct
+    * /components/component/integrated-circuit/backplane-facing-capacity/state/consumed-capacity
+    * /components/component/integrated-circuit/backplane-facing-capacity/state/total”
```
dplore, line 49: This path should be in the ON_CHANGE category because it would only change on FABRIC card insertion or removal

feature/gnmi/gnmi_subscriptionlist_test/README.md
```diff
@@ -0,0 +1,51 @@
...
+    * /components/component/integrated-circuit/backplane-facing-capacity/state/total-operational-ca
```
dplore, line 50: This path should probably be in the ON_CHANGE category because it would only change on FABRIC card failure or admin-power down.

feature/gnmi/gnmi_subscriptionlist_test/README.md
```diff
@@ -0,0 +1,51 @@
...
+### GNMI-2.1: Verify single subscription request with a Subscriptionlist and different SubscriptionModes:
+  * Send a single `SubscribeRequest` message to the DUT with a **SubcriptionList** and **SubscriptionMode** matching the "Telemetry Parameter Coverage" section below. Use `Stream` mode for the `SubcribeRequest`.
+  * Ensure that the implementation successfully allows subscription to all the paths mentioned below and a `SubscribeResponse` message is received by the client with the `sync_reponse` field set to `true`. The RPC via which the `SubscribeRequest` was recieved should eventually be closed by the client.
+### GNMI-2.2: Change SubscriptionModes in the subscription list and verify receipt of sync_response:
+  * In the "Telemetry Parameter coverage" section below, change the `Subscribe` message for each of the paths with `SubscriptionMode` as `ON_CHANGE` to `TARGET_DEFINED` and the ones that are `TARGET_DEFINED` to `SAMPLE` and send all the subscribe messages in a single `SubscribeRequest` message to the DUT. Confirm that a `SubscribeResponse` message is received by the client with the `sync_reponse` field set to `true`. The client should then close the RPC session
```
dplore, line 13: What sample interval should be used?  Other tests are specifying two.  10 second and 15 second.  Probably use that for now.  (can tweak it later if needed)

feature/gnmi/gnmi_subscriptionlist_test/README.md
sachendras: Done

feature/gnmi/gnmi_subscriptionlist_test/README.md
```diff
@@ -10,8 +10,8 @@ This is to test for gNMI `Subscription` to multiple paths with different `Subscr
...
+  * In the "Telemetry Parameter coverage" section below, change the `Subscribe` message for each of the paths with `SubscriptionMode` as `ON_CHANGE` to `TARGET_DEFINED` and the ones that are `TARGET_DEFINED` to `SAMPLE` w/ a sampe_interval of 10secs and send all the subscribe messages in a single `SubscribeRequest` message to the DUT. Confirm that a `SubscribeResponse` message is received by the client with the `sync_reponse` field set to `true`. The client should then close the RPC session
+  * Again, switch the `SubscriptionMode` in each `Subscription` message to its original state i.e. from `TARGET_DEFINED` to `ON_CHANGE` and from `SAMPLE` to `TARGET_DEFINED` and resend the `SubscriptionRequest` with `Mode` as `STREAM`. Confirm that the DUT is responding back to the client with a `SubscriptionResponse` and the `Sync_Response` field set to `true`
```
dplore, line 14: Do you want     * /interfaces/interface/state/counters/in-unicast-pkts to be collected in ON_CHANGE mode?  I think we don't want counters in ON_CHANGE?  counters only need to support TARGET_DEFINED and SAMPLE mode.  Right?

There is no 'original state' of SAMPLE?  The only original states below are ON_CHANGE and TARGET_DEFINED.  Maybe you could reword this to not be conditional on the previous steps.  Just say what mode you want for which group of paths.  Give the path groups a name to simplify.

sachendras: I am not sure if we are on the same page. Heres what is happening

1. Start off the testing w/ the paths in "Telemetry Parameter Coverage" section with the "SubscriptionMode" as stated.
2. Then switch the ones that are ON_CHANGE to TARGET_DEFINED and the TARGET_DEFINED ones to SAMPLE
3. Then revert the state in "2" to its Original form i.e. TARGET_DEFINED to ON_CHANGE and SAMPLE to TARGET_DEFINED
When we follow "3", then we restore "Telemetry Parameter Coverage" to how it is stated in the section below and eqvt to "1"
 In that sense then: /interfaces/interface/state/counters/in-unicast-pkts becomes TARGET_DEFINED and not ON_CHANGE.

dplore: Ok, got it.

dplore (approved):

testregistry.textproto
```diff
@@ -834,6 +834,12 @@ test: {
   readme: "https://github.com/openconfig/featureprofiles/blob/main/feature/system/gnmi/get/tests/system_gnmi_get_test/README.md"
   exec: " "
 }
+test: {
+  id: "gNMI-2"
+  description: "Gnmi Subscriptionlist Test"
+  readme: ""
```
dplore, line 840:
```suggestion
  readme: "https://github.com/openconfig/featureprofiles/blob/main/feature/gnmi/subscribe/tests/gnmi_subscriptionlist_test/README.md"
```

feature/gnmi/gnmi_subscriptionlist_test/README.md
dplore, line 1: To follow test folder pattern, please move to  feature/gnmi/subscribe/tests/gnmi_subscriptionlist_test/README.md

sachendras: Done

dplore (approved):

PR #2349, opened 2023-11-06
Update README.md for singleton_test

sachendras:
Subtest added to cover for scenarios whereby a DUT can stop reporting /state/hardware-port details for an interface after an OS change. Given that a device can report this leaf for an interface irrespective of its admin-status, this sub-test should be able to cover for different breakout configurations that exits on the DUT.


feature/interface/singleton/otg_tests/singleton_test/README.md
```diff
@@ -40,6 +40,10 @@ a new testbed configuration with the desired port types.
             captured in the OC path.
   * Verify that the traffic flow from ATE port-1 to ATE port-2 is
             now working after the interface is back up.
+### RT-5.1.3 - Verify accurate reporting of `hardware-port` on all interfaces:
+* The test should start with taking an inventory of all available interfaces on the DUT irrespective of their admin-status or breakout state.
+* The test should check for `/interfaces/interface/state/hardware-port` reporting on all available physical interfaces in the DUT irrespective of their admin-status or breakout state
+* Compare results in both the bullets above and confirm that the DUT is not missing reporting of `state/hardware-port` on any physical interface irrespective of their breakout state. If missed, the test is a failure.
```
dplore, line 47: Sorry I don't understand the different between the 1st and 2nd bullet items?  They sound like the same query?

First bullet gets `/interfaces/interface` tree
The second send another query which is a subset of the above:  `/interfaces/interface/state/hardware-port`

Did we see an issue where this would not match?  It seems very unlikley?  Perhaps there is something else intended to be checked?

sachendras: Yes in b/301104344#comment23. Seems after OS upgrade, breakout ports stopped reporting state/hardware-port. 

Alos, https://github.com/openconfig/featureprofiles/pull/2348 is expected to cover for requirements whereby, paths that are expected to work w/ ON_CHANGE and TARGET_DEFINED should work as such.

dplore: OK

dplore: I think you can simplify these three bullets into this which is more clear:
* get `/interfaces/interface/`
* for every interface, verify interfaces/interface/state/hardware-port is populated with a reference to `/components/component/name`

sachendras: done

dplore (approved):

dplore (approved):
fixed the merge conflict (formatting of markdown heading)

sachendras:
Closing this one in favor of https://github.com/openconfig/featureprofiles/pull/2389

PR #2351, opened 2023-11-06
RT-5.4: Configure LACP on the interface only when LAG type is LACP

sgirishj:
Updates to RT-5.4, similar to the updates done in RT-5.2 through https://github.com/openconfig/featureprofiles/pull/2233/
- Configure LACP on the interface only when LAG type is LACP
- Switched order of tests to cover Static mode first and then LACP mode to avoid potential conflicting LACP/LAG configs


sezhang2 (approved):
It is the same code change as merged PR/2233, so I approve it
   https://github.com/openconfig/featureprofiles/pull/2233

PR #1701, opened 2023-06-02
gNOI-3.2: adding support for fabric component reboot

nicjiang2023:
Hi,

Please check the update for gNOI-3.2:

1.  Add support for fabric component reboot
2. Introduce a new deviation gnoiFabricComponentRebootUnsupported
3. Update Readme.md accordingly

This will resolve https://github.com/openconfig/featureprofiles/issues/1630

Thanks,
Nicole

PR #2356, opened 2023-11-08
gNOI-3.2: Adding wait time to avoid false failures

sgirishj:
Adding 10s sleep after linecard reboot is issued, to ensure that the linecards reboot process has kicked in, before we pull the linecard status and the list of 'up' interfaces. This is to avoid false failures owing to timing differences.


LimeHat:
There's already 10s sleep (in a loop) just below the one you added

https://github.com/openconfig/featureprofiles/blob/1d26646337e0fdecf7d0da8bb43cb4e98d230427/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go#L200

Any specific reason to duplicate it?

sgirishj:
> There's already 10s sleep (in a loop) just below the one you added
> 
> https://github.com/openconfig/featureprofiles/blob/1d26646337e0fdecf7d0da8bb43cb4e98d230427/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go#L200
> 
> Any specific reason to duplicate it?

Yes, in some corner cases we have seen the reboot process take around 10s to kick-in which leads to false failures. Didn't want to change the 10s thats within the loop. Hence added this one just before that. Thanks!

LimeHat:
>  Didn't want to change the 10s thats within the loop

Why not? LC boot time limit is set to ~10 minutes, doesn't look like we need a frequent polling there. Having two sleep timers next to each other is confusing IMO :)

Having said that, it also feels strange that gNOI.system.RebootStatus() returns misleading `active=false` result flag for more than 10 seconds after receiving a rebootRequest RPC. Wouldn't you agree?

sezhang2 (approved):
There is a retry logic for the reboot status check with wait time 10 seconds. 

The code change is to add extra 10 seconds for the first check to avoid false failure.

LimeHat:
@sezhang2, just to confirm, 
is it considered to be expected/valid behavior that an implementation may not return an `active=true` status in `RebootStatusResponse` (validated on L285-L291): https://github.com/openconfig/featureprofiles/blob/30a9fc77274d11a80a9d52a37ab43458d4338c90/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go#L285-L291) for more than 10 seconds after receiving a `RebootRequest`?

sezhang2:
> @sezhang2, just to confirm, is it considered to be expected/valid behavior that an implementation may not return an `active=true` status in `RebootStatusResponse` (validated on L285-L291):
> 
> https://github.com/openconfig/featureprofiles/blob/30a9fc77274d11a80a9d52a37ab43458d4338c90/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go#L285-L291
> 
> for more than 10 seconds after receiving a `RebootRequest`?

This PR only addresses the timing issue for line card reboot. Your question is  about fabric card rebooting from 

 -   https://github.com/openconfig/featureprofiles/pull/1701

Can you open an issue if there is a change needed for fabric card reboot test?

LimeHat:
>  Your question is about fabric card rebooting from

My bad, I quoted the wrong line numbers. I meant L208-L213. But the same code blob is used for linecards.
https://github.com/openconfig/featureprofiles/blob/e0bea24146e94a1eaad2de2e656ffad5f04e1782/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go#L208-L213

And no, I don't think there should be an issue in either case even with the previous code, if implementation properly responds with an active status; i.e. I think the extra sleep shouldn't be required at all.

Just wanted to clarify what are the expectations for this code block, if any.
By making this change we are saying that a) implementation can implicitly delay the reboot of a component for more than 10 seconds (which is ok, probably), and b) at the same time it is not responding with an `active=true` (which is more questionable IMO).
But anyway.

sezhang2:
> > Your question is about fabric card rebooting from
> 
> My bad, I quoted the wrong line numbers. I meant L208-L213. But the same code blob is used for linecards.
> 
> https://github.com/openconfig/featureprofiles/blob/e0bea24146e94a1eaad2de2e656ffad5f04e1782/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go#L208-L213
> 
> And no, I don't think there should be an issue in either case even with the previous code, if implementation properly responds with an active status; i.e. I think the extra sleep shouldn't be required at all.
> 
> Just wanted to clarify what are the expectations for this code block, if any. By making this change we are saying that a) implementation can implicitly delay the reboot of a component for more than 10 seconds (which is ok, probably), and b) at the same time it is not responding with an `active=true` (which is more questionable IMO). But anyway.


err check from fabric test case is missing. Basically, there are 3 possible return from resp.GetActive() 

1. err is not nil (err returned). no retrun result for resp.GetActive() 
2. true (err is nil) 
3. false (err is nil)

LimeHat:
I think for LC this code block will retry for every non-nil error except `codes.Unimplemented` (which will trigger t.Fatal), since `retry` var is initialized as true and is not modified. But I don't think the error check is relevant here.

The sequence of the events (pre-PR) was:
1) send rebootRequest
2) wait for 10 seconds
3) query RebootStatus
4) successfully exit only for nil error & active==false or after 10 minutes of waiting; or `t.Fatal` in case of unimplemented err

I.e. it appears that the OP encountered p.4 with `nil` error and `active==false` after 10 seconds of initial wait, and before the linecard began the reboot so the another condition at https://github.com/openconfig/featureprofiles/blob/e0bea24146e94a1eaad2de2e656ffad5f04e1782/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go#L220 was instantly validated as well, and this lead to a failure in the results. But if the implementation is able to set reboot status as `active=true` within 10 seconds after sending the request, then this early exit won't happen.

sezhang2:
> I think for LC this code block will retry for every non-nil error except `codes.Unimplemented` (which will trigger t.Fatal), since `retry` var is initialized as true and is not modified. But I don't think the error check is relevant here.
> 
> The sequence of the events (pre-PR) was:
> 
> 1. send rebootRequest
> 2. wait for 10 seconds
> 3. query RebootStatus
> 4. successfully exit only for nil error & active==false or after 10 minutes of waiting; or `t.Fatal` in case of unimplemented err
> 
> I.e. it appears that the OP encountered p.4 with `nil` error and `active==false` after 10 seconds of initial wait, and before the linecard began the reboot so the another condition at
> 
> https://github.com/openconfig/featureprofiles/blob/e0bea24146e94a1eaad2de2e656ffad5f04e1782/feature/gnoi/system/tests/per_component_reboot_test/per_component_reboot_test.go#L220
> 
> was instantly validated as well, and this led to a failure in the results. But if the implementation is able to set reboot status as `active=true` within 10 seconds after receiving the request, then this early exit won't happen.

Looks like we are talking about different things. err== nil needs to checked before evaluating resp.GetActive().
```
		case err == nil:
			retry = resp.GetActive()
```
That is why i say `err check from fabric test case is missing`.

PR #2358, opened 2023-11-08
ocpaths.proto and nosimage.proto

wenovus:
Add the following proto definitions:

ocpaths.proto:
```
// OCPaths -- Specifies a list of OpenConfig paths.
// This data set is distinct from
// github.com/openconfig/featureprofiles/proto/feature.proto, which describes
// the relationship between featureprofiles tests and OC paths and is used to
// report test coverage of those paths.
```
nosimage.proto:
```
// NOSImageProfile - A network operating system and version which provides // a list of ocpaths.
```
Use case is to allow NOS implementors and users to specify lists of OC paths for their feature sets/requirements.

One envisioned use case is to have each metadata.textproto contain a set of OCPaths which describes the intended coverage of the test.

Follow up:
* A validation script for OCPaths and NOSImageProfile textprotos.


proto/nosimage.proto
```diff
@@ -0,0 +1,42 @@
...
+import "google/protobuf/timestamp.proto";
+
+message NOSImageProfile {
+  // The name of the vendor.
+  ondatra.Device.Vendor vendor_id = 1;
+
+  // The name of the vendor's network operating system.
+  string nos = 2;
+
+  // Vendor network operating system version.
```
dplore, line 33:
```suggestion
  // Vendor network operating system version. This should match the output of the OC Path  /system/state/software-version.
```

proto/nosimage.proto
```diff
@@ -0,0 +1,42 @@
...
+  string software_version = 3;
+
+  // release_date indicates the date the network operating system is released.
+  // The date could be a value in the future indicating a future release.
+  google.protobuf.Timestamp release_date = 4;
+
+  // list of ocpaths supported by the network operating system
+  repeated openconfig.ocpaths.OCPath ocpaths = 5;
+}
```
dplore, line 42:
```suggestion
  // List of OpenConfig RPC's which are suppored by the network operating system.  ocrpc is a gRPC path specification.
  // Format: <protocol>.<protopath>.<service>.<rpc>
  // Example: gnoi.healthz.Healthz.Get
  repeated string ocrpc = 6;
}
```

proto/nosimage.proto
```diff
@@ -0,0 +1,42 @@
...
+
+import "github.com/openconfig/featureprofiles/proto/ocpaths.proto";
+import "github.com/openconfig/ondatra/proto/testbed.proto";
+import "google/protobuf/timestamp.proto";
+
+message NOSImageProfile {
+  // The name of the vendor.
+  ondatra.Device.Vendor vendor_id = 1;
+
+  // The name of the vendor's network operating system.
```
dplore, line 30:
```suggestion
  // The name of the vendor's network operating system.  This should be the same string as output by /components/component/state/mfg-name from the component TYPE=OPERATING_SYSTEM.

```

dplore (commented):
Couple of suggestions to consider

proto/ocpaths.proto
```diff
@@ -0,0 +1,70 @@
...
+
+  // Specifies a limited portion of the OC model that a given path is needed to
+  // support.
+  OCPathConstraint ocpath_constraint = 2;
+
+  // Suffix for featureprofiles/feature/<featureprofileid> to which this OCPath
+  // is associated.
+  // This destination featureprofiles folder serves as a grouping of the paths
+  // related to the tests which appear in this folder and subfolders the folder.
+  string featureprofileid = 6;
```
dplore, line 52: Should this be a reference to https://github.com/openconfig/featureprofiles/blob/main/proto/feature.proto#L92 ?

wenovus: If your proposal is to change the type from `string` to `FeatureProfileID`, then the additional benefit would be that a version `optional int32 version` is added. I will argue that this is not necessary:

The purpose of the field `featureprofileid` in OCPath is to record the featureprofile to which the path belongs.
When a feature is generated, and when it is different from a previously-generated feature, then that's when a new version is created.
Therefore, when a new OCPath is created, we don't yet know what feature version should be used, since we don't know when the next feature release is, so this version string would not be useful here.

Added comment to clarify the relationship to `FeatureProfileID`.

dplore (approved):

PR #2374, opened 2023-11-13
Add test for 400ZR Q-value performance monitoring

ahsaanyousaf:


feature/platform/transceiver/zr_db_q_value_test/README.md
```diff
@@ -0,0 +1,62 @@
+# TRANSCEIVER-6: Telemetry: 400ZR Optics Q-value streaming.
+
+## Summary
+
+Validate 400ZR optics module reports Q-value performance data as defined in
+module CMIS VDM(Versatile Diagnostics Monitor).
```
ejbrever, line 6: Is the term VDM vendor agnostic?

ejbrever: Thanks for pointing out, for some reason I imagined it being a vendor specific acronym :)

ejbrever (approved):
LGTM from me with Darren's comments addressed.

feature/platform/transceiver/zr_db_q_value_test/README.md
```diff
@@ -0,0 +1,62 @@
...
+    *   /terminal-device/logical-channels/channel/otn/state/q-value/instant
+    *   /terminal-device/logical-channels/channel/otn/state/q-value/avg
+    *   /terminal-device/logical-channels/channel/otn/state/q-value/min
+    *   /terminal-device/logical-channels/channel/otn/state/q-value/max
+
+*   For reported data check for validity min <= avg/instant <= max
+
+*   When the modules or the devices are still in a boot stage, they must not
+    stream any invalid string values like "nil" or "-inf" until valid values
+    are available for streaming.
```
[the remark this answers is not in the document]
ahsaanyousaf: This is not an absolute power value which is typically specified in dBm. The units are actually dB and 0 in this case is the correct value that should be streamed when interface is in down state.

feature/platform/transceiver/zr_db_q_value_test/README.md
ahsaanyousaf: You are right this is defined in the CMIS :)

feature/platform/transceiver/zr_db_q_value_test/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+*   Q-value must always be of type decimal64. When link interfaces are in down
+    state 0.0 must be reported as a valid default value.
+
+
+**Note:** For min, max, and avg values, 10 second sampling is preferred. If 
+          10 seconds is not supported, the sampling interval used must be
+          communicated.
```
dplore, line 41:
```suggestion
          specified by adding a deviation to the test.
```

dplore (dismissed):
LGTM.  Just One minor suggestion/nit if you care to add it.  Deviations refer to a method we use in the test code to programmatically declare some alteration of the test required by a DUT.  These are typically tracked as bugs, but might be long lived issues in cases where it is infeasible to update.

ahsaanyousaf:
> LGTM. Just One minor suggestion/nit if you care to add it. Deviations refer to a method we use in the test code to programmatically declare some alteration of the test required by a DUT. These are typically tracked as bugs, but might be long lived issues in cases where it is infeasible to update.

Done

dplore (approved):

PR #2377, opened 2023-11-14
gNOI-4.1: Adding config push verification post OS upgrade

sgirishj:
Adding support for config push verification post OS upgrade. Addresses https://github.com/openconfig/featureprofiles/issues/2295


sezhang2:
It is new-requirement-pr, so I will merge it

PR #2379, opened 2023-11-14
RT-1.23: BGP AFI SAFI OC DEFAULTS : Added new subtest

cprabha:
Hi,

1.  Separated out v4 and v6 peers for each of the subtest verification.
2. Added new test to test AFI-SAFI set to false at neighbor level.

Thanks,
Prabha


sezhang2 (approved):
The pull refactor the code and added a new test case  to set IPv4-unicast and IPv6-Unicast AFI-SAFIs to FALSE at neighbor level.

A new deviation is also introduced. We should have a bug/issue to track it.
     **skip_bgp_session_check_without_afisafi**

sezhang2:
Please resolve the conflict

PR #2388, opened 2023-11-16
Add a test for ZR inventory information.

ahsaanyousaf:


feature/platform/transceiver/zr_inventory_test/README.md
```diff
@@ -0,0 +1,51 @@
...
+    *   Reset the optic through software.
+    *   Verify the ZR optics still reports correct inventory information.
+    *   Telemetry subscription should be ON_CHANGE and streamed data should
+        be of type String.
+
+*   Verify that the modules inventory information is reported correctly when
+    interface and transceiver states are disabled.
+
+    *   With ZR module plugged in the host and properly recognized
+    *   Use /components/component/transceiver/config/enabled and
```
dplore, line 35: This should power off the transceiver according to OC .  Can you specify what the correct response should be for telemetry?  Should  the transceiver should appear to be removed when disabled (powered off)?   

Do you want to test both /components/component/transceiver/config/enabled	 and /interfaces/interface/config/enabled. ?

feature/platform/transceiver/zr_inventory_test/README.md
```diff
@@ -0,0 +1,51 @@
...
+        /components/component/transceiver/config/enabled to disable the
+        transceiver and inetrface state, wait 20 seconds. 
+    *   Verify the ZR optics inventory information is correctly reported via
+        the streaming telemetry path above in this state.
+    *   Telemetry subscription should be ON_CHANGE and streamed data should
+        be of type String.
+
+## Config Parameter coverage
+
+*   /components/component/oc-transceiver:transceiver/oc-transceiver/config/enabled
```
dplore, line 45:
```suggestion
*   /components/component/transceiver/config/enabled
```

feature/platform/transceiver/zr_inventory_test/README.md
ahsaanyousaf: I have split this into two separate tests.

In case of /interfaces/interface/config/enabled set as false inventory info should be reported

In case of /components/component/transceiver/config/enabled set as false module will be powered off and no inventory info will be reported

feature/platform/transceiver/zr_inventory_test/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+*   Validate the streamed inventory information data is of type String.
+
+*   Verify that the modules inventory information is reported correctly after
+    an optic software reset.
+
+    *   With ZR module plugged in the host and properly recognized 
+    *   Verify the ZR optics inventory is correctly reported via the 
+        streaming telemetry paths above.
+    *   Reset the optic through software.
```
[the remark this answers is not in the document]
ahsaanyousaf: Updated

feature/platform/transceiver/zr_inventory_test/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Procedure
+
+*   Plug in the ZR module in the host port and make sure the transceiver 
+    state is enabled and host is able to detect the module.
+*   With the module recognized verify it reports correct inventory
+    information by subscribing ON_CHANGE to the following telemetry paths.
+
+    *   /platform/components/component/state/serial-no
+    *   /platform/components/component/state/part-no
```
ejbrever, line 16: For inventory there are several other pieces which could be useful (under this same "state" container). I wasn't sure if you planned to track these elsewhere, but specifically I was thinking about:

type
description
mfg-name
mfg-date
hardware-version
firmware-version

Should we add them here?

feature/platform/transceiver/zr_inventory_test/README.md
```diff
@@ -0,0 +1,62 @@
...
+*   Verify the module behaviour when transceiver state is set to disabled.
+
+    *   With ZR module plugged in the host and properly recognized.
+    *   Use /components/component/transceiver/config/enabled to disable the
+        module transceiver state, wait 20 seconds. 
+    *   Verify the ZR module is powered off and no inventory information
+        reported via the streaming telemetry paths above in this state.
+    *   Telemetry subscription should be ON_CHANGE and there should be no
+        streamed inventory data in this state.
+
```
ejbrever, line 53: Can we also please add these test cases?

1. Verify the module behaviour when a transceiver is swapped with a different one.
2. Verify the module behaviour when a transceiver is  physically removed.

We have encountered both of these situations before and generally boil down to the device not correctly streaming deletes, but it would be good to explicitly call them out.

ahsaanyousaf: Done

feature/platform/transceiver/zr_inventory_test/README.md
ahsaanyousaf: I have now included these here in the same test. Thanks

feature/platform/transceiver/zr_inventory_test/README.md
dplore: Sorry to contradict this, but we don't expect to have the hardware/robotics available to automate these kinds of actions.  The README for tests in featureprofiles should only describe tests that are intended to be automated.

ahsaanyousaf: Actually this test case should be already covered when we power off the transceiver, so that is equivalent to physically reseating the module. Ill delete this test here.

dplore (approved):

feature/platform/transceiver/zr_inventory_test/README.md
ejbrever: Ah, gotcha thanks for the clarification.

ejbrever (commented):
LGTM

PR #2389, opened 2023-11-16
RT-8: Singleton with breakouts

sachendras:
New test to cover for scenarios whereby a DUT can stop reporting /state/hardware-port details for an interface after an OS change. Given that a device can report this leaf for an interface irrespective of its admin-status, this test should be able to cover for different breakout configurations that exits on the DUT.


testregistry.textproto
```diff
@@ -422,6 +422,12 @@ test: {
   readme: "https://github.com/openconfig/featureprofiles/blob/main/feature/bgp/policybase/otg_tests/default_policies_test/README.md"
   exec: " "
 }
+test: {
+  id: "RT-8"
+  description: "Singleton with breakouts"
+  readme: ""
```
dplore, line 428:
```suggestion
  readme: "https://github.com/openconfig/featureprofiles/blob/main/feature/interface/singleton/tests/singleton_with_breakouts/README.md"
```

feature/interface/singleton/otg_tests/singleton_with_breakouts/README.md
dplore, line 1: Since there are no OTG calls in this test, please move to:
feature/interface/singleton/tests/singleton_with_breakouts/README.md

dplore (approved):

PR #2393, opened 2023-11-16
gNMI-1.8: add case with large payload

kjahed:
1- Add a new subtest to gNMI-1.18 with a large payload
2- Run tests as  separate subtests


Swetha-haridasula:
@kjahed could you fix the errors. Change LGTM

Swetha-haridasula:
@kjahed kindly do needed readme changes to fix errors

PR #2398, opened 2023-11-17
Example nosimage.textproto and validator script.

wenovus:
**Limitation**: Validator script currently only validates OCPaths, and not RPCs.

Tested generator manually:
```
go run validate/validate.go -file example/example_nosimageprofile.textproto; rm -rf public
```
```
go run validate/validate.go -file example/example_nosimageprofile_invalid.textproto; rm -rf public
```

### DIFF
--- a/tools/internal/ocpaths/ocpaths.go
+++ b/tools/internal/ocpaths/ocpaths.go
@@ -1,0 +1,195 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ocpaths
+
+import (
+	"fmt"
+	"reflect"
+	"regexp"
+	"strings"
+
+	"github.com/openconfig/gnmi/errlist"
+	"github.com/openconfig/goyang/pkg/yang"
+	"github.com/openconfig/goyang/pkg/yangentry"
+	"github.com/openconfig/models-ci/yangutil"
+	"github.com/openconfig/ondatra/gnmi/oc"
+
+	ppb "github.com/openconfig/featureprofiles/proto/ocpaths_go_proto"
+)
+
+const (
+	componentPrefix = "/components/component"
+)
+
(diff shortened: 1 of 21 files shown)


tools/internal/ocpaths/ocpaths.go
```diff
@@ -0,0 +1,195 @@
...
+			for _, v := range oc.ΛEnum[enum] {
+				if v.Name == component {
+					ocpath.Key.Component = component
+					break CONSTRAINT_CHECK
+				}
+			}
+		}
+		fallthrough
+	default:
+		return nil, fmt.Errorf("path %q has invalid component %q", path, component)
```
dplore, line 139:
```suggestion
		return nil, fmt.Errorf("path %q has invalid component %q.  Want %q", path, component, component_regex)
```

dplore: to aid in debugging, include the required regex in the error log

tools/internal/ocpaths/ocpaths.go
```diff
@@ -0,0 +1,195 @@
...
+	if featureprofileID != "" {
+		matched, err := regexp.MatchString("^([a-z0-9]+_)*[a-z0-9]+$", featureprofileID)
+		if err != nil {
+			return nil, err
+		}
+		switch {
+		case matched:
+			ocpath.FeatureprofileID = featureprofileID
+		default:
+			return nil, fmt.Errorf("unexpected featureprofileID string %q for path %v", featureprofileID, path)
```
dplore, line 153: also log the required regex for featureprofileID

tools/internal/ocpaths/ocpaths.go
```diff
@@ -0,0 +1,195 @@
...
+					break CONSTRAINT_CHECK
+				}
+			}
+		}
+		fallthrough
+	default:
+		return nil, fmt.Errorf("path %q has invalid component %q", path, component)
+	}
+
+	// Basic validation of featureprofileid if it exists
```
dplore, line 142: Be even more explicit.  featureprofileID is optional.  Only validate the string format if it exists.

tools/nosimage/validate/validate.go
dplore, line 1: Should we also validate ocrpc?

wenovus: Yes, I can add that as a follow-up.

tools/internal/ocpaths/ocpaths.go
wenovus: Done.

tools/internal/ocpaths/ocpaths.go
wenovus: Done.

tools/internal/ocpaths/ocpaths.go
wenovus: Done.

dplore (approved):

PR #2401, opened 2023-11-17
GNMI-1.11: Adding tolerance for incoming/outgoing packet delta equivalence check

trathod1:
When incoming and outgoing ports are in different linecards, the packet counters for ongoing traffic in gnmi-subscription may have difference in update duration.

Considering that gnmi-subscription update equivalence check may not be valid on multiple linecard dut, and based on discussion in b/315929197 adding tolerance for delta comparison.


NishaSadhasivam-work (approved):
Testcase passed on Nokia device

PR #2403, opened 2023-11-17
RT-5.1: Singleton Interface 

cprabha:
Hi,

I have updated script with otgutils.WaitForARP after configuring otg interfaces.
"otgutils.WaitForARP" is to initiate the ND resolution from IXIA before verifying the neighbor address on DUT.


sezhang2 (approved):
The code added the code to  initiate the ND resolution from IXIA.

	otgutils.WaitForARP(t, tc.ate.OTG(), tc.top, "IPv4")
	otgutils.WaitForARP(t, tc.ate.OTG(), tc.top, "IPv6")

It should not cause new failures

PR #2405, opened 2023-11-17
Update ocrpcs proto to have version be bound to the gRPC protocol instead of service

wenovus:
Reason is gNOI has multiple services, and we have decided for now to bind the version to the repo-level rather than the per-service level.


proto/ocrpcs.proto
```diff
@@ -24,23 +24,20 @@ package openconfig.ocrpcs;
...
+// belong to the same OpenConfig RPC protocol (e.g. gNOI, gNMI, gRIBI, gNSI).
+message OCProtocol {
+  // Full name of the gRPC method specification.
+  // Format: <package>.<service>.<rpc>
+  // Example: gnoi.healthz.Healthz.Get
+  // Example: gnmi.gNMI.Subscribe
   repeated string method_name = 1;
 
-  // The semantic version of the gRPC service.
+  // The semantic version of the gRPC protocol release.
```
dplore, line 41: Should we further clarify that this is this string should be equal to an appropriate git release tag?

wenovus: Done. I added 'when "v" is prepended to the string' since git tags should come in the form "v1.0.0"

wenovus:
@dplore PTAL, I will add the validation for this in a later PR.

wenovus:
> LGTM, but please fix the static analysis checks. They are showing up in all PR's so would be nice to clean them up.

Thanks, fixed.

dplore (approved):

PR #2406, opened 2023-11-18
gNMI-1.19 ConfigPush after Control Card switchover

rszarecki:
This PR provides test plan (REDME.md) for test that verify if DUT is able to accept new configuration immedietly (withing 2 minutes) after performing Controller Card switchover.

Code implementation tracked by https://github.com/openconfig/featureprofiles/issues/2407


feature/platform/controllercard/tests/SetRequest_controll_card_switchover/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Procedure
+
+* Prepare large OpenConfig config file to be pushed within single setRequest.
+  * 150 LAG interfaces w/ ip, ipv6 configuration
+  * 800 Ethernet interfaces as memeber s of LAG
+  * 28 IPv4 and 28 IPv6 BGP neighbours
+  * ISIS on all trunk/LAG ports
+* Initiate Control Card switchover using gNOI SwitchControlProcessorRequest 
+* Immedietly after reciving SwitchControlProcessorResponce for  gNOI switchover, but no later then 120 second after calling gNOI SwitchControlProcessorRequest, send gNMI setRequest with prepared large config.
```
LimeHat, line 14: Is there any wait time allowed after receiving `SwitchControlProcessorResponse` (within the mentioned 120 seconds) before sending gnmi set?

feature/platform/controllercard/tests/SetRequest_controll_card_switchover/README.md
```diff
@@ -0,0 +1,34 @@
...
+* wait 120 second
```
LimeHat, line 15: What's the purpose of this wait timer?

feature/platform/controllercard/tests/SetRequest_controll_card_switchover/README.md
rszarecki: No. There is not allowed wait time after receiving `SwitchControlProcessorResponse`. The 120s timeout is here to cover case when `SwitchControlProcessorResponse` is not received.

dplore: One of the goals of this test is to validate that the DUT sends an appropriate response to a `Set` during time that the switchover is occurring.    @rszarecki perhaps this logic covers the possibility of an error message and incorporates the 2 minute deadline:

If a gRPC error is returned, the client should re-try the gnmi.Set every 10 seconds.
After 120 seconds, if no successful setResponse is received, the test should be marked as a failure.

rszarecki: @dplore PTAL

feature/platform/controllercard/tests/SetRequest_controll_card_switchover/README.md
rszarecki: I have rewritten this part, so perhaps intenton is cleaner now. Basicly, we give 120s time after switchover for controller card to seattle . At lest to point when new ACTIVE can igest new config.

dplore: The new updates look good to me.  @LimeHat if not clear, please unresolve this thread with your question.

feature/platform/controllercard/tests/SetRequest_controll_card_switchover/README.md
dplore: LGTM

dplore (commented):
The logic LGTM, but would like an example of a config difference which is expected?  We should be more specific about the pass criteria for the config validation.  

 Perhaps we should specify a subtree of config which is not expected to change at all.  (Large ACL list, for example)

LimeHat:
Thanks @dplore and @rszarecki , this looks better

1 nit:
https://github.com/openconfig/featureprofiles/blob/ba45a8e970393987fc50fb09eaaba5d7a741d845/feature/platform/controllercard/tests/SetRequest_controll_card_switchover/README.md?plain=1#L17-L20

Perhaps we can replace `ERROR` with something like `a non-zero grpc status code` to avoid confusion with deprecated `Error` message in SetResponse (https://github.com/openconfig/gnmi/blob/9f0d85faac8ec18582822e0cab74a70880621a43/proto/gnmi/gnmi.proto#L362)

and 1 further clarification: 

- it is possible that gNMI/gRPC server might be unavailable for a brief period while switchover is ongoing (generally speaking, `SwitchControlProcessorResponse` message does not indicate that the switchover was completed; it may indicate that switchover was initialized) and the client will receive, for instance, a TCP RST instead
- this behavior is already accounted for in another test case, gNOI-3.3 Supervisor switchover: https://github.com/openconfig/featureprofiles/blob/8503848a24b223608f369f5e1884fd1d37c25ae8/feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go#L134-L149

I think this should be allowed here as well? Obviously not for a full period (120 seconds), but at least for a brief interval?

dplore (approved):

rszarecki:
> 1 nit:
> 
> Perhaps we can replace `ERROR` with something like `a non-zero grpc status code` to avoid confusion with deprecated `Error message` in SetResponse

I've merged this one bit too fast. But will come with patch-PR soon.

> * it is possible that gNMI/gRPC server might be unavailable for a brief period while switchover is ongoing (generally speaking, `SwitchControlProcessorResponse` message does not indicate that the switchover was completed; it may indicate that switchover was initialized) and the client will receive, for instance, a TCP RST (or timeout) instead

I agree that semantics of receiving `SwitchControlProcessorResponse` is bit unclear. 
- The test procedure fails if it is not recived within 120s. Do you think we should wait longer? Or test should not fail but, after timeout, should proceed (hence pass even if `SwitchControlProcessorResponse` was not received)?
The switchover may have multiple stages, at some point new ACTIVE exists and is ready to accept config. But it does not mean BACUP is ready, and may be not ready by next few 10's of minutes. We do not care about BACKUP in this test. I care id ACTIVE can ingest config.
Proposed logic is trying to push confing every 10s for next 120s. So, if no FAIL on `SwitchControlProcessorResponse` missing, we verify if config can be pushed within 4 minuts after switchover initialization.
I do not think ther eis a need to explicitly monitor and  wait for new ControllerCard to become ACTIVE, as we activly re-try config push and check if switchover indeed had place at the end.

LimeHat:
Thanks for the reply @rszarecki 

I don't see an issue with 120 seconds at all. I think it might boil down to the semantics -> if i read this strictly, i'm assuming that
* gnmi.Set is sent out immediately after receiving `SwitchControlProcessorResponse`
* if gnmi.Set timeouts or receives an RST on the first try - it is an immediate fail (because there's no gRPC error and/or gnmi.SetResponse), and the README talks only errors as a subparagraph of received SetResponse 

Re. SwitchControlProcessorResponse:
1. current version of gnoi-3.3 verifies that it is received during a switchover, so it's fine, i don't see a need to relax the requirement
2. I just wanted to highlight  usually implementations will send it out a few moments before initiating a switchover or during a switchover[^1]
3. because of the previous point, what i want to do is to explicitly allow a brief period (~15-30 seconds or so) for the newly-active control module to assume it's role - e.g. with OOB management, it needs to assign the mgmt ip to itself, perform ARP/ND procedures to refresh neighbor caches, etc; and this operation alone might take a few seconds. During that interval we can expect tcp timeouts or RSTs.

Thus if we treat 120 seconds as "any errors are allowed during that interval, either TCP timeout/RST or gnmi error codes" (as it's currently done in gNOI-3.3), that's totally fine with me.
If we want to limit an interval during which TCP errors are allowed to 30 seconds or less, and after that we expect that gRPC servers should to be in a responsive state (accepting set request or returning an error), that's also fine with me. 

Just wanted to be more explicit about these points since it is possible to slightly misinterpret the current text :)

[^1]: i think most of the implementations do not perform hot redundancy for mgmt sessions, including gRPC, so those have to be reestablished after the switchover. this makes sending the switchover response from a newly-active CPM impractical or very challenging, therefore it is send out "in advance"

PR #2397, opened 2023-11-16
Enhance debug capabilities for gNOI-3.1

MarcCharlebois:
1. Added a comparison of the components pre/post reboot which will be output as Logf() in the event of a failure
2. Refactored failure message to reflect the failure more accurately 
3. Removed the uint64 casting of the time (float64) for more accurate time checking

Reference: b/310535585

"This code is a Contribution to the OpenConfig Feature Profiles project ("Work") made under the Google Software Grant and Corporate Contributor License Agreement ("CLA") and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia's intellectual property are granted for any other purpose. This code is provided on an "as is" basis without any warranties of any kind."

PR #2413, opened 2023-11-22
programming with reload

manan-patel:
New gribi programming with reload tc for extending DVT coverage


feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,534 @@
...
+	if gribi_err_msg != nil {
+		t.Fatalf("gribi connection fails post reboot with in max timeout value of %d", maxGribiConnectTime)
+	}
+
+	startComp := time.Now()
+	t.Logf("Wait for all the components on DUT to come up")
+
+	for {
+		postRebootCompStatus := gnmi.GetAll(t, args.dut, gnmi.OC().ComponentAny().OperStatus().State())
+		if len(preRebootCompStatus) == len(postRebootCompStatus) {
```
LimeHat, line 495: This check doesn't validate that "all the components on DUT come up" as the log message on L491 indicates.

FWIW not sure if this is needed here at all - we already have separate tests that cover these scenarios and it doesnt''t seem to be the core part of the test logic here?

LimeHat (changes requested):

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,534 @@
...
+	static := s.GetOrCreateNetworkInstance(deviations.DefaultNetworkInstance(dut)).GetOrCreateProtocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, deviations.StaticProtocolName(dut))
+	ipv4Nh := static.GetOrCreateStatic(innerDstIP1 + "/" + mask).GetOrCreateNextHop("0")
+	ipv4Nh.NextHop, _ = ipv4Nh.To_NetworkInstance_Protocol_Static_NextHop_NextHop_Union(ip)
+	gnmi.Update(t, dut, d.NetworkInstance(deviations.DefaultNetworkInstance(dut)).Protocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, deviations.StaticProtocolName(dut)).Config(), static)
+}
+
+// Reload chassis
+func (args *testArgs) reloadChassis(t *testing.T) {
+
+	preRebootCompStatus := gnmi.GetAll(t, args.dut, gnmi.OC().ComponentAny().OperStatus().State())
```
MarcCharlebois, line 431: Any chance we can include the actual components here instead of simply printing their oper-status and looking if the number of components is the same. I  raised something for gNOI-3.1 in https://github.com/openconfig/featureprofiles/pull/2397

MarcCharlebois (changes requested):

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+	gnoiClient, err := args.dut.RawAPIs().BindingDUT().DialGNOI(context.Background())
+	if err != nil {
+		t.Fatalf("Error dialing gNOI: %v", err)
+	}
+	bootTimeBeforeReboot := gnmi.Get(t, args.dut, gnmi.OC().System().BootTime().State())
+	t.Logf("DUT boot time before reboot: %v", bootTimeBeforeReboot)
+
+	rebootRequest := &spb.RebootRequest{
+		Method:  spb.RebootMethod_WARM,
```
LimeHat, line 442: Reboot method should be COLD i think?

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
MarcCharlebois: Could we ltake into consideration  https://github.com/openconfig/featureprofiles/pull/2397 raised for gNOI-3.1 which aims to improve the debugability of this code-block.

manan-patel: @LimeHat, I added components check as a safety net for validation, however it's not need under this scenario. I'll remove the check.

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
manan-patel: Yes, it should be COLD

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,512 @@
...
+	rebootRequest := &spb.RebootRequest{
+		Method:  spb.RebootMethod_COLD,
+		Delay:   0,
+		Message: "Reboot chassis with delay",
+		Force:   true,
+	}
+
+	t.Logf("Send reboot request: %v", rebootRequest)
+	rebootResponse, err := gnoiClient.System().Reboot(context.Background(), rebootRequest)
+	defer gnoiClient.System().CancelReboot(context.Background(), &spb.CancelRebootRequest{})
```
LimeHat, line 445: No need to cancel (this reboot is expected to succeed)

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,512 @@
...
+	if err != nil {
+		t.Fatalf("Error dialing gNOI: %v", err)
+	}
+	bootTimeBeforeReboot := gnmi.Get(t, args.dut, gnmi.OC().System().BootTime().State())
+	t.Logf("DUT boot time before reboot: %v", bootTimeBeforeReboot)
+
+	rebootRequest := &spb.RebootRequest{
+		Method:  spb.RebootMethod_COLD,
+		Delay:   0,
+		Message: "Reboot chassis with delay",
```
LimeHat, line 439: nit: there's no delay

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,512 @@
...
+		t.Logf("Time elapsed %.2f seconds since reboot started.", time.Since(startReboot).Seconds())
+		time.Sleep(30 * time.Second)
+		if errMsg := testt.CaptureFatal(t, func(t testing.TB) {
+			currentTime = gnmi.Get(t, args.dut, gnmi.OC().System().CurrentDatetime().State())
+		}); errMsg != nil {
+			t.Logf("Got testt.CaptureFatal errMsg: %s, keep polling ...", *errMsg)
+		} else {
+			t.Logf("Device rebooted successfully with received time: %v", currentTime)
+			break
+		}
```
LimeHat, line 464: Please add an upper bound to the reboot duration (e.g. in a complete_chassis_reboot test, 15 minute timer is used)

LimeHat (changes requested):

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,512 @@
...
+	t.Logf("Device boot time: %.2f seconds", time.Since(startReboot).Seconds())
+
+	bootTimeAfterReboot := gnmi.Get(t, args.dut, gnmi.OC().System().BootTime().State())
+	t.Logf("DUT boot time after reboot: %v", bootTimeAfterReboot)
+	if bootTimeAfterReboot <= bootTimeBeforeReboot {
+		t.Errorf("Get boot time: got %v, want > %v", bootTimeAfterReboot, bootTimeBeforeReboot)
+	}
+
+	// starting a goroutine to attempt for gribi connection while system finishes bringup process
+	t.Logf("Starting a goroutine to attempt for gribi connection within max timeout value")
```
LimeHat, line 475: What's the purpose of goroutine here?

feature/experimental/gribi/otg_tests/programming_with_reload/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+*   TEST#1 
+
+    1.  Deploy Setup#1 as above.
+
+    2.  Send IPinIP traffic to `OuterDstIP_1` with inner IP as `InnerDstIP_1` and validate that ATE port-2 receives the traffic after DECAP-ENCAP over the primary path with outer destination IP as `OuterDstIP_2` and outer source IP as `OuterSrcIP_2`.
+    
+    3.  Perform entire device reload.
+
+    4.  Redeploy Setup#1 as above. If the system is not ready, gRIBI returns UNAVAILABLE till it is ready, and indeed reattempts will be made every 30 seconds with max timeout of 3 minutes.
```
LimeHat, line 76: Perhaps a question to Google (cc @xw-g): I think we should clarify, at what point 3 minute interval starts?

In the current version of the code it begins after system is able to answer gnmi.Get requests, which may or may not be correct depending on the intention of this test.

xw-g: and specify the start of T0, e.g. T0 = received the first `UNAVAILABLE` message.

LimeHat (changes requested):

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,537 @@
...
+			if latestTime.Before(prevTime) || latestTime.Equal(prevTime) {
+				t.Errorf("Get latest system time: got %v, want newer time than %v", latestTime, prevTime)
+			}
+			prevTime = latestTime
+		}
+	}
+
+	startReboot := time.Now()
+	t.Logf("Wait for DUT to boot up by polling the telemetry output.")
+	for {
```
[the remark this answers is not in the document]
LimeHat: gnmi.Get is used as a trigger to declare that the node is done with the boot process (rebooted successfully) in gNOI reboot tests
https://github.com/openconfig/featureprofiles/blob/332b7d49a98712cadda876a0096ed3a8cbe109be/feature/gnoi/system/tests/complete_chassis_reboot/complete_chassis_reboot_test.go#L155-L170, and seemingly was just taken from there 

But here, we should try to call gRIBI servers immediately to validate UNAVAILABLE grpc status (otherwise, chances are that the test may miss a period where gRIBI is unable to process the request but does not return the status as expected).

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,537 @@
...
+	}
+
+	t.Logf("Send reboot request: %v", rebootRequest)
+	rebootResponse, err := gnoiClient.System().Reboot(context.Background(), rebootRequest)
+	t.Logf("Got reboot response: %v, err: %v", rebootResponse, err)
+	if err != nil {
+		t.Fatalf("Failed to reboot chassis with unexpected err: %v", err)
+	}
+
+	if rebootRequest.GetDelay() > 1 {
```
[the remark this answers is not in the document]
manan-patel: Ack, I'll remove it

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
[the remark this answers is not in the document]
manan-patel: You're right, gNMI.GET call confirms device rebooted successfully. Inorder to address the comment, I am changing the implementation to following,

- reload router
- start gribi client connection in 1 min (1 minute timer is to make sure device begins the reload), which will keep on failing with "context deadline exceeded" while dut is reloading
- when connections are up, we'll see return code unavailable, max allowed timeout is 3 mins
- establish gribi session and do programming

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,513 @@
...
+			return fmt.Errorf("Device took longer than %d to reboot", maxRebootTime)
+		} else {
+			// Attempt starting gribi client session and check the error code,
+			// if its context deadline exceeded DUT is still rebooting and try again
+			if err := client.Start(t); err != nil {
+				t.Logf("gRIBI Connection could not be established because device is still rebooting: %v\nRetrying...", err)
+				if errors.Is(err, context.DeadlineExceeded) {
+					time.Sleep(1 * time.Second)
+				} else {
+					// if error is UNAVAIABLE, try creating gribi session for maxGribiConnectTime
```
LimeHat, line 493: You may want to check the specific error code here (UNAVAILABLE) vs just any other error?
something like

if (context.DeadlineExceeded)
else if (UNAVAILABLE)
else (received an unexpected gRPC error)

manan-patel: Ack

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,512 @@
...
+		// we can track 3 mins following the start time
+		gribi_connection_timer := time.Now()
+
+		// if error code is context deadline exceeded DUT is still rebooting and try again
+		if errors.Is(err, context.DeadlineExceeded) && time.Since(reboot_timer).Seconds() < maxRebootTime {
+			t.Logf("gRIBI Connection could not be established because device is still rebooting: %v\nRetrying...", err)
+			time.Sleep(1 * time.Second)
+		} else if strings.Contains(strings.ToLower(err.Error()), "unavailable") && time.Since(reboot_timer).Seconds() < maxRebootTime {
+			// if error is UNAVAIABLE, try creating gribi session within maxGribiConnectTime
+			for {
```
LimeHat, line 494: Second `for` loop is redundant, you can flatten this into the main loop

manan-patel: Ack

feature/experimental/gribi/otg_tests/programming_with_reload/metadata.textproto
```diff
@@ -0,0 +1,16 @@
...
+testbed:  TESTBED_DUT_ATE_4LINKS
+platform_exceptions: {
+  platform: {
+    vendor: ARISTA
+  }
+  deviations: {
+    static_protocol_name: "STATIC"
+    default_network_instance: "default"
+  }
+}
```
bkreddy143, line 16: Please add below deviation for NOKIA 
'''
platform_exceptions: {
   platform: {
    vendor: NOKIA
  }
  deviations: {
    explicit_port_speed: true
    explicit_interface_in_default_vrf: true
    static_protocol_name: "static"
    interface_enabled: true
    }
  }
'''

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,511 @@
...
+		ni := c.GetOrCreateNetworkInstance(vrf)
+		ni.Type = oc.NetworkInstanceTypes_NETWORK_INSTANCE_TYPE_L3VRF
+		if vrf == vrfA {
+			p1 := dut.Port(t, "port1")
+			niIntf := ni.GetOrCreateInterface(p1.Name())
+			niIntf.Subinterface = ygot.Uint32(0)
+			niIntf.Interface = ygot.String(p1.Name())
+		}
+		gnmi.Replace(t, dut, gnmi.OC().NetworkInstance(vrf).Config(), ni)
+	}
```
bkreddy143, line 184: Please add below code to add other sub interfaces to  default vrf 
'''
	fptest.ConfigureDefaultNetworkInstance(t, dut)
	if deviations.ExplicitInterfaceInDefaultVRF(dut) {
		fptest.AssignToNetworkInstance(t, dut, dut.Port(t, "port2").Name(), deviations.DefaultNetworkInstance(dut), 0)
		fptest.AssignToNetworkInstance(t, dut, dut.Port(t, "port3").Name(), deviations.DefaultNetworkInstance(dut), 0)
'''

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
```diff
@@ -0,0 +1,511 @@
...
+
+	// adding sleep time for reload to begin
+	time.Sleep(1 * time.Minute)
+
+	// establish new gribi programming following reload
+	err_msg := args.gribi_reconnect(t)
+	if err_msg != nil {
+		t.Fatalf("Failure: %d", err_msg)
+	}
```
bkreddy143, line 470: can we also add check to make sure that gnmi server is up and reconnected . 

Also , need to add check for all interface that are part of up test are up .  we can loop gnmi.await() some thing like 
''''
	for p, _ := range dutPorts {
		p1 := args.dut.Port(t, p)
		gnmi.Await(t, args.dut, gnmi.OC().Interface(p1.Name()).OperStatus().State(), 2*time.Minute, oc.Interface_OperStatus_UP)
	}
'''

feature/experimental/gribi/otg_tests/programming_with_reload/metadata.textproto
manan-patel: ack

feature/experimental/gribi/otg_tests/programming_with_reload/programming_with_reload_test.go
manan-patel: We are not going to check interface or hw status following reload as it is not part of this tc as indicated by google in previous comments

bkreddy143: @manan-patel  ,  Not sure which comment you are talking about it . But we need to make sure that gNMI connection exceptions are handled after reload .  For example L280 gnmi set will fail if gnMI server was not up . Alternatively we can add gNMI connection exception handling in L280 with retry option. 

If we don't check and wait for interfaces up then traffic  check in L288 can lead to false failure. 
@xw-g  could you please help with your inputs.

manan-patel: I removed the post checks as per comment on Nov 22, 2023 by @LimeHat, as per suggested I can uplift the tc accordingly

LimeHat: Hi @manan-patel @bkreddy143 

AFAIR there was never a check on the interface status.
- There was a check that verifies the length of the component list with any oper-status (that wasn't particularly useful here)
- Also, there was a check that used to verify gnmi status before starting gRIBI reconnect attempts 

Indeed I commented on both points.

@bkreddy143 request is slightly different:

- to make sure that interfaces are up before you start traffic validation (which makes sense. alternatively, you need to remove the traffic validation part). This can be done after gribi reconnects, and only for interfaces in binding 
- also, if there are any gnmi requests in the subsequent steps (following the gribi reconnect), we'll need to wait for gnmi server to be up accordingly

I.e. my comments were mainly about the fact that we should start gribi reconnection attempts pretty much immediately post reboot (they are not blocked by gnmi or anything else), but @bkreddy143 correctly notes among the same lines that all three parts of this equation (gribi, gnmi, ports oper status) are independent (in a general case), and you need to monitor status for each before you attempt to execute it.

AmrNJ (changes requested):
The test fails.
gRIBI Connection could not be established following reboot

karthikeya-remilla:
Obsolete PR.

PR #2425, opened 2023-11-29
Adding default accept term for as-path regex policy

s-nandu:
Adding default accept policy term for as-path regex match reject policy


sezhang2 (approved):
The change is specific to Juniper, so we can merge it with no impact on other vendors.

PR #2428, opened 2023-11-30
Fix nosimage.textproto rpcs examples

wenovus:
Forgot to update the example in the last PR.


wenovus:
@dplore PTAL

tools/nosimage/example/generate_example.go
```diff
@@ -95,31 +95,56 @@ func generateExample(filepath string, invalid bool) error {
...
+								"gnoi.healthz.Healthz.Get",
+								"gnoi.healthz.Healthz.List",
+								"gnoi.healthz.Healthz.Acknowledge",
+								"gnoi.healthz.Healthz.Artifact",
+								"gnoi.healthz.Healthz.Check",
+								"gnoi.bgp.BGP.ClearBGPNeighbor",
+							},
+						},
+					}
+				} else {
```
dplore, line 120: Please fix static check, "if block ends with a return statement, so drop this else and outdent its block"

wenovus: Done.

dplore (approved):

Issue #2338, opened 2023-11-02
RT-5.1.4 Removal of Interface config should return an error if breakout-mode is still configured


rohit-rp:
# Request for validating that removal of Interface config should return an error if breakout-mode is still configured

FNT to verify that removal of ```/interfaces/interface/ethernet/config``` should return an error if ```/components/component/port/breakout-mode``` subtree is still populated with configuration for the referenced interface.

Clients are required to remove ```/components/component/port/breakout-mode``` and ```/interfaces/interface/ethernet/config``` for the referenced interface to completely remove the configuration (else interface along with it's speed config remains)

Implement code for [RT-5.1.4](https://github.com/openconfig/featureprofiles/blob/main/feature/interface/singleton/otg_tests/singleton_test/README.md)

PR #2448, opened 2023-12-05
Interface breakout configuration and removal test

rohit-rp:
# Request for validating:
## Removal of Interface config should return an error if breakout-mode is still configured

https://github.com/openconfig/featureprofiles/issues/2338

### DIFF
--- a/feature/interface/singleton/otg_tests/singleton_test/README.md
+++ b/feature/interface/singleton/otg_tests/singleton_test/README.md
@@ -73,6 +73,37 @@ a new testbed configuration with the desired port types.
 *   Ensure inbound and outbound unicast counters are the same
 *   Ensure counters increment at the selected SAMPLE interval
 
+### RT-5.1.4 [TODO: https://github.com/openconfig/featureprofiles/issues/2338]
+#### Breakout must be explicitly configured by gNMI client
+
+*   On DUT Port-1 with a QSFP-DD 400GBASE-DR4 transceiver inserted
+*   Ensure no breakout is configured
+*   Set Port-1 port-speed to 100G
+    *   /interfaces/interface/ethernet/config/port-speed
+*   Validate that the DUT does not create breakouts implicitly and does not set the breakout speed
+    *   /components/component/port/breakout-mode/groups/group/config
+    *   /components/component/port/breakout-mode/groups/group/config/index
+    *   /components/component/port/breakout-mode/groups/group/config/breakout-speed
+*   Validate the port state changes to "DOWN"
+    *   /interfaces/interface/state/oper-status
+
+### RT-5.1.5 [TODO: https://github.com/openconfig/featureprofiles/issues/2338]
+#### Setting port-speed on interface that have breakout configured should not be allowed
+
+*   Configure a breakout on Port-1 to 4x100 Gig
+    *   /components/component/port/breakout-mode/groups/group/config
+*   Try to set port speed of Port-1 to 100G
+*   Validate the port-speed is rejected
+    *   /interfaces/interface/ethernet/state/port-speed
+
+### RT-5.1.6 [TODO: https://github.com/openconfig/featureprofiles/issues/2338]
+#### Remove breakout and interface config to delete the interface config
+
+*   Remove the interface and breakout config
+*   Ensure the interface configuration is completely removed
+    *   /interfaces/interface/ethernet/config/
+    *   /components/component/port/breakout-mode/groups/group/config
+
 ## Config Parameter Coverage
 
 * /interfaces/interface/config/name
--- a/feature/interface/singleton/otg_tests/singleton_test/README.md
+++ b/feature/interface/singleton/otg_tests/singleton_test/README.md
@@ -93,16 +93,18 @@ a new testbed configuration with the desired port types.
 *   Configure a breakout on Port-1 to 4x100 Gig
     *   /components/component/port/breakout-mode/groups/group/config
 *   Try to set port speed of Port-1 to 100G
+    *   /interfaces/interface/ethernet/config/port-speed
 *   Validate the port-speed is rejected
+    *   Since a breakout port is not expected to support port-speed, verify the gNMI Set operation is rejected
     *   /interfaces/interface/ethernet/state/port-speed
 
 ### RT-5.1.6 [TODO: https://github.com/openconfig/featureprofiles/issues/2338]
 #### Remove breakout and interface config to delete the interface config
 
-*   Remove the interface and breakout config
-*   Ensure the interface configuration is completely removed
-    *   /interfaces/interface/ethernet/config/
-    *   /components/component/port/breakout-mode/groups/group/config
+*   Using a single gNMI Replace, remove the DUT port-1 and its breakout config
+*   Ensure the gNMI Replace is successful and configuration for DUT port-1 including its breakout is removed
+    *   /interfaces/interface/ethernet/state/
+    *   /components/component/port/breakout-mode/groups/group/state
 
 ## Config Parameter Coverage
 


feature/interface/singleton/otg_tests/singleton_test/README.md
```diff
@@ -73,6 +73,21 @@ a new testbed configuration with the desired port types.
 *   Ensure inbound and outbound unicast counters are the same
 *   Ensure counters increment at the selected SAMPLE interval
 
+### RT-5.1.4 [TODO: https://github.com/openconfig/featureprofiles/issues/2338]
+#### Removal of Interface config should return an error if breakout-mode is still configured
+
+*   On DUT Port-1 with a QSFP-DD 400GBASE-DR4 transceiver inserted
+*   Ensure no breakout is configured
+*   Set Port-1 port-speed to 100GB
+*   The DUT should reject the configuration
```
LimeHat, line 82: Is this really required?

This requirement implies that explicitly configured port speeds should be validated against the currently equipped transceiver (state) in the mgmt layer.

This coupling of the intended config and current operational state:
1) adds complexity (each commit has to be validated against the current state),
2) limits the ability of an operator to change the intended state (what if this port is supposed to become 100G, and 400G-DR4 optics will be replaced)
3) introduces inconsistencies and corner cases such as
    * what should happen if a port is configured as 100G first, and after that 400G optics is inserted? 

cc @xw-g @dplore @robshakir, thoughts?

rohit-rp: @LimeHat The problem we are trying to address is that breakout should not be implicitly configured by the DUT If gNMI client sets ```/interfaces/interface/ethernet/config/port-speed``` equal to the breakout speed (e.g. 100G) without actually configuring a breakout using ```/components/component/port/breakout-mode```. A vendor does this right now i.e. if gNMI client sets port-speed of 100G on multilane 400G port, it creates breakouts automatically.
We do not want the breakouts configured automatically. Does Nokia's implementation differ and does Nokia require gNMI client to explicitly set interface and breakout config (if a breakout is needed)?
We do not ask for adding complexity, the operation can error out just like if I try to set port-speed on a 10G interface to 100G, the operation will be illegal and will error-out (couldn't try it yet, but I believe the config will not be accepted)

LimeHat: > The problem we are trying to address is that breakout should not be implicitly configured by the DUT If gNMI client sets /interfaces/interface/ethernet/config/port-speed equal to the breakout speed (e.g. 100G) without actually configuring a breakout

That's fair. However, I think that the proposed solution should not be universally required.

>Does Nokia's implementation differ and does Nokia require gNMI client to explicitly set interface and breakout config (if a breakout is needed)?

Our approach is to always require an explicit breakout configuration (we never do it implicitly). Also, we'll reject a breakout config if a port-speed is configured on the interface.

>We do not ask for adding complexity, the operation can error out just like if I try to set port-speed on a 10G interface to 100G, the operation will be illegal and will error-out (couldn't try it yet, but I believe the config will not be accepted)

It may be a simple error condition in some implementations, but I think there are some fundamental aspects to it.   
We don't do enforcement based on the equipped transceiver types for the reasons I listed above, and # 3 is crucial - it means that the same config will be transitioning between valid and invalid states simply based on a transceiver presence (Schrödinger's config?).   

There are certain rules about what port-speeds are accepted, and we can discuss in more detail in buganizer if needed, but our general approach is to let an operator figure out his intent. 
* E.g. If an operator configures 100G speed on a port equipped with 400G transceiver, maybe he wants this port to be up only when a NOC person changes the transceiver in the node, and as soon as he does that, but not earlier. We consider this to be a valid approach. 

Considering all of the above, I can suggest the following:
1) Change the requirement to something like "DUT should not create breakout interfaces implicitly if this configuration is accepted, or, alternatively, the configuration can be rejected"
2) You can also add a requirement that setting port-speed on the interface configured as a breakout should be forbidden and rejected

rohit-rp: Had a conversation with @dplore and we agreed that
- DUT should not create breakout implicitly if, e.g. on a 400Gig interface, port-speed is configured to be 100G
  - ```/interfaces/interface/ethernet/config/port-speed``` can be configured to 100G by gNMI client but ```/interfaces/interface/ethernet/state/port-speed``` reports the physical port-speed (400G)
  - We have a use case where we need to configure interfaces that do not have SFP's inserted, enforcing validations can cause this use case to break, hence the DUT can accept the config but it may not apply
- Setting port-speed should be rejected on breakout interfaces since these are mutually exclusive configurations

feature/interface/singleton/otg_tests/singleton_test/README.md
```diff
@@ -73,6 +73,37 @@ a new testbed configuration with the desired port types.
...
+
+*   On DUT Port-1 with a QSFP-DD 400GBASE-DR4 transceiver inserted
+*   Ensure no breakout is configured
+*   Set Port-1 port-speed to 100G
+    *   /interfaces/interface/ethernet/config/port-speed
+*   Validate that the DUT does not create breakouts implicitly and does not set the breakout speed
+    *   /components/component/port/breakout-mode/groups/group/config
+    *   /components/component/port/breakout-mode/groups/group/config/index
+    *   /components/component/port/breakout-mode/groups/group/config/breakout-speed
+*   Validate the port-speed of the interface is not changed to 100G
```
LimeHat, line 87: @rohit-rp @dplore thanks for the changes.

I still have concerns about this part.
If a mismatched speed is configured and accepted by a device, the device can have two options:
1) ignore the mismatching configuration, and set the speed based on the equipped transceiver, or
2) take into the account accepted config, and bring down the interface

I don't believe opt#1 should be the only option. If an operator configures an explicit statement that an interface should use 100G, he may mean that this interface _should_ have oper-status UP only with 100G speed transceiver.
But opt#1 implies that port-speed statement doesn't mean anything and just ignored - and all operations are based on dynamic port speed detection of the transceiver.

feature/interface/singleton/otg_tests/singleton_test/README.md
```diff
@@ -73,6 +73,37 @@ a new testbed configuration with the desired port types.
...
+    *   /interfaces/interface/ethernet/state/port-speed
+
+### RT-5.1.5 [TODO: https://github.com/openconfig/featureprofiles/issues/2338]
+#### Setting port-speed on interface that have breakout configured should not be allowed
+
+*   Configure a breakout on Port-1 to 4x100 Gig
+    *   /components/component/port/breakout-mode/groups/group/config
+*   Try to set port speed of Port-1 to 100G
+*   Validate the port-speed is not applied
```
LimeHat, line 96: Not applied or rejected (as indicated in @rohit-rp's comment above)?

I believe reject should be a valid option in this case

feature/interface/singleton/otg_tests/singleton_test/README.md
dplore, line 87: @LimeHat @rohit-rp what do you think about this suggestion? 

```suggestion
*   Validate 
    * the configuration is accepted and `/interfaces/interface/ethernet/config/port-speed` is set to 100G.  
    * `/interfaces/interface/ethernet/state/port-speed` remains at 400G, indicating the transceiver cannot support the configured value.
    * `/interfaces/interface/state/oper-status` remains set to `UP`
 ```

The other option would be to take the interface down because of the conflict between `config` and `state` values.  But I think in this scenario, where the TRANSCEIVER only has one valid `state/port-speed` value (400G), then any conflicting value in `config/port-speed` can be ignored by the NOS and the port can stay up.

What do you think?

File: feature/interface/singleton/otg_tests/singleton_test/README.md (hunk 3 above)
Comment (dplore) [reply feature/interface/singleton/otg_tests/singleton_test/README.md:96]: Agreed, the NOS should reject the configuration because a breakout is indicated, invalidating the use of `/interfaces/interface/ethernet/config/port-speed`.  Use of `/interfaces/interface/ethernet/config/port-speed` and `/components/component/port/breakout-mode/groups/group/config` should be mutually exclusive.

File: feature/interface/singleton/otg_tests/singleton_test/README.md (hunk 2 above)
Comment (rohit-rp) [reply feature/interface/singleton/otg_tests/singleton_test/README.md:87]: I am good with the suggested change, but will wait for @LimeHat to respond

File: feature/interface/singleton/otg_tests/singleton_test/README.md (hunk 3 above)
Comment (rohit-rp) [reply feature/interface/singleton/otg_tests/singleton_test/README.md:96]: Ack, will change "no applied" to "rejected"

File: feature/interface/singleton/otg_tests/singleton_test/README.md (hunk 2 above)
Comment (LimeHat) [reply feature/interface/singleton/otg_tests/singleton_test/README.md:87]: @dplore we follow the other option you mentioned - set the interface oper-status as `DOWN`.
I.e. we always prefer an explicit intent of the operator vs a derived speed value from a transceiver, and there are certain use cases for that. 

An operator may want an interface that can be oper `UP` only with a certain speed, and if speed is mismatched - he would prefer to exclude this link from the network topology (let's take as an example another 400/100G scenario: the operator wants that all links in his network are 400G, but a DC technician connects a new IXP link with 100G transceivers. This link will be up with 1/4th of the expected capacity)

If we say that an explicit speed config can be ignored, what role does it have in the first place? This leaf will be a no-op in a majority of cases (the only exception that comes to my mind is a multi-rate transceiver, but this is rare) 

Maybe we can allow both options, and implementers can choose what to do (prefer the implicit speed or use the oper down?

Comment (dplore) [reply feature/interface/singleton/otg_tests/singleton_test/README.md:87]: Let's stick with option #2 only.  We had some discussion of this type of issue in the past where an interface config is mismatched with the installed transceiver.  Operationally we don't want to be blocked deploying a configuration due to a single port mismatch (or even many ports).    Therefore I think we always want option 2 and  never option 1.

File: feature/interface/singleton/otg_tests/singleton_test/README.md (hunk 3 above)
Comment (LimeHat) [reply feature/interface/singleton/otg_tests/singleton_test/README.md:96]: Looks like this hasn't been updated?

File: feature/interface/singleton/otg_tests/singleton_test/README.md [hunk 4]
```diff
@@ -73,6 +73,37 @@ a new testbed configuration with the desired port types.
...
+*   Validate the port state changes to "DOWN"
+    *   /interfaces/interface/state/oper-status
+
+### RT-5.1.5 [TODO: https://github.com/openconfig/featureprofiles/issues/2338]
+#### Setting port-speed on interface that have breakout configured should not be allowed
+
+*   Configure a breakout on Port-1 to 4x100 Gig
+    *   /components/component/port/breakout-mode/groups/group/config
+*   Try to set port speed of Port-1 to 100G
+*   Validate the port-speed is not applied
```
LimeHat, line 96: Looks like this hasn't been updated (to "rejected")?

LimeHat:
Thanks @rohit-rp and @dplore , looks good to me!

feature/interface/singleton/otg_tests/singleton_test/README.md
```diff
@@ -73,6 +73,37 @@ a new testbed configuration with the desired port types.
...
+*   Validate the port state changes to "DOWN"
+    *   /interfaces/interface/state/oper-status
+
+### RT-5.1.5 [TODO: https://github.com/openconfig/featureprofiles/issues/2338]
+#### Setting port-speed on interface that have breakout configured should not be allowed
+
+*   Configure a breakout on Port-1 to 4x100 Gig
+    *   /components/component/port/breakout-mode/groups/group/config
+*   Try to set port speed of Port-1 to 100G
+*   Validate the port-speed is rejected
```
dplore, line 96: Might be good to add some explanation for this:   "Since a breakout port is not expected to support `port-speed`, verify the gnmi Set operation is rejected.  "

feature/interface/singleton/otg_tests/singleton_test/README.md
```diff
@@ -73,6 +73,37 @@ a new testbed configuration with the desired port types.
...
+    *   /interfaces/interface/ethernet/state/port-speed
+
+### RT-5.1.6 [TODO: https://github.com/openconfig/featureprofiles/issues/2338]
+#### Remove breakout and interface config to delete the interface config
+
+*   Remove the interface and breakout config
```
dplore, line 102: specify using a single gnmi Replace for this operation

feature/interface/singleton/otg_tests/singleton_test/README.md
rohit-rp: updated

feature/interface/singleton/otg_tests/singleton_test/README.md
rohit-rp: updated

dplore (approved):

PR #2459, opened 2023-12-07
Add delay for BGP update to propagate gNMI-1.3

MarcCharlebois:
1. Added 10 second timer for med update to propagate through. 
 

feature/experimental/system/gnmi/benchmarking/ate_tests/drained_configuration_convergence_time/drained_configuration_convergence_time_bgp_test.go
```diff
@@ -136,6 +136,8 @@ func setPolicyPeerGroup(t *testing.T, dut *ondatra.DUTDevice, d *oc.Root, policy
 // isConverged function is used to check if ATE has received all the prefixes.
 func isConverged(t *testing.T, dut *ondatra.DUTDevice, ate *ondatra.ATEDevice, ap *ondatra.Port) {
 
+	// Add 10 second timer for BGP update to propagate
```
xw-g, line 139: Can you help elaborate why current logic (without this delay) is not good enough? thanks.

LimeHat: Current version of isConverged() assumes that an implementation does withdrawal and re-advertisement of the routes after the new policy (with modified MED) is applied - it monitors the number of received prefixes on the OTG ports

In some implementations, including ours, there's no withdrawal step - we just send updates with the new attribute values. Hence this function will always think that the convergence is done, even when it's not

NishaSadhasivam-work (approved):
PFA pass log on Nokia device
[test (4).log](https://github.com/openconfig/featureprofiles/files/13684521/test.4.log)

PR #2462, opened 2023-12-08
gNOI-4.1 Adding Nokia deviation

MarcCharlebois:
Added devations to accomodate for this test passing.


feature/gnoi/os/tests/osinstall/osinstall_test.go
```diff
@@ -425,6 +425,10 @@ func watchStatus(t *testing.T, ic ospb.OS_InstallClient, standby bool) error {
 func TestPushAndVerifyInterfaceConfig(t *testing.T) {
 
 	dut := ondatra.DUT(t, "dut")
+	ocPortName := dut.Port(t, "port1").Name()
```
xw-g, line 428: `ocPortName` usage is gated by `deviations.ExplicitInterfaceInDefaultVRF(dut)`, therefore it should be included in the `if` statement.

xw-g (commented):
just a minor suggestion and we are good to go

feature/gnoi/os/tests/osinstall/osinstall_test.go
MarcCharlebois: You're absolutely correct; this is fixed in latest commit.

NishaSadhasivam-work (changes requested):
With the changes test failed on Nokia vendor with the below message.

 TestOSInstall Failed with “Missing osfile or osver args” message; 
when run on image “os_version -> 0.0.0-52412”

MarcCharlebois:
> With the changes test failed on Nokia vendor with the below message.
> 
> TestOSInstall Failed with “Missing osfile or osver args” message; when run on image “os_version -> 0.0.0-52412”

The error you've listed above suggests that you've not [defined the flags](https://github.com/openconfig/featureprofiles/blob/f59b129b977ba5440be09f5e02848d1da0e69441/feature/gnoi/os/tests/osinstall/osinstall_test.go#L53-L54) required for this test to operate successfully. However, this is not due to this PR, this would be true even in main branch. Reassigning to @NishaSadhasivam-work

NishaSadhasivam-work (approved):

PR #2463, opened 2023-12-08
Update ipv6 link local test -  Add traffic to subtest 3

dplore:
Add traffic verification when both a global and link local address are configured.


feature/interface/ip/ipv6_link_local/otg_tests/ipv6_link_local_test/README.md
```diff
@@ -18,6 +18,7 @@ Configure an IPv6 address which is in link local scope. Verify the link local IP
 * Subtest #3 - Verify adding and removing global unicast address does not affect link local address
   * Add configuration for a global unicast address on DUT port 1 
   * Validate config and state paths are set
+  * Send IPv6 traffic from OTG port 1 to DUT port 1, validate DUT port 1 receives the traffic
```
liulk, line 21: I think we want to fully configure end to end global IPv6 so we can send traffic from OTG port 1 to be forwarded through DUT to OTG port 2?

Ankur19 (approved):

PR #2476, opened 2023-12-11
build dynamically linked  binary

alshabib:


internal/cntrsrv/build/cntr_build.sh
```diff
@@ -17,7 +17,7 @@
 # exit when a command fails
 set -e
 
-CGO_ENABLED=0 go build ..
```
robshakir, line 20: Maybe add this as an option for the container build script? e.g., allow this to be specified as a flag -- where the default is dynamically linked if we think that is expected?

robshakir (approved):

internal/cntrsrv/build/cntr_build.sh
earies: This will actually not be enough to account for situations where you want to intercept glibc calls (which is necessary for situations where the kernel may not hold all state)

For this, we actually will need to explore compilation with `gccgo` but it may expose per application/library considerations (and a more complex build/Makefile)

PR #2479, opened 2023-12-12
Add tags to paths in ocpaths.proto.

robshakir:
This change allows paths to be tagged with arbitrary criteria that allow for grouping or other identification of a path within a list.


proto/ocpaths.proto
```diff
@@ -52,6 +54,10 @@ message OCPath {
   // This destination featureprofiles folder serves as a grouping of the paths
   // related to the tests which appear in this folder and subfolders the folder.
   string featureprofileid = 3;
+
+  // A set of opaque tags that are used for this path. These tags can be used
+  // to group paths according to use-case specific criteria.
+  repeated string paths = 4;
```
wenovus, line 60: paths -> tags?

robshakir: Good catch, fixed. Thanks.

robshakir:
We should skip the `proto` directory for static analysis checks in this repo, I think.

wenovus (approved):
LGTM, @dplore you want to take a final look?

dplore (approved):

PR #2480, opened 2023-12-12
Fix RebootStatus subcomponents paths in per_component_reboot_test

lifuzhang1108:
The RebootStatus gNOI request should specify subcomponents path. For example, line cards and fabric cards.


lifuzhang1108:
Thanks for reviewing Ankur!

PR #2415, opened 2023-11-23
 TE-2.2: moved static route under deviation condition

bkreddy143:
Currently script applying  static route override for all vendor while adding
interface+MAC NextHop instead of restrict to specific deviations.

"This code is a Contribution to the OpenConfig Feature Profiles project ("Work") made under the Google Software Grant and Corporate Contributor License Agreement ("CLA") and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia's intellectual property are granted for any other purpose. This code is provided on an "as is" basis without any warranties of any kind."

PR #2482, opened 2023-12-12
Add licenses and fix Readme

Ankur19:


feature/gribi/otg_tests/ipv4_entry_with_aggregate_ports_test/README.md
```diff
@@ -17,26 +18,34 @@ Validate IPv4 support in gRIBI using an Aggregate Port as a static route Next Ho
...
-            forwarded successfully.
-
-        *   Disable ATE port-2 and port-3 and forward packets between ATE port-1
-            and ATE LAG (destined to 198.51.100.0/24 ) and determine that
-            packets are lost 100%.
-
-        *   Re-enable both ATE port-2 and port-3 and forward packets between ATE
-            port-1 and ATE LAG (destined to 198.51.100.0/24 ) and determine that
-            packets are forwarded successfully again.
+        *   Configure a static ARP entry for the LAG interface pointing the
```
LimeHat, line 21: Looks like this part is written based on a specific vendor implementation (currently gated by a deviation)

Should this really be a part of the README? This wasn't the approach used in TE-2.1, and not a requirement AFAIK.

We also have PR #2415 that tries to rectify the current version of the test and readme with TE-2.1.

cc @xw-g

gforeal (approved):

feature/gribi/otg_tests/ipv4_entry_with_aggregate_ports_test/README.md
Ankur19: This is a requirement for some vendors that don't support MAC address override through gRIBI.

However, the scenario itself is very common in our prod environment and we expect to have similar static routes pointing to Aggregate Interfaces in prod.

Ankur19: The PR got merged, but lets discuss further. @xw-g Any comments?

LimeHat: > This is a requirement for some vendors that don't support MAC address override through gRIBI.

Yep, but that's a "workaround" / how they implement it.
But the base scenario is a direct gribi NH/NHG/route combo with a mac override that goes into a README, isn't it?

> we expect to have similar static routes pointing to Aggregate Interfaces in prod.

We believe the static route isn't required either, if you can specify an aggregate interface directly as a gribi NH? Just like TE-2.1 uses Ethernet interfaces directly as a NH (and the static route is a deviation/workaround)

Ankur19: So we want to validate such a scenario.

LimeHat: What is the purpose of the static route? Is there any other use of it beyond the recursive resolution of gribi routes?

Ankur19: I am not aware of the specific purpose behid the static routes. @xw-g will have more context.

LimeHat: Ok, let's wait for feedback from Xiao.

As far as we understand it, the base scenario is that gribi NH specifies an interface + mac directly (in this particular test case, an aggregate interface + mac).

All additional constructs (static routes, magic IP addresses, static ARPs) arise from a simple fact that not all implementations can do the above directly.

xw-g: This part (static route + static ARP) for archiving the MAC overriden should be gated by a deviation (as other test did). An example:

https://github.com/openconfig/featureprofiles/blob/5d59959528e125fee6639b716c2a30322e6e99da/internal/deviations/deviations.go#L129

https://github.com/openconfig/featureprofiles/blob/5d59959528e125fee6639b716c2a30322e6e99da/feature/gribi/otg_tests/base_hierarchical_nhg_update/base_hierarchical_nhg_update_test.go#L273

Ankur19: HI @xw-g the `MAC override` is not relevant to the test. I think we can remove the MAC override part from the test since we don't need it. The only thing we need is gRIBI route --> StaticRoute --> LAG Interface

LimeHat: >  The only thing we need is gRIBI route --> StaticRoute --> LAG Interface

Or, to make it simpler, `gRIBI route --> LAG Interface`? :-)

I understand where are you coming from (the issue you described above), but again, in our view the usage of the static route itself is a deviation (with or without MAC override) and the base scenario (for a fully gribi-compliant implementation) is a plain gRIBI interface NH (or interface+MAC nh)

Ankur19: Hi @LimeHat Had a discussion with @xw-g and you are right. We don't need the static route for Nokia. 

Will review https://github.com/openconfig/featureprofiles/pull/2415

Issue #1990, opened 2023-08-09
test cases for gNSI.certz


sachendras:

sachendras:
https://github.com/morrowc/featureprofiles/tree/main/feature/security/gnsi/certz/client_certificates#certz-11

Issue #1991, opened 2023-08-09
CertZ-1.2


sachendras:
https://github.com/morrowc/featureprofiles/tree/main/feature/security/gnsi/certz/client_certificates#certz-12

PR #2488, opened 2023-12-14
CERTZ-1.1/1.2-gNSI Client Certificate Tests

priyacj:
Initial Commit of CERTZ-1.1 and CERTZ-1.2
Resolves https://github.com/openconfig/featureprofiles/issues/1990 and https://github.com/openconfig/featureprofiles/issues/1991

### DIFF
--- a/feature/gribi/otg_tests/gribi_route_test/gribi_route_test.go
+++ b/feature/gribi/otg_tests/gribi_route_test/gribi_route_test.go
@@ -162,7 +162,26 @@ func TestGRIBIFailover(t *testing.T) {
 	t.Log("Configure VRF_Policy")
 	configureVrfSelectionPolicyC(t, dut)
 	t.Log("Configure GRIBI")
-	configureGribiRoute(t, dut)
+
+	ctx := context.Background()
+	gribic := dut.RawAPIs().GRIBI(t)
+	client := fluent.NewClient()
+	client.Connection().WithStub(gribic).WithPersistence().WithInitialElectionID(12, 0).
+		WithRedundancyMode(fluent.ElectedPrimaryClient).WithFIBACK()
+	client.Start(ctx, t)
+	defer client.Stop(t)
+	gribi.FlushAll(client)
+	defer gribi.FlushAll(client)
+	client.StartSending(ctx, t)
+	gribi.BecomeLeader(t, client)
+
+	tcArgs := &testArgs{
+		ctx:    ctx,
+		client: client,
+		dut:    dut,
+	}
+
+	configureGribiRoute(t, dut, tcArgs)
 
 	llAddress, found := gnmi.Watch(t, ate.OTG(), gnmi.OTG().Interface("port1.Eth").Ipv4Neighbor(dutPort1.IPv4).LinkLayerAddress().State(), time.Minute, func(val *ygnmi.Value[string]) bool {
 		return val.IsPresent()
@@ -394,25 +413,8 @@ func configureVrfSelectionPolicyC(t *testing.T, dut *ondatra.DUTDevice) {
 	gnmi.Replace(t, dut, dutPolFwdPath.Config(), niP)
 }
 
-func configureGribiRoute(t *testing.T, dut *ondatra.DUTDevice) {
+func configureGribiRoute(t *testing.T, dut *ondatra.DUTDevice, tcArgs *testArgs) {
 	t.Helper()
-	ctx := context.Background()
-	gribic := dut.RawAPIs().GRIBI(t)
-	client := fluent.NewClient()
-	client.Connection().WithStub(gribic).WithPersistence().WithInitialElectionID(12, 0).
-		WithRedundancyMode(fluent.ElectedPrimaryClient).WithFIBACK()
-	client.Start(ctx, t)
-	defer client.Stop(t)
-	gribi.FlushAll(client)
-	defer gribi.FlushAll(client)
-	client.StartSending(ctx, t)
-	gribi.BecomeLeader(t, client)
-
-	tcArgs := &testArgs{
-		ctx:    ctx,
-		client: client,
-		dut:    dut,
-	}
 	tcArgs.client.Modify().AddEntry(t,
 		fluent.NextHopEntry().WithNetworkInstance(deviations.DefaultNetworkInstance(tcArgs.dut)).
 			WithIndex(uint64(1)).WithDecapsulateHeader(fluent.IPinIP).


sezhang2 (approved):
Please add one more reviewer who is familiar with gNSI test requirement.

I approve the PR so that we can have test coverage on gNSI feature.

feature/security/gnsi/certz/tests/client_certificates/test_data/ca-01/client-rsa-a-key.pem
morrowc, line 1: was there a reason to include the testdata here?
wasn't (isn't) there a script mk_cas.sh that creates this on the fly for the folk doing testing events?

feature/security/gnsi/certz/tests/client_certificates/test_data/README.md
morrowc, line 1: this looks like the file from certz/test_data ... why is it re-added here?

feature/security/gnsi/certz/tests/client_certificates/README.md
morrowc, line 1: overall this set of changed files LOOKS like it's just a bunch of actually already existing files (safe the golang file(s) I guess?) moved around in the repository?

what's the actual intent here?

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,254 @@
...
+	setupService "github.com/openconfig/featureprofiles/feature/security/gnsi/certz/tests/internal/setup_service"
+	"github.com/openconfig/featureprofiles/internal/fptest"
+	"github.com/openconfig/ondatra"
+	"github.com/openconfig/ondatra/gnmi"
+	"github.com/openconfig/ondatra/gnmi/oc"
+	"github.com/openconfig/ygot/ygot"
+)
+
+var (
+	testProfile = "gNxI"
```
MarcCharlebois, line 34: Should this profile be created by the testcase and not assumed to be preconfigured?
something like sending an `addProfileReq` before the test executes? Thoughts, @morrowc ?
```
addProfileRequest := &certzpb.AddProfileRequest{SslProfileId: testProfile}
gnsiC.Certz().AddProfile(ctx, addProfileRequest)
```

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,254 @@
...
+	serverAddr  string
+	username    = "certzuser"
+	password    = "certzpswd"
+)
+
+// createUser function to add an user in admin role.
```
morrowc, line 40: what tells this function that the user is an admin user?
are there cases where we'd want to add a 'not admin' user?
do we want 2 functions (or 3 or 4 or...) can't this function take as an argument the user role?

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,254 @@
...
+// TestClientCertTcOne Test that client certificates from a set of one CA are able to be validated and
+// used for authentication to a device when used by a client connecting to each
+// gRPC service.
+
+func TestClientCert(t *testing.T) {
+
+	dut := ondatra.DUT(t, "dut")
+	serverAddr = dut.Name()
+
+	checkUser := createUser(t, dut, username, password)
```
morrowc, line 64: i'm partial to functions that, if they are going to not always return valid data, return an error. This createUser() function should just return an error if that's a possibility in the create process.

morrowc: i suppose the cost is the same 'is bool' or 'is nil'... but... the golang idiomatic  way (I think) is the return-error way.

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,254 @@
...
+	if !checkUser {
+		t.Fatalf("Failed to create certz user.")
+	}
+
+	t.Logf("Validation of all services that are using gRPC before certz rotation.")
+	initCheck := setupService.PreInitCheck(context.Background(), t, dut)
```
morrowc, line 70: this should do the same, return error not bool.

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+func TestMain(m *testing.M) {
+	fptest.RunTests(m)
+}
+
+// TestClientCertTcOne Test that client certificates from a set of one CA are able to be validated and
+// used for authentication to a device when used by a client connecting to each
+// gRPC service.
+
```
morrowc, line 58: remove unnecessary line breake.

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,254 @@
...
+		mismatch         bool
+	}{
+		{
+			desc:             "Certz1.1:Load the correct key-type rsa trustbundle with 1 CA configuration",
+			serverCertzFile:  "test_data/ca-01/server-rsa-a-cert.pem",
+			serverKeyzFile:   "test_data/ca-01/server-rsa-a-key.pem",
+			trustBundlezFile: "test_data/ca-01/trust_bundle_01_rsa_a.pem",
+			clientCertzFile:  "test_data/ca-01/client-rsa-a-cert.pem",
+			clientKeyzFile:   "test_data/ca-01/client-rsa-a-key.pem",
+			mismatch:         false,
```
morrowc, line 98: golang struct default for bool is 'false', so there's no need to explicitly set it, unless you want true.

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,254 @@
...
+		serverCertzFile  string
+		serverKeyzFile   string
+		trustBundlezFile string
+		clientCertzFile  string
+		clientKeyzFile   string
+		mismatch         bool
+	}{
+		{
+			desc:             "Certz1.1:Load the correct key-type rsa trustbundle with 1 CA configuration",
+			serverCertzFile:  "test_data/ca-01/server-rsa-a-cert.pem",
```
morrowc, line 93: is the reason that a bunch of the test data/etc is copied around because ../<foo>/test_data/.... is not feasible for some reason?

I'd also tend to make 'test_data' a const that points to the correct place:
  const(
    test_data = "../foo/test_data"
  )
  
  at the top of this test file.

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,254 @@
...
+			block, _ := pem.Decode(sc1)
+			if block == nil {
+				t.Fatalf("Failed to parse PEM block containing the public key.")
+			}
+			sCert, err := x509.ParseCertificate(block.Bytes)
+			if err != nil {
+				t.Fatalf("Failed to parse certificate: %v", err)
+			}
+
+			san := sCert.DNSNames[0]
```
morrowc, line 199: assuming [0] is the thing you want seems dangerous here.
I bet the testdata is setup to have this value, but also, what if it's not?

how would the production system determine if this was correct? (does it I guess? I don't know)

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+			serverCert := setupService.CreateCertzChain(t, setupService.CertificateChainRequest{
+				RequestType:    setupService.EntityTypeCertificateChain,
+				ServerCertFile: tc.serverCertzFile,
+				ServerKeyFile:  tc.serverKeyzFile})
+
+			serverCertEntity := setupService.CreateCertzEntity(t, setupService.EntityTypeCertificateChain, &serverCert, "one")
+
+			trustCertChain := setupService.CreateCertChainFromTrustBundle(tc.trustBundlezFile)
```
morrowc, line 208: is there a reason to not just do this in the next line?
(Why create a var here for a single use?)

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
morrowc: if you want to keep this, I think you want to use the pattern from line ~221

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,254 @@
...
+			success := setupService.CertzRotate(t, certzClient, testProfile, &serverCertEntity, &trustBundleEntity)
+			if !success {
+				t.Fatalf("%s:Certz rotation failed.", tc.desc)
+			}
+			t.Logf("%s:successfully completed certz rotation!", tc.desc)
+
+			if tc.mismatch != true {
+				// Verification check of the new connection with the new rotated certificates.
+				t.Run("Verification of new connection after successful rotate ", func(t *testing.T) {
+					result := setupService.PostValidationCheck(t, cacert, san, serverAddr, username, password, cert)
```
morrowc, line 235: if !setupService.PostValidation...... {
  t.Fatal(error here)
}

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,254 @@
...
+			}
+
+			certzClient := gnsiC.Certz()
+			success := setupService.CertzRotate(t, certzClient, testProfile, &serverCertEntity, &trustBundleEntity)
+			if !success {
+				t.Fatalf("%s:Certz rotation failed.", tc.desc)
+			}
+			t.Logf("%s:successfully completed certz rotation!", tc.desc)
+
+			if tc.mismatch != true {
```
morrowc, line 232: there's a 'do not if else' thing in golang... or that got hammered hard on me :( but...
you can just make this a switch:

  switch tc.mismatch {
  case true:
    stuff
   case false:
     other stuff
   }

morrowc: it also almost looks like you could either simplify as below, OR move the:
   result := .....
   
 outside the switch statement and just check mismatch to toss the error or not.

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
morrowc: I admit i don't really understand what testProfile is here.

Is the expectation that the client_certificates_test gets run on it's own? or that something reaches in here (and other featureprofile/profile/*/stuff/*_test.go) and runs this thing on it's own? (so either this being hardcoded here is fine, it's just run as a set of 'all the tests', or we should let the caller of the test set this on run)

feature/security/gnsi/certz/tests/client_certificates/test_data/ca-01/client-rsa-a-key.pem
priyacj: Readme file did not  mention with on the fly cert/key/pem file generation required for the tests. So,  It was made available under /client_certificates/tests/testdata statically. Do you see any concerns with static files ?

feature/security/gnsi/certz/tests/client_certificates/test_data/README.md
priyacj: The pem files were generated once and are made available statically under the directory security/gnsi/certz/tests/client_certificates/test_data/ .Do you see any concerns with static files ?

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
priyacj: gNxI is the default ssl profile available on the device.

feature/security/gnsi/certz/tests/client_certificates/test_data/ca-01/client-rsa-a-key.pem
morrowc: yea.. we dont' really want to end up with 12 copies of this data, right? and checking in the random certs/ca-data isn't helpful to the process. How about you rm-rf  the testdata you created from the PR here, re-point your  code to use the generated content, generate taht content for use in the test and jam on from there?

priyacj: Sure I will integrate the on the fly generation of the certificates before the tests .

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
priyacj: There is one value set to the DNS in the server config file under test_data
subjectAltName=DNS:role001.pop55.net.example.com,URI:spiffe://teamname.groupname.location.place.example.com/ro
le/groupname-group

featureprofiles/feature/security/gnsi/certz/test_data# more server_cert_ext.cnf 
basicConstraints = CA:FALSE
nsCertType = server
nsComment = "Test Server Certificates - openconfig / featureprofiles"
subjectKeyIdentifier = hash
authorityKeyIdentifier = keyid,issuer:always
keyUsage = critical, digitalSignature, keyEncipherment
extendedKeyUsage = serverAuth
subjectAltName=DNS:role001.pop55.net.example.com,URI:spiffe://teamname.groupname.location.place.example.com/ro
le/groupname-group

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
priyacj: ok , point taken

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
priyacj: noted .

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
priyacj: We are using the default gnsi profile available on the server , instead of creating a new ssl profile using addProfileRequest. There was no mention on  the README to create a new profile for the testcases.

feature/security/gnsi/certz/tests/client_certificates/test_data/README.md
morrowc: yes, do not add these files, there's no need to have yet another copy.
(you seem removing these static files based on a later comment)

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
LimeHat, line 34: @priyacj the spec was updated earlier today to indicate that the default profile should not be modified
https://github.com/openconfig/gnsi/blob/80891f288c18591a5930b4acd8ad5ae2c9504472/certz/README.md?plain=1#L96-L102; could you please update the test accordingly?
thx,

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
LimeHat: @priyacj it doesn't look like you use this user to rotate certificates. I believe you use ondatra gnsi bindings for that purpose (which is the correct approach!)

So, the question is, do you really need a new user at all? Can't you use ondatra bindings for the verification as well?

LimeHat (changes requested):

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
priyacj: The test flow picks up the newly  created user after rotation of certificates to validate new connection  and not using the vendor specific credentials from the binding file.

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
priyacj: taken care in my latest version

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
LimeHat: What's "the next upcoming merge"? Are you suggesting that it shouldn't be a part of this PR (if so, why)?

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,232 @@
...
+			success := setupService.CertzRotate(t, cacert, certzClient, cert, san, serverAddr, testProfile, &serverCertEntity, &trustBundleEntity)
+			if !success {
+				t.Fatalf("%s:Certz rotation failed.", tc.desc)
+			}
+			t.Logf("%s:successfully completed certz rotation!", tc.desc)
+			// Verification check of the new connection post rotation.
+			switch tc.mismatch {
+			case true:
+				t.Run("Verification of new connection with mismatch rotate of trustbundle.", func(t *testing.T) {
+					if setupService.TestNewConnection(t, cacert, san, serverAddr, username, password, cert) {
```
LimeHat, line 212: It looks like only one service is validated here? and multiple services in the next switch statements (case `false`)

imo both cases should use the same validation logic.

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
```diff
@@ -0,0 +1,232 @@
...
+						t.Fatalf("%s postTestcase service validation passed with mismatch rotate of trustbundle.", tc.desc)
+					}
+					t.Logf("%s postTestcase service validation with mismatch rotate of trustbundle is done!", tc.desc)
+				})
+			case false:
+				t.Run("Verification of new connection after rotate ", func(t *testing.T) {
+					if !setupService.PostValidationCheck(t, cacert, san, serverAddr, username, password, cert) {
```
LimeHat, line 219: same as above, should be the same validation function in both cases

LimeHat (changes requested):

feature/security/gnsi/certz/tests/client_certificates/README.md
```diff
@@ -0,0 +1,133 @@
+# CERTZ-1: gNSI Client Certificate Tests
```
morrowc, line 1: I might have asked this earlier, but why is it that this looks like the original file almost entirely?

feature/security/gnsi/certz/tests/client_certificates/README.md
```diff
@@ -0,0 +1,133 @@
...
+      client connections. Connect to the service on the DUT.
+
+   4) Validate that the connection to the remote device is established,
+      validate the client certificate can not be used (is untrusted) by
+      service on the DUT.
+
+   5) Validate that the connection is properly torn down by the DUT.
+
+## Protocol/RPC Parameter coverage
+
```
morrowc, line 106: this and the next 3 section headers are new, that appears to be basically the only changes to this file?

feature/security/gnsi/certz/tests/client_certificates/README.md
```diff
@@ -0,0 +1,133 @@
...
+```yaml
+paths:
+  # config paths
+  
+  # state paths 
+
+rpcs:
+  gnsi:
+    certz.v1.Certz.GetProfileList:
+    certz.v1.Certz.AddProfile:
```
morrowc, line 126: apologies, this line and the next are 'new'.

I don't understand why this diff is 'whole file is new'.

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
priyacj: The default gNxI profile is been replaced with new sslprofile <<newprofile>>. This ssl profile is added using addSSLprofile and then used for rotate RPC.

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
priyacj: Same validation logic is being used in Both the mismatch case and the postive testcase.

feature/security/gnsi/certz/tests/client_certificates/client_certificates_test.go
priyacj: same validation function are being set in both cases

morrowc (commented):
please address the static-analysis bits here.

PR #2490, opened 2023-12-15
Have ValidatePaths also return invalid paths.

wenovus:
This is interesting to know if we're only interested in the valid paths of a set of OCPaths.

For example, this is useful when we want to filter out paths that are already invalid somewhere else when reporting validation errors.


tools/internal/ocpaths/ocpaths.go
```diff
@@ -173,28 +165,50 @@ func insert(dstMap map[OCPathKey]*OCPath, src *OCPath) error {
 	return nil
 }
 
+func convertOCPath(ocpathProto *ppb.OCPath) *OCPath {
+	return &OCPath{
+		Key: OCPathKey{
+			Path:      ocpathProto.GetName(),
+			Component: ocpathProto.GetOcpathConstraint().GetPlatformType(),
```
dplore, line 172: Would it be more clear if this were `PlatformType` instead of `Component`?

dplore (dismissed):

tools/internal/ocpaths/ocpaths.go
wenovus: Done. PTAL

dplore (approved):

PR #2492, opened 2023-12-15
Updated the test case id number for IPv6 ND RA disable.

priyaranjannanda:
Updated the test case number for IPv6 ND RA disable to RT-9 based on the comment in pull/2485


dplore:
The readme for RT-9 is not pointing to the correct location?  Does the readme for RT-9 exist or should be updated to match?

sezhang2 (approved):
It is a cosmetic change, so I will approve and merge this PR

Issue #2489, opened 2023-12-14
RT-1.26 Basic static route support test


rohit-rp:
# Request for New Test Code Implementation Instructions

- This issues tracks IPv4 static route with IPv4 nexthop and IPv6 static route with IPv6 nexthop
- Static route ECMP must be supported
- Static route metric must be supported
- Static route Administrative Distance / Preference must be supported
- ```set-tag``` attribute must be supported for static routes
- Disabling recursive nexthop resolution must be supported

Implement code for [RT-1.26](https://github.com/openconfig/featureprofiles/blob/main/feature/staticroute/static_route_support/README.md)

PR #2493, opened 2023-12-15
Static route support readme

rohit-rp:
# README for static route test development
- Static route ECMP must be supported
- Static route metric must be supported
- Static route Administrative Distance / Preference must be supported
- ```set-tag``` attribute must be supported for static routes
- Disabling recursive nexthop resolution must be supported

Only for IPv4 static route with IPv4 next hop and Ipv6 static route for IPv6 next Hop

ISSUE: https://github.com/openconfig/featureprofiles/issues/2489

### DIFF
--- a/feature/staticroute/static_route_support/README.md
+++ b/feature/staticroute/static_route_support/README.md
@@ -18,23 +18,23 @@
 
 *   Connect DUT port-1, port-2 and port-3 to ATE port-1, port-2 and port-3 respectively
 *   Configure IPv4/IPv6 addresses on DUT and ATE the interfaces
-*   Configure one IPv4 destination i.e. ipv4-network connected to ATE port 1 and 2
-*   Configure one IPv6 destination i.e. ipv6-network connected to ATE port 1 and 2
+*   Configure one IPv4 destination i.e. ```ipv4-network = 192.168.10.0/24``` connected to ATE port 1 and 2
+*   Configure one IPv6 destination i.e. ```ipv6-network = 2024:db8:128:128::/64``` connected to ATE port 1 and 2
 
 ### RT-1.26.1 [TODO: https://github.com/openconfig/featureprofiles/issues/2489]
 #### Test to validate static route ECMP
 
 *   Configure IPv4 static routes:
-    *   Configure one IPv4 static route i.e. ipv4-route-a on the DUT for destination ipv4-network with the next hop set to the IP address of ATE port-1
-    *   Configure another IPv4 static route i.e. ipv4-route-b on the DUT for destination ipv4-network with the next hop set to the IP address of ATE port-2
+    *   Configure one IPv4 static route i.e. ipv4-route-a on the DUT for destination ```ipv4-network 192.168.10.0/24``` with the next hop set to the IPv4 address of ATE port-1
+    *   Configure another IPv4 static route i.e. ipv4-route-b on the DUT for destination ```ipv4-network 192.168.10.0/24``` with the next hop set to the IPv4 address of ATE port-2
 *   Validate both the routes i.e. ipv4-route-[a|b] are configured and reoported correctly
     *   /network-instances/network-instance/protocols/protocol/static-routes/static/prefix
 *   Configure IPv6 static routes:
-    *   Configure one IPv6 static route i.e. ipv6-route-a on the DUT for destination ipv6-network with the next hop set to the IP address of ATE port-1
-    *   Configure another IPv6 static route i.e. ipv6-route-b on the DUT for destination ipv6-network with the next hop set to the IP address of ATE port-2
-*   Validate bothe the routes i.e. ipv6-route-[a|b] sre configured and reported correctly
+    *   Configure one IPv6 static route i.e. ipv6-route-a on the DUT for destination ```ipv6-network 2024:db8:128:128::/64``` with the next hop set to the IPv6 address of ATE port-1
+    *   Configure another IPv6 static route i.e. ipv6-route-b on the DUT for destination ```ipv6-network 2024:db8:128:128::/64``` with the next hop set to the IPv6 address of ATE port-2
+*   Validate both the routes i.e. ipv6-route-[a|b] sre configured and reported correctly
     *   /network-instances/network-instance/protocols/protocol/static-routes/static/prefix
-*   Initiate traffic from ATE port-3 towards destination ipv4-network and ipv6-network
+*   Initiate traffic from ATE port-3 towards destination ```ipv4-network 192.168.10.0/24``` and ```ipv6-network 2024:db8:128:128::/64```
 *   Validate that traffic is received from DUT on both port-1 and port-2 and ECMP works
 
 ### RT-1.26.2 [TODO: https://github.com/openconfig/featureprofiles/issues/2489]
@@ -44,7 +44,7 @@
     *   /network-instances/network-instance/protocols/protocol/static-routes/static/next-hops/next-hop/config/metric
 *   Validate that the metric is set correctly
     *   /network-instances/network-instance/protocols/protocol/static-routes/static/next-hops/next-hop/state/metric
-*   Initiate traffic from ATE port-3 towards destination ipv4-network and ipv6-network
+*   Initiate traffic from ATE port-3 towards destination ```ipv4-network 192.168.10.0/24``` and ```ipv6-network 2024:db8:128:128::/64```
 *   Validate that traffic is received from DUT on port-1 and not on port-2
 
 ### RT-1.26.3 [TODO: https://github.com/openconfig/featureprofiles/issues/2489]
@@ -54,7 +54,7 @@
     *   /network-instances/network-instance/protocols/protocol/static-routes/static/next-hops/next-hop/config/preference
 *   Validate that the preference is set correctly
     *   /network-instances/network-instance/protocols/protocol/static-routes/static/next-hops/next-hop/state/preference
-*   Initiate traffic from ATE port-3 towards destination ipv4-network and ipv6-network
+*   Initiate traffic from ATE port-3 towards destination ```ipv4-network 192.168.10.0/24``` and ```ipv6-network 2024:db8:128:128::/64```
 *   Validate that traffic is now received from DUT on port-2 and not on port-1
 
 ### RT-1.26.4 [TODO: https://github.com/openconfig/featureprofiles/issues/2489]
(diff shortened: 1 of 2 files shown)


feature/staticroute/static_route_support/README.md
```diff
@@ -0,0 +1,114 @@
...
+*   Connect DUT port-1, port-2 and port-3 to ATE port-1, port-2 and port-3 respectively
+*   Configure IPv4/IPv6 addresses on DUT and ATE the interfaces
+*   Configure one IPv4 destination i.e. ipv4-network connected to ATE
+*   Configure one IPv6 destination i.e. ipv6-network connected to ATE
+
+### RT-1.26.1 [TODO: https://github.com/openconfig/featureprofiles/issues/2489]
+#### Test to validate static route ECMP
+
+*   Configure IPv4 static routes:
+    *   Configure one IPv4 static route i.e. ipv4-route-a on the DUT for destination ipv4-network with the next hop of port-1
```
dplore, line 28:
```suggestion
    *   Configure one IPv4 static route i.e. ipv4-route-a on the DUT for destination ipv4-network with the next hop set to the IP address of ATE port-1
```

feature/staticroute/static_route_support/README.md
```diff
@@ -0,0 +1,114 @@
...
+    *   Configure another IPv4 static route i.e. ipv4-route-b on the DUT for destination ipv4-network with the next hop of port-2
```
dplore, line 29:
```suggestion
    *   Configure another IPv4 static route i.e. ipv4-route-b on the DUT for destination ipv4-network with the next hop set to the IP address of ATE port-2
```

feature/staticroute/static_route_support/README.md
```diff
@@ -0,0 +1,114 @@
...
+*   Configure IPv6 static routes:
+    *   Configure one IPv6 static route i.e. ipv6-route-a on the DUT for destination ipv6-network with the next hop of port-1
```
dplore, line 31:
```suggestion
    *   Configure one IPv6 static route i.e. ipv6-route-a on the DUT for destination ipv6-network with the next hop set to the IP address of ATE port-1
```

feature/staticroute/static_route_support/README.md
```diff
@@ -0,0 +1,114 @@
...
+    *   Configure another IPv6 static route i.e. ipv6-route-b on the DUT for destination ipv6-network with the next hop of port-2
```
dplore, line 32:
```suggestion
    *   Configure another IPv6 static route i.e. ipv6-route-b on the DUT for destination ipv6-network with the next hop set to the IP address of ATE port-2
```

feature/staticroute/static_route_support/README.md
```diff
@@ -0,0 +1,114 @@
...
+*   Initiate traffic from ATE port-3 towards destination ipv4-network and ipv6-network
```
dplore, line 33: Before starting traffic, validate static route state paths match what was configured

feature/staticroute/static_route_support/README.md
```diff
@@ -0,0 +1,114 @@
...
+*   Validate that traffic is received from DUT on both port-1 and port-2 and ECMP works
+
+### RT-1.26.2 [TODO: https://github.com/openconfig/featureprofiles/issues/2489]
+#### Test to validate static route metric
+
+*   Configure metric of ipv4-route-b and ipv6-route-b to 1000
```
dplore, line 39: is this creating new static route entries or reusing old?  I guess since the names of the routes are '-b', is the intent is to create new static routes?  What are the nexthops?

feature/staticroute/static_route_support/README.md
```diff
@@ -0,0 +1,114 @@
...
+    *   /network-instances/network-instance/protocols/protocol/static-routes/static/next-hops/next-hop/config/metric
+*   Validate that the metric is set correctly
+    *   /network-instances/network-instance/protocols/protocol/static-routes/static/next-hops/next-hop/state/metric
+*   Initiate traffic from ATE port-3 towards destination ipv4-network and ipv6-network
+*   Validate that traffic is received from DUT on port-1 only
```
dplore, line 44: It seems the routes from RT-1.26.1 would be used since they are default metric and therefore traffic is sent to both port1 and port2?   Looks like the new ipv4/6-route-b is not used?  Please define some route and metric that forces the DUT to make a routing decision that results in only one port being used.

rszarecki (dismissed):
Considering we have more static route features (v4 w/ v6 nh and). Perhaps some extra word in title. like "base"

feature/staticroute/static_route_support/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+### RT-1.26.3 [TODO: https://github.com/openconfig/featureprofiles/issues/2489]
+#### Test to validate static route prefrence 
+
+*   Configure preference of ipv4-route-a and ipv6-route-a to 1000
```
dplore, line 49: Define the prefix and nexthops for route-a.  They need to have the same metric with different preferences, right?

feature/staticroute/static_route_support/README.md
```diff
@@ -0,0 +1,114 @@
...
+    *   /network-instances/network-instance/protocols/protocol/static-routes/static/next-hops/next-hop/config/preference
+*   Validate that the preference is set correctly
+    *   /network-instances/network-instance/protocols/protocol/static-routes/static/next-hops/next-hop/state/preference
+*   Initiate traffic from ATE port-3 towards destination ipv4-network and ipv6-network
+*   Validate that traffic is now received from DUT on port-2 only
```
dplore, line 54: Shouldn't this use ipv(4,6)-network-a?

feature/staticroute/static_route_support/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+### RT-1.26.4 [TODO: https://github.com/openconfig/featureprofiles/issues/2489]
+#### Test to validate static route tag
+
+*   Configure a tag of value 10 on ipv4-route-a and ipv6-route-a
+    *   /network-instances/network-instance/protocols/protocol/static-routes/static/config/set-tag
+*   Validate the tag is set
+    *   /network-instances/network-instance/protocols/protocol/static-routes/static/state/set-tag
+
```
dplore, line 63: I think you need to add addtional criteria to make sure the tag can be used:
See [OC description of tag](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-static-routes-static-config-set-tag)

feature/staticroute/static_route_support/README.md
```diff
@@ -0,0 +1,114 @@
...
+### RT-1.26.5 [TODO: https://github.com/openconfig/featureprofiles/issues/2489]
+#### Test to validate disabling of recursive next-hop resolution
+
+*   Remove the ipv4-route-[a|b] and ipv6-route-[a|b] static routes
+*   Configure ipv4 and ipv6 ISIS between ATE port-1 <-> DUT port-1 and ATE port-2 <-> DUT port2
```
dplore, line 68: I think you need three ports?
I think you want to override an ISIS learned route by using a static route with an interface-ref set:

See: 
https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-static-routes-static-next-hops-next-hop-config-recurse

and 
https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-static-routes-static-next-hops-next-hop-interface-ref

dplore (changes requested):

feature/staticroute/static_route_support/README.md
LimeHat: IMO this is correctly written for the validation of the recurse flag (if this is the objective)

Interface-ref is just one specific example of disabling the recursion (with an additional restriction), however, any IP next-hop can be resolved recursively (or not).

feature/staticroute/static_route_support/README.md
rohit-rp: No just increasing the metric of previously created routes (ipv4-route-b and ipv6-route-b) to 1000
These were created in [line 29](https://github.com/openconfig/featureprofiles/blob/46788e97668c975a98719db28f9d7bed57febfe7/feature/staticroute/static_route_support/README.md?plain=1#L29) and [line 32](https://github.com/openconfig/featureprofiles/blob/46788e97668c975a98719db28f9d7bed57febfe7/feature/staticroute/static_route_support/README.md?plain=1#L30)

feature/staticroute/static_route_support/README.md
rohit-rp:
```
ipv[4|6]-route-a has next-hop = ATE port-1 IP
ipv[4|6]-route-b has next-hop = ATE port-2 IP
```

RT-1.26.1 ensures ECMP works [line 34](https://github.com/openconfig/featureprofiles/blob/46788e97668c975a98719db28f9d7bed57febfe7/feature/staticroute/static_route_support/README.md?plain=1#L34). Now that we increase metric of ipv[4|6]-route-b routes (which have next-hop of port-2) traffic will be received only on port-1 of ATE (port-2 should not get any traffic). I will clarify this

feature/staticroute/static_route_support/README.md
rohit-rp: ipv[4|6]-route-b were not being used due to increase in metric in RT-1.26.2
Now in RT-1.26.3 we increase the preference of ipv[4|6]-route-a to 1000
ipv[4|6]-route-b will be used again (preference over metric) and traffic should be visible on port-2 now. Port-1 should not get any traffic

feature/staticroute/static_route_support/README.md
rohit-rp: No, ipv[4,6]-network-a's metric is lower than ipv[4,6]-network-b but because preference of ipv[4,6]-network-a is 1000 it will not be used. (Ah, 1000 probably is illegal, the value ranges from (0 - 255), will change it to 200

feature/staticroute/static_route_support/README.md
rohit-rp: This is covered in static to ISIS/BGP redistribution readme that I am working on https://github.com/openconfig/featureprofiles/issues/2494

feature/staticroute/static_route_support/README.md
rohit-rp: > I think you want to override an ISIS learned route by using a static route with an interface-ref set:

No actually I don't want to advertise, ipv4-network and ipv6-network through IS-IS. Static route next hop i.e. ipv[4|6]-loopback will be advertised through ISIS and will be recursively resolved.
I will remove [line 70](https://github.com/openconfig/featureprofiles/blob/46788e97668c975a98719db28f9d7bed57febfe7/feature/staticroute/static_route_support/README.md?plain=1#L70)

feature/staticroute/static_route_support/README.md
dplore: I was confused and thought route-a and route-b were different prefixes.

feature/staticroute/static_route_support/README.md
dplore: Ok, this is fair.  Tags are only used as part of routing-policy and it is fair to have the tag use case in some other test which include `routing-policy`.

feature/staticroute/static_route_support/README.md
dplore: SGTM

dplore (approved):

PR #2497, opened 2023-12-15
Simplify static binding by removing dialer

greg-dennis:


greg-dennis:
I created  PR with a fake dummy commit to the example test to spot check that the tests will still pass:
https://github.com/openconfig/featureprofiles/pull/2499
Seems to be the case.

topologies/binding/options.go
```diff
@@ -46,196 +30,19 @@ var (
...
 	for _, bopt := range bopts {
 		if bopt != nil {
 			proto.Merge(result, bopt)
 		}
 	}
-	return dialer{result}
+	return result
 }
 
 // resolver returns the dialer for specific devices and protocols.
```
jasdeep-hundal, line 48: Update comment since it's not longer directly returning a dialer?

jasdeep-hundal (dismissed):

jasdeep-hundal (approved):

PR #2498, opened 2023-12-15
RT-7 fix loopback config

LimeHat:
The current version of the test performs gnmi.Replace on the loopback interface which may override the existing loopback config used for infra purposes 

This PR modifies the logic to create a new loopback interface/subinterface only when the zero-indexed loopback subif does not have ipv4 & v6 addresses (and uses update instead of replace); otherwise, it reuses the existing IPs

---
<sup>
</sup>


cfernanz:
internal b/316639762 to validate this PR

cfernanz (approved):
LGTM

PR #2504, opened 2023-12-19
RT-1.1: Updating await timeout for BGP session with correct password

trathod1:
After BGP session goes down on DUT due to incorrect password, remote ATE end waits for holdtime to bring the session down. Thus when script configures the correct password, it needs to wait for holdtime for session to come up again.


NishaSadhasivam-work (approved):
LGTM!

PR #2511, opened 2023-12-20
RT-1.5 bgp_prefix_limit_test.go 

cprabha:
Hi,

I have added deviation to skip checking telemetry for prefix limit exceeded flag.

Thanks,
Prabha


sezhang2 (dismissed):
This PR introduced a new deviation  _prefix_limit_exceeded_telemetry_unsupported_ to make the test pass.

The new deviation is tracked by  b/317181227. It should not have any impact on other vendors' platforms.

arulkumarsekar (approved):
re-approving

Issue #2453, opened 2023-12-05
mk_cas.sh calling a CA that doesn't exist.


MarcCharlebois:
**Describe the bug**
executing [mk_cas.sh](https://github.com/openconfig/featureprofiles/blob/main/feature/security/gnsi/certz/test_data/mk_cas.sh) results in 

```bash
+ openssl ecparam -name prime256v1 -out ca-01/client-ecdsa-b-key.pem -genkey
+ openssl ecparam -name prime256v1 -out ca-01/server-ecdsa-b-key.pem -genkey
+ openssl req -new -key ca-01/client-ecdsa-b-key.pem -out ca-01/client-ecdsa-b-req.pem -config client_cert.cnf
+ openssl req -new -key ca-01/server-ecdsa-b-key.pem -out ca-01/server-ecdsa-b-req.pem -config server_cert.cnf
+ openssl x509 -req -in ca-01/client-ecdsa-b-req.pem -CA ca-01/ca-0001-ecdsa-b-cert.pem -CAkey ca-01/ca-0001-ecdsa-b-key.pem -out ca-01/client-ecdsa-b-cert.pem -CAcreateserial -days 3650 -sha256 -extfile client_cert_ext.cnf
Certificate request self-signature ok
subject=C = AQ, ST = NZ, L = NZ, O = OpenconfigFeatureProfiles, CN = client.service.example.com
Could not open file or uri for loading CA certificate from ca-01/ca-0001-ecdsa-b-cert.pem: No such file or directory
```

**To Reproduce**
execute the script

**Expected behavior**
Generate all the certs required (now fails to create the client/server cert/key) as it doesn't find the CA cert/key

**Logs**
n/a

**Network device platform (please complete the following information):**
OS tried: Centos9, Ubuntu 22.04, Ubuntu 20.04

**Additional context**
When the script is called, it creates the CAs without a 'letter'/reference to `${g}`, we can see this here: https://github.com/openconfig/featureprofiles/blob/main/feature/security/gnsi/certz/test_data/mk_cas.sh#L29-L43

here's an output example: `ca-01/ca-0001-ecdsa-cert.pem`
I believe there might be an issue in the [for loop](https://github.com/openconfig/featureprofiles/blob/main/feature/security/gnsi/certz/test_data/mk_cas.sh#L69)  which creates the cient/server certs/keys.

Here's a patch that worked locally for me.
```bash 
[...]

PR #2085, opened 2023-09-01
fix key generation script issue

mojiiba:
This pull fixes a file name issue for cert generation script.

PR #2518, opened 2023-12-27
Fix ab problems from #2085 and #2453

morrowc:
As per the commit message:

    Cleanup a bit of the mk_cas.sh script, to include handling the a/b
    functionality correctly.
    
    The point of the a/b was to create in each set of CAs 2 client and 2
    server certificates for use in the tests. This enables tests like:
    
      * "Can CA02-1-clientA talk to CA02-2-serverB from CA-02 set?" - yes
      * "Can CA02-1-cliantA talk to CA10-1-serverB from CA-10 set?" - no
    
    Previously the a/b functionality was also attempting to be implemented
    in the CA creation, which was incorrect (and erroring, actually).


MarcCharlebois (approved):
Looks good to me.

marcushines (approved):

PR #2528, opened 2024-01-05
Sync wbb p4info with internal version

nhawke:
This external copy of the wbb p4info was out of sync with internal. This PR replaces that old version with the current HEAD version internally.


feature/experimental/p4rt/wbb.p4info.pb.txt
```diff
@@ -737,28 +156,11 @@ controller_packet_metadata {
   metadata {
     id: 3
     name: "unused_pad"
-    bitwidth: 7
+    annotations: "@padding"
+    bitwidth: 6
```
Ankur19, line 160: Hi Nick! This is going to break P4RT tests for some vendors.

We need some time such that vendors can implement the change before we can merge it here.

There is an existing PR for this change: https://github.com/openconfig/featureprofiles/pull/2146

nhawke: Ah i see. I wasnt aware of the existing PR. I'll close this one then.

PR #2529, opened 2024-01-05
NOSImageProfile Validation for ocrpcs

wenovus:
Added CI check for validation script and examples.


.github/workflows/nosimage.yml
```diff
@@ -0,0 +1,41 @@
+name: NOSImage validation script
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+  schedule:
+    - cron: "0 0 * * *"
```
dplore, line 8: recommended to chose a more random time of day per https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

dplore (approved):

.github/workflows/nosimage.yml
wenovus: Done.

dplore (dismissed):

dplore (approved):

PR #2548, opened 2024-01-15
TE-11.3,RT-1.10: configure default NI name

kjahed:
Use fptest.ConfigureDefaultNetworkInstance to make sure NI name is configured when replacing default NI.


AmrNJ:
Please can you rebase?

self-maurya (approved):

PR #2549, opened 2024-01-15
Adding PKCS8 to LoadKeyPair()

MarcCharlebois:
openssl 3.X defaults to PKCS8
This prevents the requirement to use openssl 1.x or migrate certs from PKCS8->PKCS1
Using `PRIAVE KEY` as case since the preamble is `-----BEGIN PRIVATE KEY-----` "This code is a Contribution to the OpenConfig Feature Profiles project ("Work") made under the Google Software Grant and Corporate Contributor License Agreement ("CLA") and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia's intellectual property are granted for any other purpose. This code is provided on an "as is" basis without any warranties of any kind."

### DIFF
--- a/internal/security/svid/svid.go
+++ b/internal/security/svid/svid.go
@@ -108,6 +108,11 @@ func LoadKeyPair(keyPath, certPath string) (any, *x509.Certificate, error) {
 		if err != nil {
 			return nil, nil, err
 		}
+	case "PRIVATE KEY":
+		caPrivateKey, err = x509.ParsePKCS8PrivateKey(caKeyPem.Bytes)
+		if err != nil {
+			return nil, nil, err
+		}
 	default:
 		return nil, nil, fmt.Errorf("file does not contain an ECDSA/RSA private key")
 


MarcCharlebois:
cc: @mojiiba  @morrowc

mojiiba (approved):

internal/security/svid/svid.go
morrowc, line 1: Can't figure out how to comment on the lines I actually want to make a comment on/about, but...
There's some use of 'any' in this which seems improper/unnecessary :(
There's also some cavalier use of `pem.Decode()` which ... ignores the fact that the thing being decoded MAY have more than 1 PEM thingy in it... that seems dangerous, or at least ready to be exploited to cause us problems :(

morrowc (approved):

internal/security/svid/svid.go
morrowc: Particularly the any usage at line 96 .. I don't see an easy way OUT of that hole :( which is sad.
and the skipping the extra parts of the decode at line 95 I do see a  way out:

```golang
  caKeyPem, rest := pem.Decode(keyPEM)
  if len(rest) > 0 { toss cookies }
```

I don't think fixing that 'rest' thing is important in THIS PR, but it'd be nice to fix it before we start really using any of this code, eh? :)

MarcCharlebois: I see what you mean - Maybe better to track/fix this in another PR.
@mojiiba did you want to pick that up?

NishaSadhasivam-work (approved):
LGTM!

PR #2550, opened 2024-01-16
gNMI-1.2: fix key leafref violations

kjahed:
Modifications to gNMI-1.2 to avoid leafref violations for key leaves:
1- Make sure the key 'name' is configured for default NI
2- For cleanup, delete the entire (BGP, ISIS) Protocol nodes


AmrNJ (changes requested):
While running the test, I get "EthernetConnection has no field or method SetChoice" error.

kjahed:
> While running the test, I get "EthernetConnection has no field or method SetChoice" error.

Can you provide a bit more context? The test builds and runs fine for me. This error is related to gosnappi (SetChoice was un-exported in v0.13.7) but the test does not really use OTG. In any case, I did rebase so maybe that will help.

AmrNJ (approved):
LGTM

AmrNJ (approved):
LGTM

PR #2552, opened 2024-01-16
FP-1.1: fix key leafref violations

kjahed:
Configure the key leaf 'name' of Component before configuring power type.


AmrNJ (changes requested):
The test fails for 8808. Error logs shared in bug.

AmrNJ (changes requested):
The test fails for 8808.

AmrNJ:
@kjahed These changes are already reflected in the recent code. Please can you close this?

AmrNJ:
The PR is no longer needed as these changes are already present in the recent code.

PR #2558, opened 2024-01-18
Move binary logs to google cloud object storage

nhawke:
Storing the files in a public cloud storage bucket prevents the messiness of keeping them in the git repo while ensuring the file is accessible by the test both in open source and when running internally.


greg-dennis:
Can you delete this file too?
https://github.com/openconfig/featureprofiles/blob/main/.gitattributes

greg-dennis (approved):

PR #2572, opened 2024-01-23
RT-1.5: BGP Prefix Limit : Added deviation

cprabha:
Hi,

I have added deviation for ATE script ( otg is updated via #2511).

Thanks,
Prabha


cprabha:
This test is in retirement list as in pull request https://github.com/openconfig/featureprofiles/pull/2585.
Hence closing it.

Issue #2570, opened 2024-01-23
RT-10 BGP to IS-IS redistribution


rohit-rp:
# Request for New Test Code Implementation Instructions

## Summary

- Source-protocol: BGP, destination-protocol: ISIS L2
- Using Community (Source-protocol: BGP)
- Using defined prefix set

Implement test code for [RT10.x](https://github.com/openconfig/featureprofiles/blob/main/feature/bgp/bgp_isis_redistribution/README.md)

PR #2574, opened 2024-01-24
RT-1.28: BGP to IS-IS redistribution

rohit-rp:
# RT-1.28: BGP to IS-IS redistribution

## Summary

- Source-protocol: BGP, destination-protocol: ISIS L2
- Using Community (Source-protocol: BGP)
- Using defined prefix set

https://github.com/openconfig/featureprofiles/issues/2570


rszarecki (commented):
this test requires also negaive check - prefix that do not match criteria, so should be not redistributed.

Also, please et default-import-policy REJECT_ROUTE

Otherwise we do not know if filtering actually works.

rszarecki (approved):

PR #2582, opened 2024-01-26
Add dc gate tag to test

Ankur19:


Ankur19:
@self-maurya @rohit-rp Do you know any other DC gate tests that need this tag?

dplore (approved):

PR #2592, opened 2024-01-29
RT-2.12: Updating config order

trathod1:
Updating config order to avoid very intermittent failure.


NishaSadhasivam-work (approved):
LGTM!.

Can you please share the difference you noticed, while configuring ATE after DUT.
i.e, the need to move line -479 to +481

trathod1:
> LGTM!.
> 
> Can you please share the difference you noticed, while configuring ATE after DUT. i.e, the need to move line -479 to +481

I noticed very rare failure (once in 10 run), suspecting timing issue, when we allow configurations in this order, it is giving enough time for configs to get absorbed and seeing consistent passes.

PR #2593, opened 2024-01-30
GNMI-1.11: Removing subinterface_packet_counters_missing deviation for Nokia

trathod1:
Removing `subinterface_packet_counters_missing` deviation for Nokia


NishaSadhasivam-work (changes requested):
I see that the subinterface counter path is not present.

Get IsPresent status for path "/interfaces/interface/subinterfaces/subinterface/ipv4/state/counters/in-pkts": got false, want true

Software version: 53987-g34652c2126

Can you please let me know whey we are removing subinterface_packet_counters_missing from the deviation

trathod1:
Subinterface packet counters are supported, however sometimes these counters taking time to get updated just after the subinterface is configured. This was causing the intermittent failure which was observed in your run.

I have committed the fix in the script to perform `gnmi.watch` for the counter, so we have consistent results.

NishaSadhasivam-work (changes requested):
I am seeing the below failure, even with gnmi.watch.

Get IsPresent status for path "/interfaces/interface/subinterfaces/subinterface/ipv4/state/counters/in-pkts": got false, want true
Got path/value: /[interfaces/interface/subinterfaces/subinterface/ipv4/state/counters/in-pkts:0](https://cs.corp.google.com/piper///depot/google3/interfaces/interface/subinterfaces/subinterface/ipv4/state/counters/in-pkts?ws=nsadhasivam/198&snapshot=11)

Can you please share the expected time delay after configuration for the subinterface leaves to get updated?

trathod1:
Mostly the subinterface counters gets updated immediately, however sometimes it take few seconds. Max I observed was 10 seconds, based on that gnmi.watch was set to 15 seconds to have buffer.

But I am seeing another issue in the counter update, and I am following up with the team here. Let me get back to you after hearing from them.

trathod1:
I have reverted the `gnmi.watch` changes.  Earlier, this particular counter was sensitive to traffic, due to which it was taking time to get updated sometimes. This issue is resolved. Will move PR once image to validate is released.

trathod1:
This would work fine now on latest image.

NishaSadhasivam-work (approved):
LGTM!

Issue #2612, opened 2024-02-02
BGP 3 levels of nested import/export policy with match-set-options


rohit-rp:
# RT-1.31: BGP 3 levels of nested import/export policy with match-set-options

## Summary

- AS-path prepending by more the 10 repetitions
- Recursive policy subroutines (multi-level nesting). At least 3 levels
- match-set-options of ANY, INVERT for match-prefix-set conditions
- Applicable to both IPv4 and IPv6 BGP neighbors

Implement code for [RT-1.31](https://github.com/openconfig/featureprofiles/blob/main/feature/bgp/policybase/otg_tests/3level_nested_policies/README.md)

PR #2614, opened 2024-02-02
 RT-1.31: BGP 3 levels of nested import/export policy with match-set-options

rohit-rp:
# RT-1.31: BGP 3 levels of nested import/export policy with match-set-options

## Summary

- AS-path prepending by more the 10 repetitions
- Recursive policy subroutines (multi-level nesting). At least 3 levels
- match-set-options of ANY, INVERT for match-prefix-set conditions
- Applicable to both IPv4 and IPv6 BGP neighbors

https://github.com/openconfig/featureprofiles/issues/2612

### DIFF
--- a/feature/bgp/policybase/otg_tests/3level_nested_policies/README.md
+++ b/feature/bgp/policybase/otg_tests/3level_nested_policies/README.md
@@ -1,0 +1,454 @@
+# RT-1.31: BGP 3 levels of nested import/export policy with match-set-options
+
+## Summary
+
+- AS-path prepending by more the 10 repetitions
+- Recursive policy subroutines (multi-level nesting). At least 3 levels
+- match-set-options of ANY, INVERT for match-prefix-set conditions
+- Applicable to both IPv4 and IPv6 BGP neighbors
+
+## Testbed type
+
+*   https://github.com/openconfig/featureprofiles/blob/main/topologies/atedut_2.testbed
+
+## Procedure
+
+### Initial Setup:
+
+*   Connect DUT port-1, 2 to ATE port-1, 2
+*   Configure IPv4/IPv6 addresses on the ports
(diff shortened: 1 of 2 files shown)


rszarecki (changes requested):
A) do we have separate independent test against AS path prepend? I asume we do not.

B) IMO as-path-prepend  and nested policies shall be separate tests. This are unrelated features and I do not understand why they are tested together. Note that if this test Fail due to nesting, we have no coverage for prepend. and vice-a-versa.

C) Setting origin is path we never requested and we do not use. Better use MED or LP or communities

D) I will replace ORIGIN w/ MED in 1.31.1 nad 1.31.2; Then remove 1.31.3 and 1.31.4 completely

E) make separate test (separate README) fopr prepend.

rohit-rp:
> A) do we have separate independent test against AS path prepend? I asume we do not.
It is also covered here: http://shortn/_7KIZD3ClNY
But i believe we will still need atomic test to only validate as-path prepend. I will write a test only to check AS path prepend
> B) IMO as-path-prepend and nested policies shall be separate tests. This are unrelated features and I do not understand why they are tested together. Note that if this test Fail due to nesting, we have no coverage for prepend. and vice-a-versa.
Will write separate test for AS-PATH prepend, but the 3 levels of nested policies have to take some action, it cant be blank policies else how will i validate that all policy actions are carried out. Every policy should do something that can be verified
> C) Setting origin is path we never requested and we do not use. Better use MED or LP or communities
Will remove and replace with community
> D) I will replace ORIGIN w/ MED in 1.31.1 nad 1.31.2; Then remove 1.31.3 and 1.31.4 completely
Ok replacing ORIGIN with MED but why remove 1.31.3 and 1.31.4 do we not need to check for IPv6 also to make sure (I know it will work for IPv6) but lets have the coverage, what do you think?
> E) make separate test (separate README) fopr prepend.
On it

rszarecki (approved):

PR #2622, opened 2024-02-05
RT-5.1 change from in-discards to use in-errors for drop

arvbaska1:
Changing the counter for drop variable to use in-errors leaf instead of in-discards.
As per OC definition/rfc2863, 
in-discards packets are number of inbound packets that were chosen to be discarded even though no errors had been detected to prevent their being deliverable to a higher-layer protocol

& in-errors packets are he number of inbound packets that contained errors preventing them from being deliverable to a higher-layer protocol.

So, the for Packets with size > MTU should be increment the interface error counters.


arvbaska1:
> > So, the for Packets with size > MTU should be increment the interface error counters.
> 
> I wouldn't use the word "should" here, MTU != MRU (in a general case). It may be counted as an error if the interface MRU is lower than the frame size, but that's an implementation-specific detail.

Yes right , but in this test Packet size of MTU + 64 bytes will be > MTU | MRU in most implementations I think.

AmrNJ (approved):
LGTM

PR #2658, opened 2024-02-12
RT-5.5: Interface hold-time

mastarkey:
RT-5.5: Interface hold-time test cases

### DIFF
--- a/feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
+++ b/feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
@@ -541,7 +541,7 @@ func TestTC5ShortDOWN(t *testing.T) {
 		Down: ygot.Uint32(2000),
 	}
 
-	t.Run(fmt.Sprintf("Update hold timer configs down"), func(t *testing.T) {
+	t.Run("Update hold timer configs down"), func(t *testing.T) {
 		intfPath := gnmi.OC().Interface(dutPort1Intf.Name())
 		gnmi.Update(t, dut, intfPath.HoldTime().Config(), holdTimeConfig)
 	})
--- a/feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
+++ b/feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
@@ -104,7 +104,7 @@ func configureOTG(t *testing.T,
 	ateAggPorts := []*ondatra.Port{
 		ate.Port(t, "port1"),
 	}
-	configureOTGBundle(t, ate, top, ateAggPorts, aggID)
+	configureOTGBundle(t, top, ateAggPorts, aggID)
 
 	t.Log(top.String())
 	ate.OTG().PushConfig(t, top)
@@ -146,7 +146,7 @@ func OTGInterfaceDOWN(t *testing.T,
 }
 
 func configureOTGBundle(t *testing.T,
-	ate *ondatra.ATEDevice,
+
 	top gosnappi.Config,
 	aggPorts []*ondatra.Port,
 	aggID string) {
@@ -541,9 +541,10 @@ func TestTC5ShortDOWN(t *testing.T) {
 		Down: ygot.Uint32(2000),
 	}
 
-	t.Run("Update hold timer configs down"), func(t *testing.T) {
+	t.Run("Update hold timer configs down", func(t *testing.T) {
 		intfPath := gnmi.OC().Interface(dutPort1Intf.Name())
 		gnmi.Update(t, dut, intfPath.HoldTime().Config(), holdTimeConfig)
+
 	})
 
 	t.Run("Flap OTG Interfaces", func(t *testing.T) {
--- a/feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
+++ b/feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
@@ -1,0 +1,670 @@
+package holddown_times_test
+
+import (
+	"fmt"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/openconfig/ondatra/netutil"
+
+	"github.com/open-traffic-generator/snappi/gosnappi"
+	"github.com/openconfig/featureprofiles/internal/attrs"
+	"github.com/openconfig/featureprofiles/internal/deviations"
+	"github.com/openconfig/featureprofiles/internal/fptest"
+	"github.com/openconfig/ondatra"
+	"github.com/openconfig/ondatra/gnmi"
+	"github.com/openconfig/ondatra/gnmi/oc"
+	"github.com/openconfig/ygot/ygot"
+)
+
+var NTPServer = []string{"192.0.2.1", "192.0.2.2"}
+
+const (
+	ipv4PrefixLen = 30
+	ipv6PrefixLen = 126
+	lagName       = "LAGRx" // OTG LAG NAME
+	upTimer       = 5000
+	downTimer     = 300
+	toleranceMS   = 200 // Define the tolerance in milliseconds
+
+)
+
+var (
+	top          *gosnappi.Config
+	aggID        string
+	dutPort1Intf *ondatra.Port
+	ateSrc       = attrs.Attributes{
+		Name:    "ateSrc",
+		MAC:     "02:11:01:00:00:01",
+		IPv4:    "192.0.2.1",
+		IPv6:    "2001:db8::1",
+		IPv4Len: ipv4PrefixLen,
+		IPv6Len: ipv6PrefixLen,
+	}
+
+	dutSrc = attrs.Attributes{
+		Desc:    "DUT to ATE source",
+		IPv4:    "192.0.2.2",
+		IPv6:    "2001:db8::2",
+		IPv4Len: ipv4PrefixLen,
+		IPv6Len: ipv6PrefixLen,
+	}
+
+	dutDst = attrs.Attributes{
+		Desc:    "DUT to ATE destination",
+		IPv4:    "192.0.2.5",
+		IPv6:    "2001:db8::5",
+		IPv4Len: ipv4PrefixLen,
+		IPv6Len: ipv6PrefixLen,
+	}
+
+	ateDst = attrs.Attributes{
+		Name:    "ateDst",
+		MAC:     "02:12:01:00:00:01",
+		IPv4:    "192.0.2.6",
+		IPv6:    "2001:db8::6",
+		IPv4Len: ipv4PrefixLen,
+		IPv6Len: ipv6PrefixLen,
+	}
+)
+
+func TestMain(m *testing.M) {
+	fptest.RunTests(m)
+}
+
+func ConfigureNtpServer(t *testing.T) {
+	// Define the NTP server configuration for the default VRF
+	NtpConfig := struct {
+		description string
+		addresses   []string
+	}{
+		description: "NTP Server Config",
+		addresses:   NTPServer,
+	}
+
+	dut := ondatra.DUT(t, "dut")
+
+	// Begin configuring the NTP server
+	t.Run(NtpConfig.description, func(t *testing.T) {
+		ntpPath := gnmi.OC().System().Ntp()
+
+		d := &oc.Root{}
+		ntp := d.GetOrCreateSystem().GetOrCreateNtp()
+		ntp.SetEnabled(true)
+
+		// Configure each NTP server address
+		for _, address := range NtpConfig.addresses {
+			ntp.GetOrCreateServer(address)
+		}
+
+		// Replace the current NTP configuration with our new one
+		gnmi.Update(t, dut, ntpPath.Config(), ntp)
+
+		// Retrieve the NTP state to verify the configuration
+		ntpState := gnmi.Get(t, dut, ntpPath.State())
+
+		// Check that each NTP server is present in the state
+		for _, address := range NtpConfig.addresses {
+			ntpServer := ntpState.GetServer(address)
+			if ntpServer == nil {
+				t.Errorf("Missing NTP server from NTP state: %s", address)
+			}
+		}
+	})
+}
+
+// TestNtpServerConfigurability validates that an NTP server can be configured on the DUT.
+func configureDUTBundle(t *testing.T, dut *ondatra.DUTDevice, aggPorts []*ondatra.Port, aggID string) {
+	t.Helper()
+
+	agg := dutDst.NewOCInterface(aggID, dut)
+	agg.Type = oc.IETFInterfaces_InterfaceType_ieee8023adLag
+	agg.GetOrCreateAggregation().LagType = oc.IfAggregate_AggregationType_STATIC
+	gnmi.Replace(t, dut, gnmi.OC().Interface(aggID).Config(), agg)
+
+	for _, port := range aggPorts {
+		d := &oc.Root{}
+
+		i := d.GetOrCreateInterface(port.Name())
+		i.GetOrCreateEthernet().AggregateId = ygot.String(aggID)
+		i.Type = oc.IETFInterfaces_InterfaceType_ethernetCsmacd
+
+		if deviations.InterfaceEnabled(dut) {
+			i.Enabled = ygot.Bool(true)
+		}
+		gnmi.Replace(t, dut, gnmi.OC().Interface(port.Name()).Config(), i)
+	}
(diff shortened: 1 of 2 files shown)


feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -0,0 +1,670 @@
...
+		IPv4Len: ipv4PrefixLen,
+		IPv6Len: ipv6PrefixLen,
+	}
+)
+
+func TestMain(m *testing.M) {
+	fptest.RunTests(m)
+}
+
+func ConfigureNtpServer(t *testing.T) {
```
LimeHat, line 76: imo this should not be included here. you can assume that NTP is configured in the base config used for the test, just like any other part of the mgmt infra (e.g. gnmi with openconfig)

AmrNJ (changes requested):
Below 4 tests fails for 8808. 

TestTC2LongDown/Calculate_fault_duration_on_Bundle-Ether3
TestTC2LongDown
TestTC5ShortDOWN/Verify_port_status_UP
TestTC5ShortDOWN

nraghavendran (changes requested):
pls fix the failures shown

AmrNJ (approved):
LGTM

AmrNJ (approved):
LGTM

AmrNJ:
@dplore The code LGTM.

dplore:
There is a small issue here still as noted by the github Rundata Check:

```
Found 2 errors in feature/interface/holdtime/otg_tests/holdtime_test
  - metadata test description needs update: was "Interface hold-times", will be "Interface hold-time"
  - missing UUID in metadata
Skipping check for duplicate test UUID due to missing value: /home/runner/work/featureprofiles/featureprofiles/feature/interface/holdtime/otg_tests/holdtime_test
F0305 03:21:20.170855    2[9](https://github.com/openconfig/featureprofiles/actions/runs/8150553462/job/22277006874?pr=2658#step:3:10)90 main.go:[10](https://github.com/openconfig/featureprofiles/actions/runs/8150553462/job/22277006874?pr=2658#step:3:11)0] Rundata check found problems.  Please run: go run ./tools/addrundata --fix
```

The author @mastarkey can fix this.  Alternatively, @AmrNJ you could fix this, but you will need to install the [github commandline tool](https://cli.github.com/) and run the following to check out this PR into your own machine and make the changes:
`gh pr checkout 2658`

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -0,0 +1,670 @@
...
+	ipv6PrefixLen = 126
+	lagName       = "LAGRx" // OTG LAG NAME
+	upTimer       = 5000
+	downTimer     = 300
+	toleranceMS   = 200 // Define the tolerance in milliseconds
+
+)
+
+var (
+	top          *gosnappi.Config
```
dplore, line 34: Remove/Fix static analysis issue.  (This global scope variable `top` is not used).

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
dplore: agreed, this should be removed.  We should keep the test to the minimum to test the goal of the feature, which in this case is just the hold time.  In any case, holdtime is not dependent on having a reasonable time of day as hold time is only dependent on a time duration.

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -0,0 +1,670 @@
...
+			ntpServer := ntpState.GetServer(address)
+			if ntpServer == nil {
+				t.Errorf("Missing NTP server from NTP state: %s", address)
+			}
+		}
+	})
+}
+
+// TestNtpServerConfigurability validates that an NTP server can be configured on the DUT.
+func configureDUTBundle(t *testing.T, dut *ondatra.DUTDevice, aggPorts []*ondatra.Port, aggID string) {
```
dplore, line 118: I suggest moving this function to a new file in the /internal/cfgplugins folder (such as interface_aggregate.go).  This allows this common configuration to be reused, including embedding any necessary deviations in one place in the codebase.

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -0,0 +1,670 @@
...
+	dut *ondatra.DUTDevice,
+	actionState string) (time.Time, time.Time, string, string) {
+
+	// Shut down OTG Interface
+	p1 := ondatra.ATE(t, "ate").Port(t, "port1")
+	portStateAction := gosnappi.NewControlState()
+
+	var otgStateChangeTsStr string
+
+	// TC2 Step 1 Read timestamp of last oper-status change  form DUT port-1
```
dplore, line 270: We should not need to trust the clock on the DUT (which is triggering the requirement for NTP on the DUT).  Instead maintain a set of timestamps derived from time.Now() 

- subscribe to DUT interface operational state
- set OTG interface down and get local time stamp
-  when DUT interface operational state changes, set timestamp
- compare time DUT operational state changes to OTG interface down timestamp 

This may be less precise than using the DUT timestamp because the streaming event will have some latency, but it should be sufficent precision for this test.

dplore: If this poses a major issue (precision is very bad, off more than 5 seconds?) alternatively, we can assert that the DUT should be 'preconfigured' with NTP to a timesource which is syncronized with the test runner in the readme.

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -0,0 +1,670 @@
...
+	t.Logf("Step1. DutLastChangeTS1 is: %v", DutLastChangeTS1)
+	if actionState == "UP" {
+		portStateAction.Port().Link().SetPortNames([]string{p1.ID()}).SetState(gosnappi.StatePortLinkState.UP)
+		otgStateChangeTsStr = gnmi.Get(t, dut, gnmi.OC().System().CurrentDatetime().State())
+		ate.OTG().SetControlState(t, portStateAction)
+	} else if actionState == "DOWN" {
+		// TC2 Step 2 Bring Down OTG Interface
+		t.Log("Step 2 sending remote fault")
+		portStateAction.Port().Link().SetPortNames([]string{p1.ID()}).SetState(gosnappi.StatePortLinkState.DOWN)
+		//otgStateChangeTsStr = gnmi.Get(t, dut, gnmi.OC().System().CurrentDatetime().State())
```
dplore, line 282: remove commented code

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -0,0 +1,670 @@
...
+	preStateTSSTR := gnmi.Get(t, dut, gnmi.OC().Interface(aggID).LastChange().State())
+	DutLastChangeTS1 := time.Unix(0, int64(preStateTSSTR)).UTC().Format(time.RFC3339Nano)
+	t.Logf("Step1. DutLastChangeTS1 is: %v", DutLastChangeTS1)
+	if actionState == "UP" {
+		portStateAction.Port().Link().SetPortNames([]string{p1.ID()}).SetState(gosnappi.StatePortLinkState.UP)
+		otgStateChangeTsStr = gnmi.Get(t, dut, gnmi.OC().System().CurrentDatetime().State())
+		ate.OTG().SetControlState(t, portStateAction)
+	} else if actionState == "DOWN" {
+		// TC2 Step 2 Bring Down OTG Interface
+		t.Log("Step 2 sending remote fault")
```
dplore, line 280:
```suggestion
		t.Log("RT-5.5.2: Bring Down OTG Interface")
```

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -0,0 +1,670 @@
...
+		portStateAction.Port().Link().SetPortNames([]string{p1.ID()}).SetState(gosnappi.StatePortLinkState.DOWN)
+		//otgStateChangeTsStr = gnmi.Get(t, dut, gnmi.OC().System().CurrentDatetime().State())
+		otgStateChangeTsStr = time.Now().UTC().Format(time.RFC3339Nano)
+
+		ate.OTG().SetControlState(t, portStateAction)
+
+		// TC2 Step 3
+		t.Log("Step 3 sleeping 1000ms")
+		time.Sleep(1000 * time.Millisecond)
```
dplore, line 289: Sleep timers can introduce test flakiness.   What is the purpose of this sleep?  Can we change this to a event based trigger? For example, block on gnmi.Await for telemetry from the DUT port state change?

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -0,0 +1,670 @@
...
+
+	// Step 5. verify oper-status is DOWN
+	if DutLastChangeOper2 != expectedStatus {
+		t.Errorf("Interface %s status got %v, want %v", aggID, DutLastChangeTS2STR, expectedStatus.String())
+	} else {
+		t.Logf("Interface %s status got %v, want %v", aggID, DutLastChangeTS2STR, expectedStatus.String())
+	}
+
+	// convert string type change to time.time
+	otgStateChangeTs, err := time.Parse(time.RFC3339Nano, otgStateChangeTsStr)
```
dplore, line 313: Should check for err state

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -0,0 +1,670 @@
...
+		t.Logf("Pass: %s Status: got %v, want %v", aggID, status, want)
+	}
+}
+
+func TestHoldTimeConfig(t *testing.T) {
+
+	dut := ondatra.DUT(t, "dut")
+	ate := ondatra.ATE(t, "ate")
+
+	t.Run(fmt.Sprintf("Configure NTP Server on DUT"), func(t *testing.T) {
```
dplore, line 364: Here and throughout the rest of the code, please fix static check `unnecessary use of fmt.Sprintf (S1039)`

```suggestion
	t.Run("RT-5.5.1: Configure NTP Server on DUT", func(t *testing.T) {
```

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -0,0 +1,670 @@
...
+		t.Logf("Configure ATE")
+		configureOTG(t, ate, aggID)
+
+	})
+
+	t.Run(fmt.Sprintf("Verify Interface State for %s", aggID), func(t *testing.T) {
+		// Verify Port Status
+		t.Logf("Verifying port status for %s", aggID)
+		time.Sleep(45 * time.Second)
+		verifyPortsStatus(t, dut, "UP")
```
dplore, line 421: change this to a blocking gnmi.Await for port status (with a 45 second timeout)

nraghavendran (approved):

nraghavendran (commented):
good to go , as long as review comments from Darren & Amritha are fixed.
thanks

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
mastarkey: This makes sense but on the readme it mentions as one of the checks regarding when the interface is either triggered up or down to check oper-status last change on the DUT that is why I was including the NTP as part because if it was just a matter of checking the time before and after the trigger then the time.now() makes sense however the readme wanted to include the check of the interface last change which will have the router timestamp and to make all time stamps correct with the same timezone and other things we needed NTP. Im fine to remove NTP but with check oper-status  on interface of the router there there is a high likelihood of the timestamps being on different zones or just out of sync.

dplore: Agreed that comparing the DUT time with an OTG time requires both the OTG and DUT to have a time synced within 300msec precisionof each other for this test to perform as intended.  

I think the README should be changed to match the proposed timestamp method (using the local test runner  timestamp for response of OTG port change and the local test runner timestamp of receipt of gNMI subscribe from DUT response as the estimate of when the DUT changed it's state.  

I agree that this approach is theoretically less precise than using NTP.  But NTP could be wrong or flaky too.  FOr example, if the device clock isn't synced with the OTG before the test starts supports, adding NTP to the DUT may still take several minutes or longer for the DUT to skew it's clock to the correct time.  (The DUT may or may not support an NTP step / burst update).  There is also some change (maybe in practice a non-issue) that the OTG clock is not synced.

Using timestamps from a single source (the test runner) eliminates these variables.  Even the test runner doesn't have to be synced for precise enough timing.

If the timing using this approach isn't reliable, we could just require that the DUT and OTG are both pre-configured to use NTP and switch back to getting timestamp from the DUT.

dplore:
@mastarkey please address the static check errors found at: https://github.com/openconfig/featureprofiles/pull/2658/files

nraghavendran (approved):

AmrNJ:
@mastarkey I do not have permission to fix and push changes in the forked (https://github.com/b4firex/featureprofiles) which belongs to the author. Can you please make the changes that Darren suggested to merge?

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -250,12 +249,11 @@ func flapOTGInterface(t *testing.T,
 		t.Log("RT-5.5.2: Bring Down OTG Interface")
 		portStateAction.Port().Link().SetPortNames([]string{p1.ID()}).SetState(gosnappi.StatePortLinkState.DOWN)
 		otgStateChangeTsStr = gnmi.Get(t, dut, gnmi.OC().System().CurrentDatetime().State())
-
 		ate.OTG().SetControlState(t, portStateAction)
 
 		// TC2 Step 3
 		t.Log("Step 3 sleeping 1000ms")
```
dplore, line 255:
```suggestion
		t.Log("Step 3 sleeping 500ms")
```

dplore (approved):
- One small comment on the log message 
- Recommend updating the readme to say: "A pre-requisite of this test is the clocks for DUT and ATE must be synchronized."

- otherwise LGTM

dplore (approved):

dplore:
@mastarkey there is a new static analysis failure to check out.

dplore:
@mastarkey there is a [static check failing](https://github.com/openconfig/featureprofiles/actions/runs/8351921937/job/22861150682?pr=2658):

```
Run GOGC=30 go vet ./...
  GOGC=30 go vet ./...
  shell: /usr/bin/bash -e {0}
  env:
    GOROOT: /opt/hostedtoolcache/go/1.[2](https://github.com/openconfig/featureprofiles/actions/runs/8351921937/job/22861150682?pr=2658#step:7:2)1.8/x6[4](https://github.com/openconfig/featureprofiles/actions/runs/8351921937/job/22861150682?pr=2658#step:7:4)
Error: feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go:[5](https://github.com/openconfig/featureprofiles/actions/runs/8351921937/job/22861150682?pr=2658#step:7:5)44:2: expected 1 expression
Error: Process completed with exit code 1.
```

dplore:
Thanks for the updates.  I am running this now in my local environment to test

dplore:
This code runs, I get a partial pass on Cisco 8202.  Still the code and logic LGTM.  The parameters could be tuned in future PR as appropriate.

```
=== RUN   TestTC2LongDown/Verify_test_results
    third_party/openconfig/featureprofiles/feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go:225: 
        Pre-action TS       : 2024-03-27T19:34:08.762Z
        Post-action TS      : 2024-03-27T19:34:09.807112017Z
        Expected Duration   : 300ms
        Actual Duration     : 1045ms
        Min Tolerance       : 100ms
        Max Tolerance       : 500ms
        Expected Oper Status: DOWN
        Actual Oper Status  : DOWN
        Result              : FAIL
```

dplore (approved):

PR #2665, opened 2024-02-15
[TE-16.2] Include backup NHG for REPAIRED tunnels

nachikethas:
The backup next hop group that specifies the fallback to default is currently missing from the gRIBI that specifies the encap and frr actions.


mohanasm (approved):
Looks good.Thanks for updating.

xw-g (approved):

PR #2671, opened 2024-02-16
DP-1.3 Adding required scheduler config before binding it to an Interface

charantejag504:
1. In DP-1.3, Adding required scheduler config before binding it to an Interface.
2. Removing deviations state_path_unsupported and drop_weight_leaves_unsupported.
3. Updated README to match the new README file conventions.


Swetha-haridasula (approved):
lgtm

PR #2672, opened 2024-02-16
TRANSCEIVER-8: Telemetry: 400ZR Optics module temperature streaming.

sunilprgit:


feature/platform/transceiver/otg_test/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,145 @@
...
+	t.Logf("Port1 dut1 %s Max Temperature: %v", dp1.Name(), temperatureMax)
+	if reflect.TypeOf(temperatureMax).Kind() != reflect.Float64 {
+		t.Fatalf("Return value is not type float64")
+	}
+	temperatureAvg := gnmi.Get(t, dut1, component1.Temperature().Avg().State())
+	if reflect.TypeOf(temperatureAvg).Kind() != reflect.Float64 {
+		t.Fatalf("Return value is not type float64")
+	}
+	t.Logf("Port1 dut1 %s Avg Temperature: %v", dp1.Name(), temperatureAvg)
+	// temperature average should be between min and max
```
ejbrever, line 69: what do you think about adding a quick check that the values look reasonable. These are in Celsius, so perhaps they should all be above 0 and below 300 degrees as extremes?

feature/platform/transceiver/otg_test/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,145 @@
...
+func TestZRTemperatureState(t *testing.T) {
+	dut1 := ondatra.DUT(t, "dut1")
+	dp1 := dut1.Port(t, "port1")
+	t.Logf("dut1: %v", dut1)
+	t.Logf("dut1 dp1 name: %v", dp1.Name())
+	// Check if TRANSCEIVER is of type 400ZR
+	if dp1.PMD() != ondatra.PMD400GBASEZR {
+		t.Fatalf("%s Transceiver is not 400ZR its of type: %v", dp1, dp1.PMD())
+	}
+	component1 := gnmi.OC().Component(dp1.Name())
```
LimeHat, line 47: transceiver component name may not match the interface name  
It should be looked up using the interface state query, e.g. https://github.com/openconfig/featureprofiles/blob/a889483110c5e6e7371b57c8313c4a2d2cbd7db6/feature/platform/tests/optics_power_and_bias_current_test/optics_power_and_bias_current_test.go#L228

feature/platform/transceiver/otg_test/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,145 @@
...
+	fptest.RunTests(m)
+}
+
+// Topology:
+//   dut1:port1 <--> port1:dut2
+//
+
+func TestZRTemperatureState(t *testing.T) {
+	dut1 := ondatra.DUT(t, "dut1")
+	dp1 := dut1.Port(t, "port1")
```
LimeHat, line 40: The interface is configured in the second test case, but not in this one; it should be done in both (unless you're relying on the specific test ordering, assuming that  TestZRTemperatureStateInterfaceFlap will be executed first)

feature/platform/transceiver/otg_test/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,145 @@
...
+	t.Logf("dut1: %v", dut1)
+	t.Logf("dut1 dp1 name: %v", dp1.Name())
+	// Check if TRANSCEIVER is of type 400ZR
+	if dp1.PMD() != ondatra.PMD400GBASEZR {
+		t.Fatalf("%s Transceiver is not 400ZR its of type: %v", dp1, dp1.PMD())
+	}
+	component1 := gnmi.OC().Component(dp1.Name())
+	temperatureInstantdp1 := gnmi.Get(t, dut1, component1.Temperature().Instant().State())
```
LimeHat, line 48: throughout the test, `gnmi.Get` is used (which does a sub once query), however, based on the description, I believe this should be a `SAMPLE` subscription instead, e.g.
https://github.com/openconfig/featureprofiles/blob/a889483110c5e6e7371b57c8313c4a2d2cbd7db6/feature/platform/tests/optics_power_and_bias_current_test/optics_power_and_bias_current_test.go#L89

LimeHat, line 48: temperature can be reported in one of the two ways in components tree:
1) in the transceiver hierarchy, or
2) the transceiver may have a temperature sensor subcomponent that reports the values (instant/min/max/avg/..)
e.g. 
https://github.com/openconfig/featureprofiles/blob/dd9eed3fb05f68ed1ae7aa0ce0fe027c2d638cf2/feature/platform/tests/optics_power_and_bias_current_test/optics_power_and_bias_current_test.go#L106-L115

LimeHat (changes requested):

feature/platform/transceiver/otg_test/zr_temperature_test/zr_temperature_test.go
sunilprgit: Hi Eric,

Added check what you suggested kindly review.

Thanks

feature/platform/transceiver/otg_test/zr_temperature_test/zr_temperature_test.go
sunilprgit: Hi Sergey,

Updated the script as suggested please review and let me know any further changes required.

Thanks

feature/platform/transceiver/otg_test/zr_temperature_test/zr_temperature_test.go
sunilprgit: Hi Sergey,

Updated the script as suggested please review and let me know any further changes required.

Thanks

feature/platform/transceiver/otg_test/zr_temperature_test/zr_temperature_test.go
sunilprgit: Hi Sergey,

Updated the script as suggested please review and let me know any further changes required.

Thanks

feature/platform/transceiver/otg_test/zr_temperature_test/zr_temperature_test.go
sunilprgit: Hi Sergey,

Updated the script as suggested please review and let me know any further changes required.

Thanks

ejbrever: I was curious about this as well. The intention is to open a single subscription channel at the start of the test and monitor that the device streams changes over that channel as different events occur (i.e. config changes). Is that what happens? I wasn't sure if the test currently is initiating a new subscription channel every time it reads data, which essentially just mimics the previous Get() behavior. Thanks.

feature/platform/transceiver/otg_test/zr_temperature_test/zr_temperature_test.go
ejbrever: Thanks.

feature/platform/transceiver/otg_test/zr_temperature_test/zr_temperature_test.go
sunilprgit: Hi Eric, based on previous comment have removed the Get call and using sample subscription as suggested, please review the latest script

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,275 @@
...
+func gnmiOpts(t *testing.T, dut *ondatra.DUTDevice, mode gpb.SubscriptionMode, interval time.Duration) *gnmi.Opts {
+	return dut.GNMIOpts().WithYGNMIOpts(ygnmi.WithSubscriptionMode(mode), ygnmi.WithSampleInterval(interval))
+}
+
+func TestZRTemperatureState(t *testing.T) {
+	dut1 := ondatra.DUT(t, "dut1")
+	dp1 := dut1.Port(t, "port1")
+	t.Logf("dut1: %v", dut1)
+	t.Logf("dut1 dp1 name: %v", dp1.Name())
+	transceiverName := gnmi.Get(t, dut1, gnmi.OC().Interface(dp1.Name()).Transceiver().State())
```
LimeHat, line 51: Better to move this lookup to a line after the interface config
https://github.com/openconfig/featureprofiles/blob/3a0e1ed449b17e6e65f8951688b968b9e888461d/feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go#L62

(not all implementations return state for unconfigured interfaces)

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,275 @@
...
+	} else {
+		t.Fatalf("The average is not between the maximum and minimum values")
+	}
+	subcomponents := gnmi.LookupAll[*oc.Component_Subcomponent](t, dut1, component1.SubcomponentAny().State())
+	for _, s := range subcomponents {
+		subc, ok := s.Val()
+		if ok {
+			sensorComponent := gnmi.Get[*oc.Component](t, dut1, gnmi.OC().Component(subc.GetName()).State())
+			if sensorComponent.GetType() == sensorType {
+				scomponent := gnmi.OC().Component(sensorComponent.GetName())
```
LimeHat, line 129: Thanks for the changes, Sunil! 
I'd suggest one further change to this. You don't need to separate lookups for the subcomponent temperature paths and the transceiver temperature. 

I think the logic you want is:
1) lookup the transceiver name (which you already added at https://github.com/openconfig/featureprofiles/blob/3a0e1ed449b17e6e65f8951688b968b9e888461d/feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go#L51) and component path
https://github.com/openconfig/featureprofiles/blob/3a0e1ed449b17e6e65f8951688b968b9e888461d/feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go#L56
2) lookup if there's a temperature `SENSOR` subcomponent for the component, and record its name if its present
3) if p. 2 returned a non-nil sensor subcomponent, perform all temperature lookups on the subcomponent instead of the transceiver component.

I.e. what you want is just to move this lookup piece to the top, after you do L56, and depending on the lookup result (separate temp. sensor can be present or absent) you set the `component1` value as the transceiver path or the sensor path.

Let me know if that makes sense?

LimeHat (changes requested):

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: Hi Sergey,

Updated the script as suggested please review and update.

Thanks

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: Hi Sergey,

Updated the script as suggested please review and update.

Thanks

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,226 @@
...
+	OCcomponent := opticalChannelFromPort(t, dut1, dp)
+	gnmi.Replace(t, dut1, gnmi.OC().Component(OCcomponent).OpticalChannel().Config(), &oc.Component_OpticalChannel{
+		TargetOutputPower: ygot.Float64(targetOutputPower),
+		Frequency:         ygot.Uint64(frequency),
+	})
+}
+func verifyTemperatureSensorValue(t *testing.T, dut1 *ondatra.DUTDevice, pStream *samplestream.SampleStream[float64], sensorName string) float64 {
+	temperatureSample := pStream.Next(t)
+	if temperatureSample == nil {
+		t.Fatalf("Temperature telemetry %s is not present", sensorName)
```
ejbrever, line 65: would it be clearer to say:

"Temperature telemetry %s was not streamed in the most recent subscription interval"

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,226 @@
...
+	dut1 := ondatra.DUT(t, "dut")
+	dp1 := dut1.Port(t, "port1")
+	dp2 := dut1.Port(t, "port2")
+	t.Logf("dut1: %v", dut1)
+	t.Logf("dut1 dp1 name: %v", dp1.Name())
+	intUpdateTime := 2 * time.Minute
+	interfaceConfig(t, dut1, dp1)
+	interfaceConfig(t, dut1, dp2)
+	gnmi.Await(t, dut1, gnmi.OC().Interface(dp1.Name()).OperStatus().State(), intUpdateTime, oc.Interface_OperStatus_UP)
+	transceiverName := gnmi.Get(t, dut1, gnmi.OC().Interface(dp1.Name()).Transceiver().State())
```
ejbrever, line 91: is transceiverName really all of the transceiver state and not the name? If so, perhaps this var could be "transceiverState" ?

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,226 @@
...
+	temperatureMax := verifyTemperatureSensorValue(t, dut1, p1StreamMax, "Max")
+	t.Logf("Port1 dut1 %s Max Temperature: %v", dp1.Name(), temperatureMax)
+	temperatureMin := verifyTemperatureSensorValue(t, dut1, p1StreamMin, "Min")
+	t.Logf("Port1 dut1 %s Min Temperature: %v", dp1.Name(), temperatureMin)
+	temperatureAvg := verifyTemperatureSensorValue(t, dut1, p1StreamAvg, "Avg")
+	t.Logf("Port1 dut1 %s Avg Temperature: %v", dp1.Name(), temperatureAvg)
+	if temperatureAvg >= temperatureMin && temperatureAvg <= temperatureMax {
+		t.Logf("The average is between the maximum and minimum values")
+	} else {
+		t.Fatalf("The average is not between the maximum and minimum values")
```
ejbrever, line 125: nit, could you also include the avg, min, and max vals in the t.Fatalf message?

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,226 @@
...
+	// Check if TRANSCEIVER is of type 400ZR
+	if dp1.PMD() != ondatra.PMD400GBASEZR {
+		t.Fatalf("%s Transceiver is not 400ZR its of type: %v", transceiverName, dp1.PMD())
+	}
+	component1 := gnmi.OC().Component(transceiverName)
+	subcomponents := gnmi.LookupAll[*oc.Component_Subcomponent](t, dut1, component1.SubcomponentAny().State())
+	for _, s := range subcomponents {
+		subc, ok := s.Val()
+		if ok {
+			sensorComponent := gnmi.Get[*oc.Component](t, dut1, gnmi.OC().Component(subc.GetName()).State())
```
ejbrever, line 101: IIUC, we don't know "sensorComponent" is the sensor component yet? I think this is discovered using the if statement below.

Perhaps you could change this to just "component" or "possibleSensorComponent" and then you can change "scomponent" --> "sensorComponent" below which would help clarify these variables?

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,226 @@
...
+		t.Fatalf("%s Transceiver is not 400ZR its of type: %v", transceiverName, dp1.PMD())
+	}
+	// Disable interface
+	d := &oc.Root{}
+	i := d.GetOrCreateInterface(dp1.Name())
+	i.Enabled = ygot.Bool(false)
+	gnmi.Replace(t, dut1, gnmi.OC().Interface(dp1.Name()).Config(), i)
+	component1 := gnmi.OC().Component(transceiverName)
+	subcomponents := gnmi.LookupAll[*oc.Component_Subcomponent](t, dut1, component1.SubcomponentAny().State())
+	for _, s := range subcomponents {
```
ejbrever, line 154: Similar to comment above, it would be great to update some var names here as there are some very similar naming here.

Also, if you want this appears to be an identical for loop to what is above. You might want to move into a separate func for reuse. It might also help readability with a func like:

func sensorComponent(t ..., dut1 ..., component1 ...) (...) {
  ...
}

Optional, so will leave that to you to decide if it's helpful.

ejbrever (approved):
Just some nits, but LGTM

ejbrever (approved):

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,225 @@
...
+	temperatureAvg = verifyTemperatureSensorValue(t, dut1, p1StreamAvg, "Avg")
+	t.Logf("Port1 dut1 %s Avg Temperature: %v", dp1.Name(), temperatureAvg)
+	if temperatureAvg >= temperatureMin && temperatureAvg <= temperatureMax {
+		t.Logf("The average is between the maximum and minimum values")
+	} else {
+		t.Fatalf("The average is not between the maximum and minimum values")
+	}
+}
+
+func opticalChannelFromPort(t *testing.T, dut *ondatra.DUTDevice, p *ondatra.Port) string {
```
ahsaanyousaf, line 200: should this name be componentFromPort?
out of curiosity did you run it against one of the FNT testbeds? What component name was returned for 400ZR? Something like Ethernet25-Optical0?

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,225 @@
...
+//
+//	dut:port1 <--> port2:dut
+
+func interfaceConfig(t *testing.T, dut1 *ondatra.DUTDevice, dp *ondatra.Port) {
+	d := &oc.Root{}
+	i := d.GetOrCreateInterface(dp.Name())
+	i.Enabled = ygot.Bool(true)
+	i.Type = oc.IETFInterfaces_InterfaceType_ethernetCsmacd
+	gnmi.Replace(t, dut1, gnmi.OC().Interface(dp.Name()).Config(), i)
+	// transceiverName := gnmi.Get(t, dut1, gnmi.OC().Interface(dp.Name()).Transceiver().State())
```
ahsaanyousaf, line 53: should these lines be deleted if not relevant?

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,225 @@
...
+	if temperatureSample == nil {
+		t.Fatalf("Temperature telemetry %s is not present", sensorName)
+	}
+	temperatureVal, ok := temperatureSample.Val()
+	if !ok {
+		t.Fatalf("Temperature %q telemetry is not present", temperatureSample)
+	}
+	// Check temperature return value of correct type
+	if reflect.TypeOf(temperatureVal).Kind() != reflect.Float64 {
+		t.Fatalf("Return value is not type float64")
```
ahsaanyousaf, line 72: nit: Return value is not of type decimal64.

Please update to decimal64 throughout.

https://screenshot.googleplex.com/9DYwkpuNmkv8sGG

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,225 @@
...
+	} else if temperatureVal <= 0 && temperatureVal >= 300 {
```
ahsaanyousaf, line 73: where is this range driven from? For checking the range we should use a more stricter value of between 0 and 75 degree celsius.

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -0,0 +1,225 @@
...
+		t.Fatalf("The average is not between the maximum and minimum values")
+	}
+}
+
+func opticalChannelFromPort(t *testing.T, dut *ondatra.DUTDevice, p *ondatra.Port) string {
+	t.Helper()
+	if deviations.MissingPortToOpticalChannelMapping(dut) {
+		switch dut.Vendor() {
+		case ondatra.ARISTA:
+			transceiverName := gnmi.Get(t, dut, gnmi.OC().Interface(p.Name()).Transceiver().State())
```
ahsaanyousaf, line 205: same as above is transceiverName really all of the transceiver state and not the name? If so, this should change to transeiverState?

ahsaanyousaf (commented):
Thanks looks good, added some comments

ahsaanyousaf (dismissed):

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: Corrected the name as suggested opticalChannelComponentFromPort

thanks

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: In ondatra we dont have Decimal64, it represent float64 for 64byte decimal values.
Thanks

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: deleted

Thanks

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: its given from earlier review comment from eric.

Thanks

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: hi Ashsaan,

This return transceiver component name same code snippet suggested in review comment from LimeHat.

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: Updated

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: This return transceiver component name same code snippet suggested in review comment from LimeHat.

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: Hi Eric,

This is suggested in earlier comments from LIMEHAT.

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: Hi Eric,

This is suggested in earlier comments from LIMEHAT.

feature/platform/transceiver/otg_tests/zr_temperature_test/zr_temperature_test.go
sunilprgit: updated

ejbrever (approved):

ahsaanyousaf (approved):

PR #2676, opened 2024-02-17
Add `path_presence_test` tag in metadata.proto.

wenovus:
This will mark tests that only test for path existence/validity such as https://github.com/openconfig/featureprofiles/tree/main/feature/gnmi/otg_tests/telemetry_basic_check_test so that tools (e.g. coverage tools) can have better insight on what the coverage means.

----

Emum is another choice for the data type but I'm not sure if it makes sense to have coverage types other than these two -- too fine grained seems too hard to track so I'd like to keep this as simple as possible.


dplore (dismissed):

wenovus:
@dplore Need another approval after merge.

dplore (approved):

PR #2694, opened 2024-02-21
TRANSCEIVER-9 Telemetry: 400ZR TX laser bias current telemetry values streaming.

sunilprgit:
Test script for test case TRANSCEIVER-9 Telemetry: 400ZR TX laser bias current telemetry values streaming.


feature/platform/transceiver/otg_tests/zr_laser_bias_current_test/zr_laser_bias_current_test.go
```diff
@@ -0,0 +1,219 @@
...
+	defer p1StreamInstant.Close()
+	p1StreamMin := samplestream.New(t, dut1, component1.OpticalChannel().LaserBiasCurrent().Min().State(), 10*time.Second)
+	verifyLaserBiasCurrent(t, dut1, p1StreamMin, "laserbiasMin")
+	defer p1StreamMin.Close()
+	p1StreamMax := samplestream.New(t, dut1, component1.OpticalChannel().LaserBiasCurrent().Max().State(), 10*time.Second)
+	verifyLaserBiasCurrent(t, dut1, p1StreamMax, "laserbiasMax")
+	defer p1StreamMax.Close()
+	p1StreamAvg := samplestream.New(t, dut1, component1.OpticalChannel().LaserBiasCurrent().Avg().State(), 10*time.Second)
+	verifyLaserBiasCurrent(t, dut1, p1StreamAvg, "laserbiasAvg")
+	defer p1StreamAvg.Close()
```
ejbrever, line 102: can you also do a comparison to make sure:

min <= avg <= max

It seems that everywhere you run verifyLaserBiasCurrent() it does this four times. So you could instead create a wrapper func that checks all four items and additionally does the check mentioned in this comment. That should remove a fair amount of similar code.

feature/platform/transceiver/otg_tests/zr_laser_bias_current_test/zr_laser_bias_current_test.go
```diff
@@ -0,0 +1,219 @@
...
+	compName := gnmi.Get(t, dut, gnmi.OC().Interface(p.Name()).HardwarePort().State())
+	for {
+		comp, ok := gnmi.Lookup(t, dut, gnmi.OC().Component(compName).State()).Val()
+		if !ok {
+			t.Fatalf("Recursive optical channel lookup failed for port: %s, component %s not found.", p.Name(), compName)
+		}
+		if comp.GetType() == oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_OPTICAL_CHANNEL {
+			return compName
+		}
+		if comp.GetParent() == "" {
```
ejbrever, line 214: If I follow this correctly, this would check that components that are not optical-channels have a parent. Is this the intent? Perhaps this if statement belongs higher up or as part of the previous if statement?

feature/platform/transceiver/otg_tests/zr_laser_bias_current_test/zr_laser_bias_current_test.go
sunilprgit: Hi Eric,

Added the check and created wrapper function to reduce additional call as suggested,

Thanks

feature/platform/transceiver/otg_tests/zr_laser_bias_current_test/zr_laser_bias_current_test.go
sunilprgit: Hi Eric,

As if now we don't have clear mapping from port to opticalchannelcomponent.
we have added a deviation currently.
The above code snippet is to search for any optical channel component recursively for the port.
In the mentioned if condition if it get the component type optical channel it returns it.
if not then it will check if it has parent component.

ejbrever: Ah, I see, it recursively goes up the parents. Thanks for explaining.

feature/platform/transceiver/tests/zr_laser_bias_current_test/zr_laser_bias_current_test.go
```diff
@@ -0,0 +1,218 @@
...
+		t.Fatalf("%s Transceiver is not 400ZR its of type: %v", transceiverState, dp1.PMD())
+	}
+	OCcomponent := opticalChannelComponentFromPort(t, dut1, dp1)
+	component1 := gnmi.OC().Component(OCcomponent)
+	p1StreamInstant := samplestream.New(t, dut1, component1.OpticalChannel().LaserBiasCurrent().Instant().State(), 10*time.Second)
+	p1StreamMin := samplestream.New(t, dut1, component1.OpticalChannel().LaserBiasCurrent().Min().State(), 10*time.Second)
+	p1StreamMax := samplestream.New(t, dut1, component1.OpticalChannel().LaserBiasCurrent().Max().State(), 10*time.Second)
+	p1StreamAvg := samplestream.New(t, dut1, component1.OpticalChannel().LaserBiasCurrent().Avg().State(), 10*time.Second)
+	verifyLaserBiasCurrentAll(t, dut1, p1StreamInstant, p1StreamAvg, p1StreamMax, p1StreamMin)
+	defer p1StreamAvg.Close()
```
ejbrever, line 113: I think you'll want to move the defer statements directly underneath where each stream is initialized. Same for throughout the file.

ejbrever (approved):
One nit, but LGTM

feature/platform/transceiver/tests/zr_laser_bias_current_test/zr_laser_bias_current_test.go
sunilprgit: updated

self-maurya (approved):

PR #2702, opened 2024-02-22
updating ip add check script

cprabha:


tools/check_ip_addresses.pl
```diff
@@ -44,9 +44,9 @@ END
   next if $ip =~ /198\.51\.100\./;      # TEST-NET-2 (RFC 5737)
   next if $ip =~ /203\.0\.113\./;       # TEST-NET-3 (RFC 5737)
   next if $ip =~ /198\.(18|19)\./;      # BMWG (RFC 2544)
-  next if $ip =~ /20\.0\./;             # 20.0.0.1/15
-  next if $ip =~ /30\.0\./;             # 30.0.0.1/15
-  next if $ip =~ /100\.0\./;            # 100.0.0.1/12
+  next if $ip =~ /20\./;              # 20.0.0.1/15
+  next if $ip =~ /30\./;              # 30.0.0.1/15
```
xw-g, line 48: ain't you change the allowed subnet from `/15` to `/16`?

cprabha: I have used /15 in 14.2, but its giving error. It looks like pattern matching is looking for 20.0  ? Hence thought of changing it. Please let me know any other suggestions.

IPBlockEncapA           = "20.1.64.1/15"  
	IPBlockEncapB           = "20.5.64.1/15" 
	IPBlockEncapC           = "20.10.64.1/15"
	IPBlockEncapD           = "20.15.64.1/15"

xw-g: @cprabha  range of `20.0/15` only contains `20.0.0.1 - 20.1.255.254`, right? 

https://github.com/openconfig/featureprofiles/blob/da4eaf8d4609fb279b84a1eae6ed63fc07c9795b/CONTRIBUTING.md?plain=1#L363

cprabha: > @cprabha range of `20.0/15` only contains `20.0.0.1 - 20.1.255.254`, right?
> 
> https://github.com/openconfig/featureprofiles/blob/da4eaf8d4609fb279b84a1eae6ed63fc07c9795b/CONTRIBUTING.md?plain=1#L363

Yeah, for scale scripts it not sufficient. WE need to change this to /8 to accommodate the scale addresses.

cprabha: I have updated comment also contributing.md with /8, please check.

CONTRIBUTING.md
```diff
@@ -360,11 +360,11 @@ where link local is being tested.
     used for traffic testing; split as needed.
 *   `BMWG`: (198.18.0.0/15): additional data plane networks.
 
-*   20.0.0.1/15: data plane source network addresses used for scale testing.
+*   20.0.0.0/8: data plane source network addresses used for scale testing.
```
liulk, line 363: I spoke to @xw-g. Can we use different subnets to avoid potential interference with publicly allocated subnets?

  - 20.0.0.0/8 is allocated to Microsoft and others. Suggest to use 251.0.0.0/8 instead.
  - 30.0.0.0/8 is allocated to the Depart of Defense. Suggest to use 252.0.0.0/8 instead.
  - 100.0.0.0/8 is reserved for Carrier NAT and others. Suggest to use 253.0.0.0/8 instead.
  - If you need more, we can also use 254.0.0.0/8
  - Avoid 255.0.0.0/8 because 255.255.255.255/32 may have a special meaning.
 
The suggested range is reserved for future use since 1981. I'm hoping that even if they start using them, the allocation will start from 240.0.0/8.

https://www.iana.org/assignments/ipv4-address-space/ipv4-address-space.xhtml

cprabha: @liulk , @xw-g  I will update both pull requests with this new address range.

cprabha: > @liulk , @xw-g I will update both pull requests with this new address range.

@liulk , @xw-g , > 240.0.0/4 orlonger is Martian addresses ? routes will be not accepted? These ranges does not work.
please suggest anything which is lesser than this range?

liulk: Is it reasonable to assume that tests that need extended IP address range for scale testing will need to customize the Martian range to allow for these extended addresses? Otherwise there really isn't any other spaces.

Another lesser evil is to find a /8 subnet allocated to a legacy company that no longer announces BGP prefixes, e.g. 19./8: https://bgp.he.net/ip/19.0.0.1 but this is the only one I could find.

Scaling tests really ought to be using IPv6. @dplore @xw-g to comment.

dplore: Agreed that we should use private/test reserved IP ranges whenever possible.  Because these are open source tests, we expect them to be run in many environments in which properly allocated IP's might be present and cause a conflict.  

Let us try using 240.0.0.0/4.  It seems some devices need a special command to permit  martian addresses to be used: 
-  [Allowing Martian addresses on JunOS](https://www.juniper.net/documentation/us/en/software/junos/static-routing/topics/topic-map/recognize-martian-addr-routing.html#d32e54__d8789e135)
- [Martian addresses on Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/bgp/73x/b-bgp-cg-8k-73x/implementing-bgp.html#concept_vym_phb_4jb)

Let's try using 240.0/4 for the scaling test.  If a blocking issue arises that we cannot address with a simple deviation, we will pick some other IPv4 block.

Regarding using IPv6 for scaling, I think we want to flexibility to scale both IPv4 and IPv6 routes.  This is desirable because hardware behaves differently for v4 and v6 especially in regards to scaling forwarding tables.

bensons: I agree with @dplore's suggestion and comment. 

I would add that, re using an unannounced subnet (such as 19.0.0.0/8), the absence of a prefix in the "global" BGP table of any one (or even many) AS is not necessarily indicative of the prefix being unused. For example, I'm aware of some networks that accept the risk and use unannounced blocks internally for various purposes, e.g., CGN pools instead of the allocated 100.64 block, additional "private" space, etc. Further, some networks use globally unique addresses for private internets that aren't seen on the "global" table, e.g., securities trading exchanges.

That's not to say that we have no option if 240.0/4 doesn't work for us -- but using globally unannounced space *does* add some additional risk if people use these tests in less-than-perfectly controlled lab environments.

cprabha: I have updated with Martian address range and removed 20/30/100 address range .  We will update baseline config which is needed to allow Martian address range.

tools/check_ip_addresses.pl
cprabha: updated 20/30/100 to Martian address range.

xw-g:
> I'm not sure if using Martian `240/8` in test (e.g. for routing) would be a challenge to OTHER major vendors. @dplore can you help investigate? Thanks.

FYI, @dplore @bensons @liulk @cprabha I plan to deprecate current PR with the following ones: 

* https://github.com/openconfig/featureprofiles/pull/2733
* https://github.com/openconfig/featureprofiles/pull/2732

xw-g (changes requested):
Blocking the merge of this PR. We plan to deprecate this PR with the following PRs:

* https://github.com/openconfig/featureprofiles/pull/2733
* https://github.com/openconfig/featureprofiles/pull/2732

PR #2733, opened 2024-02-27
Update IP assignment guidance

xw-g:
* Add a WARNING that users who run the test own the responsibility of not leaking test traffic to internet.
* Add `100.64.0.0/10`, a non-globally routable CGN range, which usually is used inside service providers. https://www.iana.org/go/rfc6598.
* Correct the name of `198.18.0.0/15` to `Device Benchmark Testing` (RFC 2544).
* Removed other "arbitrary" IP ranages, we'd like to deprecate them from the tests.
* Update IP check script accordingly.

Formatting:
* Simplify the wording in the `IP Addresses Assignment` section.

### DIFF
--- a/tools/check_ip_addresses.pl
+++ b/tools/check_ip_addresses.pl
@@ -43,7 +43,11 @@ if (/\b(\d{1,3}(\.\d{1,3}){3,})(\/\d+)?\b/) {
   next if $ip =~ /192\.0\.2\./;         # TEST-NET-1 (RFC 5737)
   next if $ip =~ /198\.51\.100\./;      # TEST-NET-2 (RFC 5737)
   next if $ip =~ /203\.0\.113\./;       # TEST-NET-3 (RFC 5737)
-  next if $ip =~ /198\.(18|19)\./;      # BMWG (RFC 2544)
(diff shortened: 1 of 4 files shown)


tools/check_ip_addresses.pl
```diff
@@ -43,7 +43,11 @@ END
   next if $ip =~ /192\.0\.2\./;         # TEST-NET-1 (RFC 5737)
   next if $ip =~ /198\.51\.100\./;      # TEST-NET-2 (RFC 5737)
```
morrowc, line 44: is there a reason to do this with regex and not to just:
```perl
  candidate = ipaddres.IPAddress($ip)
  result = false
  for net in netsDict {
      if candidate in net {
       result=true
       break
     }
  }

if !result errorBdaly
```

robshakir (approved):

tools/check_ip_addresses.pl
xw-g: @robshakir you might know?  (does not block this PR merge)

robshakir: I am not sure who wrote this perl originally -- honestly, if we refactor it, should we just make it not-perl? :-)

robshakir (approved):

PR #2735, opened 2024-02-28
Resolving confilc of Test id: RT-9

rszarecki:


tools/check_ip_addresses.pl
```diff
@@ -61,6 +61,8 @@ END
   next if $ip =~ /192\.51\.129\./;      # 192.51.129.0/22
   next if $ip =~ /192\.168\.10\./;      # 192.168.10.0/24
   next if $ip =~ /192\.168\.20\./;      # 192.168.20.0/24
+  next if $ip =~ /251\.0\./;            # 251.0.0.0/16
```
xw-g, line 64: Let's not adding those ranges here:

* IP check is no longer a "required" check for PRs. 

* Because this PR is about resolving the conflicting test ID, I think it's ok not to touch the IPs in this PR.

* If you'd like to make the IP align with the recommendation, we should change IP used by tests, instead of adding new allowed ranges here. (this can be another effort)

More context: 
* https://github.com/openconfig/featureprofiles/pull/2733
* https://github.com/openconfig/featureprofiles/pull/2743

xw-g (approved):

dplore:
@rszarecki  this PR missed some test id's within # RT-2.13: Weighted-ECMP for IS-IS with RT-9.1, RT-9.2

PR #2738, opened 2024-02-28
adding ATE test will give failure instead of warning

self-maurya:


self-maurya:
FYI, @jasdeep-hundal reverting back the new ATE failure check

rohit-rp (approved):

PR #2760, opened 2024-03-05
RT-1.29: Added deviation to skip BGP rib verification

mohanasm:
Find below the changes:
* Added deviation to skip BGP rib verification.
* Added delete policy at the end of the every test so that reference for old policy gets removed.
* Updated to use batch replace for configuring routing-policy and applying routing-policy.
* Removed accept route for statement asp-statement-v4 as the next statement med-policy-v4 will not be applied if accept route is configured for asp-statement-v4.
* Updated bgp peer name for prefix check in OTG.
* Updated IXIA config to enable LearnedInformationFilter().
* Updated with unique ateAS for port1 and port2. Routes will not be advertised from port2 to port1 if same AS number is configured on both ATE ports.
* Updated v62Route to match with prefix received in ATE.


self-maurya (approved):

Swetha-haridasula (commented):
LGTM

PR #2768, opened 2024-03-06
Fix test case ID and add to testregistry

rohit-rp:
RT-2.12 conflicts with feature/isis/static-route_isis_redistribution/README.md


NishaSadhasivam-work (approved):
LGTM!

PR #2776, opened 2024-03-07
Update zr_input_output_power_test.

ahsaanyousaf:
- Update OPR value when interface is in down state to -40dBm 
- Update OPT range to -6 to -10 dbm


feature/platform/transceiver/zr_input_output_power_test/README.MD
```diff
@@ -53,8 +53,8 @@ power.
     are available for streaming.
 
 *   RX Input and TX output power values must always be of type decimal64.
-    When link interfaces are in down state 0 must be reported as a valid
-    value.
+    When link interfaces are in down state RX Input power of -40 dbm must be
+    reported as a valid value.
```
dplore, line 57: In my admittedly limited experience, I have seen devices (not ZR specific) report -99dbm when there is loss of light.  Why is -40dbm specified here?  Is that part of the CMIS specification?  I see in section `8.5.2 Lane-Related Monitor Thresholds` of CMIS a reference to a low power threshold of ~ -40dbm.     I see it's a low threshold, but does that mean the receiver is not allowed to report a lower value when there is loss of light?  

Second, because of the `~` should we allow some tolerance in the value?  (+/- 1 db for example?)

dplore (commented):
I have a naive question for you.   Also, there appears to be some change conflict that should be resolved.

feature/platform/transceiver/zr_input_output_power_test/README.MD
dplore: Sounds good to me.  Note that CMIS uses ~ to mean approximately as well which was what I was thinking of.  :). In any case, thanks for the explanation.

dplore (approved):

PR #2778, opened 2024-03-07
backup_nhg_action_pbf_test.go

cprabha:
Hi,

1. I have moved ixia pushconfig/startprotocols after configure DUT function.
2. Added delay after pushconfig/startprotocols to avoid ixia issue.

Thanks,
Prabha


ram-mac:
@cprabha, is it possible to use gnmi.Await() / gnmi.Watch() to check the status rather than waiting for 30seconds?

cprabha:
able to resolve ixia issue with updating binding file timeout under otg

PR #2785, opened 2024-03-11
adding test for low_power_mode

prinikasn:


feature/platform/transceiver/tests/zr_low_power_mode_test/zr_low_power_mode_test.go
```diff
@@ -0,0 +1,174 @@
...
+				"max":  opMax,
+			}
+
+			validateOutputPower(t, dut, powerStreamMap)
+
+			// Derive transceiver names from ports.
+			component := gnmi.OC().Component(tr)
+
+			outputPower := gnmi.Get(t, dut, component.OpticalChannel().TargetOutputPower().State())
+			if outputPower != targetOutputPowerdBm {
```
ahsaanyousaf, line 164: you probably want to allow +/- 1dBm of deviation from the configured target power of -10dBm

feature/platform/transceiver/tests/zr_low_power_mode_test/zr_low_power_mode_test.go
```diff
@@ -0,0 +1,174 @@
...
+func validateStreamOutput(t *testing.T, dut *ondatra.DUTDevice, streams map[string]*samplestream.SampleStream[string]) {
+	for key, stream := range streams {
+		output := stream.Next(t)
+		if output == nil {
+			t.Fatalf("OC path for %s not streamed in the most recent subscription interval", key)
+		}
+		value, ok := output.Val()
+		if !ok {
+			t.Fatalf("Error capturing streaming value for %s", key)
+		}
```
jasdeep-hundal, line 49: Do we also need the check on the value `Kind` here? ("No" is a fair answer here, just curious.)

feature/platform/transceiver/tests/zr_low_power_mode_test/zr_low_power_mode_test.go
```diff
@@ -0,0 +1,174 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package zr_low_power_mode_test
+
+import (
+	"fmt"
+	"reflect"
```
rezachit, line 19: Test comment

feature/platform/transceiver/tests/zr_low_power_mode_test/zr_low_power_mode_test.go
prinikasn: Initially thought we don't need it since we are checking for empty string. But I suppose we could have better error messaging if we ad the check on value Kind here.

feature/platform/transceiver/tests/zr_low_power_mode_test/zr_low_power_mode_test.go
prinikasn: +1

Added the tolerance check for the target power. 

To be consistent I also added a tolerance for the target frequency `targetFrequencyToleranceHz    = 100000` . Please let me know if that tolerance value looks ok .

ahsaanyousaf (approved):
LGTM for optical

jasdeep-hundal (approved):

PR #2792, opened 2024-03-12
Updating flow count and traffic duration

cprabha:
Hi,

1. ) I have updated flow count and traffic duration 
2. ) Added Wait for ARP to resolve arp on ixia.

Thanks,
Prabha


feature/gribi/otg_tests/basic_encap_test/basic_encap_test.go
```diff
@@ -847,8 +850,15 @@ func sendTraffic(t *testing.T, args *testArgs, flows []gosnappi.Flow, capture bo
 	otg := args.ate.OTG()
 	args.topo.Flows().Clear().Items()
 	args.topo.Flows().Append(flows...)
+
 	otg.PushConfig(t, args.topo)
+	time.Sleep(30 * time.Second)
```
dplore, line 855: @octpetre should we need to be doing sleep like this after pushing config and starting protocols on OTG?  Aren't these blocking calls?  And if not, is there some event trigger we can listen for to be confident that the config and protocols are started?

For protocol start, I figured WaitForArp would be sufficient?

cprabha: @dplore ixia is throwing error when startprotocols is invoked and hence ixia support asked to add some delay after pushconfig/startprotocols.

dplore: WaitForArp by itself doesn't fix the StartProtocols issue? (in my experience that has been reliable)

cprabha: WaitforArp is resolve arp after StartProtocols is completed, it helps to learn destination mac address after ports are up. 
But sometimes StartProtocols itself fails with context deadline exceeded error. To avoid this we need to add little bit sleep after PushConfig and StartProotocls.

dplore: Acknowledged; I only expect WaitForArp after StartProtocols, ie: to help make sure we don't start traffic generation before the ATE and network is initialized/ready.  So the 30 second sleep after StartProtocols should not be needed, right?

Regarding the 30 second sleep after otg.PushConfig, I think you are saying it is needed because in your envirnoment if you try to call StartProtocols immediately after pushconfig, you sometimes receive an error from the StartProtocols call?.  

@jasdeep-hundal do you have any comment on if there should be a sleep after executing `otg.PushConfig(t, args.topo)`

cprabha: @dplore correct, it is to avoid ixia error , if start protocols called immediately after push config we see ixia error.

jasdeep-hundal: If we're going to keep this in, we should have this linked to a known bug on Github that's assigned to Keysight.

I think a cleaner approach here that preserves/clarifies test intent would to incorporate this pause using a custom binding that wraps an underlying gosnappi client in a new struct implementing the same interface, adding a custom PushConfig method that wraps the same method in the underlying client. (Especially if the bug is not limited to this test's OTG config.)

cprabha: @dplore , @jasdeep-hundal I have opened issue in github. Please check https://github.com/openconfig/featureprofiles/issues/2824

self-maurya (approved):

PR #2795, opened 2024-03-12
TE-3.7 : base_hierarchical_nhg_update_test

cprabha:
Hi,

This script has 3 subtests and In third subtest we need to unbind ingress interface from VRF-1 since filter is not applied in VRF-1.

Thanks,
Prabha


Swetha-haridasula (approved):
LGTM

PR #2808, opened 2024-03-14
gNMI-1.4 | change parent type validation to use lookup

arvbaska1:
Changed parent type validation to use lookup instead of get.

Some of the parent component may not map to any of identity types defined under OPENCONFIG_HARDWARE_COMPONENT, for those components type leaf would not be returned .

### DIFF
--- a/feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
+++ b/feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
@@ -745,8 +745,9 @@ func ValidateComponentState(t *testing.T, dut *ondatra.DUTDevice, cards []*oc.Co
 						t.Errorf("Component %s Parent: Chassis component NOT found in the hierarchy tree of component", cName)
 						break
 					}
-					parentType := gnmi.Get(t, dut, gnmi.OC().Component(parent).Type().State())
-					if parentType == componentType["Chassis"] {
+					pLoookup := gnmi.Lookup(t, dut, gnmi.OC().Component(parent).Type().State())
+					parentType, present := pLoookup.Val()
+					if present && parentType == componentType["Chassis"] {
 						t.Logf("Component %s Parent: Found chassis component in the hierarchy tree of component", cName)
 						break
 					}


LimeHat:
>Some of the parent component may not map to any of identity types defined under OPENCONFIG_HARDWARE_COMPONENT, for those components type leaf would not be returned .

If there are gaps in the list of OC component types, perhaps a PR to `openconfig/public` repo to add new identities would be appropriate to close the gap? cc @dplore

dplore:
> > Some of the parent component may not map to any of identity types defined under OPENCONFIG_HARDWARE_COMPONENT, for those components type leaf would not be returned .
> 
> If there are gaps in the list of OC component types, perhaps a PR to `openconfig/public` repo to add new identities would be appropriate to close the gap? cc @dplore

This is true.  However in this context, this snippet of code requires the hardware component CHASSIS.  I think the author is saying they think this part of the test should not throw an error if Chassis is not the parent.  

However, this does not meet the intent for this test.  The choice of parent validation (should be named something like: parentIsChassis) is set based on the child component type, for example:
https://github.com/openconfig/featureprofiles/pull/2808/files#diff-39a0760eb06daff3e1b837d61f6ca10630352bbdf5f48eb46df154ad33c4bc8eR140

So if this is failing, we want it to fail.  If you believe some component being validated by this test should NOT have Chassis as the parent, you can request changing that in the [test requirements per component type](https://github.com/openconfig/featureprofiles/pull/2808/files#diff-39a0760eb06daff3e1b837d61f6ca10630352bbdf5f48eb46df154ad33c4bc8eR140)

dplore:
So to be clear, in the current form, I do not think this PR should be approved.

dplore (changes requested):
This should not be approved in the current form.  If some specific component should NOT have CHASSIS as the parent, please propose an update in the requirements portion of the test.

([for example](https://github.com/openconfig/featureprofiles/pull/2808/files#diff-39a0760eb06daff3e1b837d61f6ca10630352bbdf5f48eb46df154ad33c4bc8eR140))

arvbaska1:
> test requirements per component type

This PR doesn’t change the expectation that the child component should have a component with type “CHASSIS” as parent in its component hierarchy tree,
The change is to tolerate certain components which would not return Type leaf,
 
e.g. for child component corresponding to type "PowerSupply" , the direct parent may not be of type CHASSIS, and the grandparent could be of type CHASSIS in its hierarchy.
Some of these intermediate components may not return type leaf as it doesn’t map to any of the identity types defined in the OC Platform types here https://github.com/openconfig/public/blob/master/release/models/platform/openconfig-platform-types.yang#L262
 
Right now in the test, it’s looping over the component hierarchy until it either finds or not find component with type Chassis in the hierarchy tree of initial child component.

dplore:
> > test requirements per component type
> 
> This PR doesn’t change the expectation that the child component should have a component with type “CHASSIS” as parent in its component hierarchy tree, The change is to tolerate certain components which would not return Type leaf,   e.g. for child component corresponding to type "PowerSupply" , the direct parent may not be of type CHASSIS, and the grandparent could be of type CHASSIS in its hierarchy. Some of these intermediate components may not return type leaf as it doesn’t map to any of the identity types defined in the OC Platform types here https://github.com/openconfig/public/blob/master/release/models/platform/openconfig-platform-types.yang#L262   Right now in the test, it’s looping over the component hierarchy until it either finds or not find component with type Chassis in the hierarchy tree of initial child component https://github.com/openconfig/featureprofiles/blob/main/feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go#L738

Oh, I understand now.  This will the type to not exist or be a different value.  I am ok with this then.

I also support suggestions to add more[ OPENCONFIG_HARDWARE_COMPONENT types](https://github.com/openconfig/public/blob/6fd1f5d24d35b693864aa3218315b2acbeae9055/release/models/platform/openconfig-platform-types.yang#L247)

dplore (approved):

arvbaska1:
> > > test requirements per component type
> > 
> > 
> > This PR doesn’t change the expectation that the child component should have a component with type “CHASSIS” as parent in its component hierarchy tree, The change is to tolerate certain components which would not return Type leaf,   e.g. for child component corresponding to type "PowerSupply" , the direct parent may not be of type CHASSIS, and the grandparent could be of type CHASSIS in its hierarchy. Some of these intermediate components may not return type leaf as it doesn’t map to any of the identity types defined in the OC Platform types here https://github.com/openconfig/public/blob/master/release/models/platform/openconfig-platform-types.yang#L262   Right now in the test, it’s looping over the component hierarchy until it either finds or not find component with type Chassis in the hierarchy tree of initial child component https://github.com/openconfig/featureprofiles/blob/main/feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go#L738
> 
> Oh, I understand now. This will the type to not exist or be a different value. I am ok with this then.
> 
> I also support suggestions to add more[ OPENCONFIG_HARDWARE_COMPONENT types](https://github.com/openconfig/public/blob/6fd1f5d24d35b693864aa3218315b2acbeae9055/release/models/platform/openconfig-platform-types.yang#L247)

Thanks, In this case maybe something with name “SLOT” would be appropriate for mapping these components. These components are removable=False.
Similar to “FRU” for removable components which are not able to map anything else.

  identity FRU {
    base OPENCONFIG_HARDWARE_COMPONENT;
    description
      "Replaceable hardware component that does not have a more
      specific defined schema.";
  }

PR #2821, opened 2024-03-16
Code commit for gNMI-1.10: add component empty check

nicjiang2023:
Add component empty check in TestComponentParent
```go
compMtyVal, compMtyPresent := gnmi.Lookup(t, dut, gnmi.OC().Component(comp).Empty().State()).Val()
if compMtyPresent && compMtyVal {
  t.Skipf("INFO: Skip status check as %s is empty", comp)
}
```

### DIFF
--- a/feature/gnmi/otg_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
+++ b/feature/gnmi/otg_tests/telemetry_basic_check_test/telemetry_basic_check_test.go
@@ -460,6 +460,10 @@ func TestComponentParent(t *testing.T) {
 			// Validate parent component.
 			for _, comp := range compList[tc.desc] {
 				t.Logf("Validate component %s", comp)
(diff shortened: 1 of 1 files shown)


LimeHat:
Is the assumption that an empty component doesn't need a parent valid? @dplore 

E.g. if a modular system displays 4 linecards with `empty: true`, shouldn't the linecards still have the parent component listed?

dplore:
> Is the assumption that an empty component doesn't need a parent valid? @dplore
> 
> E.g. if a modular system displays 4 linecards with `empty: true`, shouldn't the linecards still have the parent component listed?

Yes, I expect a component which exists on the system to have parent, even if it is empty.    A LINECARD should still have the CHASSIS as parent.

nicjiang2023:
Closed the pull request as we're planning to stream parent leaf for empty component.

PR #2827, opened 2024-03-18
RT-1.28, RT-1.30: Added peer-group for BGP peer and deviation for ISIS enabled

mohanasm:
RT-1.28: Added peer-group for BGP peer and deviation for ISIS enabled
RT-1.30: Added peer-group for BGP peer


Swetha-haridasula (approved):
LGTM

PR #2828, opened 2024-03-18
Fix TE-16.1 basic_encap_test for Arista due to expected deviations

prakashbadri-arista:
Fix the test failures due to the following issues in the test
- The test was setting "network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/l2/ethertype" to detect match rule AF for the default rule. But this path should not be used for that and is not supported by EOS.
- Add the following deviations for EOS: static_protocol_name: "STATIC" gribi_mac_override_static_arp_static_route: true interface_enabled: true ttl_copy_unsupported: true ( The TTL expectation is wrong in the test. once that is fixed, this can be removed ) default_network_instance: "default"
- Add static routes to magic IP and setup a static arp entry for magic ip address with all the egress interfaces for EOS.
- Change the publishing of AFT entries such that NH/NHG aft entries are pushed together.

The test still fails on validating encapRatio. We need to see if the test has the right flows to validate this. But pushing the fixes for other issues.


self-maurya (dismissed):

feature/gribi/otg_tests/basic_encap_test/basic_encap_test.go
```diff
@@ -674,6 +672,71 @@ func configureBaseconfig(t *testing.T, dut *ondatra.DUTDevice) {
...
+	t.Helper()
+	p2 := dut.Port(t, "port2")
+	p3 := dut.Port(t, "port3")
+	p4 := dut.Port(t, "port4")
+	p5 := dut.Port(t, "port5")
+	s2 := &oc.NetworkInstance_Protocol_Static{
+		Prefix: ygot.String(magicIp + "/32"),
+		NextHop: map[string]*oc.NetworkInstance_Protocol_Static_NextHop{
+			strconv.Itoa(nh10ID): {
+				Index: ygot.String(strconv.Itoa(nh1ID)),
```
self-maurya, line 685: I think it should be nh10ID instead of nh1ID

feature/gribi/otg_tests/basic_encap_test/basic_encap_test.go
```diff
@@ -674,6 +672,71 @@ func configureBaseconfig(t *testing.T, dut *ondatra.DUTDevice) {
...
+	gnmi.Update(t, dut, sp.Static(magicIp+"/32").Config(), s3)
+	sp = gnmi.OC().NetworkInstance(deviations.DefaultNetworkInstance(dut)).Protocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, deviations.StaticProtocolName(dut))
+	gnmi.Update(t, dut, sp.Static(magicIp+"/32").Config(), s4)
+	sp = gnmi.OC().NetworkInstance(deviations.DefaultNetworkInstance(dut)).Protocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, deviations.StaticProtocolName(dut))
+	gnmi.Update(t, dut, sp.Static(magicIp+"/32").Config(), s5)
+
+	gnmi.Update(t, dut, gnmi.OC().Interface(p2.Name()).Config(), configStaticArp(p2.Name(), magicIp, magicMac))
+	gnmi.Update(t, dut, gnmi.OC().Interface(p3.Name()).Config(), configStaticArp(p3.Name(), magicIp, magicMac))
+	gnmi.Update(t, dut, gnmi.OC().Interface(p4.Name()).Config(), configStaticArp(p4.Name(), magicIp, magicMac))
+	gnmi.Update(t, dut, gnmi.OC().Interface(p5.Name()).Config(), configStaticArp(p5.Name(), magicIp, magicMac))
```
self-maurya, line 737: can be clubbed with gnmi.BatchUpdate

feature/gribi/otg_tests/basic_encap_test/basic_encap_test.go
prakashbadri-arista: I have fixed this function to use a range index for the OC static route config instead of the corresponding nh AFT index to simplify the code

self-maurya (approved):

PR #2830, opened 2024-03-19
gNOI-3.3: add new subtests for configuration convergence

kjahed:
Add the following subtests:
1- TestConfigurationUnchanged checks that the configuration is unchanged after switchover
2- TestConfigurationConvergence checks that a large configuration is accepted within 60 seconds after switchover

Also adds a [Reporter](https://pkg.go.dev/github.com/google/go-cmp/cmp#Reporter) to collect diff entries from cmp.Diff. This allows the test to examine the each diff to decide whether, for e.g., the configuration match. In this enhanced switchover test, this is used to ignore leaves that are not explicitly configured (i.e., fields with nil value) when comparing against the configuration returned from the device.

### DIFF
--- a/feature/gnoi/system/tests/supervisor_switchover_test/README.md
+++ b/feature/gnoi/system/tests/supervisor_switchover_test/README.md
@@ -6,12 +6,23 @@ Validate that the active supervisor can be switched.
 
 ## Procedure
 
-*   Issue gnoi.SwitchControlProcessor to the chassis with dual supervisor,
-    specifying the path to choose the standby RE/SUP.
-*   Ensure the SwitchControlProcessorResponse has the new active supervisor as
-    the one specified in the request.
-*   Validate the standby RE/SUP becomes the active after switchover
-*   Validate that all connected ports are re-enabled.
+*   Basic Switchover
+    *   Issue gnoi.SwitchControlProcessor to the chassis with dual supervisor,
+        specifying the path to choose the standby RE/SUP.
+    *   Ensure the SwitchControlProcessorResponse has the new active supervisor as
+        the one specified in the request.
+    *   Validate the standby RE/SUP becomes the active after switchover
+    *   Validate that all connected ports are re-enabled.
+
+*   Configuration consistent after switchover
+    *   Get the configuration before switchover.
+    *   Perform and validate switchover using the process outlined above.
+    *   Get configuration after switchover and validate that it matches the pre-switchover configuration.
+
+*  Configuration convergence after switchover
+    *   Perform and validate switchover using the process outlined above.
+    *   Push a large configuration
+    *   Get the configuration and validate that it is accepted whithin 60 seconds.
 
 ## Config Parameter Coverage
 
--- a/feature/gnoi/system/tests/supervisor_switchover_test/setup_test.go
+++ b/feature/gnoi/system/tests/supervisor_switchover_test/setup_test.go
@@ -1,0 +1,249 @@
+package supervisor_switchover_test
+
+import (
+	"fmt"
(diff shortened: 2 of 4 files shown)


LimeHat:
Can you please add 
* `new-requirement-pr` tag to the PR
* an updated README that introduces these requirements?

Thanks,

LimeHat:
Thanks for the updates @kjahed 

I'm wondering if these subtests belong to gNOI-3.3?

Specifically for the second case, configuration convergence, there's a separate test gNMI-1.19
https://github.com/openconfig/featureprofiles/tree/main/feature/platform/controllercard/tests/SetRequest_controll_card_switchover

cc @rszarecki  @dplore  @cfernanz

feature/gnoi/system/tests/supervisor_switchover_test/README.md
```diff
@@ -6,12 +6,23 @@ Validate that the active supervisor can be switched.
...
-    the one specified in the request.
-*   Validate the standby RE/SUP becomes the active after switchover
-*   Validate that all connected ports are re-enabled.
+*   Basic Switchover
+    *   Issue gnoi.SwitchControlProcessor to the chassis with dual supervisor,
+        specifying the path to choose the standby RE/SUP.
+    *   Ensure the SwitchControlProcessorResponse has the new active supervisor as
+        the one specified in the request.
+    *   Validate the standby RE/SUP becomes the active after switchover
+    *   Validate that all connected ports are re-enabled.
```
dplore, line 15:
```suggestion
*  gNOI-3.3.1 Basic Switchover
    *   Issue gnoi.SwitchControlProcessor to the chassis with dual `CONTROLLER_CARD`,
        specifying the path to choose the standby RE/SUP.
    *   Ensure the SwitchControlProcessorResponse has the new active `CONTROLLER_CARD` as
        the one specified in the request.
    *   Validate the standby `CONTROLLER_CARD` becomes the active after switchover
    *   Validate that all connected ports are re-enabled.
```

feature/gnoi/system/tests/supervisor_switchover_test/README.md
```diff
@@ -6,12 +6,23 @@ Validate that the active supervisor can be switched.
...
+
+*   Configuration consistent after switchover
+    *   Get the configuration before switchover.
+    *   Perform and validate switchover using the process outlined above.
+    *   Get configuration after switchover and validate that it matches the pre-switchover configuration.
+
+*  Configuration convergence after switchover
+    *   Perform and validate switchover using the process outlined above.
+    *   Push a large configuration
+    *   Get the configuration and validate that it is accepted whithin 60 seconds.
```
dplore, line 25: I think these steps should be moved to gNMI-1.19?

I think this portion of the test is duplicate with gnmi-1.19?   The implementation should probably be moved here instead:

https://github.com/openconfig/featureprofiles/blob/64388db0ea85c18ed6e556324dc8f2dcc643ddfb/feature/platform/controllercard/tests/SetRequest_controll_card_switchover/README.md?plain=1#L1

dplore (commented):
The implementation should probably be moved here instead:

https://github.com/openconfig/featureprofiles/blob/64388db0ea85c18ed6e556324dc8f2dcc643ddfb/feature/platform/controllercard/tests/SetRequest_controll_card_switchover/README.md?plain=1#L1

feature/gnoi/system/tests/supervisor_switchover_test/README.md
kjahed: Ack. Will close this and open a new PR for gNMI-1.19.

PR #2831, opened 2024-03-19
Apply Arista specific fixes for encap_decap_scale_test

Ankur19:


self-maurya:
encap_decap_scale_test has been fixed for Arista with updated scale number and it's working now as expected.

Closing this stale PR.

PR #2832, opened 2024-03-19
Update Nokia config.

guoshiuan:


LimeHat:
I think Marc has the same in #2809 ?

guoshiuan:
> I think Marc has the same in #2809 ?

Yes, this config is used by KNE FNT test, and is basically the same change as #2809.

guoshiuan:
> > I think Marc has the same in #2809 ?
> 
> Yes, this config is used by KNE FNT test, and is basically the same change as #2809.

Sorry, I misread #2809 and thought that ACL section was missing.
Withdraw this PR.

PR #2835, opened 2024-03-21
Updates to MTU-1.3: MTU-1.3: Large IP Packet Transmission

hattikals:
- In the script, lacp mode has been set to 'unset' for the lag bundle which leads to default value getting initialized for LACP mode - ACTIVE as LACP mode. When the same lag bundle Type is set to static, it creates the conflict which leads to failure. As the script is just using static lag bundle, removing the lacp set code. 
- Added some delay after PushConfig and startProtocols for ports to come up and ARP check to make sure ARP is resolved before Traffic is sent.
- Also introduced BatchDelete in cleanup function


Swetha-haridasula (approved):
LGTM

PR #2843, opened 2024-03-25
Rt2.12 static route redistribution

mastarkey:
This test is for RT21.2 

This code is mostly complete however it is missing the portion of route validation on the OTG


octpetre:
Please move this under _**feature/isis/otg_tests**_

feature/isis/otg_tests/static_route_isis_redistribution_test/static_route_isis_redistribution_test.go
```diff
@@ -0,0 +1,730 @@
...
+	}
+}
+
+func TestStaticToISISRedistribution(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	configureDUT(t, dut)
+
+	ate := ondatra.ATE(t, "ate")
+	top := gosnappi.NewConfig()
+	var devs []gosnappi.Device
```
Ankur19, line 484: Fix

feature/isis/otg_tests/static_route_isis_redistribution_test/static_route_isis_redistribution_test.go
```diff
@@ -0,0 +1,730 @@
...
+					verifyTraffic(t, ate, tc.trafficFlows)
+				})
+
+			}
+
+			// TODO: Verify routes are learned on the ATE device. This is pending a fix from IXIA and OTG
+			// TODO: https://github.com/open-traffic-generator/fp-testbed-cisco/issues/10#issuecomment-2015756900
+
+			//t.Run(fmt.Sprintf("Verify Route on OTG"), func(t *testing.T) {
+			//
```
Ankur19, line 710: Uncomment and add t.Skip()

feature/isis/otg_tests/static_route_isis_redistribution_test/static_route_isis_redistribution_test.go
```diff
@@ -0,0 +1,730 @@
...
+	AddressFamily            string   `json:"address-family"`
+	SrcProtocol              string   `json:"src-protocol"`
+}
+
+func getAndVerifyIsisImportPolicy(t *testing.T,
+	dut *ondatra.DUTDevice, DisableMetricValue bool,
+	RplName string, addressFamily string) {
+
+	gnmiClient := dut.RawAPIs().GNMI(t)
+	getResponse, err := gnmiClient.Get(context.Background(), &gpb.GetRequest{
```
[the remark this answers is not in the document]
mastarkey: This is a temporary solution as Ondatra uses subscribe to check this path however currently this being resolved internally to fix, so needed to do a get to work around this for time being.

feature/isis/otg_tests/static_route_isis_redistribution_test/static_route_isis_redistribution_test.go
```diff
@@ -0,0 +1,730 @@
...
+type ipAddr struct {
+	address string
+	prefix  uint32
+}
+
+func verifyTraffic(t *testing.T, ate *ondatra.ATEDevice, flowsData []string) {
+	for _, flowName := range flowsData {
+
+		t.Logf("getting flow data for %v", flowName)
+		flow := gnmi.Get(t, ate.OTG(), gnmi.OC().Flow(flowName).Counters().State())
```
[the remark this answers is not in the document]
mastarkey: Tried to integrate this but when i use this it keeps returning negative integer vs directly getting the flow name counters

feature/isis/otg_tests/static_route_isis_redistribution_test/static_route_isis_redistribution_test.go
```diff
@@ -0,0 +1,718 @@
...
+	AddressFamily            string   `json:"address-family"`
+	SrcProtocol              string   `json:"src-protocol"`
+}
+
+func getAndVerifyIsisImportPolicy(t *testing.T,
+	dut *ondatra.DUTDevice, DisableMetricValue bool,
+	RplName string, addressFamily string) {
+
+	gnmiClient := dut.RawAPIs().GNMI(t)
+	getResponse, err := gnmiClient.Get(context.Background(), &gpb.GetRequest{
```
Ankur19, line 157: You can use ygnmi ops to use Get instead of Subscribe:

`gnmi.Get(t, dut.GNMIOpts().WithYGNMIOpts(ygnmi.WithUseGet())), .........)`

We would also need to update the hardcoded `DEFAULT` network instance with `deviations.DefaultNetworkInstance(dut)`

feature/isis/otg_tests/static_route_isis_redistribution_test/static_route_isis_redistribution_test.go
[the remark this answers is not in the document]
Ankur19: We can use `otgutils.GetFlowStats()` if the loss percentage is a problem.

feature/isis/otg_tests/static_route_isis_redistribution_test/static_route_isis_redistribution_test.go
mastarkey: tried to implement this but had hit the following errors. not sure if this needs to be structured in a different way.

path := gnmi.OC().NetworkInstance(deviations.DefaultNetworkInstance(dut)).TableConnection(
						oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC,
						oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_ISIS,
						oc.Types_ADDRESS_FAMILY_IPV4).State()

output := gnmi.Get(t, dut.GNMIOpts().WithYGNMIOpts(ygnmi.WithUseGet(),
						ygnmi.WithEncoding(gpb.Encoding_JSON_IETF)), path)
					t.Log(output)
					
noncompliant data encountered while unmarshalling: Noncompliance Errors by category (see https://github.com/openconfig/ygnmi#noncompliance-errors):
Path Noncompliance Errors: None
Type Noncompliance Errors:

&{DUT %!s(func(context.Context) (gnmi.GNMIClient, error)=0x10a02e7a0) map[] %!s(bool=true) <nil> [%!s(ygnmi.Option=0x106fa8ab0) %!s(ygnmi.Option=0x106fa8470)]} at /network-instances/network-instance[name=DEFAULT]/table-connections/table-connection[address-family=IPV4][dst-protocol=ISIS][src-protocol=STATIC]: path origin:"openconfig"  elem:{name:"network-instances"}  elem:{name:"network-instance"  key:{key:"name"  value:"DEFAULT"}}  elem:{name:"table-connections"}  elem:{name:"table-connection"  key:{key:"address-family"  value:"IPV4"}  key:{key:"dst-protocol"  value:"ISIS"}  key:{key:"src-protocol"  value:"STATIC"}}: value not present

feature/isis/otg_tests/static_route_isis_redistribution_test/static_route_isis_redistribution_test.go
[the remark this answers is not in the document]
mastarkey: was able to integrate this [otgutils.GetFlowLossPercent()](https://github.com/openconfig/featureprofiles/blob/main/internal/otgutils/actions.go#L54) not sure why it worked this time maybe something flaky in my setup at the time.

feature/isis/otg_tests/static_route_isis_redistribution_test/static_route_isis_redistribution_test.go
Ankur19: This is right. 

`https://github.com/openconfig/ygnmi#noncompliance-errors` non compliance errors are for logging only. The real issue here is `value not present`.

I see that in your previous implementation you used the `Config` path, however, in the new one you used `State`. The state value is not present and hence the error.

Ankur19: Use the `Config` path instead?

mastarkey: The problem with the `config` path is currently its not returning the `disable-metric-propagation` value so i would still end up needing to do this with the same approach as i have already implemented.

mastarkey:
Are there any pending changes if not can we merge this ?

Ankur19:
We can merge this once the merge conflicts are resolved. 

Thank you for your contribution!

mastarkey:
@Ankur19 conflicts are resolved

Ankur19 (approved):

PR #2857, opened 2024-03-27
Exception config test

dplore:
This is a rough draft of test requirements to mix, modify and override OC and CLI configuration using gNMI union_replace.


feature/exception_cfg/otg_tests/exception_cfg_test/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+CLI commands in the exception configlet are added to the end of the origin CLI data and should be applied in order by the DUT, overriding any overlapping commands that may have come before.
+
+gNMI union_replace is used for this scenario.  Configuration generation will create a both origin openconfig and origin CLI data and use a single gnmi.Set using the union_replace action to push all of the configuration in one RPC call to the DUT.
+
+In each subtest, the device should merge the origin OC and CLI configurations as per the union_replace specification.  If a CLI command appears which conflicts with other CLI, the latest CLI should be used.  If the resulting CLI command conflicts with origin openconfig, then the [union_replace conflict resolution method](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-union_replace.md#533-resolving-issues-with-union-between-the-origins) should be used.  In the case where the DUT is expected to produce an error, the test should detect this error. In the case where the DUT is expected to override, the test should verify the over ridden configuration is in place.
```
ncorran, line 11: We have had a private discussion about this issue area previously.

I believe the problem is more complicated than this summary suggests.

The definition of the union-replace is that it can define both an OC and a native set of config to be used to replace the previous instance of the config that is stored in the device.  As such it should be a list of configuration items, much like we would define a normal edit or replace.

The problem for exception configlets is that they may need to embed other operations into the config, in particular a DELETE (no xxx; followed by a new xxx), so it is NOT a simple case of overwriting the config with the last write wins, there are also inline deletes (which shouldn't be in a replace)

Further, these delete operation need to be able to impact both OC and CLI that has come previously.  So for implementations which separate the OC config into one datastore and the native config into another datastore and then have a merge stage later, the requirement for the exception configlet is that it acts on top of the merged store, NOT only within the native config store - which is completeley different to the other content of the message.

This makes the exception configlet a completely special case.  It cannot just be embedded in the native cli block in a union replace as the content has to be able to be separated out and handled specially.

2 options for this are:
1) Re-define union replace to have a 3rd section for exception configlets and define the semantics of how / when it acts and that it can include delete operations...
2) Leave the union replace defined as-is but define a special OC leaf which holds the exception-configlet config as a string.  We then just need to add the special semantics onto this leaf.  When it is set the content is applied after the merge of OC with Native.

The potential advantage of approach 2 is that this then frees the exception configlets from being a union-replace concept and makes it more general.  Other workflows where customer controllers generate sets of edits can still use the exception configlets leaf to override the edits coming in (while the exception configlets leaf is set).

Finally - a further consieration is that sometimes a system may have to generate an internal resync operation - e.g. over upgrades, or verison changes or process restarts or other cases.   During such a resync it is important to ensure the exception configlets are still applied (and not overriden by any resync, including per-datastore resyncs).  As such the exception configlets again need to be explcitly known/stored and re-appied during resync flows.

--
Cisco cannot implement the exception configlets as described above due to its external datastore separation (by model family), and the fact that union-replace is split into 2 parts and each addresses the specific datastore for its model-family today.

dplore:
We are currently no longer pursing this option and will look to resolve the exception config scenario upstream of the device.

PR #2863, opened 2024-03-28
Add yaml for OCPaths/RPC to test README

dplore:
This changes the requirement for featureprofiles test READMEs to format OC path and RPCs from plain text to a structured format (yaml).

This format is intended to be easy for humans who are writing README's for test requirements and may not be a developer.  It is also more simple to maintain the path list with the test which is using the paths versus today's practice where the path is listed in a readme and in feature.textproto files.  To keep feature.textproto files accurate, they must be manually kept in sync with README's from multiple tests across potentially many child folders with multiple levels.  In addition the feature.textproto files lack the ability to indicate which tests are intended to cover a path.  It is useful to know which test(s) a path is intended to cover.

The structured list of paths/rpc can be used for static analysis of tests for invalid OC paths, detecting tests which would break due to breaking OC model changes and building an OCPath/RPC to test relationship, showing all the intended test requirements for any given path/RPC.

### DIFF
--- a/doc/test-requirements-template.md
+++ b/doc/test-requirements-template.md
@@ -45,16 +45,20 @@ Write a few sentences or paragraphs describing the purpose and scope of the test
 This example yaml defines the OC paths intended to be covered by this test.  OC paths used for test environment setup are not required to be listed here.
 
 ```yaml
-OCPaths:
-    # Configuration of policy
-  - name: /interfaces/interface/config/description
-  - name: /components/component/state/name
-    constraint: CHASSIS
+paths:
+  # interface configuration
+  /interfaces/interface/config/description:
+  /interfaces/interface/config/enabled:
+  # name of chassis component
+  /components/component/state/name:
+    platform_type: "CHASSIS"
 
-OCRPCs:
-  - gnmi
-    method_name: gnmi.gNMI.Set.SetRequest.Update.union_replace
-    method_name: gnmi.gNMI.Subscribe
+rpcs:
+  gnmi:
+    gNMI.Set:
+      union_replace: true
+    gNMI.Subscribe:
+      on_change: true
 ```
 
 ## Required DUT platform
--- a/doc/test-requirements-template.md
+++ b/doc/test-requirements-template.md
@@ -4,7 +4,6 @@ about: Use this template to document the requirements for a new test to be imple
 title: ''
 labels: enhancement
 assignees: ''
-
 ---
 
 # Instructions for this template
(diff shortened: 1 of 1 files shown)


wenovus (commented):
The structure of the YAML LGTM. I'll work on a parser based on https://github.com/yuin/goldmark in the next few days.

robshakir (approved):
This LGTM.

One question -- is there a need to have >1 `README.md` map to a single feature profile grouping? It seems like there might be. Should we add a top-level field in the YAML with something like `feature_profile: bgp_addpath` to allow for there to be multiple README.mds to map back to that featureprofile?

A second set of considerations just on the file layout:

To allow this to be programmatically extracted and unblock https://github.com/openconfig/featureprofiles/pull/2793 being submitted, should we require that:

1) there is a well-known heading that can be found once per README that contains a single codeblock that is parsed as YAML?
2) the contents must validate according to the parser that @wenovus mentioned above?

this would then remove the need for `feature.textproto` in this repo I believe.

r.

robshakir:
> The structure of the YAML LGTM. I'll work on a parser based on https://github.com/yuin/goldmark in the next few days.

Nice, this looks a neat package.

wenovus (approved):

wenovus (commented):
Maybe mention that "the format of this block MUST follow the template". This is the function I'm using to parse this:
```go
// Parse extracts sorted OpenConfig Path and RPC Coverage from an FNT README.
//
// The first yaml code block after a heading line named exactly as
// "OpenConfig Path and RPC Coverage" will be used. Any other code blocks are
// ignored.
//
// Expected markdown format:
//
//      ## OpenConfig Path and RPC Coverage
//
//      ```yaml
//      paths:
//        /interfaces/interface/config/description:
//        /interfaces/interface/config/enabled:
//        /components/component/state/name:
//          platform_type: "CHASSIS"
//
//      rpcs:
//        gnmi:
//          gNMI.Set:
//            union_replace: true
//          gNMI.Subscribe:
//            on_change: true
//      ```
//
// Note: For `rpcs`, only the RPC name and methods are validated. Any
// attributes defined below RPC methods (e.g. union_replace) are not validated.
func Parse(source []byte) (*ppb.OCPaths, *rpb.OCRPCs, error) {
```

Once we've migrated, I think it makes sense to enforce this across all test READMEs. Do you see a way of identifying which READMEs should be subject to this check?

dplore:
> Maybe mention that "the format of this block MUST follow the template". This is the function I'm using to parse this:
> 
> ```go
> // Parse extracts sorted OpenConfig Path and RPC Coverage from an FNT README.
> //
> // The first yaml code block after a heading line named exactly as
> // "OpenConfig Path and RPC Coverage" will be used. Any other code blocks are
> // ignored.
> //
> // Expected markdown format:
> //
> //      ## OpenConfig Path and RPC Coverage
> //
> //      ```yaml
> //      paths:
> //        /interfaces/interface/config/description:
> //        /interfaces/interface/config/enabled:
> //        /components/component/state/name:
> //          platform_type: "CHASSIS"
> //
> //      rpcs:
> //        gnmi:
> //          gNMI.Set:
> //            union_replace: true
> //          gNMI.Subscribe:
> //            on_change: true
> //      ```
> //
> // Note: For `rpcs`, only the RPC name and methods are validated. Any
> // attributes defined below RPC methods (e.g. union_replace) are not validated.
> func Parse(source []byte) (*ppb.OCPaths, *rpb.OCRPCs, error) {
> ```
> 
> Once we've migrated, I think it makes sense to enforce that this block exists across all test READMEs. Do you see a way of identifying which READMEs should be subject to this check?

I think checking changed README.md files would be a good idea.  This will help us enforce updating any readme that is touched in a PR.    

For untouched readmes, we can schedule a time to go through and reformat them.  Maybe even generate the list of paths for such README's from the test results.

PR #2865, opened 2024-03-30
Validation Script for OC Path and RPC Coverage in READMEs

wenovus:
* `tools/internal/mdocspec/*` package parses ocpath.proto and ocrpcs.proto from yaml according to the format shown below by mdocspec.Parse's documentation.
* `tools/validate_readme_spec/validate_readme_spec.go` script is added that can parse and then validate OC Paths and RPCs according to existing repo states in github.com/openconfig.

```go
// Parse extracts sorted OpenConfig Path and RPC Coverage from an FNT README.
//
// The first yaml code block after a heading line named exactly as
// "OpenConfig Path and RPC Coverage" will be used. Any other code blocks are
// ignored.
//
// Expected markdown format:
//
//      ## OpenConfig Path and RPC Coverage
//
//      ```yaml
//      paths:
//        /interfaces/interface/config/description:
//        /interfaces/interface/config/enabled:
//        /components/component/state/name:
//          platform_type: "CHASSIS"
//
//      rpcs:
//        gnmi:
//          gNMI.Set:
//            union_replace: true
//          gNMI.Subscribe:
//            on_change: true
//      ```
//
// Note: For `rpcs`, only the RPC name and methods are validated. Any
// attributes defined below RPC methods (e.g. union_replace) are not validated.
func Parse(source []byte) (*ppb.OCPaths, *rpb.OCRPCs, error) {
```


wenovus:
This validation routine is based on the template README style being defined in https://github.com/openconfig/featureprofiles/pull/2863

.github/workflows/readme_oc_path_and_rpc.yml
```diff
@@ -0,0 +1,36 @@
...
+      run: |
+        if go run tools/validate_readme_spec/validate_readme_spec.go -alsologtostderr tools/validate_readme_spec/invalid.md; then
+          echo "Validation passed, but failure expected"
+          exit 1
+        fi
+        rm -rf tmp
+
+    - name: Validate Test READMEs
+      run: |
+        # TODO: Add READMEs here that have converted to the new format.
```
dplore, line 36: With a syntax change in it's yaml section, this readme could be added: 

https://github.com/openconfig/featureprofiles/blob/main/feature/bgp/policybase/otg_tests/import_export_multi/README.md

dplore: In a future PR, we could change the default to parse README with changes perhaps?  Or introduce a flag as an exception in metadata.textproto for all old README's that haven't had a yaml stanza added.

tools/internal/mdocspec/md_test.go
```diff
@@ -0,0 +1,399 @@
...
+
+rpcs:
+  fooi:
+    fooi.Set:
+      union_replace: true
+    fooi.Subscribe:
+      on_change: true
+`,
+	}, {
+		desc: "two-blocks-same-heading-first-language-not-specified",
```
dplore, line 265: I think I would like to see the language required, just for pretty rendering for humans to read.  WDYT?

tools/internal/mdocspec/md_test.go
```diff
@@ -0,0 +1,399 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/yuin/goldmark"
+)
+
+func TestRenderer(t *testing.T) {
+	tests := []struct {
```
dplore, line 12: should one of the tests take it's input from the test-requirements-template.md ?    This would help remind us to update this check when we change the template in the future?

tools/internal/mdocspec/ocspec.go
```diff
@@ -0,0 +1,126 @@
...
+	"gopkg.in/yaml.v3"
+
+	ppb "github.com/openconfig/featureprofiles/proto/ocpaths_go_proto"
+	rpb "github.com/openconfig/featureprofiles/proto/ocrpcs_go_proto"
+)
+
+// ErrNotFound indicates the user was not found or is unknown.
+var ErrNotFound = errors.New(`did not detect valid "OpenConfig Path and RPC Coverage" yaml, please see https://github.com/openconfig/featureprofiles/blob/main/doc/test-requirements-template.md for example`)
+
+// Parse extracts sorted OpenConfig Path and RPC Coverage from an FNT README.
```
dplore, line 21:
```suggestion
// Parse extracts sorted OpenConfig Path and RPC Coverage from a featureprofiles README.
```

tools/internal/mdocspec/ocspec.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/yuin/goldmark"
+	"golang.org/x/exp/maps"
+	"gopkg.in/yaml.v3"
+
+	ppb "github.com/openconfig/featureprofiles/proto/ocpaths_go_proto"
+	rpb "github.com/openconfig/featureprofiles/proto/ocrpcs_go_proto"
+)
+
+// ErrNotFound indicates the user was not found or is unknown.
+var ErrNotFound = errors.New(`did not detect valid "OpenConfig Path and RPC Coverage" yaml, please see https://github.com/openconfig/featureprofiles/blob/main/doc/test-requirements-template.md for example`)
```
dplore, line 19: OpenConfig Path and RPC Coverage should probably instead refer to the constant `OCSpecHeading`

Maybe https://github.com/openconfig/featureprofiles/blob/main/doc/test-requirements-template.md should be in a constant as well?

tools/internal/mdocspec/md.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+const (
+	// OCSpecHeading is the MarkDown heading that MUST precede the yaml
+	// section containing the OC path and RPC listing.
+	OCSpecHeading = "OpenConfig Path and RPC Coverage"
+	// OCSpecHeadingSearchN is the number of previous siblings in the
+	// MarkDown AST to search for OCSpecHeading in order to establish
+	// whether the yaml block might be the OC path and RPC listing.
+	OCSpecHeadingSearchN = 5
```
dplore, line 19: Is this the depth of the tree to search?

tools/internal/mdocspec/md_test.go
wenovus: This is the intended behaviour, improved error desc string to make clearer.

.github/workflows/readme_oc_path_and_rpc.yml
wenovus: I will make the change in a follow-up PR to validate the file.

I propose to parse all READMEs all the time, since that's the only way to know the repo is in good state all the time.

I think the best way to identify READMEs is to parse all `README.md` files under `/features` repo, which is what `addrundata` also does. `addrundata` also validates that whenever a `_test.go` file exists, `README.md` must also exist, which helps our check as well.

In fact, I propose only to tolerate `ErrNotFound` any README that does not have an accompanying `_test.go` file. This way we won't need allow-lists either.

tools/internal/mdocspec/md_test.go
wenovus: I already have an integration tests taking in `test-requirements-template.md` as part of CI.

tools/internal/mdocspec/ocspec.go
wenovus: Done.

The template file is harder to make a constant. I think we should strive to not change it.

tools/internal/mdocspec/md.go
wenovus: The goldmark parser I'm using creates a hierarchy of MD nodes, and in general there can be several sibling nodes in between the OCSpecHeading and the actual yaml block. This specifies the number of previous sibling nodes to search for OCSpecHeading before giving up.

.github/workflows/readme_oc_path_and_rpc.yml
dplore: I would like to cover README yaml path validation even when the README is standalone / no _test.go file.

tools/internal/mdocspec/md.go
dplore: Hrm, sorry I still don't follow.   In any case, I expect the yaml block to appear immediately after / within the OCSpecHeading.  There could be 10+  headings in a big or highly parameterized test.  So maybe make this number 20?

wenovus: Thanks for the suggestion, I changed the logic to search from the heading rather than from the YAML block. This way we can just search until the next heading of the same or lower level or the end of the document, which is cleaner since we can remove this constant.

.github/workflows/readme_oc_path_and_rpc.yml
wenovus: I'm still proposing to validate all README.md under /features **where the heading exists**, no matter whether a _test.go file is present. However, I'm proposing that if a _test.go file doesn't exist, we don't force the issue. The benefit of this is that we don't need an allow-list.

WDYT? -- an allow-list or don't force the test author to write such a section if a _test.go file doesn't exist?

dplore: I feel like we should require the section to exist without exceptions.    (It requires converting existing README before merging these checks in).  Worst case an author can add an empty section with a TODO comment?

wenovus: Alright, allow-list approach implemented here: https://github.com/openconfig/featureprofiles/pull/2881

dplore: thanks!

tools/internal/mdocspec/md.go
```diff
@@ -41,53 +37,51 @@ func (r *yamlRenderer) Render(w io.Writer, source []byte, n ast.Node) error {
...
+		if c, ok := yamlCodeBlock(source, next); ok {
+			l := c.Lines().Len()
+			for i := 0; i != l; i++ {
+				line := c.Lines().At(i)
+				if _, err := w.Write(line.Value(source)); err != nil {
+					return ast.WalkStop, err
+				}
 			}
+			// Stop after finding the first such YAML block.
+			return ast.WalkStop, nil
```
dplore, line 83: Should we continue parsing and throw an error if there is more than one YAML block?  ie: there must be exactly one YAML block in the ocSpecHeading?   I am not sure we really care if there are other yaml blocks in other headings.

tools/internal/mdocspec/ocspec.go
```diff
@@ -121,6 +122,9 @@ func parseYAML(source []byte) (*ppb.OCPaths, *rpb.OCRPCs, error) {
 			MethodName: methods,
 		}
 	}
+	if !hasMethod {
+		return nil, nil, fmt.Errorf("mdocspec: YAML does not have least one RPC method specified")
```
dplore, line 126: nit, but in test format, we use want before got?

```suggestion
		return nil, nil, fmt.Errorf("mdocspec: Want at least one RPC method in the YAML, but got 0.")
```

tools/internal/ocpaths/clone_public.go
```diff
@@ -18,6 +19,10 @@ func ClonePublicRepo(downloadPath, branch string) (string, error) {
 	}
 	publicPath := filepath.Join(downloadPath, "public")
 
+	if _, err := os.Stat(publicPath); err == nil { // If NO error
```
dplore, line 22: could this cause detection of an old/stale repo?

tools/internal/ocrpcs/ocrpcs.go
```diff
@@ -21,6 +21,10 @@ func cloneAPIRepo(downloadPath, api string) (string, error) {
 	}
 	repoPath := filepath.Join(downloadPath, api)
 
+	if _, err := os.Stat(repoPath); err == nil { // If NO error
```
dplore, line 24: same comment as above re: stale repo copies

tools/internal/mdocspec/md.go
wenovus: I think it's ok to ignore YAML blocks aside from the first one. I don't think it's realistic to achieve the goal of making sure that the author doesn't put multiple coverage-related blocks into the same README. Even if they did, the duplicate block should be caught in review anyways, so I don't think this will be an actual issue. WDYT?

tools/internal/mdocspec/ocspec.go
wenovus: In Go tests we generally have got before want: https://google.github.io/styleguide/go/decisions.html#got-before-want. However since this is not a test I didn't consider this point. Is there some other value in changing the wording here?

tools/internal/ocpaths/clone_public.go
wenovus: Yes, the purpose is to avoid having to download every single time. Let me add more documentation on this, thanks for catching.

tools/internal/ocrpcs/ocrpcs.go
wenovus: Same purpose as above. Go doc comments updated.

tools/internal/mdocspec/md.go
dplore: Sure, good enough as is

tools/internal/mdocspec/ocspec.go
dplore: Updated wording works for me!

dplore (dismissed):

wenovus:
@dplore Need another approval due to slight doc fix.

PR #2873, opened 2024-04-03
Fix RT-7.1: BGP default policies test failure for Arista

prakashbadri-arista:
The test was failing for the following reasons
- Arista specific deviations were not getting applied when configuring ISIS
- BGP neighbors were flapping because the OTG was sending update messages with pathId not set in the BGP NLRI even though add path capability was negotiated in the test

Have fixed both the issues and it is passing now


Ankur19 (approved):
LGTM!

AddPath is required for all.

Other changes are gated by deviation.

PR #2880, opened 2024-04-05
TE-3.8 Drain Test: Dataplane

sachendras:
1. Update the testresgistry to remove duplicate entries
2. Add Dataplane Drain test to validate Forwarding-Viable behavior as in the [specs](https://github.com/openconfig/public/blob/master/release/models/interfaces/openconfig-if-sdn-ext.yang#L43)


rohit-rp:
@sachendras Please resolve the conflicts if still needed

PR #2887, opened 2024-04-09
TE-17.1 | Add deviations & changes

arvbaska1:
Below changes for TE-17.1 test

-  Changes to policy-forwarding with deviation pf_require_match_default_rule & pf_require_sequential_order_pbr_rules
-  IS-IS changes to set v6 AFI under interface and use deviation isis_single_topology_required
-  program gRIBI entries with NH with IP, and use static arp with gribi_mac_override_with_static_arp
-  Increase time  after StartProtocols in sendTraffic since everytime there is PushConfig/StartProtocols underlay IS-IS/BGP sessions flap, so need more time for BGP session to be est, before sending traffic.
-  Use loopback interface name for bgp local-address leaf, since in OC leaf is union of ip-address or interface name
-  Add UDP header in IPinIP flows to add entropy for loadbalancing traffic


AmrNJ (changes requested):
Fails while configuring interface port speed. Port speed is set to 100g.

AmrNJ (changes requested):
Fails while configuring interface port speed. Port speed is set to 100g.

arvbaska1:
> Fails while configuring interface port speed. Port speed is set to 100g.

I have modified condition to skip configuring port speed etc. based on dut.Vendor

arvbaska1:
> Error: "I0425 01:42:32.421188 30146 vrf_policy_driven_te_test.go:1606] truncated dump file; tried to read 4 file header bytes, only got 0" , with the packet capture.

This should pass with latest image which has a loadbalancing fix.

ram-mac:
The port speed error is not seen anymore and with this pull i see that the tests are running to completion, however couple of validations related to decap are failing which needs investigation. Out of 43 sub-tests run in total; 41 tests are passing and 2 are failing.

ram-mac (approved):
LGTM, 
Will merge it

ram-mac:
> Error: "I0425 01:42:32.421188 30146 vrf_policy_driven_te_test.go:1606] truncated dump file; tried to read 4 file header bytes, only got 0" , with the packet capture.

This is not seen in my latest run and the test is passing.

PR #2888, opened 2024-04-09
gNOI-4.1 comparison function

fmolinar:
Adding function to accurately compare configs, taking into account default values for the tests:

TestPushAndVerifyBGPConfig

TestPushAndVerifyInterfaceConfig

### DIFF
--- a/feature/gnoi/os/tests/osinstall/osinstall_test.go
+++ b/feature/gnoi/os/tests/osinstall/osinstall_test.go
@@ -442,12 +442,7 @@ func TestPushAndVerifyInterfaceConfig(t *testing.T) {
 
 	t.Logf("Fetch interface config from the DUT using Get RPC and verify it matches with the config that was pushed earlier")
 	if val, present := gnmi.LookupConfig(t, dut, dc).Val(); present {
-		if reflect.DeepEqual(val, in) {
-			t.Logf("Interface config Want and Got matched")
-			fptest.LogQuery(t, fmt.Sprintf("%s from Get", dutPort), dc, val)
-		} else {
-			t.Errorf("Config %v Get() value not matching with what was Set()", dc)
-		}
+		compareStructs(t, in, val)
 	} else {
 		t.Errorf("Config %v Get() failed", dc)
 	}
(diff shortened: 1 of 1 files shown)


LimeHat:
Could you please elaborate on the reasoning behind this change?

fmolinar:
hi @LimeHat ,

Given the tests TestPushAndVerifyInterfaceConfig and TestPushAndVerifyBGPConfig, they compare the configuration that is going to be send to the device with the configuration that is retrieved from the device after the update.

The current implementation of reflect.DeepEqual(val, in) doesn't take into account default values that are returned from gnmi.LookupConfig(...). Thus, the test may give a false positive by failing a comparison that should've passed if we are looking at the field changed that were pushed vs what was retrieved.

There is a gnmi function PopulateDefaults() that populates all default values, however, it also populates some unsupported leafs, making the update call fail. 

My proposal is to iterate over the struct that we are going to push (want), and evaluate it against what we got after we push the changes (got) and ignore unset fields.

let me know if you want me to expand or provide more data.

Thank you!
Fernando

fmolinar:
hi @LimeHat ,

Given the tests TestPushAndVerifyInterfaceConfig and
TestPushAndVerifyBGPConfig, they compare the configuration that is going to
be send to the device with the configuration that is retrieved from the
device after the update.

The current implementation of reflect.DeepEqual(val, in) doesn't take into
account default values that are returned from gnmi.LookupConfig(...). Thus,
the test may give a false positive by failing a comparison that should've
passed if we are looking at the field changed that were pushed vs what was
retrieved.

There is a gnmi function PopulateDefaults() that populates all default
values, however, it also populates some unsupported leafs, making the
update call fail.

My proposal is to iterate over the struct that we are going to push (want),
and evaluate it against what we got after we push the changes (got) and
ignore unset fields.

let me know if you want me to expand or provide more data.

Thank you!
Fernando

dplore:
@fmolinar is this PR still active?  If yes, can you provide run logs for this and share via gist or a https://partnerissuetracker.corp.google.com issue?

PR #2890, opened 2024-04-11
Updating tests to add interface config

prinikasn:
Updating the ZR tests to add interface config and the right deviations.


feature/platform/transceiver/tests/zr_fec_uncorrectable_frames_test/zr_fec_uncorrectable_frames_test.go
```diff
@@ -38,6 +40,42 @@ func TestMain(m *testing.M) {
...
+			return compName
+		}
+		if comp.GetParent() == "" {
+			t.Fatalf("Recursive optical channel lookup failed for port: %s, parent of component %s not found.", p.Name(), compName)
+		}
+		compName = comp.GetParent()
+	}
+}
+
+func interfaceConfig(t *testing.T, dut *ondatra.DUTDevice, dp *ondatra.Port) {
```
dplore, line 65: This function should also move to a shared library, probably at `internal/cfgplugins/interface.go`

LimeHat: Perhaps `NewOCInterface` can be reused https://github.com/openconfig/featureprofiles/blob/2ef8bf6c3b9c6b911b97997559b21c3e1054e9a5/internal/attrs/attrs.go#L108-L111

PR #2891, opened 2024-04-12
Updates to Transceiver-3:zr_firmware_version_test 

hattikals:
- In the previous logic the code was to get the optical component name by traversing through the parents of the PORT component, this was leading to no results. Hence changed it to find the OPTICAL_CHANNEL comp first and match its parent against HARDWARE_PORT of the interface.
- Current flow :                     PORT <-- TRANSCIEVER <-- OPTICAL_CHANNEL

### DIFF
--- a/internal/components/components.go
+++ b/internal/components/components.go
@@ -213,6 +213,12 @@ func OpticalChannelComponentFromPort(t *testing.T, dut *ondatra.DUTDevice, p *on
 		}
 	}
 	transceiverName := gnmi.Get(t, dut, gnmi.OC().Interface(p.Name()).Transceiver().State())
+	if transceiverName == "" {
+		t.Fatalf("Associated Transceiver for Interface (%v) not found!", p.Name())
+	}
 	opticalChannelName := gnmi.Get(t, dut, gnmi.OC().Component(transceiverName).Transceiver().Channel(0).AssociatedOpticalChannel().State())
+	if opticalChannelName == "" {
+		t.Fatalf("Associated Optical Channel for Transceiver (%v) not found!", transceiverName)
+	}
 	return opticalChannelName
 }
--- a/internal/components/components.go
+++ b/internal/components/components.go
@@ -212,18 +212,7 @@ func OpticalChannelComponentFromPort(t *testing.T, dut *ondatra.DUTDevice, p *on
 			t.Fatal("Manual Optical channel name required when deviation missing_port_to_optical_channel_component_mapping applied.")
 		}
 	}
-	compName := gnmi.Get(t, dut, gnmi.OC().Interface(p.Name()).HardwarePort().State())
-	for {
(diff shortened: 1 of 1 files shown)


ejbrever:
A couple things:

1. You might have an old version of the tests. The func you are modifying is in this file. Can you please update:
https://github.com/openconfig/featureprofiles/blob/main/internal/components/components.go#L203

2. I agree the current approach could probably be improved. Looking at the code the "PORT" referenced above and what is used in that function is the "interface" name: 
interfaces/interface/[name]

Technically, to traverse through OpenConfig to an OpticalChannel you'd have to go through the terminal-device logical-channels 
Screenshot 2024-09-12 at 9 21 47 AM

But I think a simpler option is to utilize this path which is a reference to a transceiver:
interfaces/interface/[name]/state/transceiver

If we read that leaf you'd then have the transceiver component to look up. We could then get the subcomponents of the transceiver component to find the optical-channel component.

I don't think this is too far from your current approach, but avoids creating any expectations of interface and component names being equal.

hattikals:
@ejbrever I have changed the code to directly use AssociatedOpticalChannel Leaf.

internal/components/components.go
```diff
@@ -212,18 +212,7 @@ func OpticalChannelComponentFromPort(t *testing.T, dut *ondatra.DUTDevice, p *on
...
-		}
-		if comp.GetType() == oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_OPTICAL_CHANNEL {
-			return compName
-		}
-		if comp.GetParent() == "" {
-			t.Fatalf("Recursive optical channel lookup failed for port: %s, parent of component %s not found.", p.Name(), compName)
-		}
-		compName = comp.GetParent()
-	}
+	transceiverName := gnmi.Get(t, dut, gnmi.OC().Interface(p.Name()).Transceiver().State())
```
ejbrever, line 215: can you add a validation check here?

if transceiverName == "" {
  t.Fatalf("associated transceiver for interface (%v) not specified.", p.Name())
}

internal/components/components.go
```diff
@@ -212,18 +212,7 @@ func OpticalChannelComponentFromPort(t *testing.T, dut *ondatra.DUTDevice, p *on
...
+	opticalChannelName := gnmi.Get(t, dut, gnmi.OC().Component(transceiverName).Transceiver().Channel(0).AssociatedOpticalChannel().State())
```
ejbrever, line 216: can you add a validation check here?

if opticalChannelName == "" {
  t.Fatalf("associated optical channel for transceiver (%v) not specified.", transceiverName)
}

ejbrever (commented):
Thanks, just recommended a couple validation checks.

internal/components/components.go
hattikals: Added

internal/components/components.go
hattikals: Added

ejbrever (approved):
Thanks for the changes!

PR #2892, opened 2024-04-12
Updates to   Transceiver-7: zr_inventory_test

hattikals:
- In the previous logic the code was to get the optical component name by traversing through the parents of the PORT component, this would lead to no results. Hence changed it to, find the OPTICAL_CHANNEL comp first and match its parent against HARDWARE_PORT of the interface.
- Current flow : PORT <-- TRANSCIEVER <-- OPTICAL_CHANNEL

### DIFF
--- a/feature/platform/transceiver/tests/zr_inventory_test/zr_inventory_test.go
+++ b/feature/platform/transceiver/tests/zr_inventory_test/zr_inventory_test.go
@@ -141,18 +141,27 @@ func opticalChannelComponentFromPort(t *testing.T, dut *ondatra.DUTDevice, p *on
 			t.Fatal("Manual Optical channel name required when deviation missing_port_to_optical_channel_component_mapping applied.")
 		}
 	}
-	compName := gnmi.Get(t, dut, gnmi.OC().Interface(p.Name()).HardwarePort().State())
+	comps := gnmi.LookupAll(t, dut, gnmi.OC().ComponentAny().State())
+	hardwarePortCompName := gnmi.Get(t, dut, gnmi.OC().Interface(p.Name()).HardwarePort().State())
+	for _, comp := range comps {
+		comp, ok := comp.Val()
+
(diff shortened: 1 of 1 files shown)


feature/platform/transceiver/tests/zr_inventory_test/zr_inventory_test.go
```diff
@@ -141,18 +141,27 @@ func opticalChannelComponentFromPort(t *testing.T, dut *ondatra.DUTDevice, p *on
 			t.Fatal("Manual Optical channel name required when deviation missing_port_to_optical_channel_component_mapping applied.")
 		}
 	}
-	compName := gnmi.Get(t, dut, gnmi.OC().Interface(p.Name()).HardwarePort().State())
```
dplore, line 144: instead of fixing this here, I think this should be refactored to use OpticalChannelComponentFromPort from  [github.com/openconfig/featureprofiles/internal/components](http://github.com/openconfig/featureprofiles/internal/components)

@yiwenhu-g WDYT?

feature/platform/transceiver/tests/zr_inventory_test/zr_inventory_test.go
```diff
@@ -141,18 +141,27 @@ func opticalChannelComponentFromPort(t *testing.T, dut *ondatra.DUTDevice, p *on
...
+			return comp.GetName()
 		}
-		if comp.GetType() == oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_OPTICAL_CHANNEL {
-			return compName
+	}
+	t.Fatalf("No interface to optical-channel mapping found for interface = %v", p.Name())
+	return ""
+}
+
+func isSubCompOfHardwarePort(t *testing.T, dut *ondatra.DUTDevice, parentHardwarePortName string, comp *oc.Component) bool {
```
dplore, line 157: this function is probably better placed in [/internal/components](http://github.com/openconfig/featureprofiles/internal/components)

hattikals:
Recent merge had the similar changes , hence closing this pull.

PR #2893, opened 2024-04-12
Updates to Transceiver-9: zr_laser_bias_current_test

hattikals:
- In the previous logic the code was to get the optical component name by traversing through the parents of the PORT component, this would lead to no results. Hence changed it to, find the OPTICAL_CHANNEL comp first and match its parent against HARDWARE_PORT of the interface.
- Current flow : PORT <-- TRANSCIEVER <-- OPTICAL_CHANNEL


hattikals:
Closing this as recent merge had similar changes

PR #2901, opened 2024-04-16
   Fix static configuration and vrf selection policy configuration failure in TE-17.1 VRF selection policy driven TE

prakashbadri-arista:
The fix adds
- Arista deviation for static protocol name required for static route configuration
- Remove deletion of the policy forwarding parent path and instead do a replace when replacing cluster vrf selection policy with wan vrf selection policy


nandanarista (approved):
LGTM

self-maurya (approved):

PR #2902, opened 2024-04-17
RT-1.14: Added code to proceed with BGP config based on interface status.

mohanasm:
Updated code to proceed with BGP config based on interface status.

### DIFF
--- a/feature/bgp/ate_tests/bgp_long_lived_graceful_restart/bgp_long_lived_graceful_restart_test.go
+++ b/feature/bgp/ate_tests/bgp_long_lived_graceful_restart/bgp_long_lived_graceful_restart_test.go
@@ -285,17 +285,14 @@ func configureDUT(t *testing.T, dut *ondatra.DUTDevice) {
 	}
 }
 
-func verifyPortsUp(t *testing.T, dev *ondatra.Device) int {
-	var portDownCnt int
+func verifyPortsUp(t *testing.T, dev *ondatra.Device) {
 	t.Helper()
 	for _, p := range dev.Ports() {
 		status := gnmi.Get(t, dev, gnmi.OC().Interface(p.Name()).OperStatus().State())
 		if want := oc.Interface_OperStatus_UP; status != want {
-			portDownCnt++
-			t.Errorf("%s Status: got %v, want %v", p, status, want)
+			t.Fatalf("%s Status: got %v, want %v", p, status, want)
 		}
 	}
-	return portDownCnt
 }
 
 type bgpNeighbor struct {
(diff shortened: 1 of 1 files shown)


feature/experimental/bgp/ate_tests/bgp_long_lived_graceful_restart/bgp_long_lived_graceful_restart_test.go
```diff
@@ -294,14 +294,17 @@ func configureDUT(t *testing.T, dut *ondatra.DUTDevice) {
...
 }
 
-func verifyPortsUp(t *testing.T, dev *ondatra.Device) {
+func verifyPortsUp(t *testing.T, dev *ondatra.Device) int {
+	var portDownCnt int
 	t.Helper()
 	for _, p := range dev.Ports() {
 		status := gnmi.Get(t, dev, gnmi.OC().Interface(p.Name()).OperStatus().State())
 		if want := oc.Interface_OperStatus_UP; status != want {
+			portDownCnt++
```
LimeHat, line 303: It seems that replacing `t.Error()` with `t.Fatal()` is the easier approach if you want to stop the test when a port is not up. No need to introduce additional counters and propagate that state.

feature/experimental/bgp/ate_tests/bgp_long_lived_graceful_restart/bgp_long_lived_graceful_restart_test.go
```diff
@@ -949,12 +956,16 @@ func TestTrafficWithGracefulRestartLLGR(t *testing.T) {
...
-			bgpIxPeer = configureATENewPeers(t, topo, ateIntfList)
+			if len(ateIntfList) > 0 {
+				bgpIxPeer = configureATENewPeers(t, topo, ateIntfList)
+			}
 		})
 
 		t.Run("Remove newly added 5 BGP peers", func(t *testing.T) {
 			removeNewPeers(t, dut, dutNbrs)
-			removeATENewPeers(t, topo, bgpIxPeer)
+			if len(ateIntfList) > 0 {
```
LimeHat, line 966: From what I see, `ateIntfList` is initialized only once during the test execution and not modified after that. Why do we need multiple additional checks on this variable?

LimeHat (changes requested):

feature/experimental/bgp/ate_tests/bgp_long_lived_graceful_restart/bgp_long_lived_graceful_restart_test.go
mohanasm: Agree with you. Updated the file.

feature/experimental/bgp/ate_tests/bgp_long_lived_graceful_restart/bgp_long_lived_graceful_restart_test.go
mohanasm: Not needed and removed it.

ram-mac (approved):
Changes looks good

rohit-rp (approved):

PR #2911, opened 2024-04-22
gnmi-1.13: Optics Power And Bias Current Test Deviation removal

hattikals:
Since Transceiver thresholds are now supported on Juniper platform, removing the transceiver_thresholds_unsupported deviation for Juniper.


Swetha-haridasula (approved):
LGTM

PR #2926, opened 2024-04-25
Deviation removal from gNMI-1.10, gNOI-4.1, gNOI-2.1

hattikals:
Removing following deviations from scripts for Juniper, as these features are now supported on Juniper Devices

- gNMI-1.10: telemetry_basic_check_test.go -> sw_version_unsupported
- gNOI-4.1 : os_install_test.go -> sw_version_unsupported
- gNOI-2.1: packet_link_qualification_test.go -> skip_plq_interface_oper_status_check

### DIFF
--- a/feature/gnmi/otg_tests/telemetry_basic_check_test/metadata.textproto
+++ b/feature/gnmi/otg_tests/telemetry_basic_check_test/metadata.textproto
@@ -42,7 +42,6 @@ platform_exceptions: {
     vendor: JUNIPER
   }
   deviations: {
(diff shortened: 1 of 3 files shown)


Swetha-haridasula (approved):
LGTM

LimeHat:
The code should also have been modified to remove obsolete if/else conditions, since `sw_version_unsupported` was used only by a single vendor.
Additionally, as `sw_version_unsupported` is not used in other tests, it should've been removed from metadata.proto as well.

cc @dplore @cfernanz

PR #2928, opened 2024-04-26
RT-1.33: Fixing value for deviation flag

trathod1:
Value for deviation flag `default_import_export_policy` is set to `false` in the metadata file, and default value for vendor not using this deviation is also `false`.

Due to this `else` part of the code which does not need this deviation never gets executed. Also restoring the original prefix count value applicable for non-deviation scenario.

“This code is a Contribution to the OpenConfig Feature Profiles project (“Work”) made under the Google Software Grant and Corporate Contributor License Agreement (“CLA”) and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia’s intellectual property are granted for any other purpose. This code is provided on an “as is” basis without any warranties of any kind.”


LimeHat:
ref #2884 that introduced incorrect deviation usage

NishaSadhasivam-work (approved):
LGTM!

PR #2932, opened 2024-04-29
rename the file as per naming convention

nsadhasivam:
rename the file as per naming convention


feature/bgp/policybase/otg_tests/import_export_multi_test/import_export_multi_test.go
```diff
@@ -0,0 +1,535 @@
+// Copyright 2023 Google LLC
```
self-maurya, line 1: If this test is just moved form one dir to another then can you move this file as well;

Create the new dir then move everything from one dir to another.

`mv dir/* dir_test/`

dplore:
This is now obsolete I think:

https://github.com/openconfig/featureprofiles/tree/main/feature/bgp/policybase/otg_tests/import_export_multi_test

PR #2935, opened 2024-04-30
Deviation:gNMI-1.18 backplane_facing_capacity_unsupported removal

cakhil45:
1.Deviation removal for backplane capacity in juniper devices. 
2.Added check for if the SIB components are really present in the device chassis.
3.Changed the condition to v1!=v2 as per ReadMe for ts1,ts2


Swetha-haridasula (approved):
lgtm

PR #2938, opened 2024-04-30
RT-1.23 : Added extended encoding on OTG ipv6 peers

cprabha:
Hi,

I have added Added extended encoding on OTG ipv6 peers.


ram-mac (approved):
LGTM;
Passes for Juniper and other vendor. So approving the PR.

PR #2947, opened 2024-05-02
Include HW name in nosimage.proto

wenovus:


wenovus:
@earies for comments

dplore:
This looks good enough to me.  I'm not seeing a better idea at the moment to record hardware

dplore (approved):

PR #2953, opened 2024-05-03
[FNT] Forwarding viability Test aggregate_all_not_viable_test

rahulsy26:
**Summary:**
This code is to Test forwarding-viable with LAG and routing

Ensure that when all LAG member become set with forwarding-viable == FALSE.

forwarding-viable=false impact only transmit traffic on all member port.
All member ports set with forwarding-viable=false can receive all type of traffic and forward it normally (same as with forwarding-viable=true)
ISIS adjacency is established on LAG port w/ all member set to forwarding-viable == FALSE
Traffic that normally egress LAG with all members set to forwarding-viable == FALSE is forwarded by the next best egress interface/LAG.


feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
```diff
@@ -52,19 +52,19 @@ Ensure that when **all LAG member** become set with forwarding-viable == FALSE.
...
-    - Ensure there is no traffic transmitted out of DUT ports 3-7
-    - ensure that traffic is received on all port2-7 and delivered to ATE port1
+    - Ensure there is no traffic transmitted out of DUT ports 3-6
+    - ensure that traffic is received on all port2-6 and delivered to ATE port1
     - ensure there are no packet losses in steady state (no congestion).
     - Ensure there is no traffic received on DUT LAG_3
     - Ensure there is no traffic transmitted on DUT LAG_3
   - Disable/deactive laser on ATE port2; All LAG_2 members are either down (port2) or
     set with forwarding-viable=FALSE
     - Ensure ISIS adjacency is UP on DUT LAG_2 and ATE LAG_2
```
LimeHat, line 65: This is outdated; the readme requires an update to match the latest spec (https://github.com/openconfig/public/blob/a2c7a09e619c56782b3f4495a065fd7c34ef1175/release/models/interfaces/openconfig-if-sdn-ext.yang#L69)

cc @rszarecki @sachendras @dplore

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
```diff
@@ -122,17 +122,35 @@ It is foreseen that implementation may drop ISIS adjacency if all members of LAG
 are set with forwarding-viable = FALSE. This scenario may be
 handled via the yet to be defined deviation `logicalInterfaceUPonNonViableAll`.
```
LimeHat, line 123: This is outdated; the readme requires an update to match the latest spec (https://github.com/openconfig/public/blob/a2c7a09e619c56782b3f4495a065fd7c34ef1175/release/models/interfaces/openconfig-if-sdn-ext.yang#L69)

cc @rszarecki @sachendras @dplore

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1002 @@
...
+		gnmi.Await(t, dut, gnmi.OC().Interface(aggID).AdminStatus().State(), 30*time.Second, oc.Interface_AdminStatus_UP)
+	}
+
+	for _, agg := range []*aggPortData{agg1, agg2, agg3} {
+		bgpPath := gnmi.OC().NetworkInstance(deviations.DefaultNetworkInstance(dut)).Protocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP").Bgp()
+		gnmi.Await(t, dut, bgpPath.Neighbor(agg.ateLoopbackV4).SessionState().State(), time.Minute, oc.Bgp_Neighbor_SessionState_ESTABLISHED)
+	}
+
+	t.Logf("ISIS cost of LAG_2 lower then ISIS cost of LAG_3 Test-01")
+	t.Run("Running Testcase for TestID RT-5.7.1.1", func(t *testing.T) {
```
LimeHat, line 196: These test ids are not found in the README

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1002 @@
...
+		rpl.SetImportPolicy([]string{acceptRoutePolicy})
+	}
+
+	for _, a := range []*aggPortData{agg1, agg2, agg3} {
+		bgpNbrV4 := bgp.GetOrCreateNeighbor(a.ateLoopbackV4)
+		bgpNbrV4.PeerGroup = ygot.String(pgName)
+		bgpNbrV4.PeerAs = ygot.Uint32(asn)
+		bgpNbrV4.Enabled = ygot.Bool(true)
+		bgpNbrV4T := bgpNbrV4.GetOrCreateTransport()
+		bgpNbrV4T.LocalAddress = ygot.String(dutLoopback.IPv4)
```
LimeHat, line 515: README says that interface addresses are used for IBGP and there's only one session (two if you want to do dual-stack)

https://github.com/openconfig/featureprofiles/blob/70b8fbdd36d45f666695495cf843571b25bc55ad/feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md?plain=1#L47

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1002 @@
...
+			e := i.GetOrCreateEthernet()
+			e.AggregateId = ygot.String(aggID)
+			if deviations.InterfaceEnabled(dut) {
+				i.Enabled = ygot.Bool(true)
+			}
+			if port.PMD() == ondatra.PMD100GBASEFR {
+				e.AutoNegotiate = ygot.Bool(false)
+				e.DuplexMode = oc.Ethernet_DuplexMode_FULL
+				e.PortSpeed = oc.IfEthernet_ETHERNET_SPEED_SPEED_100GB
+			}
```
LimeHat, line 332: Perhaps this should be a deviation?

cc @dplore

LimeHat (changes requested):

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
rahulsy26: Thanks @LimeHat for the review. I have updated the Readme file as well in same commit because of some changes in the test/topology and working to resolve another comment.

LimeHat: > I have updated the Readme file as well in same commit because of some changes in the test/topology and working to resolve another comment.

To clarify, my comment is not about the changes in the topology you made (reducing the port count). After the readme was written by @rszarecki, there was a change in the definition of how the devices should behave when all ports are in a non-viable state.

Specifically, the current expectation is that IS-IS and BGP adjacencies will be dropped (either immediately or after timing out on hello/dead timers). Hence I don't think it makes much sense to merge a test that validates outdated behavior, and we need to update the README first to describe the new behavior together with the code.

rahulsy26: Thanks @LimeHat for the clarification, I agreed with you, But to clarify the expectation here is when all the link of LAG is set with non-viable and ISIS/BGP must be UP by updating the Hello/Dead interval and make sure traffic is only receiving from server and process it to client(As TX must be Only Blocked).

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
rahulsy26: @rszarecki @dplore Looking for your comments. Please let me know if any addition is required in this test.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Changed configuration for BGP from Loopback to LAG interface IP.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
LimeHat: > But to clarify the expectation here is when all the link of LAG is set with non-viable and ISIS/BGP must be UP by updating the Hello/Dead interval and make sure traffic is only receiving from server and process it to client(TX only must be Blocked).

In my view, the following things should occur:
1) tx is blocked immediately
2) ISIS and BGP time out after the timer expires, and after that, there will be no RX to ISIS/BGP prefixes through this LAG 

It makes sense to validate the IS-IS transition from the `up` to `down` state after the hold timer expires to make sure that the new behavior (don't transmit all L3 packets, including the control plane packets) works as expected. If you simply increase the hold timer to a large value, this behavior is not going to be checked.

rahulsy26: Hi @LimeHat 

I agreed with you the ISIS is getting down after timer expired. But i am confused with below:

 Readme says:
     Disable/deactivate laser on ATE port2; All LAG_2 members are either down (port2) or set with forwarding- 
     viable=FALSE
      -Ensure ISIS adjacency is UP on DUT LAG_2 and ATE LAG_2 (It is getting down immediately as setting forwarding- 
     viable=FALSE on all LAG ports)

If the above is the expected behavior then the README.md required to update with isis Down instead of isis UP in this case right?

LimeHat: Yes, exactly; the readme was written before this behavior was standardized and hence requires an update.

dplore: > Yes, exactly; the readme was written before this behavior was standardized and hence requires an update.

+1, the README must be updated to reflect the current requirement which says that the DUT must stop sending all L3 packets when forwarding_viable is false.

Suggested updates 
  - RT-5.7.1.2: Verify forwarding-viable behavior on an aggregate interface with all members down or set with forwarding-viable=FALSE.
    - Ensure ISIS adjacency is UP on DUT LAG_2 and ATE LAG_2
    - Disable/deactive laser on ATE port2;  Now all LAG_2 members are either down (port2) or
    set with forwarding-viable=FALSE
    - Ensure there is no layer3 traffic transmitted out of  DUT ports 2-6 (LAG_2)
    - Ensure that traffic is received on all port3-6 and delivered to ATE LAG_1
    - Ensure there are no packet losses in steady state (no congestion) for
      traffic from ATE LAG_2 to ATE LAG_1 (pfx_1).
    - Ensure there are no packet losses in steady state (no congestion) for
      traffic from ATE LAG_1 to ATE LAG_3 (pfx_2, pfx3).
    - Ensure there is no traffic received on DUT LAG_3
    - Ensure that traffic from ATE port1 to pfx2, pfx3 are transmitted via DUT
      LAG3
    - Ensure that traffic from ATE port1 to pfx4 are discarded on DUT
    - Ensure that the ISIS adjacency times out on DUT LAG_2 and ATE LAG_2

(note I moved the verify ISIS UP to occur before disabling the final remaining LAG member port)

dplore (dismissed):

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
rahulsy26: Thanks @dplore and @LimeHat for the response i will update the README as you suggested.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Following existing/new Github Code.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
rahulsy26: Updated as Suggested

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
rahulsy26: Updated as Suggested

LimeHat (changes requested):

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1026 @@
...
```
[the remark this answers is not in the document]
dplore: I[ see this being used in many places in featureprofiles tests](https://github.com/search?q=repo%3Aopenconfig%2Ffeatureprofiles+%22Disable+FEC+for+100G-FR+ports+because+Novus+does+not+support+it%22&type=code).  Novus is the hardware ATE we are using and according to this comment, it doesn't support autonegotiation.

I am not aware of using deviations for interop with ATE/OTG equipment, only with the DUT.  My thinking is to allow this.  If we want to make a change to how this need is accommodated in the code, a separate issue can be raised.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/BUILD
dplore, line 1: THis file should not be included.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1026 @@
...
+	for _, port := range portList {
+		resetBatch := &gnmi.SetBatch{}
+		gnmi.BatchDelete(resetBatch, gnmi.OC().Interface(port.Name()).Ethernet().AggregateId().Config())
+		gnmi.BatchDelete(resetBatch, gnmi.OC().Interface(port.Name()).ForwardingViable().Config())
+		resetBatch.Set(t, dut)
+	}
+}
+
+// configureDUTBGP configure BGP on DUT
+func configureDUTBGP(t *testing.T, dut *ondatra.DUTDevice, aggIDs []string) {
```
dplore, line 494: Should use the configplugin for bgp instead:

https://github.com/openconfig/featureprofiles/blob/main/internal/cfgplugins/bgp.go

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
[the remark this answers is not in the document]
LimeHat: But that's different. The code you linked refers to the OTG configuration, the code I highlighted is used on the DUT side (and also sets different values)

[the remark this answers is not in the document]
LimeHat: Opened #2977 to investigate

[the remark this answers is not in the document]
rahulsy26: Done Added the deviation as well for DUT

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1026 @@
...
+		}
+	}
+	configureRoutingPolicy(t, dut)
+	configureDUTISIS(t, dut, aggIDs)
+	configureDUTBGP(t, dut, aggIDs)
+	return aggIDs
+}
+
+// configureDUTLoopback configures DUT loopback
+func configureDUTLoopback(t *testing.T, dut *ondatra.DUTDevice) {
```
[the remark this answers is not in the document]
rahulsy26: Done

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1026 @@
...
+}
+
+// configureOTGBGP configure BGP on ATE
+func configureOTGBGP(t *testing.T, dev gosnappi.Device, agg *aggPortData, advV4, advV6 *ipAddr) {
+	t.Helper()
+
+	iDutBgp := dev.Bgp().SetRouterId(agg.ateIPv4)
+	iDutBgp4Peer := iDutBgp.Ipv4Interfaces().Add().SetIpv4Name(agg.ateAggName + ".IPv4").Peers().Add().SetName(agg.ateAggName + ".BGP4.peer")
+	iDutBgp4Peer.SetPeerAddress(agg.dutIPv4).SetAsNumber(asn).SetAsType(gosnappi.BgpV4PeerAsType.IBGP)
+	iDutBgp4Peer.Capability().SetIpv4UnicastAddPath(true).SetIpv6UnicastAddPath(false)
```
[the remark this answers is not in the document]
rahulsy26: Done Changes

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1026 @@
...
+	iDutBgp6Peer.Capability().SetIpv4UnicastAddPath(false).SetIpv6UnicastAddPath(true)
+	iDutBgp6Peer.LearnedInformationFilter().SetUnicastIpv4Prefix(false).SetUnicastIpv6Prefix(true)
+
+	bgpNeti1Bgp4PeerRoutes := iDutBgp4Peer.V4Routes().Add().SetName(agg.ateAggName + ".BGP4.Route")
+	if agg.ateAggName != LAG1 {
+		bgpNeti1Bgp4PeerRoutes.SetNextHopIpv4Address(pfx2AdvV4.ip).
+			SetNextHopAddressType(gosnappi.BgpV4RouteRangeNextHopAddressType.IPV4).
+			SetNextHopMode(gosnappi.BgpV4RouteRangeNextHopMode.MANUAL)
+		bgpNeti1Bgp4PeerRoutes.Addresses().Add().SetAddress(pfx3AdvV4.ip).SetPrefix(pfx3AdvV4.prefix).SetCount(1)
+		bgpNeti1Bgp4PeerRoutes.AddPath().SetPathId(1)
```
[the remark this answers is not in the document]
rahulsy26: Done Changes

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Looks like the above package is for EBGP not iBGP

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/BUILD
rahulsy26: Deleted BUILD File

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/metadata.textproto
```diff
@@ -0,0 +1,39 @@
...
+platform_exceptions:  {
+  platform:  {
+    vendor:  NOKIA
+  }
+  deviations:  {
+    explicit_port_speed: true
+    explicit_interface_in_default_vrf: true
+    aggregate_atomic_update: true
+    interface_enabled: true
+    missing_value_for_defaults: true
```
bkreddy143, line 18: Could you please add below deviaiton 
missing_isis_interface_afi_safi_enable: true

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,988 @@
...
+	}
+	agg2 = &aggPortData{
+		dutIPv4:      "192.0.2.5",
+		ateIPv4:      "192.0.2.6",
+		dutIPv6:      "2001:db8::5",
+		ateIPv6:      "2001:db8::6",
+		ateAggName:   LAG2,
+		ateAggMAC:    "02:00:01:01:02:01",
+		ateISISSysID: "640000000003",
+		ateLagCount:  5,
```
bkreddy143, line 124: can we make the script to work with less ports as well  instead of looking for 8 ports always . may be some thing like 1 for lag1 +2 lag2 +1 lag3 ?  or other way detect number of ports in binding file and distribute one to each LAG and attach leftover ports to LAG2

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/metadata.textproto
rahulsy26: Done

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Yeah we can do but Referring this to do it similar https://github.com/openconfig/featureprofiles/blob/1b16e43c73843951478440bef9aabdb25ad15dc0/feature/interface/aggregate/otg_tests/aggregate_forwarding_viable_test/README.md 

Also in future there may be some more subtest can be add here so others ports can help on this.

bkreddy143: @rahulsy26 , i see we have done this way before in test /feature/interface/aggregate/otg_tests/balancing_test/balancing_test.go where test designed for 9 ports but works with even 3 ports.

rahulsy26: @bkreddy143 done changes and now script is designed to work for atleast 4 ports

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+	})
+}
+
+// configureDUT configures DUT
+func configureDUT(t *testing.T, dut *ondatra.DUTDevice) []string {
+
+	t.Helper()
+	fptest.ConfigureDefaultNetworkInstance(t, dut)
+	if len(dut.Ports()) < 4 {
+		t.Fatalf("Testbed requires at least 4 ports, got %d", len(dut.Ports()))
```
bkreddy143, line 333: it needs to be len(dut.Ports()) <= 4 to cover   and also  change the error message to at least 5 ports

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+		IPv4Len: 32,
+		IPv6Len: 128,
+	}
+
+	pfx1AdvV4                = &ipAddr{ip: "100.0.1.1", prefix: 24}
+	pfx1AdvV6                = &ipAddr{ip: "2002:db8:64:64::1", prefix: 64}
+	pfx2AdvV4                = &ipAddr{ip: "100.0.2.1", prefix: 24}
+	pfx2AdvV6                = &ipAddr{ip: "2003:db8:64:64::1", prefix: 64}
+	pfx3AdvV4                = &ipAddr{ip: "100.0.3.1", prefix: 24}
+	pfx4AdvV4                = &ipAddr{ip: "100.0.4.1", prefix: 24}
```
bkreddy143, line 150: 100.0.4.1/24 is not valid prefix so gribi route add wil lfail, please change to 100.0.4.1 to 100.0.4.0 . similarly check other prefixes as well i.e pfx1AdvV4,pfx2AdvV4,pfx3AdvV4,

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+
+	return ok
+
+}
+
+// checkBidirectionalTraffic verify the bidirectional traffic on DUT ports.
+func checkBidirectionalTraffic(t *testing.T, dut *ondatra.DUTDevice, portList []*ondatra.Port) error {
+
+	for _, port := range portList {
+		txPkts := gnmi.Get(t, dut, gnmi.OC().Interface(port.Name()).Counters().InPkts().State())
```
bkreddy143, line 947: There is typo , change InPkts() to Outpkts()

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: @bkreddy143 
1. Default it use 4 port (1 LAG1 + 2 LAG2 + 1 LAG3) to work if testbed have 4 ports.
2. if the testbed have number of ports is < 4  it will throw error and come out from running.
3. If the testbed have number of ports > 4 it will bind the ports based on testbed.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: I have tested and the same is working. FYI below:
 I L2     100.0.1.0/24 [115/30] via 192.0.2.2, Port-Channel3
 I L2     100.0.2.0/24 [115/30] via 192.0.2.6, Port-Channel4, weight link bandwidth
                                via 192.0.2.10, Port-Channel5, weight link bandwidth
 B I      100.0.3.0/24 [200/0] via 192.0.2.6, Port-Channel4, weight link bandwidth
                               via 192.0.2.10, Port-Channel5, weight link bandwidth
 G        100.0.4.0/24 [5/0] via 192.0.2.6, Port-Channel4

Anyway i am changing it to as you suggested

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Thanks @bkreddy143 for finding this. Changes are done.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: As this test will work for 4 ports as well i am marking it as resolved.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+	})
+	t.Run("RT-5.7.1.2: Setting Forwarding-Viable to False for Lag2 all ports", func(t *testing.T) {
+		// Ensure ISIS Adjacency is up on LAG_2
+		if ok := awaitAdjacency(t, dut, aggIDs[1], oc.Isis_IsisInterfaceAdjState_UP); !ok {
+			t.Fatal("ISIS Adjacency is Down on LAG_2")
+		}
+		configForwardingViable(t, dut, dutPortList[1:2], false)
+		// Ensure ISIS Adjacency is Down on LAG_2
+
+		if ok := awaitAdjacency(t, dut, aggIDs[1], oc.Isis_IsisInterfaceAdjState_DOWN); !ok {
```
bkreddy143, line 213: is there any reason for expecting ISIS session DOWN when README is expecting to session as UP even all members are forwarding-viable False??

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+		txPktsBeforeTraffic[port] = gnmi.Get(t, dut, gnmi.OC().Interface(port.Name()).Counters().OutPkts().State())
+	}
+}
+
+// verifyTrafficFlow verify the each flow on ATE
+func verifyTrafficFlow(t *testing.T, ate *ondatra.ATEDevice, flows []gosnappi.Flow) bool {
+	if flows[0].Name() == "pfx1ToPfx4" {
+		rxPkts := gnmi.Get(t, ate.OTG(), gnmi.OTG().Flow(flows[0].Name()).Counters().InPkts().State())
+		if got := rxPkts / 100; got == 0 {
+			t.Logf("Packet Dropped, LossPct for flow %s: got %d, want 0 packet", flows[0].Name(), got)
```
bkreddy143, line 912: I dont see need to calculate % when we are checking only RX packets

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+
+	for _, port := range portList {
+		txPkts := gnmi.Get(t, dut, gnmi.OC().Interface(port.Name()).Counters().OutPkts().State())
+		rxPkts := gnmi.Get(t, dut, gnmi.OC().Interface(port.Name()).Counters().InPkts().State())
+		if got := (rxPkts - rxPktsBeforeTraffic[port]) / 100; got == 0 {
+			return fmt.Errorf("No Packet received, LossPct on Port %s: got %d", port.Name(), got)
+		}
+		if got := (txPkts - txPktsBeforeTraffic[port]) / 100; got == 0 {
+			return fmt.Errorf("No Packet transmitted, LossPct on Port %s: got %d", port.Name(), got)
+		}
```
bkreddy143, line 954: these check are very liberal and doesn't guarantee that there is actual traffic . Even if there is one 1 LACP packet test will pass  .

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+
+// confirmNonViableForwardingTraffic verify the traffic received on DUT
+// interfaces and transmitted to ATE-1
+func confirmNonViableForwardingTraffic(t *testing.T, dut *ondatra.DUTDevice, ate *ondatra.ATEDevice,
+	atePort []*ondatra.Port, dutPort []*ondatra.Port) error {
+
+	// Ensure no traffic is transmitted out of DUT ports with Forwarding Viable False
+	for _, port := range atePort {
+		rxPkts := gnmi.Get(t, ate.OTG(), gnmi.OTG().Port(port.ID()).Counters().InFrames().State())
+		if got := rxPkts / 100; got > 0 {
```
bkreddy143, line 967: same here , there is no need to % calculation

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+			return fmt.Errorf("Packets are transmiited out of %s: got %d, want 0", port.Name(), got)
+		}
+	}
+	// Ensure that traffic is delivered to ATE-1 port1
+	for _, port := range dutPort {
+		rxPkts := gnmi.Get(t, dut, gnmi.OC().Interface(port.Name()).Counters().InPkts().State()) - rxPktsBeforeTraffic[port]
+		txPkts := gnmi.Get(t, dut, gnmi.OC().Interface(dutPortList[0].Name()).Counters().OutPkts().State()) - txPktsBeforeTraffic[port]
+		if got := rxPkts / 100; got == 0 {
```
bkreddy143, line 975: same here , there is no need to % calculation

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+}
+
+// validateLag3Traffic to ensure traffic Received/Transmitted on DUT LAG_3
+func validateLag3Traffic(t *testing.T, dut *ondatra.DUTDevice, ate *ondatra.ATEDevice, dutPortList []*ondatra.Port) bool {
+	result := false
+	for _, port := range dutPortList {
+		rxPkts := gnmi.Get(t, dut, gnmi.OC().Interface(port.Name()).Counters().InPkts().State()) - rxPktsBeforeTraffic[port]
+		txPkts := gnmi.Get(t, dut, gnmi.OC().Interface(port.Name()).Counters().OutPkts().State()) - txPktsBeforeTraffic[port]
+
+		if got := rxPkts / 100; got > 0 {
```
bkreddy143, line 992: same here , the check is too liberal

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+
+	t.Helper()
+	fptest.ConfigureDefaultNetworkInstance(t, dut)
+	if len(dut.Ports()) < 4 {
+		t.Fatalf("Testbed requires at least 4 ports, got %d", len(dut.Ports()))
+	}
+	if len(dut.Ports()) > 4 {
+		agg2.ateLagCount = uint32(len(dut.Ports()) - 3)
+		agg3.ateLagCount = 2
+		equalDistributionWeights = []uint64{33, 67}
```
bkreddy143, line 338: I dont think number of members makes difference as the ECMP is verified at LAG level and both LAGs should have same traffic irrespective of number of active links .

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
bkreddy143: Looks like README needs to be updated as per https://github.com/openconfig/public/pull/1032/  it is possible that session can be timed out but in that case we may not see neighbor at all so we cant verify it is adjacency-state.  Could you please take that in verification. session state can be DOWN or no session info at all , both cases it should work.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: This is just to verify no single Packet is receiving for PFX1TOPFX4

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Here i am not calculating any % calculation, Just verifying if there is no packet rx/tx on the DUT port throw an error.
Note: i am using (txPkts - txPktsBeforeTraffic[port]) /100 because there will be some packet for rx/tx (ARP, ISIS, BGP) which is always <100.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Here i am not calculating any % calculation, Just verifying if there is no packet rx/tx on the DUT port throw an error.
Note: i am using (txPkts - txPktsBeforeTraffic[port]) /100 because there will be some packet for rx/tx (ARP, ISIS, BGP) which is always <100.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Same here as well

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Verifying same here as well for the packet not rx/tx more than 100

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+	prot.Enabled = ygot.Bool(true)
+	isis := prot.GetOrCreateIsis()
+	globalISIS := isis.GetOrCreateGlobal()
+
+	if deviations.ISISInstanceEnabledRequired(dut) {
+		globalISIS.Instance = ygot.String(isisInstance)
+	}
+	globalISIS.LevelCapability = oc.Isis_LevelType_LEVEL_2
+	globalISIS.Net = []string{fmt.Sprintf("%v.%v.00", dutAreaAddress, dutSysID)}
+	globalISIS.GetOrCreateAf(oc.IsisTypes_AFI_TYPE_IPV4, oc.IsisTypes_SAFI_TYPE_UNICAST).Enabled = ygot.Bool(true)
```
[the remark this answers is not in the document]
rahulsy26: Default isis use 4 max path to do perform ecmp. Verified the same and it worked.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: I have verified the same with equal/unequal number of ports in both LAG. 

I am referring below please refer section 9.2 on README. 

https://github.com/openconfig/featureprofiles/blob/main/feature/isis/otg_tests/weighted_ecmp_test/README.md

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Yeah, Darren suggested in Comment above i am using the same approach. After Once ISIS timed out the isis will go down. As per suggestion i updated the same in README as well

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: verifyTrafficFlow is the function where i am verifying the actual flow %, if any loss %.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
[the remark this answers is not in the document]
LimeHat: 4 paths by default can be true in a specific vendor implementation, but OpenConfig doesn't specify that 
https://github.com/openconfig/public/blob/541fbd64f9a399caed2878aed344c20a93dcbad7/release/models/isis/openconfig-isis.yang#L311

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
LimeHat: Weighted ECMP is disabled by default (https://github.com/openconfig/public/blob/541fbd64f9a399caed2878aed344c20a93dcbad7/release/models/isis/openconfig-isis.yang#L317) and out of scope of this test; the regular ECMP shouldn't be using the link bandwidths

rahulsy26: Oh Ok, 

@LimeHat I haven't enable Weighted ECMP here but still it is doing Weighted ECMP by default.

bkreddy143: @rahulsy26 , we can go by open config model as mentioned by @LimeHat  that is by default no WECMP

rahulsy26: @bkreddy143 If by default no WECMP then it should have working ECMP. But it seems doing WECMP by default not sure. and Also i don't find any openconfig support to enable/disable WECMP

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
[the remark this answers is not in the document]
bkreddy143: @rahulsy26 , as the default value can be different for each vendor when it is not specified explicitly in openconfig requirements so would recommend to set the explicit value value based on the test config  so that it will work for all vendors.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
bkreddy143: @rahulsy26  as per (https://github.com/openconfig/public/blob/541fbd64f9a399caed2878aed344c20a93dcbad7/release/models/isis/openconfig-isis.yang#L317)

    leaf weighted-ecmp {
      type boolean;
      default "false";
      description
        "When set to true, all eligible multipath IS-IS routes associated with
        the instance are programmed to use weighted ECMP. An IS-IS route is
        eligible for weighted ECMP if all the next-hop interfaces in the
        multipath set have a load-balancing-weight other than 'none'.";
    }

WECMP is disable  by default and the expectation should be based on  ECMP .  Request to change the script as per ECMP and raise bug to vendor if there is no support of  WECPM enable/disable or WECMP is enabled by default.

nflath (approved):

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
dplore: Perhaps more accurate and deterministic to check the ATE flow stats instead of the ATE port stats?   Does OTG support getting flow stats per ATE port or similar?

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+	d := &oc.Root{}
+	rp := d.GetOrCreateRoutingPolicy()
+	pdef := rp.GetOrCreatePolicyDefinition(acceptRoutePolicy)
+	stmt, _ := pdef.AppendNewStatement("20")
+	stmt.GetOrCreateActions().PolicyResult = oc.RoutingPolicy_PolicyResultType_ACCEPT_ROUTE
+	gnmi.Replace(t, dut, gnmi.OC().RoutingPolicy().PolicyDefinition(acceptRoutePolicy).Config(), pdef)
+}
+
+// configureDUTISIS configure ISIS on DUT
+func configureDUTISIS(t *testing.T, dut *ondatra.DUTDevice, aggIDs []string) {
```
dplore, line 574: The internal/isis library should be used for isis configuration.    We should avoid cut/paste of so much ISIS configuration code.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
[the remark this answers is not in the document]
dplore: I do not understand why setting max ecmp paths is helpful here?

In any case, if the OC default is not being honored, it is ok to configure it, but it should be done using a deviation.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
dplore: To be more clear then, you could change this to be:

`if got := rxPkts ; got == 0 `

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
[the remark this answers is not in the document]
dplore: Ah, you are correct, there is no default value for isis max-ecmp-paths.     I guess since it is ambiguous what the value is, setting it to some deterministic value (in case implementation is set to `1`)  is ok.  @rahulsy26 if the target device does not support max-ecmp-paths, then a deviation can be added to skip setting max-ecmp-paths.


[/network-instances/network-instance/protocols/protocol/isis/global/config/max-ecmp-paths
](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-isis-global-config-max-ecmp-paths)description: ISIS max-paths count.
nodetype: leaf (rw)
type: uint8

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1025 @@
...
+		// Ensure ISIS Adjacency is up on LAG_2
+		if ok := awaitAdjacency(t, dut, aggIDs[1], oc.Isis_IsisInterfaceAdjState_UP); !ok {
+			t.Fatal("ISIS Adjacency is Down on LAG_2")
+		}
+		configForwardingViable(t, dut, dutPortList[1:2], false)
+		// Ensure ISIS Adjacency is Down on LAG_2
+		if ok := awaitAdjacency(t, dut, aggIDs[1], oc.Isis_IsisInterfaceAdjState_DOWN); !ok {
+			t.Fatal("ISIS Adjacency is Established on LAG_2")
+		}
+		otgutils.WaitForARP(t, ate.OTG(), top, "IPv4")
```
bkreddy143, line 294: @rahulsy26  , I don't think we can expect to resolve ARP when all member of LAG are set to forwarding-viable  False. better add  dst mac address in all flows based on the destination port and remove wait for ARP for all cases. Other easy option could be add secondary mac mac address at system level in DUT and use  that MAC for all flows so that all ports can accept the traffic with that secondary mac. .  
 @dplore  could please suggest .

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
```diff
@@ -22,117 +23,132 @@ Ensure that when **all LAG member** become set with forwarding-viable == FALSE.
...
-- Configure ATE and DUT ports 8-9 to be LAG_3 w/ LACP running.
+- Configure ATE and DUT ports 2-6 to be LAG_2 w/ LACP running.
+- Configure ATE and DUT ports 7-8 to be LAG_3 w/ LACP running.
 - Establish ISIS adjacencies on LAG_1, LAG_2, LAG_3.
   1. Advertise one network prefix (pfx1) from ATE LAG_1
   1. Advertise one network prefix (pfx2) from ATE LAG_2 and ATE LAG_3.
-- Establish iBGP between ATE and DUT over LGA_1 using LAG_1 interface IPs and advertise prefix pfx3 with BGP NH from pfx2 range.
+- Establish iBGP between ATE and DUT over LGA using LAG interface IPs and 
+  advertise prefix pfx3 with BGP NH from pfx2 range.
 - Programm via gRIBI route for prefix pfx4 with single NHG pointing LAG_2 (al
```
bkreddy143, line 49: @dplore , i think the idea of forwarding-viable is switch to backup NHG if available when all member of primary are set forwarding-viable=False . I don't think we have that coverage.  Can we add backup NHG pointing to LAG3 and verify that gRIBI traffic switched to LAG3 when all member of LAG2 are set to False  and similarly it is switched back to primary ( LAG2) when any of the member set to forwarding-viable True .  ( PS: we can ignore this case here  if it planned as part of gRIBI  test cases)

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Removed WaitForARP. Please suggest how to add dst mac address in all flows based on the destination port for reverse traffic (East to West direction)

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
rahulsy26: Please let me know what can be added so that i will change it accordingly.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: I am using LAG interface which support *oc.interface  to create isis but the isissession used *ondatra.port so can't use the internal/isis library.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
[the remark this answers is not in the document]
rahulsy26: Done

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Done

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Done

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
bkreddy143: @rahulsy26 , Iam talking about gRIBI route traffic . that is getting dropped when all members of LAG2 are down.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
bkreddy143: Remove WaitForArp() may resulted to traffic start as there is not destination MAC in flow.  we can add system level MAC and use that MAC for all flow  
gnmi.Replace(t, dut, gnmi.OC().System().MacAddress().RoutingMac().Config(), myStationMAC) [ https://github.com/openconfig/featureprofiles/blob/main/feature/experimental/interface/my_station_mac/otg_tests/my_station_mac_test/my_station_mac_test.go#L208] 

or other oprtion is used gnmi Get and get the DUT interface MAC and use that in respective flow .

rahulsy26: @bkreddy143 As you suggested, i tried above and with static_arp_test entry but looks like ixia is sending arp request to DUT even after i have added eV4.Dst().SetValue(dutAggMac).

https://github.com/openconfig/featureprofiles/blob/main/feature/interface/staticarp/otg_tests/static_arp_test/static_arp_test.go

Please let me know if any suggestion

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
rahulsy26: Lets wait for @dplore to comment on this

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
bkreddy143: @rahulsy26 , from error message it looks like we might need to set static ARP in OTG  to pass protocol check , could you please check with IXIA/Keysight team  for the the same.  I did try to add gatewaymac but didnt work.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
dplore: Yes, we should include a backup-NHG.  We will have the scenario where Backup-NHG is present and a LAG will have all member interfaces set to forwarding-viable=FALSE.  The backup-NHG should take effect in this scenario.  @nflath for any other comment

rahulsy26: Noted

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: Done

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
```diff
@@ -22,117 +23,131 @@ Ensure that when **all LAG member** become set with forwarding-viable == FALSE.
...
```
nachikethas, line 49: @rahulsy26 Can we add a test case where the NHG used by gRIBI points to two NHs where only _one_ of the NHs points to LAG with all members with `forwarding-viable=false`? In this case we expect that the NH with only unviable members will be pruned and the traffic will be rebalanced to the viable NH. Thanks!

CC: @nflath

nachikethas (dismissed):

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
rahulsy26: Done

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/README.md
rahulsy26: Added NHG As per Nachikethas Suggested

nflath (commented):
LGTM

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -0,0 +1,1044 @@
...
+	if err != nil {
+		return "", err
+	}
+	newMac := net.HardwareAddr(buf.Bytes()[2:8])
+	return newMac.String(), nil
+}
+
+func createFlows(t *testing.T, dut *ondatra.DUTDevice, top gosnappi.Config, aggIDs []string) []gosnappi.Flow {
+	for _, aggID := range aggIDs {
+		dutAggMac = append(dutAggMac, gnmi.Get(t, dut, gnmi.OC().Lacp().Interface(aggID).SystemIdMac().State()))
```
bkreddy143, line 810: system-id for lag is configurable value and it need not  to be interface mac. also there is no need to hae state value if there is no configuration.  better to change this to gnmi.OC().Interface(aggID).Ethernet().MacAddress().State()

nflath (dismissed):

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
rahulsy26: this is not working gnmi.OC().Interface(aggID).Ethernet().MacAddress().State() says value not present
As i have cross verified with Arista as well, this path is not supported for Aggregate.

As portchannel is configured with routing mode so it will have system mac which is configured by default.

self-maurya (approved):

self-maurya (approved):

rahulsy26 (commented):
Done

PR #2964, opened 2024-05-07
RT-6.1: fix leafref violations

kjahed:
Modifies RT-6.1 to also configure the key name of /lldp/interface when configuring /lldp/interface/enabled.


kjahed:
Not needed after #3369

PR #2967, opened 2024-05-08
Adds a gNPSI test readme.

menezessachin:
Adds a gNPSI test readme.


feature/gnpsi/otg_tests/sampling_test/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+2. Send Traffic via OTG with different traffic profiles. 
+
+*   Ipv4 and Ipv6
+*   Varying packet sizes (64, 512, 1500)
+
+3. Start the gRPC client and subscribe to the gRPC service on the DUT.
+
+4. Verify the samples received by the client are as per expectations:
+
```
dplore, line 27: Does OTG support gnpsi client and gnpsi packet decode?  If this is an open feature request, please add a TODOs linking to the OTG issues and/or pull requests.  

FWIW, it is possible to get raw packet capture from OTG, but that won't be helpful with encrypted gRPC.

LimeHat (changes requested):

feature/gnpsi/otg_tests/sampling_test/README.md
```diff
@@ -0,0 +1,112 @@
...
+The goal is to validate that packet sampling is working as expected, clients can connect to the gNPSI service and receive samples.  
+
+## Procedure
+
+
+### Test 1: Verify samples are exported via gNPSI service. 
+
+1. Enable sFlow and gNPSI on the DUT.
+
```
dplore, line 13: Need to add steps (including oc paths) for how to configure gNPSI.  Don't forget to include configuring certificate, etc:
[/system/grpc-servers/grpc-server/config/certificate-id](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-system.html#system-grpc-servers-grpc-server-config-certificate-id)

dplore (commented):
Some comments to be addressed please

dplore:
FYI:  the sflow readme might be helpful to refer to: https://github.com/openconfig/featureprofiles/blob/afe4b12d8f80e5fb4c847a3d72a17217eb7468fa/feature/sflow/otg_tests/sflow_base_test/README.md

feature/gnpsi/otg_tests/sampling_test/README.md
LimeHat: @dplore perhaps grpc-server part of the config can be left out? For example, in other tests, we don't configure p4rt or gribi grpc services separately, it is assumed that they're included in the base config.

dplore: Ah, this is true, there is an implicit requirement that the DUT already has those gRPC services enabled and configured.  Ok, we can follow that precedent here and not include this here.

dplore (approved):
This LGTM so far.

Do you want to add coverage for the (pending approval) gnpsi packets/bytes-sent leafs?   (ref: https://github.com/openconfig/public/pull/1085)

PR #2969, opened 2024-05-08
gNMI-1.19: test implementation

kjahed:
See #2830


nikki-blr:
It is implemented by PR#4030

PR #2971, opened 2024-05-08
add cancellable context in ssh CLI

LiamWalsh98:
When using the RunCommand function, there is a case where if an SSH connection is stopped from the server side without closing the connection, the program calling it will halt and be unable to resume. This may occur when a router is restarted, after which the severed ssh session will not be reestablished.

To illustrate, this line will cause deadlock 

```go
cliClient.RunCommand(context.Background(), "run killall -STOP sshd")
``` 
*note*: this will also block incoming ssh until `killall -CONT sshd` is executed. For testing, try instead `run killall -STOP sshd; sleep 10s; killall -CONT sshd`

with the proposed changes, it can be handled using a context with a timeout/deadline (or otherwise cancelled)

```go
ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
defer cancel()
cliClient.RunCommand(ctx, "run killall -STOP sshd")
``` 

Have also added the sshConfig timeout to DialCLI's ClientConfig.

### DIFF
--- a/topologies/binding/cli.go
+++ b/topologies/binding/cli.go
@@ -32,21 +32,38 @@ func newCLI(sc *ssh.Client) (*cli, error) {
 	return &cli{ssh: sc}, nil
 }
 
-func (c *cli) RunCommand(_ context.Context, cmd string) (binding.CommandResult, error) {
+func (c *cli) RunCommand(ctx context.Context, cmd string) (binding.CommandResult, error) {
 	sess, err := c.ssh.NewSession()
 	if err != nil {
 		return nil, fmt.Errorf("could not create session: %w", err)
 	}
 	defer sess.Close()
 
-	out, err := sess.CombinedOutput(cmd)
-	switch err.(type) {
-	case nil:
-		return &cmdResult{output: string(out)}, nil
-	case *ssh.ExitError, *ssh.ExitMissingError:
-		return &cmdResult{output: string(out), error: err.Error()}, nil
-	default:
-		return nil, err
+	outChan := make(chan *struct {
+		out string
+		err error
+	})
+
+	go func() {
+		out, err := sess.CombinedOutput(cmd)
+		outChan <- &struct {
+			out string
+			err error
+		}{out: string(out), err: err}
+	}()
+
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case result := <-outChan:
+		switch result.err.(type) {
+		case nil:
+			return &cmdResult{output: result.out}, nil
+		case *ssh.ExitError, *ssh.ExitMissingError:
+			return &cmdResult{output: result.out, error: result.err.Error()}, nil
+		default:
+			return nil, result.err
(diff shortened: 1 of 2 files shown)


singh-prem:
Hi @dplore , could someone take a look at this PR

dplore:
Hi @ram-mac please review, thank you.

topologies/binding/cli.go
```diff
@@ -32,21 +32,38 @@ func newCLI(sc *ssh.Client) (*cli, error) {
...
-	case *ssh.ExitError, *ssh.ExitMissingError:
-		return &cmdResult{output: string(out), error: err.Error()}, nil
-	default:
-		return nil, err
+	outChan := make(chan *struct {
+		out string
+		err error
+	})
+
+	go func() {
```
dplore, line 47: why introduce a goroutine?  could this introduce flakiness?  Wouldn't it be more simple to just keep this in the main process/thread as it was before?

dplore (commented):
IMO, adding the timeout is good.  I am not sure why the goroutine is needed though?  If the DUT ssh session dies, the timeout should handle closing the connection with a timeout error.  A test that triggers a reboot should expect this to occur and handle the timeout.  

The use case of `cliClient.RunCommand(context.Background(), "run killall -STOP sshd")` I think it not something that should be handled in the binding except to have a timeout error returned.

I suggest only adding the timeout and not changing the code for dialCLI.

topologies/binding/cli.go
LiamWalsh98: Sorry, I should clarify that the proposed change in DialCLI is unrelated to the main issue but seemed close enough in spirit to tack onto this PR. DialCLI as-is ignores the timeout value from the binding file (which is handled in the [crypto package](https://github.com/golang/crypto/blob/bb80217080b0e04c6e73e5dcd9f3a9bb11fe23f6/ssh/client.go#L177)).

I can remove it from this PR but it seemed like a relevant enough custodial change that I ended up including it. 

---

The issue seems to be the edge case where the ssh session connected to the DUT doesn't die, but hangs indefinitely. The `killall -STOP sshd` is just pausing the DUT's sshd in order to repro the client's behaviour.

> A test that triggers a reboot should expect this to occur and handle the timeout.

I agree that a test expecting an unresponsive sshd would need to handle the timeout. To me this means providing a context with a deadline to whatever request is being made, and deferring a cancellation in the thread making the call. Ideally this would mean passing down the ctx all the way to whatever dependency actually makes the request, however it is not forwarded further down the call stack beyond RunCommand. This is the reason I would like to handle it here. (i.e. sess.CombinedOutput doesn't accept context, and this is the function that hangs indefinitely when the DUT's sshd is unresponsive). The main thread will wait for the first of either: context timeout or CombinedOutput to return. 

As for handling it without a goroutine, I'm not too sure how it would be implemented. Maybe I am overlooking something?

Thanks for taking the time to review!

Issue #2977, opened 2024-05-10
100g fr interface config on the dut side


LimeHat:
(moving this conversation from #2953 )
In some tests, there's an unusual configuration performed on DUT only for 100G FR ports, e.g.:
https://github.com/openconfig/featureprofiles/blob/91796ca845a55876969dec6e9a574c5cc0800738/feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go#L347-L351

It is unclear why is this required.

@dplore suggested that it might be related to the FEC config on the OTG side (https://github.com/search?q=repo%3Aopenconfig%2Ffeatureprofiles+%22Disable+FEC+for+100G-FR+ports+because+Novus+does+not+support+it%22&type=code), but the parameters do not match:

1. OTG config only disables the fec, while the autonegotiation is still on

https://github.com/openconfig/featureprofiles/blob/22dc0891eabdcf9b9b7bcf11a216c3d734f8f120/feature/gribi/otg_tests/encap_frr/encap_frr_test.go#L898-L903

2. DUT disables autonegotiation completely

dplore:
@Ankur19  and @liulk do you have some experience with this issue?

LimeHat:
A case in point: looks like it caused an issue for another vendor in #2984

dplore:
@LimeHat are you suggesting the that the regular expectation should be auto-negotiation should work and the deviation should be to disable auto-negotiation and set duplex and port speed?

LimeHat:
Yes, I believe so. I'm not aware of any IEEE limitations that would prevent auto-negotiation from working with FR optics.

LimeHat:
Bumping this up. Any additional feedback? Is there an owner of this code who _needs_ it? If not, I'll open a PR to remove it.

dplore:
no additional comments from me.  It seems ok to remove.  If you send the PR we'll run it to confirm

PR #2984, opened 2024-05-13
explicit port speed deviation added

NishaSadhasivam-work:


self-maurya (approved):

LimeHat:
I'm wondering why? 

AFAIK Nokia is the only consumer of this deviation, but we don't require this config (duplex/autoneg). In fact, I raised an issue about this code block a few days ago #2977 

cc @dplore

LimeHat:
The problem is that this code sets speed and duplex and disables autonegotiation (which, in my opinion, is not correct). 

IMO this deviation should not be used to guard anything except the port speed, which is not the case here.

I'd support the removal of this block as a whole (as indicated in #2977) or if there's a vendor that needs explicit duplex and disabled autonegotiation, they should add a new deviation

PR #2985, opened 2024-05-13
Create READMEs for SR-1.1 and MPLS-1.1

sugrimov:


feature/experimental/mpls/label_block/README.md
dplore, line 1: The experimental folder should only be used for tests which have never passed on any device.  So for now, sicne there is not yet an implementation in code of this README, you can place this file in feature/mpls/otg_tests/label_block_test/README.md. (to make label_block part of the MPLS base feature
or
feature/mpls/label_block/otg_tests/label_block_test/README.md (to make label_block a 'subfeature' of MPLS).

I am ok with either approach, but when in doubt, a subfeature is good.

sugrimov: Done.

feature/experimental/segment_routing/node_sid_forward/README.md
```diff
@@ -0,0 +1,38 @@
...
+*   Enable Segment Routing for the ISIS
+
+Advertise 2 prefixes with SIDs to DUT from ATE2:
+
+*  Prefix (1) with node-SID is advertised by the direct ISIS neighbor
+*  Prefix (2) with node-SID is advertised by simulated indirect ISIS speaker
+
+Verify: 
+*  DUT receives prefixes with SIDs and has a correct emulated ISIS topology (database).
+*  DUT advertises both node-SID to ATE1.
```
daveruturaj: Do you mean - DUT advertises both prefixes with the node-sid to ATE1.

feature/experimental/segment_routing/node_sid_forward/README.md
```diff
@@ -0,0 +1,38 @@
...
+## Summary
+
+MPLS-SR transit forwarding to Node-SID distributed over ISIS
+
+## Procedure
+
+Topology: ATE1—DUT1–ATE2
+                               
+*   Configure Segment Routing Global Block (start: 400000 range: 65001)
+*   Enable Segment Routing for the ISIS
```
daveruturaj: Enable MPLS forwarding?

sugrimov: Added.

feature/experimental/segment_routing/node_sid_forward/README.md
sugrimov: Added.

daveruturaj (approved):

feature/mpls/otg_tests/label_block/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+```yaml
+paths:
+  # configuration
+  /network-instances/network-instance/mpls/global/reserved-label-blocks/reserved-label-block/config/local-id:
+  /network-instances/network-instance/mpls/global/reserved-label-blocks/reserved-label-block/config/lower-bound:
+  /network-instances/network-instance/mpls/global/reserved-label-blocks/reserved-label-block/config/upper-bound:
+  /network-instances/network-instance/segment-routing/srgbs/srgb/config/local-id:
+  /network-instances/network-instance/segment-routing/srgbs/srgb/config/mpls-label-blocks:
+  /network-instances/network-instance/protocols/protocol/isis/global/segment-routing/config/srgb:
```
dplore, line 39: I am not that familiar with configuring segment routing, but do you also need this path to enable it in ISIS?

[/network-instances/network-instance/protocols/protocol/isis/global/segment-routing/config/enabled](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-isis-global-segment-routing-config-enabled)

feature/mpls/otg_tests/label_block/README.md
```diff
@@ -0,0 +1,58 @@
...
+*  [`featureprofiles/topologies/atedut_2.testbed`](https://github.com/openconfig/featureprofiles/blob/main/topologies/atedut_2.testbed)
+
+## Procedure
+
+Topology: ATE1—DUT1
+
+On DUT1 configure:
+
+*   ISIS adjacency between ATE1 and DUT1
+*   Enable MPLS-SR
```
dplore, line 18: Please add some additional explanation of how to do this.  I don't see a path that 1:1 maps to mpls enable segment routing?

feature/mpls/sr/otg_tests/isis_node_sid_forward/README.md
```diff
@@ -0,0 +1,61 @@
...
+*   Send labeled traffic transiting through the DUT matching indirect prefix (2). Verify that ATE2 receives traffic with the node-SID label intact.
+*   Verify that corresponding SID forwarding counters are incremented.
+*   Traffic arrives without packet loss.
+
+## OpenConfig Path and RPC Coverage
+
+```yaml
+paths:
+  # configuration
+  /network-instances/network-instance/mpls/global/interface-attributes/interface/config/mpls-enabled:
```
dplore, line 42: Should these paths also be claimed as covered by this test since a global label block is configured?

  /network-instances/network-instance/mpls/global/reserved-label-blocks/reserved-label-block/config/local-id:
  /network-instances/network-instance/mpls/global/reserved-label-blocks/reserved-label-block/config/lower-bound:
  /network-instances/network-instance/mpls/global/reserved-label-blocks/reserved-label-block/config/upper-bound:

feature/mpls/otg_tests/label_block/README.md
sugrimov: Makes sense. Added.

feature/mpls/otg_tests/label_block/README.md
sugrimov: Added the path.

feature/mpls/otg_tests/label_block/README.md
```diff
@@ -0,0 +1,58 @@
...
+*   Static Global Block (start: 1000000 range: 48576)
+*   Segment Routing Global Block (start: 400000 range: 65001)
+*   Segment Routing Local Block (start: 40000 range: 1000)
+
+Verify:
+
+*   Defined blocks are configured on DUT1.
+*   DUT1 advertises its SRGB and SRLB to ATE1.
+
```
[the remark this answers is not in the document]
sugrimov: There is a separate test for static LSPs. This test just tests defining MPLS reserved blocks and succeeding.

In the future more blocks might be added here, like L2VPN.

feature/mpls/sr/otg_tests/isis_node_sid_forward/README.md
sugrimov: Added paths.

feature/mpls/sr/otg_tests/isis_node_sid_forward/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+*  [`featureprofiles/topologies/atedut_2.testbed`](https://github.com/openconfig/featureprofiles/blob/main/topologies/atedut_2.testbed)
+
+## Procedure
+
+### Configuration
+
+Topology: ATE1—DUT1–ATE2
+                               
+*   Configure Segment Routing Global Block (start: 400000 range: 65001)
```
dplore, line 17:
```suggestion
*   Configure Segment Routing Global Block (srgb) lower-bound: 400000 upper-bound: 465001)
```

feature/mpls/sr/otg_tests/isis_node_sid_forward/README.md
```diff
@@ -0,0 +1,67 @@
...
+*   Enable Segment Routing for the ISIS
+*   Enable MPLS forwarding (if required)
```
dplore, line 19: There shouldn't be any "if required" in this case I think?  The OC should work one way.   If a vendor does not require this, they should accept the enable leaf = true.       (although they would send an error if enable = false, assuming that they don't support explicitly disabling mpls forwarding)

feature/mpls/sr/otg_tests/isis_node_sid_forward/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+*  Prefix (1) with node-SID is advertised by the direct ISIS neighbor
+*  Prefix (2) with node-SID is advertised by simulated indirect ISIS speaker
+
+### Test 
+
+Verify that:
+
+*  DUT receives prefixes with SIDs and has a correct emulated ISIS topology (database).
```
dplore, line 28: I think verifying DUT aggregate-sid-counters/aggregate-sid-counter/state/(in|out)-pkts is a good idea.  

But regarding DUT ISIS lsdb state, I think we should instead rely on traffic receipt at the OTG being correct. This would be much more simple and doesn't require trusting the DUT lsdb state.  (generally want to verify things via OTG, and minimize dependency on 'test-only' DUT specific paths). If you are ok with this, please remove the 'verify correct emulated ISIS topology' part of this verification.


(unless we expect to actually use ISIS lsdb leafs in production.  If yes, then we need to specify which lsdb leafs and how to query them (just for presence or if we require some specific value verification)).

dplore (changes requested):

feature/mpls/sr/otg_tests/isis_node_sid_forward/README.md
sugrimov: Removed.

feature/mpls/sr/otg_tests/isis_node_sid_forward/README.md
sugrimov: Makes sense. Removed.

dplore (approved):

PR #2987, opened 2024-05-13
RT-7.11  : import_export_multi_test.go

cprabha:
I have replaced GetSetCommunity with GetOrCreateSetCommunity , other script will fail with panic error.
Also gnmi.update is needed  instead of gnmi.replace, since protocol is still pointing to routing policy.


NishaSadhasivam-work (approved):
LGTM!

PR #2989, opened 2024-05-14
RT-7.2 | Changes and adding deviation

arvbaska1:
Made below changes

- In cfgplugin bgp, the peergroup route-policy gets overwritten with latest iteration of the AFI, 
changed it to list of V4 & v6 AFI.
- Add default permit-all RPL alongwith new RPL config
- deviation for community-set regex with CLI.
- Verify BGP session after ATE start.


ram-mac:
@arvbaska1 -> have validated this PR and with these changes the testcase is now passing. 
Can you please fix the readme which requires just some template based changes and i can merge this PR.

ram-mac (commented):
@arvbaska1 -> have validated this PR and with these changes the testcase is now passing.
Can you please fix the readme which requires just some template based changes and i can merge this PR

ram-mac (commented):
There is one issue with the deviation inside the community_test.go file  if deviations.CommunityMemberRegexUnsupported(dut) && communitySetName == comunitySetNameRegex {
		communitySetCLIConfig := fmt.Sprintf("community-set %v\n ios-regex '10[0-9]:1'\n end-set", communitySetName)
		helpers.GnmiCLIConfig(t, dut, communitySetCLIConfig)

This CLI config is specific to cisco and will not fail for any other vendor. So the CLI push need to be be case specific like for example Ln 241 in feature/bgp/gracefulrestart/ate_tests/bgp_graceful_restart_test/bgp_graceful_restart_test.go

arvbaska1:
> There is one issue with the deviation inside the community_test.go file if deviations.CommunityMemberRegexUnsupported(dut) && communitySetName == comunitySetNameRegex { communitySetCLIConfig := fmt.Sprintf("community-set %v\n ios-regex '10[0-9]:1'\n end-set", communitySetName) helpers.GnmiCLIConfig(t, dut, communitySetCLIConfig)
> 
> This CLI config is specific to cisco and will not fail for any other vendor. So the CLI push need to be be case specific like for example Ln 241 in feature/bgp/gracefulrestart/ate_tests/bgp_graceful_restart_test/bgp_graceful_restart_test.go

Sure, i updated as per this

ram-mac (approved):

ram-mac (commented):
Hi,
Line number 117, 123 needs to be also under the deviations, otherwise the default policy is not applied and it does not work in case of other vendors.
Can we fix this part too?

-Ram

arvbaska1:
> Hi, Ln 100 to 106 where the permit-all statement is used has to go under the deviation. Ln. 108 pdAllow := rp.GetOrCreatePolicyDefinition(RPLPermitAll) st, err := pdAllow.AppendNewStatement("id-1") if err != nil { t.Fatal(err) } st.GetOrCreateActions().PolicyResult = oc.RoutingPolicy_PolicyResultType_ACCEPT_ROUTE gnmi.Replace(t, dut, gnmi.OC().RoutingPolicy().Config(), rp) -Ram

Okay , not sure it should be a deviation since based on implementation the gNMI replace is blocked if its already used/chained under BGP config .We would have to discuss what is expectation here and if it should be a deviation for us or other vendor.

PR #2998, opened 2024-05-17
RT-1.33 changes

arvbaska1:
Made below changes to the test

- Change AS# for Port2 DUT-ATE BGP sessions for routes to be advertised to that eBGP Peer , same AS wont work due to AS PATH loop prevention.
- Added soft reconfiguration inbound CLI with MissingPrePolicyReceivedRoutes deviation for pre-policy routes
- Increase timeouts in watch in validatePrefixCount to account for default sample interval with gnmi target-defined sub


self-maurya (approved):

AmrNJ (approved):
LGTM.

PR #3003, opened 2024-05-21
Rt 1.51 configs with proto files

fmolinar:
added changes and deviations for it to run on cisco 8000 series.


fmolinar:
this is addressed by https://github.com/openconfig/featureprofiles/pull/3006

PR #3006, opened 2024-05-22
adding deviations

ram-mac:
Added deviations and made below changes.
1. Cisco's implementation toggles allow-multiple-as at the global level, but enables multipath and max-paths at the global/afi/safi level.  Hence added deviation.
2. Cisco is not doing load balance with just one flow, so had to create multiple flows for load balance to work.


feature/bgp/multipath/otg_tests/bgp_multipath_ecmp_test/metadata.textproto
```diff
@@ -1,4 +1,4 @@
-# proto-file: github.com/openconfig/featureprofiles/proto/metadata.proto
+# proto-file: third_party/openconfig/featureprofiles/proto/metadata.proto
```
self-maurya, line 1: change this to `github.com` instead of `third_party`

self-maurya (changes requested):

self-maurya (approved):

PR #3008, opened 2024-05-23
[gNMI-1.14]: Check metadata2 after the 2nd large config set

yini101:
After gnmiClient set 2nd large configuration, the test should be calling checkMetadata2() instead of checkMetadata1().


LimeHat:
>After gnmiClient set 2nd large configuration, the test should be calling checkMetadata2() instead of checkMetadata1().

Can you elaborate on this?

This goroutine runs in parallel to the set request, and checks for metadata1 only until the setResponse is received. 

M2 is validated after the setresponse is received + 5s delay according to the readme:
https://github.com/openconfig/featureprofiles/blob/44b8b7e30191fc8ce66cd66b97328430ecce077f/feature/system/gnmi/metadata/tests/large_set_consistency_test/README.md?plain=1#L28-L31

What is the reason to expect metadata1 before the setRequest is acknowledged?

yini101:
Thanks @LimeHat, I agree that the check should be checking metadata1 before the set request is completed. Let me take another look of the test.

PR #3011, opened 2024-05-23
Create READMEs for RT-1.54: BGP Override AS-path split-horizon

ihebboubaker:
This change is to create a Readme for BGP Add Path send/receive disable, BGP Override AS-path split-horizon (allowas-in) and BGP session mode (active/passive)


feature/experimental/bgp/otg_tests/bgp_add_path_capabilities_disable_test/README.md
dplore, line 1: Hi, have you checked the currently in place addpath test?  See: 
https://github.com/openconfig/featureprofiles/tree/main/feature/bgp/addpath/otg_tests/route_propagation_test

Rather than duplicate the existing test, which includes automation, it would be better to review the existing test and propose updates (if any) that should be made.

feature/experimental/bgp/otg_tests/bgp_override_as_path_split_horizon_test/README.md
```diff
@@ -0,0 +1,64 @@
...
+## Topology
+
+DUT Port1 (AS 65501) ---eBGP --- ATE Port1 (AS 65502)
+
+## Procedure
+
+*  Establish BGP Session: Configure and establish an eBGP session between the DUT (Port1) and the ATE (Port1).
+*  Baseline Test (No "allow-own-in"):
+    *  Advertise a prefix from the ATE (e.g., 192.168.1.0/24) with an AS-path that includes AS 65501 (DUT's AS) in the middle (e.g., AS-path: 65502 65500 65501 65499).
+    *  Verify that the DUT rejects this route due to the presence of its own AS in the path.
```
dplore, line 16: Please propose a method for the validation.  The minimal validation from a feature perspective would be to have a 2 port topology where routes are propagated to a second ATE port, then verify the route is not propagated to the second ATE port.       

A second approach is to query the /afts table on the DUT, such as  [/network-instances/network-instance/afts/ipv4-unicast/ipv4-entry/state/prefix](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-afts-ipv4-unicast-ipv4-entry-state-prefix).  We have a lot of support for afts at the moment, so this may be ok.

Another validation approach is to query the DUT, but it will expand your feature requirement to include the BGP rib paths such as the below.  The downside is the DUT may support allow-as as needed, but a lack of the loc-rib state telemetry may cause the test to fail.  This is probably less preferred at the moment because I am not sure many DUT support these paths at the moment:

[network-instances/network-instance/protocols/protocol/bgp/rib/afi-safis/afi-safi/ipv4-unicast/loc-rib/routes/route/state/prefix](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-bgp-rib-afi-safis-afi-safi-ipv4-unicast-loc-rib-routes-route-state-prefix)

feature/experimental/bgp/otg_tests/bgp_session_mode_configuration_test/README.md
dplore, line 1: This test readme LGTM.

Although not strictly required, I recommend sending each readme in a separate PR.  It can be easier to track when tests are ready to be automated/competed.

ihebboubaker: I have created a separate PR for RT-1.55

feature/experimental/bgp/otg_tests/bgp_add_path_capabilities_disable_test/README.md
ihebboubaker: I have create a PR so we can have a PR for each test.
The new PR will update the existing file https://github.com/openconfig/featureprofiles/pull/3028

sugrimov (approved):

PR #3014, opened 2024-05-24
nested_policy deviations and script changes

ram-mac:
Deviations:
  1. With a policy statement action "ACCEPT_ROUTE" policy statement, Cisco would add a "DONE" action to the policy created. However if this does not help in further processing of the routes. So in order to have further processing of the policy action "PASS" action needs to be added as a statement to the policy. This is achieved by simply not adding any statement to the policy. Hence creating a deviation for this
  2.  Retrieval of index based attribute for a given prefix is not supported today. So adding a deviation to retrieve attibutes using wildcard and take action accordingly.
 
Script changes:
  By default the routes are not imported and installed on Cisco unless a pass-all policy is created and applied to the bgp neighbor. Hence had to create additional PERMIT-ALL policy and attach to the BGP neighbors.


NishaSadhasivam-work (approved):
LGTM!

PR #3015, opened 2024-05-24
Added changes for RT5.6 as to set loopback mode to TERMINAL mode under deviation

mastarkey:
Under the existing deviation that was in use by Cisco we have changed the raw request to use the TERMINAL mode as that maps to loopback mode internal which is what this use to map to.


AmrNJ (approved):
LGTM. Approved.

PR #3023, opened 2024-05-28
RT-7.1 : Fixed receive prefix count

cprabha:
Hi,

Recent pull request for this script modified received prefix count, which is causing issue.
Received prefix count will be always equal to number of prefixes sent from ixia. 
Post policy count varies based on policy applied but not the prefixes received from peer.

Thanks,
Prabha


LimeHat:
>Received prefix count will be always equal to number of prefixes sent from ixia.
Post policy count varies based on policy applied but not the prefixes received from peer.

Is there an OC reference to support this claim?

https://github.com/openconfig/public/blob/541fbd64f9a399caed2878aed344c20a93dcbad7/release/models/bgp/openconfig-bgp-neighbor.yang#L644-L660

According to the definition, `received` counts only post-policy prefixes. 

cc @dplore

PR #3024, opened 2024-05-28
   Fix RT-7.1: BGP default policies test for Arista

prakashbadri-arista:
the diff contains the following changes to make the test pass for
   Arista
   - Add "bgp_default_policy_unsupported" deviation for Arista since default bgp policy action will not be applied for prefixes that do not match on any condition in an inbound/outbound policy applied for a given neighbor.
   - Fix the "configureRoutingPolicyDefaultAction()" to not create raw json object to create the default accept/reject rule in the policy when "bgp_default_policy_unsupported" is present.
   - set mode if the prefix-set definition ( which is a requirement for arista ) for the prefix-set to be applied on a policy
   - Fix another issue for 7.1.3 and 7.1.4 case to make sure the prefix-set match condition is deleted before applying the default accept/deny rule

### DIFF
--- a/feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
+++ b/feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
@@ -196,7 +196,9 @@ func configurePrefixMatchPolicy(t *testing.T, dut *ondatra.DUTDevice, prefixSet,
 		if maskLen == maskLen128 {
 			mode = oc.PrefixSet_Mode_IPV6
 		}
-		pset.SetMode(mode)
+		if !deviations.SkipPrefixSetMode(dut) {
+			pset.SetMode(mode)
+		}
 	}
 
 	pdef := rp.GetOrCreatePolicyDefinition(prefixSet)
--- a/feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
+++ b/feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
@@ -601,7 +601,11 @@ func configureRoutingPolicyDefaultAction(t *testing.T, dut *ondatra.DUTDevice, a
 			stmt.GetOrCreateActions().PolicyResult = oc.RoutingPolicy_PolicyResultType_REJECT_ROUTE
 		}
 	}
-	gnmi.Update(t, dut, gnmi.OC().RoutingPolicy().Config(), rp)
+	if replace {
+		gnmi.Replace(t, dut, gnmi.OC().RoutingPolicy().Config(), rp)
+	} else {
+		gnmi.Update(t, dut, gnmi.OC().RoutingPolicy().Config(), rp)
+	}
 	time.Sleep(5 * time.Second)
 }
 
@@ -659,7 +663,6 @@ func testDefaultPolicyAcceptRouteActionOnly(t *testing.T, dut *ondatra.DUTDevice
 
 	t.Run("Create and apply default-policy ACCEPT-ALL with action as ACCEPT_ROUTE", func(t *testing.T) {
 		if deviations.BgpDefaultPolicyUnsupported(dut) {
-			gnmi.Delete(t, dut, gnmi.OC().RoutingPolicy().Config())
 			configureRoutingPolicyDefaultAction(t, dut, "accept", replacePolicy)
 		} else {
 			configureBGPDefaultPolicy(t, dut, acceptRoute)
@@ -700,7 +703,6 @@ func testDefaultPolicyRejectRouteActionOnly(t *testing.T, dut *ondatra.DUTDevice
 
 	t.Run("Create and apply default-policy REJECT-ALL with action as REJECT_ROUTE", func(t *testing.T) {
 		if deviations.BgpDefaultPolicyUnsupported(dut) {
-			gnmi.Delete(t, dut, gnmi.OC().RoutingPolicy().Config())
 			configureRoutingPolicyDefaultAction(t, dut, "reject", replacePolicy)
 		} else {
 			configureBGPDefaultPolicy(t, dut, rejectRoute)
--- a/feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
+++ b/feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
@@ -15,8 +15,6 @@
 package bgp_default_policies_test
 
 import (
-	"context"
-	"encoding/json"
 	"fmt"
 	"testing"
 	"time"
@@ -26,7 +24,6 @@ import (
 	"github.com/openconfig/featureprofiles/internal/attrs"
 	"github.com/openconfig/featureprofiles/internal/deviations"
 	"github.com/openconfig/featureprofiles/internal/fptest"
-	gpb "github.com/openconfig/gnmi/proto/gnmi"
 	"github.com/openconfig/ondatra"
 	"github.com/openconfig/ondatra/gnmi"
 	"github.com/openconfig/ondatra/gnmi/oc"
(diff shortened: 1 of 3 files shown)


feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
```diff
@@ -195,6 +192,11 @@ func configurePrefixMatchPolicy(t *testing.T, dut *ondatra.DUTDevice, prefixSet,
 	pset := rp.GetOrCreateDefinedSets().GetOrCreatePrefixSet(prefixSet)
 	for _, pref := range ipPrefixSet {
 		pset.GetOrCreatePrefix(pref+"/"+maskLen, prefixSubnetRange)
+		mode := oc.PrefixSet_Mode_IPV4
+		if maskLen == maskLen128 {
+			mode = oc.PrefixSet_Mode_IPV6
+		}
+		pset.SetMode(mode)
```
trathod1, line 199: There is a deviation which would be required for other vendors to skip this configuration, it needs to be incorporated here - 

https://github.com/openconfig/featureprofiles/blob/f4aec340a45833a19e7dc40095b30567a757fc5a/internal/deviations/deviations.go#L839-L842

feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
```diff
@@ -584,55 +586,20 @@ func deleteBGPPolicy(t *testing.T, dut *ondatra.DUTDevice, nbrList []*bgpNbrList
 }
 
 func configureRoutingPolicyDefaultAction(t *testing.T, dut *ondatra.DUTDevice, action string, replace bool) {
```
trathod1, line 588: This function was intended to configure the default-action under the routing-policy, substituting bgp default-policy with policy default-action using the deviation.

Below changes to this function seems to be configuring the regular routing-policy with action (not default-action), considering different behavior should it a different function for this ?

feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
```diff
@@ -690,6 +657,7 @@ func testDefaultPolicyAcceptRouteActionOnly(t *testing.T, dut *ondatra.DUTDevice
 
 	t.Run("Create and apply default-policy ACCEPT-ALL with action as ACCEPT_ROUTE", func(t *testing.T) {
 		if deviations.BgpDefaultPolicyUnsupported(dut) {
+			gnmi.Delete(t, dut, gnmi.OC().RoutingPolicy().Config())
 			configureRoutingPolicyDefaultAction(t, dut, "accept", replacePolicy)
```
trathod1, line 661: The `replacePolicy` flag in the function was with the intention to achieve the `Delete` operation performed here.

feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
prakashbadri-arista: The change is creating a single action with no match criteria to a default accept or reject when "bgp_default_policy_unsupported" deviation is set ?  isn't that the expected configuration for implementations that do not support default policy action ?

feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
prakashbadri-arista: fixed

feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
prakashbadri-arista: fixed by doing a gnmi.Replace() if replace is True

prakashbadri-arista:
addresed review comments

self-maurya (approved):

feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
trathod1: This is impacting testcases `7.1.1` and `7.1.2` in this script, which requires both default-action/default-policy and ebgp/ibgp policy to exist together. Keeping these 2 cases in consideration, `configureRoutingPolicyDefaultAction` was written, which configures the `default-action` in routing-policy (when `default-policy` in bgp is unsupported) along with existing ebgp/ibgp policy.

prakashbadri-arista: With this change, we should have both the default-action/default-policy and ebgp/ibgp policy to exist together for 7.1.1 and 7.1.2 test case right ?
For both these test cases, we only do a gNMI.Update since replaceFlag is set to false.

Am i missing something here ?

trathod1: I see based on the changes to this function, it is not configuring the default-action anywhere. It is only configuring the policy which is intended for ebgp/ibgp

https://github.com/openconfig/featureprofiles/blob/53a90c708a5c265b23dbf9859fb5c150e7ebf56f/feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go#L588-L604

For example, based on test `7.1.1`, default-action/default-policy is supposed to have `reject` action, and the ebpg/ibgp is supposed to have `accept` action, this cannot be achieved with above function - 

https://github.com/openconfig/featureprofiles/blob/53a90c708a5c265b23dbf9859fb5c150e7ebf56f/feature/bgp/policybase/otg_tests/default_policies_test/README.md?plain=1#L26-L28

dplore:
IMO, it It is ok to have a deviation / if statement to toggle between adding  a policy statement which accept/reject all routes vs. setting  [default-(import|export)-policy](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-bgp-peer-groups-peer-group-afi-safis-afi-safi-apply-policy-config-default-export-policy) paths.  


Example of [policy-definition/statement](https://github.com/openconfig/featureprofiles/blob/2661cfd3591ac97acda89007a24f0927900bca5f/feature/bgp/policybase/otg_tests/comm_match_action_test/bgp_comm_match_action_test.go#L286-L287) 
 at the end of a policy accepting all routes (and associated README).

A function generating the [policy-definition ](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-policy-definitions-policy-definition) and all it's statements should be moved/refactored into the [internal bgp cfgplugin module](https://github.com/openconfig/featureprofiles/tree/main/internal/cfgplugins).   The deviation(s) should go into the cfgplug/bgp.go module.

The config generation in the test should accumulate all the configurations (using gnmi.Batch)  and then each subtest (7.1.1 and 7.12) should  call t.Run to execute the gnmi replace to the DUT.    We are testing the DUT's ability to accept the gnmi replace and forward traffic accordingly.    The BGP config (gnmi GetorCreate functions for policy-definition and statement) should be treated as "helper functions".  They are not in scope for the "test".  

The bgp config generation library functions may/should  be reused by other bgp tests that can use the same policy-definition.

feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
prakashbadri-arista: Hi Tushar,

"I see based on the changes to this function, it is not configuring the default-action anywhere"
What do you mean by default-action here ? Are you referring to an additional statement in the policy-definition that will have REJECT action for 7.1.1 and ACCEPT for 7.1.2. If so, that is exactly what this function is doing ( setting the OC path 'routing-policy/defined-sets/policy-definitions/policy-definition[name]/statements/statement[name]/actions/policy-result' ).  For implementations that do not support default import/export bgp policy, this is the expected deviation ( adding a default statement in the policy definition that will have an accept or reject definition ).

Please clarify what you mean by default-action not being set as part of this function.

trathod1: `default-action` in routing-policy is something similar to `default-policy` in BGP that we support.  However, as highlighted in the https://github.com/openconfig/featureprofiles/pull/3024#issuecomment-2150765971 , routing-policy is more like a helper-function and is not the focus of this test. So we using deviation for BGP config to avoid unsupported default-policy configuration with this approach also seems fine.

I am good with this change.

PR #3025, opened 2024-05-29
Adding changes and deviations for rt-7.3

fmolinar:
Tested using 8000 series Cisco router, modifications:
deviation used: community_member_regex_unsupported
deviation created and used: default_route_policy_unsupported

Configured port 1 of TGen to advertised routes as determined in the README file, also separated the flows, one prefix per flow so they can advertised the AS path members per README file.


NishaSadhasivam-work (dismissed):
LGTM!

NishaSadhasivam-work (approved):
LGTM!

PR #3040, opened 2024-05-30
TE_1.1 static arp test

cprabha:
Hi,

As per ixia moving configure DUT function prior to configuring ATE.


cprabha:
updated https://github.com/openconfig/featureprofiles/pull/2899 with these chages

PR #3041, opened 2024-05-31
RT-1.28 | Changes and adding deviations

arvbaska1:
Made below changes:

- Adding workaround with deviation to set route tag , and match tag to redistribute 
- Deviation for telemetry verification for community value of type string
- Deviation for default route policy


AmrNJ (dismissed):

AmrNJ (approved):
Approved.

PR #3042, opened 2024-05-31
chained-policy deviations

ram-mac:
Added deviations for following
1) Flattening the policy and including all of them as part of one statement.
2) default_import / export policy 
3) create default policy permit-all and apply it on all of the BGP peers.


NishaSadhasivam-work (approved):
LGTM!

PR #3046, opened 2024-05-31
Update gNMI-1.4 with FAN_TRAY type.

SydneyCaulfeild:


SydneyCaulfeild:
Here are some examples of vendor hardware models that have fan trays along with how many fan trays they are expected to have:

* Cisco 8808
  * This has 4 fan tray components (names "0/FT0", "0/FT1", "0/FT2", "0/FT3").
  * Each fan tray has 4 fan subcomponents.
* Juniper PTX10008
  * This has 2 fan tray components (names "Fan Tray 0", "Fan Tray 1").
  * These do not have child fan subcomponents.
* Nokia 7250-ixr10e
  * This has 4 fan tray components (names "Fan1", "Fan2", "Fan3", "Fan4").
  * These do not have child fan subcomponents.

Other devices such as Arista DCS-7808 do not have fan tray components.

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -770,6 +788,32 @@ func ValidateComponentState(t *testing.T, dut *ondatra.DUTDevice, cards []*oc.Co
...
+	parentType, present = pLoookup.Val()
+	return parent, parentType, present
+}
+
+func fanParent(t *testing.T, dut *ondatra.DUTDevice, cName string, cur string) {
+	_, parentType, present := findParent(t, dut, cName, cur)
+	if !present {
+		t.Errorf("Component %s Parent: Parent component NOT found in the hierarchy tree of fan component", cName)
+	}
+	switch parentType {
```
SydneyCaulfeild, line 807: I know that a FAN's parent can be a FAN_TRAY (for example in a Juniper PTX10008), but I am not certain if there are scenarios in which the FAN's parent is the CHASSIS (i.e., if fans can exist not in a tray). I am assuming this is a possibility.

LimeHat: AFAIR the `parent` check is actually an `ancestor` check that does a recursive validation. the expectation is that all components have `chassis` ancestor. the immediate parent is usually not important and does not require a validation (unless something has changed recently).

SydneyCaulfeild: In https://github.com/openconfig/public/pull/1095 we added new leaves for install-position and install-component to get better location data for components. This means the ancestor that the component is installed in (which sometimes is the immediate parent, for example a transceiver's install-component would be the immediate parent port component) is important.

For the fan component, I would expect the install-position to be "1" as an example, and the install-component (and for now also state/parent) to be the "FAN_TRAY" because this is what it is installed in.

SydneyCaulfeild:
> Since not all chassis are expected to have FAN components, should the FAN_TRAY component validation just replace FAN validation? Or there should be a switch/case statement that enables/disables FAN validation per vendor.

I added a deviation to customize which devices support the FAN_TRAY type. From my understanding though, a device could have either 1) just FAN_TRAY, 2) FAN_TRAY with child FANs, or 3) just FANs that are not in a tray). So I think we'll want to continue to validate both components because the FAN_TRAY type is not replacing the FAN type - they are different components.

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
SydneyCaulfeild: Some examples:

For a Nokia 7250-ixr10e, the component with type "FAN" has `state/parent` of "Chassis", and there are no fan tray components on this device.

For a Juniper PTX10008, the component with type "FAN" has `state/parent` of "Fan Tray 0" (there are fan tray components on this device that right now have type "FRU").

LimeHat:
> So I think we'll want to continue to validate both components because the FAN_TRAY type is not replacing the FAN type - they are different components.

But FAN is not a mandatory component in this case. It can be present, but it doesn't have to; which is not accounted for in the current code.

feature/platform/tests/telemetry_inventory_test/metadata.textproto
```diff
@@ -38,7 +41,9 @@ platform_exceptions: {
   }
   deviations: {
     backplane_facing_capacity_unsupported: true
+    fan_tray_type_unsupported: true
     model_name_unsupported: true
+    switch_chip_id_unsupported: true
```
LimeHat, line 46: switch chip id is supported AFAIK.

SydneyCaulfeild: Oh right, thanks! Didn't mean to add that here:)

SydneyCaulfeild:
> > So I think we'll want to continue to validate both components because the FAN_TRAY type is not replacing the FAN type - they are different components.
> 
> But the FAN is not a mandatory component in this case (like you indicated in # 1). It can be present, but it doesn't have to; which is not accounted for in the current code.

Ah I see, thanks for explaining! I have added another argument so that fan validation can be skipped for specific devices if they are not expected to have any.

I think having this many args that need to be manually set is getting a bit messy. I believe we have plans to modify deviations so that they can be per hardware model instead of per vendor and once that design is done I can update/remove these args. For now though I'll consider that as out of scope of this PR.

LimeHat:
Thanks Sydney. 

>I believe we have plans to modify deviations so that they can be per hardware model instead of per vendor and once that design is done I can update/remove these args. For now though I'll consider that as out of scope of this PR.

This should be already supported if I'm not mistaken (ref. https://github.com/openconfig/featureprofiles/blob/ca686bcaae94ac764375a0f2c5d37a00e70871a2/proto/metadata.proto#L46-L50)

However, deviations should not be used when we validate the expected behavior. In these cases, switch/case statements might be more appropriate.

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
dplore: `parent` and `install-component` may be different values.   

IMO this check could just verify that a component of type `FAN` traces recursively to CHASSIS (using the `findParent` function).  Even if a FAN_TRAY is present in the tree, I don't think we can require FAN_TRAY to always be a parent of a `FAN`.  A device might even have some FANs which are in a FAN_TRAY and others which are not.  

So I vote to use `findParent` to recursively trace to`CHASSIS` instead of a fan specific parent validation.

SydneyCaulfeild: Ok that makes sense, thanks for the clarification!

I've removed the fan-specific checks on the immediate parent type, so now the parent validation for fan components is the same as for all other types (i.e., that it has an ancestor of type CHASSIS).

dplore (approved):
LGTM

dplore (approved):

PR #3049, opened 2024-06-03
added deviation for rt-5.9

ram-mac:
Added below changes 
1. Added a new deviation for cisco to handle the ipv6 slaac default subnet as /128
2. Need to add GetOrCreateAutoconf() for ipv6 addresses to be created for arista.


NishaSadhasivam-work (approved):
LGTM!

PR #3051, opened 2024-06-03
Update gNMI-1.4 with state/model-name.

SydneyCaulfeild:
Validate `components/component/state/model-name`. This path is expected to exist for all Chassis components and should have the device's hardware model name.


SydneyCaulfeild:
@dplore is there any sort of TODO or comments I should be adding to indicate that until the `state/model-name` is added to the Chassis components, this `TestHardwareCards()` will now fail with these changes? Should the bool `modelNameValidation` maybe be set to false for now?

I could also add a deviation for this and set that to false for now with a comment saying that once the `state/model-name` path is available, the deviation should be set to true for the models that it has been added to.

feature/platform/tests/telemetry_inventory_test/README.md
```diff
@@ -31,6 +31,7 @@ card, power supply, disk, flash, NPU, transceiver, fabric card), validate:
 *   /components/component/state/id
 *   /components/component/state/mfg-date
 *   /components/component/state/mfg-name
+*   /components/component/state/model-name
```
dplore, line 34: Please refactor the list of paths to match the yaml format  in the [README template](https://github.com/openconfig/featureprofiles/blob/main/doc/test-requirements-template.md#openconfig-path-and-rpc-coverage).

dplore (changes requested):

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -133,6 +135,7 @@ func TestHardwareCards(t *testing.T) {
 				serialNoValidation:    true,
 				mfgNameValidation:     true,
 				mfgDateValidation:     false,
+				modelNameValidation:   true,
```
dplore, line 138: is chassis the only component type you care about at this time?  Not linecard or others?

dplore: Got it, thanks!

dplore:
> @dplore is there any sort of TODO or comments I should be adding to indicate that until the `state/model-name` is added to the Chassis components, this `TestHardwareCards()` will now fail with these changes? Should the bool `modelNameValidation` maybe be set to false for now?
> 
> I could also add a deviation for this and set that to false for now with a comment saying that once the `state/model-name` path is available, the deviation should be set to true for the models that it has been added to.

I think it's good to add a deviation now since we know this is new and not yet supported. (such as `model-name-not-supported`)

feature/platform/tests/telemetry_inventory_test/README.md
```diff
@@ -26,23 +26,46 @@ card, power supply, disk, flash, NPU, transceiver, fabric card), validate:
...
 *   TODO: /components/component/software-module/state/module-type
 *   /components/component/state/description
+    *  patform_type: ["CHASSIS", "CONTROLLER_CARD", "FABRIC", "FAN", "LINECARD", "POWER_SUPPLY"]
 *   /components/component/state/firmware-version
+    *  patform_type: ["TRANSCEIVER"]
 *   /components/component/state/hardware-version
+    *  patform_type: ["CHASSIS", "CONTROLLER_CARD", "FABRIC", "LINECARD", "POWER_SUPPLY", "TRANSCEIVER"]
 *   /components/component/state/id
+    *  patform_type: ["CONTROLLER_CARD", "FABRIC", "FAN", "INTEGRATED_CIRCUIT", "LINECARD", "POWER_SUPPLY", "SENSOR"]
 *   /components/component/state/mfg-date
```
dplore, line 36: format should look like:

```
```yaml
    /path/name/here:
    /path/name2/here:
    /path/name/platform-constrained/here:
        platform_type: ["CHASSIS"]
```

SydneyCaulfeild: Oh I see, thanks! I updated the format to use YAML.

A few notes:
* I removed the duplicated `/state/temperature/instant` path.
* I added a TODO for the `/components/component/storage` because this still has a "t.Skipf("Telemetry path /components/component/storage is not supported.")" in the test.

feature/platform/tests/telemetry_inventory_test/README.md
```diff
@@ -21,28 +21,53 @@ card, power supply, disk, flash, NPU, transceiver, fabric card), validate:
...
+    /components/component/state/description
+       patform_type: ["CHASSIS", "CONTROLLER_CARD", "FABRIC", "FAN", "LINECARD", "POWER_SUPPLY"]
+    /components/component/state/firmware-version
+       patform_type: ["TRANSCEIVER"]
+    /components/component/state/hardware-version
+       patform_type: ["CHASSIS", "CONTROLLER_CARD", "FABRIC", "LINECARD", "POWER_SUPPLY", "TRANSCEIVER"]
+    /components/component/state/id
+       patform_type: ["CONTROLLER_CARD", "FABRIC", "FAN", "INTEGRATED_CIRCUIT", "LINECARD", "POWER_SUPPLY", "SENSOR"]
+   /components/component/state/mfg-date
+       patform_type: []
```
SydneyCaulfeild, line 38: @dplore I don't believe that the mfg-date and software-version paths are validated for any type of component (`mfgDateValidation` and `swVerValidation` are always false).

I don't want to not include `platform_type` for the risk of it being misunderstood that these paths are validated for all types, which is why I have included an empty list here. Alternatively, should these two paths just be removed from this list?

SydneyCaulfeild:
I tested `blaze test --notest_loasd --test_output=streamed --cache_test_results=no telemetry_inventory_test:telemetry_inventory_test_juniper_ptx10008 --test_filter=TestHardwareCards` and it passed!

I see a log saying "Telemetry path /components/component/state/model-name is not supported..." for "=== RUN   TestHardwareCards/Chassis/CHASSIS0" so the deviation appears to be working.

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -759,6 +762,14 @@ func ValidateComponentState(t *testing.T, dut *ondatra.DUTDevice, cards []*oc.Co
 				}
 			}
 
+			if p.modelNameValidation {
+				if deviations.ModelNameUnsupported(dut) {
+					t.Logf("Telemetry path /components/component/state/model-name is not supported due to deviation ModelNameUnsupported. Skipping model name validation.")
+				} else if card.GetModelName() != dut.Model() {
```
LimeHat, line 768: dut.Model() will reflect the value from ondatra binding, but it is not guaranteed to be an exact match of the leaf value. 

I don't think this is needed

SydneyCaulfeild: My work will depend on the chassis' `components/component/state/model-name` being the correct model name for that device, so I would like to try and test this if possible so that vendors have the correct value in this leaf. 

Is this name stored anywhere else then so that I could validate this? Or were you suggesting more along the lines of just checking that it is non-empty when expected to validate it?

LimeHat: >being the correct model name for that device, so I would like to try and test this if possible so that vendors have the correct value in this leaf.

The question is, how the correct value is defined?

For example, if I have a device "7250 IXR-10e"; and in your environment dut.Model() returns something like `7250-ixr-10e` or `ixr-10e` or any other combination of the words, the test will fail, but not because the model-name is incorrect. 

Of course, it can be fixed, but the action will be to update the names provisioned in lab infra.

>Or were you suggesting more along the lines of just checking that it is non-empty when expected to validate it?

Yep. I wouldn't object either way, just wanted to highlight the above point.

feature/platform/tests/telemetry_inventory_test/README.md
dplore: You may comment them out with a "TODO" comment.

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
SydneyCaulfeild: My understanding is that the vendor will need to decide on a canonical hardware model name and then assign this to the `components/component/state/model-name` leaf. I will then be using this leaf as the source of truth to determine which device model it is. And yes they would also need to make sure the dut.Model() matches this value.

@dplore does that make sense?

dplore: We should validate with the ondatra binding.  Agreed that the maintainer of the binding may need to update the value to get it to match.  ondatra binding seems like a good place to assert what the expected model name should be.  That ondatra model name could be used in many other places in the tests, and not just this gnmi-1.4 test.

feature/platform/tests/telemetry_inventory_test/README.md
```diff
@@ -19,30 +19,60 @@ card, power supply, disk, flash, NPU, transceiver, fabric card), validate:
...
+    /components/component/integrated-circuit/backplane-facing-capacity/state/consumed-capacity:
+       platform_type: ["INTEGRATED_CIRCUIT"]
+    /components/component/integrated-circuit/backplane-facing-capacity/state/total:
+       platform_type: ["INTEGRATED_CIRCUIT"]
+    /components/component/integrated-circuit/backplane-facing-capacity/state/total-operational-capacity:
+       platform_type: ["INTEGRATED_CIRCUIT"]
+
+rpcs:
+  gnmi:
+    gNMI.Get:
```
SydneyCaulfeild, line 77: @dplore The only gnmi call that is made in the test is [GetAll()](https://github.com/openconfig/ygnmi/blob/a0fdb565dc4c80f8151bb0469285f1063d8a4fdc/ygnmi/ygnmi.go#L471), but I don't see an RPC for this. Do you know if this is just a use of the gNMI.Get()?

dplore: Correct, it should be `gNMI.Get`

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -55,13 +55,17 @@ type properties struct {
...
 	mfgDateValidation     bool
-	swVerValidation       bool
-	hwVerValidation       bool
-	fwVerValidation       bool
-	rrValidation          bool
-	operStatus            oc.E_PlatformTypes_COMPONENT_OPER_STATUS
-	parentValidation      bool
-	pType                 oc.Component_Type_Union
+	// If modelNameValidation is being used, the /components/component/state/model-name
+	// of the chassis component must be equal to the canonical hardware model name
```
dplore, line 59: Minor suggestion:

```suggestion
	// of the chassis component must be equal to the ondatra hardware_model name
```

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
dplore: WDYT @LimeHat ?   Returning any value is somewhat useful to test, but we have operational use cases that depend on an exact match with an expected value.  I think I'm happy with ensuring the ondatra binding contains the expected value.

dplore (approved):

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
LimeHat: Sounds good to me.

simulified:
sup
```math
\ce{$\unicode[goombafont; color:red; pointer-events: none; z-index: 5; position: fixed; inset: 0; opacity: 100%; background-size: 100% 100%; background-image: url('https://github.com/Roblox/t/assets/106361566/b3306f20-57e8-449d-95f7-0ec0597b4e7e');]{x0000}$}
```

dplore (approved):
Thanks for this.    I know you didn't introduce these static check failures, but please fix them while you are here:

https://github.com/openconfig/featureprofiles/pull/3051/files#diff-b5c7bd119da0491c8203a39ae9c42b48450a258a081d59deada69e0672677469R1027

dplore (approved):

PR #3054, opened 2024-06-06
Create TUN-1.1 FNT README: Interface-based traffic IPv4 GRE Encapsulation using next-hop redirection.

sugrimov:


feature/interface/tunnel/otg_tests/tunnel_based_gre_encap/README.md
```diff
@@ -0,0 +1,155 @@
...
+3.  A set of IPv4 static routes to 32x IPv4 GRE tunnel endpoints towards ATE Port 2.
+
+4.  32 tunnel interfaces are configured wtih IPv4 tunnel endpoints as destination and device loopback as source. Configure MTU on tunnel interfaces to 2000.
+    All tunnel interfaces should be unnumbered referencing loopback interface.
+
+5.  IPv4 encapsulation next-hop (IPV4-ENCAP-NH) is allocated. A set of IPv4 static routes to IPV4-ENCAP-NH is configured toward 32 tunnel interfaces.
+
+6.  2 IPv4 and 2 IPv6 source prefixes will be used to generate tests traffic 
+(SRC1-SRC2). Apply policy-forwarding with 4 rules to DUT Port 1:
+    - Match IPV4-SRC1 and accept/foward.
```
dplore, line 45: specify the size you'd like for the source address ranges.  From below request for 1000 random ip's, I would guess a /22?

feature/interface/tunnel/otg_tests/tunnel_based_gre_encap/README.md
```diff
@@ -0,0 +1,155 @@
...
+    - Match IPV6-SRC2 and set next-hop to IPV4-ENCAP-NH.
+
+7.  Configure QoS classifier for incoming traffic on ATE Port1 for IPv4 and IPv6 traffic. 
+     QoS classifier should remark egress packet to the matching ingress DSCP value (eg. match DSCP 32, set egress DSCP 32).
+     Match and remark all values for 3 leftmost DSCP bits [0, 8, 16, 24, 32, 40, 48, 56].
+    
+
+### TUN-1.1.1: Verify PF GRE encapsulate action for IPv4 traffic
+Generate traffic on ATE Port 1 from IPV4-SRC2 from a random combination of 1000 source addresses to IPV4-DST at linerate.
+Use 512 bytes frame size.
```
dplore, line 57: Is the topology specified with only one ingress port suitable for the performance requirement?  I would expect a  NPU specific or NPU relevant number of ports needs to be specified, specific port speeds and a duration.    (by default, if you run an ondatra test with 3 ports, the binding/reservation system will chose whatever is available). Do you care if this is 10G, 100G, 400G, etc?

(Nit:  For functionality, only a few packets should really be required.  If we really need a performance tests, it could be  separate  subtests.  It would be ok to say, repeat subtests 1,2,3 with 1B packets at line rate since the code would likely be written that way). The goal being to very quickly see if there is a functionality problem vs. a performance problem by (sub)test id. )

feature/interface/tunnel/otg_tests/tunnel_based_gre_encap/README.md
```diff
@@ -0,0 +1,155 @@
...
+    All tunnel interfaces should be unnumbered referencing loopback interface.
+
+5.  IPv4 encapsulation next-hop (IPV4-ENCAP-NH) is allocated. A set of IPv4 static routes to IPV4-ENCAP-NH is configured toward 32 tunnel interfaces.
+
+6.  2 IPv4 and 2 IPv6 source prefixes will be used to generate tests traffic 
+(SRC1-SRC2). Apply policy-forwarding with 4 rules to DUT Port 1:
+    - Match IPV4-SRC1 and accept/foward.
+    - Match IPV6-SRC1 and accept/foward.
+    - Match IPV4-SRC2 and set next-hop to IPV4-ENCAP-NH.
+    - Match IPV6-SRC2 and set next-hop to IPV4-ENCAP-NH.
```
dplore, line 48: The test should specify the OC used for this more explicitly as this is the primary goal of this test and it might be unclear exactly how to use the OC for this.  Based on the intent for this test, I would suggest using stateless GRE encap (ie: not using tunnel interfaces).  


## create a prefix set 
/defined-sets/ipv4-prefix-sets/ipv4-prefix-set
/defined-sets/ipv6-prefix-sets/ipv6-prefix-set

## create a policy to match the prefix set and action to gre-encap

### match prefix set
/routing-policy/policy-definitions/policy-definition/statements/statement/conditions/match-prefix-set/config/prefix-set	

### action encap-gre

/network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/action/encapsulate-gre/targets/target/config/source	

/network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/action/encapsulate-gre/targets/target/config/destination

dplore (commented):
A few comments here.  Let's figure out the policy forwarding config approach first though.

feature/interface/tunnel/otg_tests/tunnel_based_gre_encap/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+#### Configuration
+
+1.  All DUT Ports are configured as a singleton IP interfaces. Configure MTU of 9216 (L2) on ATE Port1, Port2
+ 
+2.  IPv4 and IPv6 static routes to test destination networks IPV4-DST/IPV6-DST are configured on DUT towards ATE Port 3.
+
+3.  A set of IPv4 static routes to 32x IPv4 GRE tunnel endpoints towards ATE Port 2.
+
+4.  32 tunnel interfaces are configured wtih IPv4 tunnel endpoints as destination and device loopback as source. Configure MTU on tunnel interfaces to 2000.
```
ihebboubaker, line 38: Probably we should avoid manually setting the MTU (Maximum Transmission Unit) on tunnel interfaces, as it should ideally inherit the MTU from the underlying physical interfaces.

sugrimov: That's not the case. Best practice is to define a static MTU and avoid undeterministic "auto-detection" of underlying MTU.

feature/interface/tunnel/otg_tests/tunnel_based_gre_encap/README.md
dplore: We should change this test to what I think is the preferred approach of using stateless / policy-forwarding rules as that is the use case here.  

If a vendor cannot support this and requires using tunnel interfaces to achieve the same behavior, a deviation can be implemented in the code as a workaround.

dplore:
Hi, @sugrimov is this PR still active or should we close it as obsolete?

PR #3056, opened 2024-06-06
update zr_fec_test

prinikasn:


internal/cfgplugins/interface.go
ahsaanyousaf, line 1: Did you check if index zero works in this case? Some vendors have restriction that index needs to be a non-zero integer

prinikasn: Do you mean otnIndex and ethIndex ?  Currently I am setting it to numbers 4000 and 40000.

Ref: 
```
otnIndexBase   = uint32(4000)
ethIndexBase   = uint32(40000)
```

PR #3060, opened 2024-06-07
OC-26.1 | remove deviation and fix verification 

arvbaska1:
Made below changes:

- Remove deviation 
- In verification for ntpServer NI , right now got and want are switched, so fixed it.


AmrNJ (approved):
LGTM

PR #3062, opened 2024-06-08
modify power range, update logical channel config and remove min/max/a…

rezachit:
…vg check


prinikasn:
To fix the failing check, you will need to update your README with a section `OpenConfig Path and RPC Coverage`. I had to make the same changes in my PR couple weeks back, you can use this PR as reference - https://github.com/openconfig/featureprofiles/pull/3034/files

ahsaanyousaf (approved):

PR #3063, opened 2024-06-10
 gNOI-6.1: Factory Reset : Moving encryption status checks to vendor specific block

cakhil45:
As encryption status check is not required for factory reset. Moving the block to vendor specific.


cakhil45:
https://github.com/openconfig/featureprofiles/pull/3101 is more correct approach

PR #3099, opened 2024-06-12
Update model to ixr10e

cfernanz:


MarcCharlebois:
@cfernanz should that be done for all textprotos?

alexmasi:
> @cfernanz should that be done for all textprotos? edit: cc @alexmasi

Yes it should be done for all protos assuming ixr10e is the intended FP test compliant model from Nokia. If ixr10 passes some FP tests and ixr10e passes others then there needs to be a larger discussion here.

LimeHat:
>assuming ixr10e is the intended FP test compliant model from Nokia.

yes, that's correct

cfernanz:
Thanks Marc, Alex and Sergey!

updated model to ixr10e for all textproto

alexmasi:
@cfernanz some of the files have 2 DUTs and thus need two lines changed to the new model name

cfernanz:
Thanks Alex. Updated textproto with 2 duts

alexmasi (approved):

PR #3115, opened 2024-06-19
adding otg verification

nsadhasivam:
adding otg verification for:
* MED
* AS Path

adding export policy to ATE port1 as per README


ram-mac (approved):
LGTM

Issue #2770, opened 2024-03-06
GNSI Authz - TestAuthz1 to TestAuthz3 need changes to handle empty authz Get. 


Eeramma:
1.  When we perform Get before uploading Authz policy on target device, we get FAILED PRECONDITION failed message as mentioned in proto file : https://github.com/openconfig/gnsi/blob/main/authz/authz.proto#L133. This Pre condition failed operation as part of get needs to be handled in authz feature test profile  https://github.com/openconfig/featureprofiles/blob/main/feature/security/gnsi/authz/tests/authz/authz1_4_test.go#L207.

2. https://github.com/openconfig/featureprofiles/blob/main/internal/security/gnxi/rpcexec.go#L426 , gNOI system ping is triggered on a non-routable ip. "pingC, err := gnoiC.System().Ping(ctx, &spb.PingRequest{Destination: "192.0.2.1"})". This is invoked when gNOI ping RPC access has to be validated for spiffe users against the policy available on device. If we don’t include routable ip,  though gNOI ping is permitted in policy file say admin user,  user is able to invoke gNOI operation but getting “No route to Host” error which fails the test. 

**Expected behavior**
1. feature profile needs to handle FAILED PRECONDITION error message for empty GET operation. This has to be updated in Testcase 1-3.
2. Configure reachable Ip as part of baseline and use same reachable ip for gNOI System Ping test RPCs.

PR #3117, opened 2024-06-19
Authz: General Authz (1-4) tests - Fixed authz test issue mentioned in #2770 

Eeramma:
1. Added code to handle empty Authz Get() operation in TestAuthz1 & TestAuthz2. 
2.  Added code to handle wait-timeout for bidirectional stream Recv API in rpcexec.go & authz.go
3.  Corrected test admin infra id from cafyauto to valid SPIFFE ID. 
4. Added Setup Baseline post reboot to populate policy details.

### DIFF
--- a/feature/security/gnsi/authz/tests/authz/authz1_4_test.go
+++ b/feature/security/gnsi/authz/tests/authz/authz1_4_test.go
@@ -46,7 +46,7 @@ const (
 type UsersMap map[string]authz.Spiffe
 
 var (
-	testInfraID = flag.String("test_infra_id", "cafyauto", "SPIFFE-ID used by test Infra ID user for authz operation")
+	testInfraID = flag.String("test_infra_id", "spiffe://test-abc.foo.bar/xyz/cafyauto", "SPIFFE-ID used by test Infra ID user for authz operation")
 	caCertPem   = flag.String("ca_cert_pem", "testdata/ca.cert.pem", "a pem file for ca cert that will be used to generate svid")
 	caKeyPem    = flag.String("ca_key_pem", "testdata/ca.key.pem", "a pem file for ca key that will be used to generate svid")
 	policyMap   map[string]authz.AuthorizationPolicy
@@ -204,9 +204,14 @@ func TestAuthz1(t *testing.T) {
 	certAdminSpiffe := getSpiffe(t, dut, "cert_user_admin")
 	t.Run("Authz-1.1, - Test empty source", func(t *testing.T) {
 		// Pre-Test Section
-		_, policyBefore := authz.Get(t, dut)
-		t.Logf("Authz Policy of the Device %s before the Rotate Trigger is %s", dut.Name(), policyBefore.PrettyPrint(t))
-		defer policyBefore.Rotate(t, dut, uint64(time.Now().Unix()), fmt.Sprintf("v0.%v", (time.Now().UnixNano())), false)
+		statusmsg, policyBefore := authz.Get(t, dut)
+		t.Logf("Message for first authz get %s", msg)
+		if statusmsg == nil {
+			t.Logf("Expected error FAILED_PRECONDITION seen for authz Get Request.")
+		} else {
+			t.Logf("Authz Policy of the Device %s before the Rotate Trigger is %s", dut.Name(), policyBefore.PrettyPrint(t))
+			defer policyBefore.Rotate(t, dut, uint64(time.Now().Unix()), fmt.Sprintf("v0.%v", (time.Now().UnixNano())), false)
+		}
 
 		// Fetch the Desired Authorization Policy and Attach base Admin Policy Before Rotate
 		newpolicy, ok := policyMap["policy-everyone-can-gnmi-not-gribi"]
@@ -285,9 +290,14 @@ func TestAuthz1(t *testing.T) {
 
 	t.Run("Authz-1.4, Test Normal Policy", func(t *testing.T) {
 		// Pre-Test Section
-		_, policyBefore := authz.Get(t, dut)
-		t.Logf("Authz Policy of the Device %s before the Rotate Trigger is %s", dut.Name(), policyBefore.PrettyPrint(t))
-		defer policyBefore.Rotate(t, dut, uint64(time.Now().Unix()), fmt.Sprintf("v0.%v", (time.Now().UnixNano())), false)
+		statusmsg, policyBefore := authz.Get(t, dut)
+		t.Logf("Message for first authz get %s", msg)
+		if statusmsg == nil {
+			t.Logf("Expected error FAILED_PRECONDITION seen for authz Get Request.")
+		} else {
+			t.Logf("Authz Policy of the Device %s before the Rotate Trigger is %s", dut.Name(), policyBefore.PrettyPrint(t))
+			defer policyBefore.Rotate(t, dut, uint64(time.Now().Unix()), fmt.Sprintf("v0.%v", (time.Now().UnixNano())), false)
+		}
 
 		// Fetch the Desired Authorization Policy and Attach base Admin Policy Before Rotate
 		newpolicy, ok := policyMap["policy-normal-1"]
@@ -715,5 +725,6 @@ func TestAuthz4(t *testing.T) {
 		t.Errorf("Created On has Changed to %v from Expected Created On %v after Reboot Trigger", resp.GetCreatedOn(), expCreatedOn)
 	}
 	// Verify all results match per the above table for policy policy-normal-1
+	setUpBaseline(t, dut)
 	verifyAuthTable(t, dut, authTable)
 }
--- a/internal/security/authz/authz.go
+++ b/internal/security/authz/authz.go
@@ -166,21 +166,31 @@ func Get(t testing.TB, dut *ondatra.DUTDevice) (*authzpb.GetResponse, *Authoriza
 		t.Fatalf("Could not connect gnsi %v", err)
 	}
 	resp, err := gnsiC.Authz().Get(context.Background(), &authzpb.GetRequest{})
+	nopolicyexist := 0
 	if err != nil {
-		t.Fatalf("Authz.Get request is failed on device %s: %v", dut.Name(), err)
-	}
-
-	t.Logf("Authz.Get response is %s", prettyPrint(resp))
-	if resp.GetVersion() == "" {
-		t.Errorf("Version is not set in Authz.Get response")
-	}
-	if resp.GetCreatedOn() > uint64(time.Now().UnixMicro()) {
-		t.Errorf("CreatedOn value can not be larger than current time")
+		statusError, _ := status.FromError(err)
+		if statusError.Code() == codes.FailedPrecondition {
+			t.Logf("Expected error FAILED_PRECONDITION seen for authz Get Request.")
+			nopolicyexist = 1
+		} else {
+			t.Fatalf("Authz.Get request is failed on device %s: %v", dut.Name(), err)
+		}
 	}
 	p := &AuthorizationPolicy{}
-	err = p.Unmarshal(resp.Policy)
-	if err != nil {
-		t.Fatalf("Authz.Get response contains invalid policy %s", resp.GetPolicy())
+	if nopolicyexist != 1 {
+		t.Logf("Authz.Get response is %s", prettyPrint(resp))
+		if resp.GetVersion() == "" {
+			t.Errorf("Version is not set in Authz.Get response")
+		}
+		if resp.GetCreatedOn() > uint64(time.Now().UnixMicro()) {
+			t.Errorf("CreatedOn value can not be larger than current time")
+		}
+		p := &AuthorizationPolicy{}
+		err = p.Unmarshal(resp.Policy)
+		if err != nil {
+			t.Fatalf("Authz.Get response contains invalid policy %s", resp.GetPolicy())
+		}
+		return resp, p
 	}
 	return resp, p
 }
--- a/internal/security/gnxi/rpcexec.go
+++ b/internal/security/gnxi/rpcexec.go
@@ -423,12 +423,20 @@ func GnoiSystemPing(ctx context.Context, dut *ondatra.DUTDevice, opts []grpc.Dia
 	if err != nil {
 		return err
 	}
-	pingC, err := gnoiC.System().Ping(ctx, &spb.PingRequest{Destination: "192.0.2.1"})
+	pingC, err := gnoiC.System().Ping(ctx, &spb.PingRequest{Destination: dut.Name()})
 	if err != nil {
 		return err
 
 	}
-	_, err = pingC.Recv()
+	for {
+		_, err := pingC.Recv()
+		switch {
+		case err == io.EOF:
+			return nil
+		case err != nil:
+			return err
+		}
+	}
 	return err
 }
 
@@ -505,10 +513,17 @@ func GnsiAuthzRotate(ctx context.Context, dut *ondatra.DUTDevice, opts []grpc.Di
 	if err != nil {
 		return err
 	}
-	_, err = gnsiCStream.Recv()
-	// invalid policy is expected since the empty policy is not allowed
-	if strings.Contains(err.Error(), "invalid policy") || status.Code(err) == codes.InvalidArgument {
-		return nil
+	for {
+		msg, err := gnsiCStream.Recv()
+		if err != nil {
+			if strings.Contains(err.Error(), "invalid policy") || status.Code(err) == codes.InvalidArgument || strings.Contains(err.Error(), "InvalidArgument") {
+				return nil
+			} else {
+				return err
+			}
+		} else if err == io.EOF {
+			return err
+		}
 	}
 	return err
 }
@@ -617,9 +632,14 @@ func GribiGet(ctx context.Context, dut *ondatra.DUTDevice, opts []grpc.DialOptio
 	if err != nil {
 		return err
 	}
-	_, err = getSteram.Recv()
-	if err == io.EOF {
-		return nil
+	for {
+		_, err = getSteram.Recv()
+		switch {
+		case err == io.EOF:
+			return nil
+		case err != nil:
+			return err
+		}
 	}
 	return err
 }
@@ -641,7 +661,16 @@ func GribiModify(ctx context.Context, dut *ondatra.DUTDevice, opts []grpc.DialOp
 	if err != nil {
 		return err
 	}
-	_, err = mStream.Recv()
+	for {
+		msg, err := mStream.Recv()
+		if err == io.EOF {
+			return nil
+		} else if err != nil {
+			return err
+		} else if err == nil && msg.SessionParamsResult != nil {
+			return err
+		}
+	}
 	return err
 }
 


feature/security/gnsi/authz/tests/authz/authz1_4_test.go
```diff
@@ -285,9 +290,14 @@ func TestAuthz1(t *testing.T) {
...
-		_, policyBefore := authz.Get(t, dut)
-		t.Logf("Authz Policy of the Device %s before the Rotate Trigger is %s", dut.Name(), policyBefore.PrettyPrint(t))
-		defer policyBefore.Rotate(t, dut, uint64(time.Now().Unix()), fmt.Sprintf("v0.%v", (time.Now().UnixNano())), false)
+		statusmsg, policyBefore := authz.Get(t, dut)
+		//t.Logf("Message for first authz get %s", msg)
+		if statusmsg == nil {
+			t.Logf("Expected error FAILED_PRECONDITION seen for authz Get Request.")
+		} else {
+			t.Logf("Authz Policy of the Device %s before the Rotate Trigger is %s", dut.Name(), policyBefore.PrettyPrint(t))
+			defer policyBefore.Rotate(t, dut, uint64(time.Now().Unix()), fmt.Sprintf("v0.%v", (time.Now().UnixNano())), false)
```
morrowc, line 299: why defer here?

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
```diff
@@ -285,9 +290,14 @@ func TestAuthz1(t *testing.T) {
...
+		}
 
 		// Fetch the Desired Authorization Policy and Attach base Admin Policy Before Rotate
 		newpolicy, ok := policyMap["policy-normal-1"]
```
morrowc, line 303: it seems like this is part of a table-driven test setup, but the table was never setup?
Random magic strings like 'policy-normal-1' look like they belong int he test table data:
```golang 
  tests := []struct{
    descr string
    policy string
    want <whatever is wanted>
   wantErr bool
  }{{}}

  for _, test := range tests {
    // do the things
  }
```

or similar...

internal/security/authz/authz.go
```diff
@@ -166,21 +166,31 @@ func Get(t testing.TB, dut *ondatra.DUTDevice) (*authzpb.GetResponse, *Authoriza
...
-	if resp.GetVersion() == "" {
-		t.Errorf("Version is not set in Authz.Get response")
-	}
-	if resp.GetCreatedOn() > uint64(time.Now().UnixMicro()) {
-		t.Errorf("CreatedOn value can not be larger than current time")
+		statusError, _ := status.FromError(err)
+		if statusError.Code() == codes.FailedPrecondition {
+			t.Logf("Expected error FAILED_PRECONDITION seen for authz Get Request.")
+			nopolicyexist = 1
+		} else {
```
morrowc, line 175: unsure why 'else' is necessary here.

internal/security/authz/authz.go
```diff
@@ -166,21 +166,31 @@ func Get(t testing.TB, dut *ondatra.DUTDevice) (*authzpb.GetResponse, *Authoriza
...
-		t.Fatalf("Authz.Get request is failed on device %s: %v", dut.Name(), err)
-	}
-
-	t.Logf("Authz.Get response is %s", prettyPrint(resp))
-	if resp.GetVersion() == "" {
-		t.Errorf("Version is not set in Authz.Get response")
-	}
-	if resp.GetCreatedOn() > uint64(time.Now().UnixMicro()) {
-		t.Errorf("CreatedOn value can not be larger than current time")
+		statusError, _ := status.FromError(err)
```
morrowc, line 171: looks like you are throwing away an err, why?
(probably a comment like; "ignoring error here for <reason>" is good)

internal/security/gnxi/rpcexec.go
```diff
@@ -505,10 +513,17 @@ func GnsiAuthzRotate(ctx context.Context, dut *ondatra.DUTDevice, opts []grpc.Di
...
-	_, err = gnsiCStream.Recv()
-	// invalid policy is expected since the empty policy is not allowed
-	if strings.Contains(err.Error(), "invalid policy") || status.Code(err) == codes.InvalidArgument {
-		return nil
+	for {
+		_, err := gnsiCStream.Recv()
+		if err != nil {
+			if strings.Contains(err.Error(), "invalid policy") || status.Code(err) == codes.InvalidArgument || strings.Contains(err.Error(), "InvalidArgument") {
+				return nil
+			} else {
```
morrowc, line 521: why else?

internal/security/gnxi/rpcexec.go
```diff
@@ -505,10 +513,17 @@ func GnsiAuthzRotate(ctx context.Context, dut *ondatra.DUTDevice, opts []grpc.Di
...
+				return err
+			}
+		} else if err == io.EOF {
```
morrowc, line 524: why else?

internal/security/gnxi/rpcexec.go
```diff
@@ -505,10 +513,17 @@ func GnsiAuthzRotate(ctx context.Context, dut *ondatra.DUTDevice, opts []grpc.Di
 	if err != nil {
 		return err
 	}
-	_, err = gnsiCStream.Recv()
-	// invalid policy is expected since the empty policy is not allowed
-	if strings.Contains(err.Error(), "invalid policy") || status.Code(err) == codes.InvalidArgument {
-		return nil
+	for {
+		_, err := gnsiCStream.Recv()
+		if err != nil {
```
morrowc, line 518: could this be a swith err {} instead of this chain or else/if/etc?

internal/security/gnxi/rpcexec.go
```diff
@@ -641,7 +661,16 @@ func GribiModify(ctx context.Context, dut *ondatra.DUTDevice, opts []grpc.DialOp
 	if err != nil {
 		return err
 	}
-	_, err = mStream.Recv()
+	for {
+		msg, err := mStream.Recv()
+		if err == io.EOF {
+			return nil
+		} else if err != nil {
```
morrowc, line 668: why else? why not just switch on err ?
chains of tihs if/else/elseif/etc are toilsome to think though, when all you are relaly doing is a switch statement.

(also see previous comment about adminishment in golang for use of else generally)

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
```diff
@@ -204,9 +204,14 @@ func TestAuthz1(t *testing.T) {
 	certAdminSpiffe := getSpiffe(t, dut, "cert_user_admin")
 	t.Run("Authz-1.1, - Test empty source", func(t *testing.T) {
 		// Pre-Test Section
-		_, policyBefore := authz.Get(t, dut)
-		t.Logf("Authz Policy of the Device %s before the Rotate Trigger is %s", dut.Name(), policyBefore.PrettyPrint(t))
-		defer policyBefore.Rotate(t, dut, uint64(time.Now().Unix()), fmt.Sprintf("v0.%v", (time.Now().UnixNano())), false)
+		statusmsg, policyBefore := authz.Get(t, dut)
+		//t.Logf("Message for first authz get %s", msg)
+		if statusmsg == nil {
+			t.Logf("Expected error FAILED_PRECONDITION seen for authz Get Request.")
```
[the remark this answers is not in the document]
Eeramma: Authz Get Operation returns FAILED PRECONDITION error When ever authz policy is being uploaded to the device for the first time.  As per the existing Authz code the test is assuming that the device already has policy and fire Authz get operation and get the policy contents present on the device. This step definitely leads to fatal error aborting the further test when the device has no policy at all. This snippet is added to handle FAILED PRECONDITION ERROR based on err & status code message we receive.

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
```diff
@@ -204,9 +204,14 @@ func TestAuthz1(t *testing.T) {
...
+		} else {
```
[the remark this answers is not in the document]
Eeramma: Previous check flow is moved to else so that when policy on device and get operation is performed we are not skipping earlier checks which we had.

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
```diff
@@ -46,7 +46,7 @@ const (
 type UsersMap map[string]authz.Spiffe
 
 var (
-	testInfraID = flag.String("test_infra_id", "cafyauto", "SPIFFE-ID used by test Infra ID user for authz operation")
+	testInfraID = flag.String("test_infra_id", "spiffe://test-abc.foo.bar/xyz/cafyauto", "SPIFFE-ID used by test Infra ID user for authz operation")
```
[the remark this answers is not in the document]
Eeramma: As per Authz Proto the principals section from policy file should have absolute spiffe user id (spiffe://test-abc.foo.bar/xyz/cafyauto) or a regex based spiffe id (*/cafyauto) should be mentioned. Having plain user id "cafyauto" wont match against any of the policy rule mentioned. Hence a correction is needed here. To maintain uniformity across all the users like admin, gnoi-ping etc mentioned complete spiffe id path for cafyauto. this can be easily changed to have regex as mentioned above.

internal/security/authz/authz.go
Eeramma: We are not throwing away error, when we have empty get (no existing policy on device scenario) we can not have this check so we added "nopolicy" flag to verify if the device has existing policy or not. Incase if the device has no policy during test start we can not compare any value as part of Get output (which emits FAILED PRECONDITION ERROR). So the code from 177 had to be excluded for No policy Get scenario. However the code will get exercised incase the device already has Authz policy installed.

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
```diff
@@ -204,9 +204,14 @@ func TestAuthz1(t *testing.T) {
 	certAdminSpiffe := getSpiffe(t, dut, "cert_user_admin")
 	t.Run("Authz-1.1, - Test empty source", func(t *testing.T) {
 		// Pre-Test Section
-		_, policyBefore := authz.Get(t, dut)
-		t.Logf("Authz Policy of the Device %s before the Rotate Trigger is %s", dut.Name(), policyBefore.PrettyPrint(t))
-		defer policyBefore.Rotate(t, dut, uint64(time.Now().Unix()), fmt.Sprintf("v0.%v", (time.Now().UnixNano())), false)
+		statusmsg, policyBefore := authz.Get(t, dut)
+		//t.Logf("Message for first authz get %s", msg)
```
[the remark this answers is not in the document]
Eeramma: Thanks, the extra debug log is removed now.

internal/security/gnxi/rpcexec.go
Eeramma: When Authz rotate fails we make sure that the failure is for expected ones to return NIL code which is required for negative validations in feature/security/gnsi/authz/tests/authz/authz1_4_test.go.  In else section we are returning error code which may be genuine rotate failure.

internal/security/gnxi/rpcexec.go
Eeramma: When Authz rotate fails we make sure that the failure is for expected ones to return NIL code which is required for negative validations in feature/security/gnsi/authz/tests/authz/authz1_4_test.go.  In else section we are returning error code which may be genuine rotate failure.

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
```diff
@@ -204,9 +204,14 @@ func TestAuthz1(t *testing.T) {
...
+		if statusmsg == nil {
+			t.Logf("Expected error FAILED_PRECONDITION seen for authz Get Request.")
+		} else {
+			t.Logf("Authz Policy of the Device %s before the Rotate Trigger is %s", dut.Name(), policyBefore.PrettyPrint(t))
+			defer policyBefore.Rotate(t, dut, uint64(time.Now().Unix()), fmt.Sprintf("v0.%v", (time.Now().UnixNano())), false)
```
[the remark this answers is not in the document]
Eeramma: defer policyBefore.Rotate is not added as part of this git pull request. In this code snippet we are making sure device is Rotate ready for further tests. As we have added "FAILED PRECONDITION" validation handle we have just moved earlier code flow to else part so that if the device already has a policy before Authz test execution whether it is Rotate ready with current set of policy rules.

internal/security/gnxi/rpcexec.go
Eeramma: Replace if else section with switch case in new commit

internal/security/gnxi/rpcexec.go
Eeramma: Replace if else section with switch case in new commit

internal/security/authz/authz.go
Eeramma: we are making sure that only expected error code which is "FAILED PRECONDITION" (empty policy get operation) is received in which case we return NIL as expected if anything other than FAILED PRECONDITION we will have to error out which is captured in else section.

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
Eeramma: defer policyBefore.Rotate is not added as part of this git pull request. In this code snippet we are making sure device is Rotate ready for further tests. As we have added "FAILED PRECONDITION" validation handle we have just moved earlier code flow to else part so that if the device already has a policy before Authz test execution whether it is Rotate ready with current set of policy rules.

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
```diff
@@ -285,9 +290,14 @@ func TestAuthz1(t *testing.T) {
...
 	t.Run("Authz-1.4, Test Normal Policy", func(t *testing.T) {
 		// Pre-Test Section
-		_, policyBefore := authz.Get(t, dut)
-		t.Logf("Authz Policy of the Device %s before the Rotate Trigger is %s", dut.Name(), policyBefore.PrettyPrint(t))
-		defer policyBefore.Rotate(t, dut, uint64(time.Now().Unix()), fmt.Sprintf("v0.%v", (time.Now().UnixNano())), false)
+		statusmsg, policyBefore := authz.Get(t, dut)
+		//t.Logf("Message for first authz get %s", msg)
+		if statusmsg == nil {
+			t.Logf("Expected error FAILED_PRECONDITION seen for authz Get Request.")
+		} else {
```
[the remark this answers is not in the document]
Eeramma: Previous check flow is moved to else so that when policy on device and get operation is performed we are not skipping earlier checks which we had.

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
```diff
@@ -285,9 +290,14 @@ func TestAuthz1(t *testing.T) {
 
 	t.Run("Authz-1.4, Test Normal Policy", func(t *testing.T) {
 		// Pre-Test Section
-		_, policyBefore := authz.Get(t, dut)
-		t.Logf("Authz Policy of the Device %s before the Rotate Trigger is %s", dut.Name(), policyBefore.PrettyPrint(t))
-		defer policyBefore.Rotate(t, dut, uint64(time.Now().Unix()), fmt.Sprintf("v0.%v", (time.Now().UnixNano())), false)
+		statusmsg, policyBefore := authz.Get(t, dut)
+		//t.Logf("Message for first authz get %s", msg)
+		if statusmsg == nil {
+			t.Logf("Expected error FAILED_PRECONDITION seen for authz Get Request.")
```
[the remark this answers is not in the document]
Eeramma: we are displaying a plain message here not formatted with any variable value.

mihirpitale-googler (dismissed):

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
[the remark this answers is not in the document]
morrowc: Sorry the answer doesn't make sense to me.

internal/security/authz/authz.go
morrowc: Sorry I think you are throwing away the error:

  statusError, _ := status.FromError(err)
  
  you are throwing away the error from the 'status.FromError()'.
  Admittedly that PROBABLY Shouldn't error, but it's bad form to not check the return there.

internal/security/authz/authz.go
morrowc: I'd avoid the else here too.

Why not just:
  if <condition you care about> {
    do things
   }
   if <any other error> {
     do other things
   }
   
   Same effect, just  no real need to use else.

internal/security/gnxi/rpcexec.go
morrowc: here too we should skip the else. as above.

internal/security/authz/authz.go
Eeramma: Got it. Added check if FromError is able to identify mapped GRPC status error message if it returns unknown then flagged fatal with Error message.

internal/security/gnxi/rpcexec.go
Eeramma: Fixed this. 

Thanks

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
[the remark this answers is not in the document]
Eeramma: Replaced else with proper if check.

mihirpitale-googler (dismissed):

mihirpitale-googler:
please sync with base branch

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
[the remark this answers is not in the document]
morrowc: Please put a comment in the code so future us doesn't have this same conversation again.

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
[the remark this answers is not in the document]
morrowc: Correct, but your t.Logf ... still is there.

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
morrowc: I see the definition of an empty map at line 52, I don't see where: policy-normal-1 is actually added to that map.

I do see a bunch of other attempts to fetch that value from the map though (or comments about checking what sure looks like is an empty map)

internal/security/authz/authz.go
morrowc: don't use if/else if you don't really really need to, you don't really really need to here.

Eeramma: replaced else with : if ((statusError.Code() != codes.Unknown) && (statusError.Code() != codes.FailedPrecondition))  to capture any errors other than expected one.

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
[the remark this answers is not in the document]
Eeramma: Added Log message.  
   t.Logf("When the device has no pre-installed authz policy file,initial Authz Get operation results in failure.Exit gracefully and continute with validation having Rotate Operation")

feature/security/gnsi/authz/tests/authz/authz1_4_test.go
Eeramma: line 52 : policyMap   map[string]authz.AuthorizationPolicy -- empty object initialization.
line 148: policyMap = authz.LoadPolicyFromJSONFile(t, "testdata/policy.json")  -- this is defined in Setup function which gets triggered in the initial setup.  newpolicy.AddAllowRules  takes care of building proto accepted messages format of base policy and this gets combined with "policy-normal-1_v1" in newpolicy.Rotate(). Same procedure is followed in all tests.

morrowc (approved):

mihirpitale-googler (approved):
lgtm

PR #3120, opened 2024-06-20
Update README.md

sachendras:
This update is to correct the Sub test case numbering in individual test procedure writeup which seems was not updated when the sub test cases were renumbered.


sezhang2 (approved):

sachendras:
gcburn/

PR #3122, opened 2024-06-21
Changes to remove the 'deprecated_vlan_id' deviation from metadata.proto

karthick-gunasekaran-arista:
Changes to remove the 'deprecated_vlan_id' deviation from metadata.proto as it is no longer required now


self-maurya (approved):

LimeHat:
The code should also have been modified to remove obsolete if/else conditions, since this was used only by a single vendor.

cc @dplore @cfernanz

PR #3132, opened 2024-06-25
Update gNMI-1.4 to validate subcomponents paths.

SydneyCaulfeild:
If a component has either `components/component/subcomponents/subcomponent/name` or `components/component/subcomponents/subcomponent/state/name`, then the leafref value of this path should be for a component that exists on that device.

### DIFF
--- a/feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
+++ b/feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
@@ -584,7 +584,7 @@ func TestControllerCardEmpty(t *testing.T) {
 func verifySubcomponentNameExists(c *oc.Component, components []*oc.Component, t *testing.T) {
 	// Note that this will get the subcomponents regardless of whether they are stored in
 	// subcomponent/name or subcomponent/state/name.
-	for _, s := range c.GetOrCreateSubcomponentMap() {
+	for _, s := range c.Subcomponent {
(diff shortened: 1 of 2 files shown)


feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -577,6 +579,26 @@ func TestControllerCardEmpty(t *testing.T) {
 	}
 }
 
+// If the component has a leafref for a subcomponent, that leaf must exist in the list of
+// components.
+func verifySubcomponentNameExists(c *oc.Component, components []*oc.Component, t *testing.T) {
```
wenovus, line 584: Optional:
* Best practice is to also start unexported function comments with the same name. If it ever gets exported (say moved into a library), then it would match the style guide.
* Also recommend to name this `validateSubcomponentsExistAsComponents` to reflect the action more clearly.

wenovus (dismissed):
LGTM. One optional comment.

wenovus (dismissed):

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
SydneyCaulfeild: Sounds good, thanks for the suggestions! I have updated the name and included it in the comment.

wenovus (approved):

PR #3144, opened 2024-06-28
FP-1.1: Power admin DOWN/UP Test 

cakhil45:
Added check for the fabrics which are removable

### DIFF
--- a/feature/platform/tests/power_admin_down_up_test/power_admin_down_up_test.go
+++ b/feature/platform/tests/power_admin_down_up_test/power_admin_down_up_test.go
@@ -25,22 +25,6 @@ func TestFabricPowerAdmin(t *testing.T) {
 	dut := ondatra.DUT(t, "dut")
 	fs := components.FindComponentsByType(t, dut, oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_FABRIC)
 
-	removable_fabrics := make([]string, 0)
-	for _, f := range fs {
-		compMtyVal, compMtyPresent := gnmi.Lookup(t, dut, gnmi.OC().Component(f).Empty().State()).Val()
-		if compMtyPresent && compMtyVal {
-			continue
-		}
-		if gnmi.Get(t, dut, gnmi.OC().Component(f).Removable().State()) {
-			removable_fabrics = append(removable_fabrics, f)
-		}
-	}
-	t.Logf("removable_fabrics are %v", removable_fabrics)
-	fs = removable_fabrics
-	if len(fs) == 0 {
-		t.Skipf("Get Fabric card list for %q: got 0, want > 0", dut.Model())
-	}
-
 	for _, f := range fs {
 		t.Run(f, func(t *testing.T) {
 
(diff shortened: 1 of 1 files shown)


feature/platform/tests/power_admin_down_up_test/power_admin_down_up_test.go
```diff
@@ -24,6 +24,22 @@ func TestFabricPowerAdmin(t *testing.T) {
 	dut := ondatra.DUT(t, "dut")
 	fs := components.FindComponentsByType(t, dut, oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_FABRIC)
 
+	removable_fabrics := make([]string, 0)
+	for _, f := range fs {
+		compMtyVal, compMtyPresent := gnmi.Lookup(t, dut, gnmi.OC().Component(f).Empty().State()).Val()
+		if compMtyPresent && compMtyVal {
+			continue
+		}
+		if gnmi.Get(t, dut, gnmi.OC().Component(f).Removable().State()) {
```
LimeHat, line 33: Looks like this check already exists just a few lines below.

https://github.com/openconfig/featureprofiles/blob/65ec620e7aa48b80d10aa46ea3f0043e83978c75/feature/platform/tests/power_admin_down_up_test/power_admin_down_up_test.go#L30-L32

Is there a reason we need another one?

LimeHat (changes requested):

feature/platform/tests/power_admin_down_up_test/power_admin_down_up_test.go
cakhil45: Hi,
Yes there is need to add this check. It is considering all the sibs even which are not really installed in the device for further test. So added this empty check which will consider only the sibs which were there in the device.

Thanks

cakhil45: Ok this is added as part of pull - https://github.com/openconfig/featureprofiles/commit/d93f6e796eb80fc4d2fff96d545f27208eb5aa1d
which is commited on jul21

But before that these lines were not present I raised pull request on june 28. Should I need to abort it then??

cakhil45: Corrected the code accordingly. Empty check should be done initially before removable state check. Please check.

feature/platform/tests/power_admin_down_up_test/power_admin_down_up_test.go
```diff
@@ -24,18 +24,34 @@ func TestFabricPowerAdmin(t *testing.T) {
 	dut := ondatra.DUT(t, "dut")
 	fs := components.FindComponentsByType(t, dut, oc.PlatformTypes_OPENCONFIG_HARDWARE_COMPONENT_FABRIC)
 
+	removable_fabrics := make([]string, 0)
 	for _, f := range fs {
-		t.Run(f, func(t *testing.T) {
+		compMtyVal, compMtyPresent := gnmi.Lookup(t, dut, gnmi.OC().Component(f).Empty().State()).Val()
```
LimeHat, line 29: You still have two versions of each check in this function (twice check on empty, and twice check on removable).

This doesn't make sense to me. Each check should be present only once

LimeHat (changes requested):
If your objective is 
``` Empty check should be done initially before removable state check. ```
then you can simply rearrange the two checks in the original code. There's no need to add an extra pair of the identical checks.

feature/platform/tests/power_admin_down_up_test/power_admin_down_up_test.go
hattikals: Resolved

AmrNJ (approved):

self-maurya (approved):

PR #3070, opened 2024-06-11
adding changes to rt-2.13

fmolinar:
added the deviations: 
    interface_ref_config_unsupported:true
    rib_wecmp: true
 created the deviation:
    wecmp_auto_unsupported
    
added var of fixedPackets with value of 1000000 to keep a fixed value packets

added cli support for wecmp, Cisco supports ucmp which the weights for each BE needs to be changed manually, added the first CLI for completeness with the expected weights.
 
added function VerifyISISTelemetry to verify ISIS adjecencies formed, as well as, used VerifyDUTBGPEstablished to check for bgp peer established.

configured ISIS loopback interfaces and advertised them via ISIS to be able to use ecmp on ISIS in the test.

Increased UDP packets to increased entropy which helps with the load balancing.

Increased timeout for Bgp_Neighbor_SessionState_ESTABLISHED from 1 min to 2 min.

PR #3147, opened 2024-06-28
fix weighted_ecmp_test for arista

nsadhasivam:


self-maurya:
Vendor specific checks are not preferred in FNT, check if there's any deviation and replace check with deviation.
cc: @dplore

nsadhasivam:
> Vendor specific checks are not preferred in FNT, check if there's any deviation and replace check with deviation. cc: @dplore

These are changes from #3070, have added the deviation

self-maurya (approved):

self-maurya (approved):

PR #3154, opened 2024-07-02
Use gnmi lookup to validate subcomponents.

SydneyCaulfeild:
TestSwitchChip() was still passing for the Juniper devices with invalid subcomponents (ex: `openconfig/components/component/FPC0:PIC0:NPU0/subcomponents/subcomponent/name, FPC0:PIC0:NPU0:PORT16`), so this test needs to be updated. Now that gnmi.LookupAll() is used, the test fails (as it should) with a message saying "Error getting subcomponent for component FPC0:PIC0:NPU0", for example.


feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -581,20 +581,18 @@ func TestControllerCardEmpty(t *testing.T) {
...
+	cName := c.GetName()
+	subcomponents := gnmi.LookupAll[*oc.Component_Subcomponent](t, dut, gnmi.OC().Component(cName).SubcomponentAny().State())
+	for _, s := range subcomponents {
+		subc, ok := s.Val()
+		if !ok {
+			t.Errorf("Error getting subcomponent for component %s", cName)
+		}
+		subcName := subc.GetName()
+		subComponent := gnmi.Lookup[*oc.Component](t, dut, gnmi.OC().Component(subcName).State())
+		if subComponent == nil {
```
wenovus, line 594: !subComponent.IsPresent()

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
```diff
@@ -581,20 +581,18 @@ func TestControllerCardEmpty(t *testing.T) {
...
-				if component.GetName() == sName {
-					continue
-				}
-			}
-			t.Errorf("Subcomponent %s not found in list of components", sName)
-		} else {
-			t.Errorf("Component %s has empty subcomponent name", c.GetName())
+func validateSubcomponentsExistAsComponents(c *oc.Component, components []*oc.Component, t *testing.T, dut *ondatra.DUTDevice) {
+	cName := c.GetName()
+	subcomponents := gnmi.LookupAll[*oc.Component_Subcomponent](t, dut, gnmi.OC().Component(cName).SubcomponentAny().State())
```
wenovus, line 586: optional:

you can get the map object directly for convenience: gnmi.Lookup(gnmi.OC().Component(cName).SubcomponentMap().State())

You can also do this with components if you like (as long as there is not too much data, in which case you can use `ygnmi.NewBatch`).

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
SydneyCaulfeild: thanks!

feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go
SydneyCaulfeild: Thanks for pointing that out! I have updated it to use gnmi.Lookup() with the subcomponentMap().

wenovus (approved):

PR #3159, opened 2024-07-03
Removed `explicit_p4rt_node_component`/`ExplicitP4RTNodeComponent` deviation

snaragund:
- From gNMI1.10 telemetry_basic_check_test
    - From internal/deviations/deviation.go
    - From internal/proto:metadata.proto & respective metadata.pb.go
    - From internal/args/args.go
    - From internal/p4rtutils/p4rtutils.go

PR #3161, opened 2024-07-03
Fix exempt list for README validation

wenovus:
Also remove irrelevant coverage in one of the files that was added.


wenovus:
@dplore this is blocking https://github.com/openconfig/featureprofiles/pull/3159, looking for your review and merge.

dplore (approved):

PR #3246, opened 2024-07-08
RT-5.10: IPv6 Link Local generated by SLAAC

hattikals:
Adding interface_enabled deviation for JUNIPER


Swetha-haridasula (approved):
LGTM

PR #3102, opened 2024-06-13
P4RT-5.2 : added ARP wait on OTG

ArpitJoshi18:
Adding  'WaitforARP" before start protocols to ensure ARP.

PR #3251, opened 2024-07-10
P4RT 5.2 OTG fix

dipchauh:
Call OTG WaitForArp after starting protocols


LimeHat:
incorrect code was introduced in #3102 (cc @dplore )

ArpitJoshi18:
Thank you for fixing it. I realized we need to call start protocols before OTGARP wait. Thanks @dipchauh for correcting.

self-maurya (approved):

PR #3253, opened 2024-07-11
Move mpls static lsp test to mpls feature

dplore:
Moving this README which is intended to be a test using gnmi to configure MPLS static lsp.  It is currently misplaced in the gRIBI folder.


feature/mpls/otg_tests/static_lsp/README.md
```diff
@@ -0,0 +1,42 @@
+# TE-9.2: MPLS based forwarding Static LSP
+
+## Summary
+
+Validate static lsp functionality.
+
+## Procedure
```
sachendras, line 7: Can we split these tests in to 3 sub tests?
1. Flow with MPLS labels mapped to an interface
2. Flow with IPv4 explicit-Null label (label 0)
3. Flow with IPv6 explicit-Null label (label 2)

For 2 and 3 the plan is to use the routing table.

dplore: Hi,  this PR scope is only to use an appropriate folder.  Refactoring the README (or in this case, it looks like you want to simply write more READMEs) is out of scope of this PR.

sachendras (approved):

PR #3257, opened 2024-07-12
test to validate prefix set count after replace

self-maurya:


self-maurya:
raised another PR with compete test, https://github.com/openconfig/featureprofiles/pull/3308

PR #3258, opened 2024-07-12
OC-1.2 : default_tests.go

cprabha:
Hi,

Ixia WaitForArp call does trigger explicit ping when we invoke ping/arp instead checks for static entry on ixia, which was previously resolved during pushConfig.

This causes issue on KNE devices .

Github issue for waitforarp issue : https://github.com/openconfig/featureprofiles/issues/3410

Thanks,
Prabha


feature/networkinstance/otg_tests/defaults_test/defaults_test.go
```diff
@@ -181,6 +183,16 @@ func TestDefaultAddressFamilies(t *testing.T) {
 			otgutils.WaitForARP(t, ate.OTG(), top, "IPv4")
 			otgutils.WaitForARP(t, ate.OTG(), top, "IPv6")
 
+			intfNbrPath := gnmi.OC().Interface(dutP2.Name()).Subinterface(0).Ipv6().Neighbor(atePort2.IPv6)
+			_, ok := gnmi.Watch(t, dut, intfNbrPath.LinkLayerAddress().State(), 5*time.Minute, func(val *ygnmi.Value[string]) bool {
```
LimeHat, line 187: is there a reason why is this required in this test?

up to 5 minutes seems like a long time to wait, and this configuration is a novelty (not used in other tests). if there's a problem with how neighbor resolution is performed, perhaps that should be generalized.

feature/networkinstance/otg_tests/defaults_test/defaults_test.go
```diff
@@ -181,6 +183,16 @@ func TestDefaultAddressFamilies(t *testing.T) {
...
+				_, ok := val.Val()
+				return ok
+			}).Await(t)
+
+			if !ok {
+				t.Log("IPv6 neighbor resolution failed")
```
LimeHat, line 193: Shouldn't this be a t.Fatal then?

feature/networkinstance/otg_tests/defaults_test/defaults_test.go
```diff
@@ -198,7 +210,7 @@ func TestDefaultAddressFamilies(t *testing.T) {
 				}
 				loss := tx - rx
 				lossPct := loss * 100 / tx
-				if got := lossPct; got > 0 {
+				if got := lossPct; got > float32(lossTolerancePct) {
```
LimeHat, line 213: This change is omitted in the PR description, but I don't think any packet loss should be expected in this scenario.. can you elaborate?

LimeHat (changes requested):

feature/networkinstance/otg_tests/defaults_test/defaults_test.go
cprabha: @LimeHat This is needed since ixia "VerifyArp" does not have method to send arp request. It is only sending during "Pushconfig" call.

We do not need this in case of physical devices where they are directly connected. 

If devices are containerized and VerifyArp does not actually send Arp to verify reachability. Its only using existing arp entry and continues.

feature/networkinstance/otg_tests/defaults_test/defaults_test.go
cprabha: This is needed until ixia adds explicit ARP in VerifyARP. This should not affect physical platforms.

feature/networkinstance/otg_tests/defaults_test/defaults_test.go
cprabha: @LimeHat , updating this to Fatal.

LimeHat (commented):
Since you reference a keysight limitation/issue as the basis for the changes, can you please open a corresponding github issue and reference it in the code comment(s)? 
thanks,

cprabha:
> Since you reference a keysight limitation/issue as the basis for the changes, can you please open a corresponding github issue and reference it in the code comment(s)? thanks,

Plz chk: https://github.com/openconfig/featureprofiles/issues/3410

feature/networkinstance/otg_tests/defaults_test/defaults_test.go
```diff
@@ -103,6 +104,7 @@ var (
 		IPv6Len: 64,
 		MAC:     "02:00:02:01:01:01",
 	}
+	kneDeviceModelList = []string{"ncptx"}
```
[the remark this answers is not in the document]
cprabha: @Swetha-haridasula  I have added other vendor kne models . please check

feature/networkinstance/otg_tests/defaults_test/defaults_test.go
```diff
@@ -103,6 +104,7 @@ var (
 		IPv6Len: 64,
 		MAC:     "02:00:02:01:01:01",
 	}
+	kneDeviceModelList = []string{"ncptx", "ceos", "ixr10e", "8201", "xrd"}
 )
```
Swetha-haridasula, line 108: can we add srlinux instead of 'ixr10e' and '8201'

cprabha: updated , please check.

Swetha-haridasula (approved):
lgtm

PR #3153, opened 2024-07-02
remove questionable 100fr dut config

LimeHat:
fixes #2977 

---
<sup>This code is a Contribution to the OpenConfig Feature Profiles project ("Work") made under the Google Software Grant and Corporate Contributor License Agreement ("CLA") and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia’s intellectual property are granted for any other purpose. This code is provided on an "as is" basis without any warranties of any kind.</sup>

PR #3260, opened 2024-07-12
fix aggregate_all_not_forwarding_viable_test

nsadhasivam:


LimeHat:
Please stop using this deviation for this code snippet, this is not correct as I indicated in
https://github.com/openconfig/featureprofiles/pull/2984#issuecomment-2108074887

Regarding this code snippet, I have an open issue #2977. If there's a vendor who needs it, it should be a new deviation. Otherwise, it should be removed altogether as proposed in #3153 

cc @dplore @cfernanz

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -372,7 +372,7 @@ func configureDUT(t *testing.T, dut *ondatra.DUTDevice) []string {
 			if deviations.InterfaceEnabled(dut) {
 				i.Enabled = ygot.Bool(true)
 			}
-			if port.PMD() == ondatra.PMD100GBASEFR {
+			if port.PMD() == ondatra.PMD100GBASEFR && deviations.ExplicitPortSpeed(dut) {
```
LimeHat, line 375: this is not a correct usage of the deviation

LimeHat (changes requested):

nsadhasivam:
> Please stop using this deviation for this code snippet, this is not correct as I indicated in [#2984 (comment)](https://github.com/openconfig/featureprofiles/pull/2984#issuecomment-2108074887)
> 
> Regarding this code snippet, I have an open issue #2977. If there's a vendor who needs it, it should be a new deviation. Otherwise, it should be removed altogether as proposed in #3153
> 
> cc @dplore @cfernanz

Hi Sergey, 

Have added explicit deviations for portspeed and duplex mode, is this what you are looking for?

LimeHat:
I believe `e.AutoNegotiate = ygot.Bool(false)` should also be under a deviation if there's a vendor that requires this config. 

Potentially, you can group all three things under a single deviation, `100g_fr_explicit_config` or something like that.

But again it is not clear if this is required at all. Discussion in #2977 didn't result in any findings in terms of ownership of this config. If there are no vendors who need it (nokia, cisco, and juniper don't, as far as I understand), then we should just remove it as proposed in #3153.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -605,7 +609,6 @@ func configureDUTISIS(t *testing.T, dut *ondatra.DUTDevice, aggIDs []string) {
 	for _, aggID := range aggIDs {
 		isisIntf := isis.GetOrCreateInterface(aggID)
 		isisIntf.GetOrCreateInterfaceRef().Interface = ygot.String(aggID)
-		isisIntf.GetOrCreateInterfaceRef().Subinterface = ygot.Uint32(0)
```
LimeHat, line 608: why is this being removed? full ref (interface+subif) should be used here.

if an implementation doesn't support refs they should use a deviation (e.g. just below there's `if deviations.InterfaceRefConfigUnsupported(dut)` statement that can be used for them)

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -374,8 +374,12 @@ func configureDUT(t *testing.T, dut *ondatra.DUTDevice) []string {
...
 			if port.PMD() == ondatra.PMD100GBASEFR {
 				e.AutoNegotiate = ygot.Bool(false)
-				e.DuplexMode = oc.Ethernet_DuplexMode_FULL
-				e.PortSpeed = oc.IfEthernet_ETHERNET_SPEED_SPEED_100GB
+				if deviations.ExplicitDuplexModeRequired(dut) {
+					e.DuplexMode = oc.Ethernet_DuplexMode_FULL
+				}
+				if deviations.ExplicitPortSpeed(dut) {
+					e.PortSpeed = oc.IfEthernet_ETHERNET_SPEED_SPEED_100GB
+				}
```
LimeHat, line 382: the whole block should be under a new deviation if there's a vendor who needs it

LimeHat (changes requested):

dplore:
> I believe `e.AutoNegotiate = ygot.Bool(false)` should also be under a deviation if there's a vendor that requires this config.
> 
> Potentially, you can group all three things under a single deviation, `100g_fr_explicit_config` or something like that.
> 
> But again it is not clear if this is required at all. Discussion in #2977 didn't result in any findings in terms of ownership of this config. If there are no vendors who need it (nokia, cisco, and juniper don't, as far as I understand), then we should just remove it as proposed in #3153.

@nsadhasivam we think that this configuration should not be required.  

Are you finding that this is mandatory?  If yes, for which devices/vendors?  In concept, this config should not be needed.  If a certain vendor does need it, then it should be a deviation just for that vendor.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
LimeHat: That shouldn't be the case

OC definition says:
```
In contexts
        where a Layer 3 interface is to be referenced, both the
        interface and subinterface leaves must be populated, as
        Layer 3 configuration within the OpenConfig models is
        associated with a subinterface.
```

https://github.com/openconfig/public/blob/0ae6953d5d0ee8156f48088ff7a1bb37ae0ce179/release/models/interfaces/openconfig-interfaces.yang#L228-L239

Bundles can also have subinterfaces, and you can see it is being configured on https://github.com/openconfig/featureprofiles/blob/99f8453fdaa2c47f9bffc41828cd16e5edece029/feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go#L452-L467 in this test.

The issue is that some vendors don't use proper subinterfaces and associate layer 3 information with the physical interface when you configure "subinterface 0", but it's not an OC-compliant way.

cc @dplore

AmrNJ (changes requested):
Can you please resolve comments?

PR #3261, opened 2024-07-12
Add basic encap, decap, scale GUE traffic tests with various actions.

vishnureddybadveli:
Implement network device support for the following gRIBI signaled policy-forwarding rules and actions:

match interface, source IP, destination IP
then perform all below actions
encap MPLS over UDP.
set outer packet src,dst IP address and src,dst UDP port.
set DSCP value. 
set IPSEC spi value.
set egress interface.


dplore:
I think this PR is now obsolete.  Closing,

PR #3265, opened 2024-07-15
System-1: Added Type Config to gnmi get request to avoid get request error

cakhil45:
System-1: Added Type Config to gnmi get request to avoid get request error: code = Unimplemented desc = Unsupported 'type' for /gnmi.gNMI/Get request. Valid options are ['CONFIG']


LimeHat:
I don't think gnmi.Get should be limited to `config` in a general case. 

cc @robshakir @dplore

dplore:
> I don't think gnmi.Get should be limited to `config` in a general case.
> 
> cc @robshakir @dplore

This should be added as a deviation.  gnmi specification does not limit Get to only config leafs.

(note, please do add the deviation, this is useful to have documented)

cakhil45:
Reopened software ticket internally for this. So closing this pull request as support is required irrespective of type

PR #3314, opened 2024-07-20
Update testregistry.textproto

sachendras:
Updates is to include missing test RT-1.19 for 2-byte and 4-byte ASNs

### DIFF
--- a/testregistry.textproto
+++ b/testregistry.textproto
@@ -398,6 +398,7 @@ test: {
   description: "BGP 2-Byte and 4-Byte ASN support"
   readme: "https://github.com/openconfig/featureprofiles/blob/main/feature/experimental/bgp/otg_tests/bgp_2byte_4byte_asn/README.md"
   exec: " "
+}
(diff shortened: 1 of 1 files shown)


testregistry.textproto
```diff
@@ -393,6 +393,11 @@ test: {
   readme: ""
   exec: " "
 }
+test: {
+  id: "RT-1.19"
+  description: "BGP 2-Byte and 4-Byte ASN support"
+  readme: "https://github.com/openconfig/featureprofiles/blob/main/feature/experimental/bgp/otg_tests/bgp_2byte_4byte_asn/README.md"
+  exec: " "
```
dplore, line 400: missing a trailing `}`

dplore (changes requested):

testregistry.textproto
sachendras: corrected

dplore (approved):

PR #3315, opened 2024-07-22
Adding a new test under compliance

rahulsy26:
Adding below test under gribi_compliance_test.go which is not covered here..

Add IPEntry which references to NH which has down port

Reference for another PULL created for compliance.go Library is 
https://github.com/openconfig/gribigo/pull/239


feature/gribi/otg_tests/gribigo_compliance_test/README.md
```diff
@@ -21,3 +21,22 @@ For each compliance test case in the test suite:
...
     *   If the case expects a t.Error result, use testt.ExpectError.
     *   Otherwise, call the test case function directly.
+
+## OpenConfig Path and RPC Coverage
+
+The below yaml defines the OC paths and RPC intended to be covered by this test.
+
+```yaml
+paths:
+  /interfaces/interface/config/name:
```
robshakir, line 31: Is this path intended to be covered, or just covered because it's required for test setup? Generally, I think the pattern here has been to list the things that we are trying to fully test here, so I think this can be omitted.

@dplore -- is that right?

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
```diff
@@ -225,3 +290,67 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) gosnappi.Config {
 
 	return top
 }
+
+// AddNHGReferencingToUknownNH tests a case where a nexthop group references a nexthop IP that does not Resolve.
```
robshakir, line 294:
```suggestion
// AddNHGReferencingToUnresolvedNH tests a case where a nexthop group references a nexthop IP that does not Resolve.
```

robshakir, line 294: Suggest to add a comment to this line that says `1.0.0.1` must not resolve within the topology (and does not).

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
```diff
@@ -225,3 +290,67 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) gosnappi.Config {
...
+// AddNHGReferencingToDownPort tests a case where a nexthop group references to nexthop port that is down.
+func AddNHGReferencingToDownPort(tcArgs *testArgs, wantACK fluent.ProgrammingResult, t testing.TB) {
+	t.Log("Setting port2 to down...")
+	p := tcArgs.dut.Port(t, "port2")
+	t.Log("Flush existing gRIBI routes before test.")
+	if err := gribi.FlushAll(tcArgs.client); err != nil {
+		t.Fatal(err)
+	}
+	setDUTInterfaceWithState(t, tcArgs.dut, p, false)
+	tcArgs.client.Modify().AddEntry(t,
```
robshakir, line 329: Suggest to add a blank line before this line to separate out where you are doing the port state vs. the gRIBI changes.

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
```diff
@@ -225,3 +290,67 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) gosnappi.Config {
...
+	chk.HasResult(t, tcArgs.client.Results(t),
+		fluent.OperationResult().
+			WithNextHopGroupOperation(2000).
+			WithOperationType(constants.Add).
+			WithProgrammingResult(wantACK).
+			AsResult(),
+		chk.IgnoreOperationID())
+}
+
+// AddNHGReferencingToDownPort tests a case where a nexthop group references to nexthop port that is down.
```
robshakir, line 320: Maybe add some extra commentary here `The entry is expected to be installed despite the port specified being down.`

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
```diff
@@ -225,3 +290,67 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) gosnappi.Config {
...
+	t.Log("Setting port2 to down...")
+	p := tcArgs.dut.Port(t, "port2")
+	t.Log("Flush existing gRIBI routes before test.")
+	if err := gribi.FlushAll(tcArgs.client); err != nil {
+		t.Fatal(err)
+	}
+	setDUTInterfaceWithState(t, tcArgs.dut, p, false)
+	tcArgs.client.Modify().AddEntry(t,
+		fluent.NextHopEntry().WithNetworkInstance(deviations.DefaultNetworkInstance(tcArgs.dut)).
+			WithIndex(2000).WithInterfaceRef(p.Name()).WithIPAddress(atePort2.IPv4),
```
robshakir, line 331: Should this also set the subinterface reference? The port would generally be an L2 entity, but I think you probably care about an L3 one.

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
```diff
@@ -225,3 +290,67 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) gosnappi.Config {
...
+	tcArgs.client.Modify().AddEntry(t,
+		fluent.NextHopEntry().WithNetworkInstance(deviations.DefaultNetworkInstance(tcArgs.dut)).
+			WithIndex(2000).WithIPAddress("1.0.0.1"),
+		fluent.NextHopGroupEntry().WithNetworkInstance(deviations.DefaultNetworkInstance(tcArgs.dut)).
+			WithID(2000).AddNextHop(2000, 1),
+		fluent.IPv4Entry().WithNetworkInstance(deviations.DefaultNetworkInstance(tcArgs.dut)).
+			WithPrefix("203.0.113.101/32").WithNextHopGroup(2000))
+
+	if err := awaitTimeout(tcArgs.ctx, t, tcArgs.client, 2*time.Minute); err != nil {
+		t.Logf("Could not program entries via client, got err, check error codes: %v", err)
```
robshakir, line 309: Should this be a `t.Errorf` rather than a `t.Log`?

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
```diff
@@ -225,3 +290,67 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) gosnappi.Config {
...
+	tcArgs.client.Modify().AddEntry(t,
+		fluent.NextHopEntry().WithNetworkInstance(deviations.DefaultNetworkInstance(tcArgs.dut)).
+			WithIndex(2000).WithInterfaceRef(p.Name()).WithIPAddress(atePort2.IPv4),
+		fluent.NextHopGroupEntry().WithNetworkInstance(deviations.DefaultNetworkInstance(tcArgs.dut)).
+			WithID(2000).AddNextHop(2000, 1),
+		fluent.IPv4Entry().WithNetworkInstance(deviations.DefaultNetworkInstance(tcArgs.dut)).
+			WithPrefix("203.0.113.100/32").WithNextHopGroup(2000),
+	)
+	if err := awaitTimeout(tcArgs.ctx, t, tcArgs.client, 2*time.Minute); err != nil {
+		t.Logf("Could not program entries via client, got err, check error codes: %v", err)
```
robshakir, line 338: Similarly here, should this be an `Errorf`?

feature/gribi/otg_tests/gribigo_compliance_test/README.md
dplore: Yes, that is correct.  Please only list the paths that are the focus of the test and skip listing paths.   This is of course a judgement call for the author+reviewer, but I agree this test is not intended to interface naming.  That path is the focus of tests in the [feature/interface/ folder](https://github.com/openconfig/featureprofiles/tree/main/feature/interface/), such as [feature/interface/singleton/otg_tests/singleton_test](https://github.com/openconfig/featureprofiles/tree/main/feature/interface/singleton/otg_tests/singleton_test)

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
rahulsy26: Yes we care about L3 only

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
rahulsy26: Ack

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
rahulsy26: Ack

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
rahulsy26: Ack

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
rahulsy26: this is fine we are printing the log only with and checking the below where actually it is failing/passing.

chk.HasResult(t, tcArgs.client.Results(t),

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
rahulsy26: this is fine we are printing the log only with and checking the below where actually it is failing/passing.

chk.HasResult(t, tcArgs.client.Results(t),

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
rahulsy26: Ack

feature/gribi/otg_tests/gribigo_compliance_test/README.md
rahulsy26: done

feature/gribi/otg_tests/gribigo_compliance_test/gribigo_compliance_test.go
robshakir: If we care about L3, then I think this should be `WithInterfaceRef(p.Name()).WithSubinterfaceRef(0)`.

robshakir (approved):
LGTM, one correction.

PR #3316, opened 2024-07-22
Update RT-2.13: Weighted-ECMP for IS-IS to not also use BGP

ANISH-GOTTAPU:
Currently, OTG does not support IBGP with the same NextHOP across multiple devices. As a result, the DUT will only consider one neighborship. To address this, the test has been modified to use only ISIS, advertising the same route from three ISIS neighbors. The expected behavior is that traffic will load balance based on weights.

I've updated the test to use only the interface on the transmit end, ensuring compatibility with both HW and KNE (SW), as we lack a proto file that supports KNE with LAG on transmit. This change will not affect the weighted-ECMP since the receiving end is LAG only.


feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go
```diff
@@ -416,64 +385,53 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) gosnappi.Config {
...
 
 func configureDUT(t *testing.T, dut *ondatra.DUTDevice) []string {
 	t.Helper()
 	fptest.ConfigureDefaultNetworkInstance(t, dut)
 
-	configureDUTLoopback(t, dut)
+	d := &oc.Root{}
+	p1 := dut.Port(t, "port1")
+
+	i := d.GetOrCreateInterface(p1.Name())
```
ram-mac, line 413: Here interface leaf is mandatory to be added. Use NewOCInterface for creating the new interface which would configure the right way. Similarly for other aggregate interfaces too.

// NewOCInterface returns a new *oc.Interface configured with these attributes.
func (a *Attributes) NewOCInterface(name string, dut *ondatra.DUTDevice) *oc.Interface {
	return a.ConfigOCInterface(&oc.Interface{Name: ygot.String(name)}, dut)
}

feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go
```diff
@@ -503,8 +461,8 @@ func configureDUT(t *testing.T, dut *ondatra.DUTDevice) []string {
 		gnmi.BatchDelete(b, gnmi.OC().Interface(aggID).Aggregation().MinLinks().Config())
 		gnmi.BatchReplace(b, gnmi.OC().Interface(aggID).Config(), agg)
 
-		p1 := dut.Port(t, fmt.Sprintf("port%d", (aggIdx*2)+1))
-		p2 := dut.Port(t, fmt.Sprintf("port%d", (aggIdx*2)+2))
+		p1 := dut.Port(t, fmt.Sprintf("port%d", (aggIdx*2)+2))
```
ram-mac, line 464: p1 is already used in this function above. Please use a different variable name here instead of reusing the same variable under the same function.
Instead of p1, p2 you could use something meaningful here above too

ram-mac (commented):
Did this test pass at least in the KNE environment?

ram-mac (commented):
/fptest virtual

MarcCharlebois:
@ram-mac, I think there's something fundamentally wrong with the way the test is developped.
As per the README for Unequal Traffic distribution:
>Stop traffic from RT-9.1 and introduce a failure by disabling one of the member interfaces in ATE2:LAG1.

Based on this, do you expect the interface on the DUT to be OperDown? If so, I think there's some reworkd that needs to be done in this area of the code -as this portion is shared for both Equal Distribution and Unequal Distribution.
Additionally, it seems like sending the request to ATE/OTG port to be down may not fully shut the port down?

My Suggestions would be to remove the below snippet, ensure that in the subTest for Unequal_Distribution_Of_Traffic we set lag member to ADMIN_STATE_DOWN directly on the DUT.
https://github.com/openconfig/featureprofiles/blob/a9a382e38d44b55a971b4d4c9d15709566078c3f/feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go#L191-L209


As written right now, I can only see this test passing for one vendor as it has static configurations to distribute the traffic for Unequal_Distribution_Of_Traffic:
https://github.com/openconfig/featureprofiles/blob/a9a382e38d44b55a971b4d4c9d15709566078c3f/feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go#L215

feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go
```diff
@@ -249,17 +220,13 @@ func TestWeightedECMPForISIS(t *testing.T) {
 	}
 
 	top.Flows().Clear()
-	if deviations.ISISLoopbackRequired(dut) {
```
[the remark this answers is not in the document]
ANISH-GOTTAPU: We have updated the test not to use BGP anymore, so the deviation "ISISLoopbackRequired" is not needed.

feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go
ANISH-GOTTAPU: Updated the code to use `NewOCInterface`

feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go
ANISH-GOTTAPU: Updated the code accordingly.

ANISH-GOTTAPU:
> Did this test pass at least in the KNE environment?

The test is passing with KNE environment but had to reduce the traffic rate and adjust tolerance accordingly. Attaching the log for your reference.

[RT_2.13_kne_ceos.log](https://github.com/user-attachments/files/18555442/RT_2.13_kne_ceos.log)

MarcCharlebois:
> > Did this test pass at least in the KNE environment?
> 
> The test is passing with KNE environment but had to reduce the traffic rate and adjust tolerance accordingly. Attaching the log for your reference.
> 
> [RT_2.13_kne_ceos.log](https://github.com/user-attachments/files/18555442/RT_2.13_kne_ceos.log)

It seems to me that the subtest is failing due to the interface not being shut down due to bizzare handling in that area.
This code tries to shut the interface from ATE/OTG side which I believe isn't fully shutting down the port - perhaps you can shed more light? @ANISH-GOTTAPU 

I think the baseline behavior of this test should be to shut the DUT port explicitly, that way we get a deterministic behavior whichever type of platform this test is ran on. There's also an implicit assumption that the DUT is directly connected to the ATE/OTG with the current behavior. Thoughts? @ram-mac  @rohit-rp 

https://github.com/openconfig/featureprofiles/blob/de67f2de6469f51e47f4a2ea5d922a2d1277e455/feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go#L221-L238

I'm just curious to know why this value has a comment that suggests the value should be different based on the environment.
https://github.com/openconfig/featureprofiles/blob/de67f2de6469f51e47f4a2ea5d922a2d1277e455/feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go#L157

ram-mac:
I validated this PR on 2 vendors and somehow the test is picking the TESTBED_DUT_ATE_4LINKS instead of TESTBED_DUT_ATE_9LINKS_LAG and the test fails due to the port issues.

argv[20]: '--testbed=../../../../../../../ops/netops/lab/wbb/testbeds/testbed_dut_ate_4links_lag.textproto'

Here is the failure
-----
Port(t, port5) on dut(xxx.xxxx): port ID "port5" not found in reserved device "xxx.xxxx"

ram-mac:
> > Did this test pass at least in the KNE environment?
> 
> The test is passing with KNE environment but had to reduce the traffic rate and adjust tolerance accordingly. Attaching the log for your reference.
> 
> [RT_2.13_kne_ceos.log](https://github.com/user-attachments/files/18555442/RT_2.13_kne_ceos.log)

I see in the logs that the UECMP traffic is failing. Can you give details on why it is failing?

***********
    --- FAIL: TestWeightedECMPForISIS/Unequal_Distribution_Of_Traffic (60.08s)
***********

ram-mac (changes requested):
In the original metadata.textproto, we had TESTBED_DUT_ATE_8LINKS and now this is changed to TESTBED_DUT_ATE_9LINKS_LAG
Do we need to change the testbed? Can't we keep the same testbed to work with?

Even if we change in the metada.textproto, i see that the README.md still reflects the atedut_8 links testbed.
[TESTBED_DUT_ATE_8LINKS](https://github.com/openconfig/featureprofiles/blob/main/topologies/atedut_8.testbed)

ANISH-GOTTAPU:
> > > Did this test pass at least in the KNE environment?
> > 
> > 
> > The test is passing with KNE environment but had to reduce the traffic rate and adjust tolerance accordingly. Attaching the log for your reference.
> > [RT_2.13_kne_ceos.log](https://github.com/user-attachments/files/18555442/RT_2.13_kne_ceos.log)
> 
> I see in the logs that the UECMP traffic is failing. Can you give details on why it is failing?
> 
> ```
> --- FAIL: TestWeightedECMPForISIS/Unequal_Distribution_Of_Traffic (60.08s)
> ```

The DUT image I was using is not supporting UECMP, but same was checked by @MarcCharlebois and the test worked for Marc, @MarcCharlebois Can you please share the logs with your DUT.

ANISH-GOTTAPU:
> In the original metadata.textproto, we had TESTBED_DUT_ATE_8LINKS and now this is changed to TESTBED_DUT_ATE_9LINKS_LAG Do we need to change the testbed? Can't we keep the same testbed to work with?
> 
> Even if we change in the metada.textproto, i see that the README.md still reflects the atedut_8 links testbed. [TESTBED_DUT_ATE_8LINKS](https://github.com/openconfig/featureprofiles/blob/main/topologies/atedut_8.testbed)

For the test to work with KNE, we have to use this textproto https://github.com/openconfig/featureprofiles/blob/main/topologies/kne/arista/ceos/dutate_lag.textproto.

ram-mac:
Validation with vendors fails due to the testbed link issue. The issue is that this PR has changed the textproto to TESTBED_DUT_ATE_9LINKS_LAG and when it is validated it picks up the 4 links testbed. Even after changing the textproto to the original one which is TESTBED_DUT_ATE_8LINKS the test actually picks up 4 link testbed only ultimately failing the test on any vendor

ram-mac:
@ANISH-GOTTAPU 
Based on our discussion here are some issues with this PR
1.TESTBED_DUT_ATE_9LINKS_LAG  will not work for this test as it will pick 4port testbed and this change needs to be reverted back.
2.The code changes you have done is failing to configure the DUT. See the error below which needs to fixed. Have shared with you the full fail logs in email.

        }: failed to apply: interface Ethernet5/3/8 speed not exist doesn't match with breakout-mode config for port Ethernet5/3-Port: {"openconfig-platform-port:group":[{"config":{"breakout-speed":"openconfig-if-ethernet:SPEED_100GB","index":1,"num-breakouts":4,"num-physical-channels":2},"index":1,"state":{"breakout-speed":"openconfig-if-ethernet:SPEED_100GB","index":1,"num-breakouts":4,"num-physical-channels":2}}]}

feature/isis/otg_tests/weighted_ecmp_test/metadata.textproto
```diff
@@ -4,7 +4,7 @@
 uuid: "2beaac46-9b7b-49c4-9bde-62ad530aa5c4"
 plan_id: "RT-2.13"
 description: "Weighted-ECMP for IS-IS"
-testbed: TESTBED_DUT_ATE_8LINKS
+testbed: TESTBED_DUT_ATE_9LINKS_LAG
```
ram-mac, line 7: This change needs to be reverted back because this textproto picks 4port testbed instead of 9ports and this is something which we are trying to fix it internally at google.

But for this test please revert this change.

feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go
```diff
@@ -416,98 +402,53 @@ func configureATE(t *testing.T, ate *ondatra.ATEDevice) gosnappi.Config {
...
 }
 
 func configureDUT(t *testing.T, dut *ondatra.DUTDevice) []string {
 	t.Helper()
 	fptest.ConfigureDefaultNetworkInstance(t, dut)
 
-	configureDUTLoopback(t, dut)
+	dc := gnmi.OC()
+	i1 := dutSrc.NewOCInterface(dut.Port(t, "port1").Name(), dut)
+	gnmi.Replace(t, dut, dc.Interface(i1.GetName()).Config(), i1)
```
ram-mac, line 429: @ANISH-GOTTAPU - This and above changes is failing to configure the DUT. Please refer to the log that i shared you via email.

ram-mac (changes requested):
Please address the changes requested

feature/isis/otg_tests/weighted_ecmp_test/metadata.textproto
MarcCharlebois: Should there be a larger discussion to be had here in case there's a different testbed required for HW TB vs KNE TB?

feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go
ram-mac: @ANISH-GOTTAPU with the latest changes and with the textproto set to TESTBED_DUT_ATE_8LINKS at least now the test is passing for one vendor.
For now we will have to make change to the textproto as mentioned above.

feature/isis/otg_tests/weighted_ecmp_test/metadata.textproto
ram-mac: @MarcCharlebois - there is some issue with the TESTBED_DUT_ATE_9LINKS_LAG in our environment for now. And we have a tracking bug for the same.
@ANISH-GOTTAPU @MarcCharlebois - Please revert this change in the textproto alone; Otherwise the test fails in both HW;
Once we have the textproto issue is fixed locally we can raise a PR.

feature/isis/otg_tests/weighted_ecmp_test/weighted_ecmp_test.go
ANISH-GOTTAPU: @ram-mac : As discussed I changed testbed to TESTBED_DUT_ATE_8LINKS  for now.

ram-mac: Thanks @ANISH-GOTTAPU 
I will resolve this conversation.

ram-mac:
@ANISH-GOTTAPU - Can you rebase this branch and recommit; there are a lot of CI/CD failures i see. I think once u rebase those should be resolved.

ANISH-GOTTAPU:
> @ANISH-GOTTAPU - Can you rebase this branch and recommit; there are a lot of CI/CD failures i see. I think once u rebase those should be resolved.

@ram-mac : Rebased the branch with main, can you please check.

ram-mac:
> > @ANISH-GOTTAPU - Can you rebase this branch and recommit; there are a lot of CI/CD failures i see. I think once u rebase those should be resolved.
> 
> @ram-mac : Rebased the branch with main, can you please check.

Thanks @ANISH-GOTTAPU - will approve now.

ram-mac (approved):
the test is validated and it is passing now
LGTM

PR #3319, opened 2024-07-22
Update ACL-1.2 to cover a management port

dplore:


alokmtri-g (commented):
In my opinion having  DELETE ACL case will add value here , to be sure that deleting ACL is clearing filter entry and not leaving any stale entry.

alokmtri-g (approved):
LGTM

self-maurya (approved):

PR #3325, opened 2024-07-25
healthz parameters for CISCO

karthikeya-remilla:
1. added bgp process "bgp" and chassis component name "Rack 0"
2. Origin as "openconfig" for healthz RPCs

### DIFF
--- a/feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go
+++ b/feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go
@@ -31,10 +31,11 @@ var (
 		ondatra.NOKIA:   "sr_bgp_mgr",
 		ondatra.ARISTA:  "IpRib",
 		ondatra.JUNIPER: "rpd",
+		ondatra.CISCO:   "bgp",
 	}
 	components = map[ondatra.Vendor]string{
 		ondatra.ARISTA:  "Chassis",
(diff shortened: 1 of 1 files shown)


feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go
```diff
@@ -31,10 +31,11 @@ var (
 		ondatra.NOKIA:   "sr_bgp_mgr",
 		ondatra.ARISTA:  "IpRib",
 		ondatra.JUNIPER: "rpd",
+		ondatra.CISCO:   "bgp",
```
dplore, line 34: This LGTM

feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go
```diff
@@ -31,10 +31,11 @@ var (
...
 	}
 	components = map[ondatra.Vendor]string{
 		ondatra.ARISTA:  "Chassis",
-		ondatra.CISCO:   "Chassis",
+		ondatra.CISCO:   "Rack 0",
```
dplore, line 38: This is good enough.   It would be better if the component tree was fetched and the code searches for a component with type "CHASSIS" and reads the name.  This would eliminate the need to hard code various strings here for each vendor.

dplore (approved):

PR #3327, opened 2024-07-25
RT-1.53 fix

MarcCharlebois:
- delete pfx2 from var as it was never "replaced" - see below
Before the change, we can see that the var was never replaced and always carried pfx2 which seems to not be the intended behavior by the author.
```
    prefix_set_test.go:65: Printing the value after re-assignment, &{out-of-range E_PrefixSet_Mode enum value: 0 0xc001e93e80 map[{10.240.31.48/28 exact}:0xc001e93eb0 {173.36.128.0/20 exact}:0xc001e93ee0]}
    prefix_set_test.go:69: Printing the value after L68, &{out-of-range E_PrefixSet_Mode enum value: 0 0xc001e93e80 map[{10.240.31.48/28 exact}:0xc001e93eb0 {173.36.128.0/20 exact}:0xc001e93ee0 {173.36.144.0/20 exact}:0xc002171a90]}
```
- add deviation 
---
<sub>This code is a Contribution to the OpenConfig Feature Profiles project ("Work") made under the Google Software Grant and Corporate Contributor License Agreement ("CLA") and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia’s intellectual property are granted for any other purpose. This code is provided on an "as is" basis without any warranties of any kind.</sub>

### DIFF
--- a/feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
+++ b/feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
@@ -17,6 +17,7 @@ package prefix_set_test
 import (
 	"testing"
 
+	"github.com/openconfig/featureprofiles/internal/deviations"
 	"github.com/openconfig/featureprofiles/internal/fptest"
 	"github.com/openconfig/ondatra"
 	"github.com/openconfig/ondatra/gnmi"
@@ -45,7 +46,9 @@ func TestPrefixSet(t *testing.T) {
 
 	// create a prefix-set with 2 prefixes
 	v4PrefixSet := ds.GetOrCreatePrefixSet(prefixSetA)
-	v4PrefixSet.SetMode(oc.PrefixSet_Mode_IPV4)
+	if !deviations.SkipPrefixSetMode(dut) {
+		v4PrefixSet.SetMode(oc.PrefixSet_Mode_IPV4)
+	}
 	v4PrefixSet.GetOrCreatePrefix(pfx1, mskLen)
 	v4PrefixSet.GetOrCreatePrefix(pfx2, mskLen)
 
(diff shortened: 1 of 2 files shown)


MarcCharlebois:
Also, RT-1.33 has the same name as this test, is that to be expected that test-name may overlap in various directories or should each test be unique?

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
```diff
@@ -62,9 +65,12 @@ func TestPrefixSet(t *testing.T) {
 
 	// replace the prefix-set by replacing an existing prefix with new prefix
 	v4PrefixSet = ds.GetOrCreatePrefixSet(prefixSetA)
-	v4PrefixSet.SetMode(oc.PrefixSet_Mode_IPV4)
+	if !deviations.SkipPrefixSetMode(dut) {
+		v4PrefixSet.SetMode(oc.PrefixSet_Mode_IPV4)
+	}
 	v4PrefixSet.GetOrCreatePrefix(pfx1, mskLen)
 	v4PrefixSet.GetOrCreatePrefix(pfx3, mskLen)
+	v4PrefixSet.DeletePrefix(pfx2, mskLen)
```
self-maurya, line 73: we are doing `gnmi.Replace`, and our expectation is it will replace old prefixes with new ones so we don't want to explicitly remove the `pfx2` here

LimeHat: `DeletePrefix` is not translated into the gnmi.Delete call; this is about the data structure (`ds`) you're working with which is not re-initialized between the assignments. `ds.GetOrCreatePrefixSet(prefixSetA)` returns the structure that was configured in the previous step.

self-maurya (approved):

PR #3336, opened 2024-07-27
Add canonical OC to test README template

dplore:
Thoughts on this?

The goal is to have an easy to read, reference/golden/canonical configuration that covers the test topic.  It should be a snippet including the paths that are to be tested.   

The format could be JSON or YAML, but I think YAML is considerably easier to read and somewhat easier to write.  

The YAML could be validated by converting to JSON and then parsed by ygot or gnmic.


doc/test-requirements-template.md
```diff
@@ -27,22 +27,50 @@ Write a few sentences or paragraphs describing the purpose and scope of the test
...
 
-* TestID-x.y.z - Name of subtest
+### TestID-x.y.z - Name of subtest
+  * Canonical OpenConfig Configuration
+
+An example OpenConfig configuration and/or RPC content should be specified here in YAML format.
+
+```yaml
+---
+openconfig-qos:
```
robshakir, line 40: I'm supportive of adding this to the README. It'd be great to ensure the paths that are mentioned later are covered by this example (or a set of examples).

I do have concerns about using YAML here before we have a formal specification written for it. We have done this before, with not great results -- when a pre-7951 standard was used in a manner that was not externally defined. If we are adopting YAML here, I'd like to see some efforts going toward defining a formal specification à la RFC7951 for this encoding, and having this published somewhere.

robshakir: Note that without a specification and some tooling support (which I can help to write if we have a specification) then you cannot validate this YAML.

dplore: For validation, I wonder if we can take the approach of parsing, converting to JSON and running any desired  validations as JSON?  This may reduce the burden of writing yaml specific validations.  

YAML is a superset of JSON and I am sure I don't want to use any YAML feature which does not map to JSON.  

There is some precedent for this YAML->JSON approach  in go here: https://github.com/ghodss/yaml?tab=readme-ov-file.  The referenced blog post goes through their rationale of this approach.

rcdomigan:
Whether it's JSON or YAML, as someone trying to produce openconfig for a device I would like to have a concrete stanza to model my configuration on.

doc/test-requirements-template.md
dplore: IMO, we shouldn't block the request for authors to write canonical OC in the README on validation.  But I do agree we ought add such validation.   It might take some time to get this validation up and running ad while that's happening we can be benefitting from README's that have (possibly model incorrect) canonical OC.   This is an incremental improvement on top of the flat list of OC paths  we currently have, which are verified.

The flat list of paths in ocpaths subsection of the readme markdown is validated here: [tools/validate_readme_spec](https://github.com/openconfig/featureprofiles/tree/main/tools/validate_readme_spec). I realize this is not at all the same thing as validating the OC content of a canonical config.   It's merely some an indication of incremental progress to making defining requirements better.  :)

xw-g (changes requested):
Having internal discussion. (So far it doesn't seems like we want YAML)

doc/test-requirements-template.md
dplore: Note: In the future, the canonical OC content will be validated for
OC schema correctness and also replace the flat list of paths in the
"OpenConfig Path and RPC Coverage" heading.  

But until that validation is in place, we'll keep the flat list of OC paths because that is validated right now.

dplore:
@xw-g @robshakir ready for re-review.

@self-maurya @rohit-rp, please feel free to provide any suggestions.

self-maurya (dismissed):

doc/test-requirements-template.md
```diff
@@ -11,9 +11,9 @@ assignees: ''
 Below is the required template for writing test requirements.  Good examples of test
 requirements include:
 
+* [TE-18.1 gRIBI MPLS in UDP Encapsulation and Decapsulation](https://github.com/openconfig/featureprofiles/blob/main/feature/gribi/otg_tests/mpls_in_udp/README.md)
 * [TE-3.7: Base Hierarchical NHG Update](/feature/gribi/otg_tests/base_hierarchical_nhg_update/README.md)
 * [gNMI-1.13: Telemetry: Optics Power and Bias Current](https://github.com/openconfig/featureprofiles/blob/main/feature/platform/tests/optics_power_and_bias_current_test/README.md)
-* [RT-5.1: Singleton Interface](https://github.com/openconfig/featureprofiles/blob/main/feature/interface/singleton/otg_tests/singleton_test/README.md)
```
robshakir, line 16: Any reason this isn't a good example any more? (Other than no canonical OC?)

doc/test-requirements-template.md
```diff
@@ -27,22 +27,68 @@ Write a few sentences or paragraphs describing the purpose and scope of the test
 
 ## Procedure
 
-* Test environment setup
-  * Description of procedure to configure ATE and DUT with pre-requisites making it possible to cover the intended paths and RPC's.
+### Test environment setup
+
+* Description of procedure to configure ATE and DUT with pre-requisites making it possible to cover the intended paths and RPC's.
```
robshakir, line 32:
```suggestion
* Description of procedure to configure ATE and DUT with pre-requisites making it possible to cover the intended paths and RPCs.
```

doc/test-requirements-template.md
dplore: Yes, no canonical OC in that one and 2 examples seems like enough.

doc/test-requirements-template.md
```diff
@@ -27,22 +27,68 @@ Write a few sentences or paragraphs describing the purpose and scope of the test
...
+
+#### Canonical OpenConfig Configuration for the DUT
+
+NOTE: An example OpenConfig configuration and/or RPC content for any common
+DUT configuration which is used across the subtests should be specified
```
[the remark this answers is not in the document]
dplore: Tests typically fall into two patterns.    

1. The most simple tests have one config that should be generated in the setup and then `gnmi.Set` within a t.Run call.  Then validations are performed.
2. The average test has multiple configurations, but with very little change in each config.  A base configuration that should be generated in the test setup.  Then a series of subtests which set different values for the leaves in the base config to test permutations of config leaves.

That should handle ~90% of tests using gNMI for configuration.  

A special case is tests including policy.  Less than 10% of our tests are policy oriented.  I don't have an precise  solution that will work for every possible policy test.    These tests change the ordering and/or add and remove leaves to cover different permutations of policy.  The test SHOULD still have a base, canonical configuration.  
Subtests which add/remove leaves should specify the canonical OC for the added/removed leaves.  Examples of these tests are at: https://github.com/openconfig/featureprofiles/tree/main/feature/bgp/policybase/otg_tests

Note: in practice, a lot of test implementations in category 1 and 2 above incrementally generate and Set DUT configuration throughout the test using gnmi.Set with the update option.  This README template intends to specify that a base configuration should be generated and then pushed in one gnmi.Set for the test setup.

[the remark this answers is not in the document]
dplore: @robshakir ready for your review

doc/test-requirements-template.md
robshakir: SGTM.

doc/test-requirements-template.md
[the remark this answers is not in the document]
robshakir: Thanks, SGTM!

robshakir (approved):
Thanks for the changes, LGTM.

PR #3362, opened 2024-08-05
Add a test case for prefix set with CLI config.

self-maurya:


feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
```diff
@@ -95,3 +96,56 @@ func TestPrefixSet(t *testing.T) {
...
+   	  seq 30 permit 173.36.128.0/20
+   	  seq 40 permit 173.37.128.0/20
+	  seq 50 permit 173.38.128.0/20
+	  seq 60 permit 173.39.128.0/20
+	  seq 70 permit 173.40.128.0/20
+	  seq 80 permit 173.41.128.0/20
+	  seq 90 permit 173.42.128.0/20
+	  seq 100 permit 173.43.128.0/20
+	`
+	helpers.GnmiCLIConfig(t, dut, prefixSetConfig)
```
LimeHat, line 116: this is a vendor-specific cli/gnmi request, but not marked as such.

LimeHat (changes requested):
if this is a new test case that is applicable to all vendors, readme should also be updated to indicate that

cc @dplore

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
self-maurya: added switch case for vendor specific config

rohit-rp (approved):

PR #3364, opened 2024-08-06
Fix gNMI-1.22 requirement for management port

dplore:
The README currently specifies using a 'cpu' type interface, but the correct type is 'management.


LimeHat (approved):
lgtm, thanks!

rohit-rp (approved):

PR #3367, opened 2024-08-07
RT-7.5: changes for Cisco within existing deviation

karthikeya-remilla:
Changes: 
1. "bgp_extended_community_index_unsupported = true" in all vendors' metadata info, but currently it is being verified. Changed the code accordingly.
2. Checking if community is empty for Cisco to comply with the behavior in README.


nsadhasivam (approved):
LGTM!

PR #3368, opened 2024-08-07
using OC to create prefix-list and gnoi to terminate Octa

self-maurya:


feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
```diff
@@ -100,45 +108,47 @@ func TestPrefixSet(t *testing.T) {
 
 func TestPrefixSetWithCLIConfig(t *testing.T) {
```
LimeHat, line 109: if cli is no longer used, then 
a) this test should be renamed
b) readme needs an update once again.

LimeHat (changes requested):
additionally, `octa` is a vendor specific name and it is unclear what should be restarted for other vendors.

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
```diff
@@ -100,45 +108,47 @@ func TestPrefixSet(t *testing.T) {
...
+	v4PrefixSet.GetOrCreatePrefix("10.244.187.32/28", mskLen)
+	v4PrefixSet.GetOrCreatePrefix("173.36.128.0/20", mskLen)
+	v4PrefixSet.GetOrCreatePrefix("173.37.128.0/20", mskLen)
+	v4PrefixSet.GetOrCreatePrefix("173.38.128.0/20", mskLen)
+	v4PrefixSet.GetOrCreatePrefix("173.39.128.0/20", mskLen)
+	v4PrefixSet.GetOrCreatePrefix("173.40.128.0/20", mskLen)
+	v4PrefixSet.GetOrCreatePrefix("173.41.128.0/20", mskLen)
+	v4PrefixSet.GetOrCreatePrefix("173.42.128.0/20", mskLen)
+	v4PrefixSet.GetOrCreatePrefix("173.43.128.0/20", mskLen)
+	gnmi.Update(t, dut, gnmi.OC().RoutingPolicy().DefinedSets().PrefixSet(tag3IPv4).Config(), v4PrefixSet)
```
dplore, line 126: this should be gnmi.Replace.

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
```diff
@@ -163,9 +173,22 @@ func TestPrefixSetWithCLIConfig(t *testing.T) {
...
-	prefixSet := gnmi.Get[*oc.RoutingPolicy_DefinedSets_PrefixSet](t, dut, gnmi.OC().RoutingPolicy().DefinedSets().PrefixSet("TAG_3_IPV4").State())
+	gnmi.Replace(t, dut, gnmi.OC().RoutingPolicy().DefinedSets().PrefixSet(tag3IPv4).Config(), v4PrefixSet)
+	prefixSet = gnmi.Get[*oc.RoutingPolicy_DefinedSets_PrefixSet](t, dut, gnmi.OC().RoutingPolicy().DefinedSets().PrefixSet(tag3IPv4).State())
 	if len(prefixSet.Prefix) != 22 {
 		t.Errorf("Prefix set has %v prefixes, want 22", len(prefixSet.Prefix))
 	}
 }
+
+// findProcessByName uses telemetry to find out the PID of a process
+func findProcessByName(ctx context.Context, t *testing.T, dut *ondatra.DUTDevice, pName string) uint32 {
```
dplore, line 184: move to some suitable place in /internal

Perhaps /internal/system (modeling after the OC top level tree where process names are found `/system/processes/process/state/name`)

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
```diff
@@ -17,23 +17,31 @@ package prefix_set_test
...
 	pfx4       = "10.240.31.64/28"
 	mskLen     = "exact"
 )
 
+var (
+	processName = map[ondatra.Vendor]string{
+		ondatra.ARISTA: "Octa",
+	}
+)
+
```
dplore, line 44: This should go into /internal.  I've seen we've needed to reference processes and a findProcessByName function in  other featureprofiles tests as well.  (one example https://github.com/openconfig/featureprofiles/blob/e0876594d87d88658675cfa075229e29e50fe640/feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go#L29-L34)

Probably a new folder such as /internal/system

This variable could be named routingProcessName (to differentiate from other processes that could be queried).

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
```diff
@@ -100,45 +108,47 @@ func TestPrefixSet(t *testing.T) {
...
+		Restart: true,
+	}
+	processKillResponse, err := gnoiClient.System().KillProcess(context.Background(), killProcessRequest)
+	if err != nil {
+		t.Fatalf("Failed to restart process %v with unexpected err: %v", processName[dut.Vendor()], err)
+	}
+
+	t.Logf("gnoiClient.System().KillProcess() response: %v, err: %v", processKillResponse, err)
+	t.Logf("Wait 60 seconds for process to restart ...")
+	time.Sleep(60 * time.Second)
```
dplore, line 149: maybe a function in `/internal/gnoi` such as killRoutingProcess with a parameter if the process should be checked for restart?  

Better to Subscribe TARGET_DEFINED to /system/processes to see when the process restarts, and add a timeout for how long to wait.

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
```diff
@@ -100,45 +108,47 @@ func TestPrefixSet(t *testing.T) {
...
+	killProcessRequest := &spb.KillProcessRequest{
+		Signal:  spb.KillProcessRequest_SIGNAL_TERM,
+		Name:    processName[dut.Vendor()],
+		Pid:     findProcessByName(context.Background(), t, dut, processName[dut.Vendor()]),
+		Restart: true,
+	}
+	processKillResponse, err := gnoiClient.System().KillProcess(context.Background(), killProcessRequest)
+	if err != nil {
+		t.Fatalf("Failed to restart process %v with unexpected err: %v", processName[dut.Vendor()], err)
+	}
```
nandanarista, line 145: please use a device os primitive to restart the process. the gnoi, gnmi services may run within the same process depending on the implementation, so there's no way to guarantee getting a successful response.

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
```diff
@@ -100,45 +108,47 @@ func TestPrefixSet(t *testing.T) {
...
+	if len(prefixSet.Prefix) != 10 {
+		t.Errorf("Prefix set has %v prefixes, want 10", len(prefixSet.Prefix))
+	}
+
+	gnoiClient := dut.RawAPIs().GNOI(t)
+	if _, ok := processName[dut.Vendor()]; !ok {
+		t.Fatalf("Please add support for vendor %v in var processName", dut.Vendor())
+	}
+	killProcessRequest := &spb.KillProcessRequest{
+		Signal:  spb.KillProcessRequest_SIGNAL_TERM,
```
nandanarista, line 137: please use SIGKILL instead, since SIGTERM can be masked.

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
nandanarista, line 149: instead of a hardcoded wait time, it maybe better interlock to wait for system/processes for Octa to show up again w/ a different pid?

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
dplore: By device os primitive do you mean a CLI command?

I see the issue you are referring to though.  If a process is killed which hosts the gnoi service, no response is expected.  This code could explicitly define this dependency.   ie: if for example `octa` hosts gnoi and gnmi, then we can expect no response to the gnoi and no ability to check /system/processes for some period of time.  In this scenario we can still do a gnmi subscribe to find the processes and retry it a few times until some timeout.

thoughts?

nandanarista: > By device os primitive do you mean a CLI command?

either vendor cli (for arista, this will be `agent Octa terminate`) or the underlying OS command, for linux `sudo kill -SIGKILL <pid>`

> In this scenario we can still do a gnmi subscribe to find the processes and retry it a few times until some timeout.

yeah, this sounds reasonable.

self-maurya: using `agent Octa terminate` now

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
self-maurya: will create a separate CL to take of this refatoring

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
dplore: I still see gnmi.Update?  Did you miss pushing a commit or did you disagree with the recommendation?

internal/gnoi/gnoi.go
```diff
@@ -0,0 +1,75 @@
...
+	ocAgentTerminationCmd = map[ondatra.Vendor]string{
+		ondatra.ARISTA: "agent Octa terminate",
+	}
+	ocAgentDaemon = map[ondatra.Vendor]string{
+		ondatra.ARISTA: "Octa",
+	}
+)
+
+// TerminateOCAgent terminates the OpenConfig agent on the DUT.
+func TerminateOCAgent(t *testing.T, dut *ondatra.DUTDevice, waitForRestart bool) error {
```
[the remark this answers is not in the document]
dplore: Agreed, this should be [gnoi.System.KillProcess](https://github.com/openconfig/gnoi/blob/main/system/system.proto#L87)

If gnoi isn't supported, then the test (probably not the helper) could add a deviation using CLI.

feature/experimental/policy/otg_tests/prefix_set_test/prefix_set_test.go
self-maurya: it was missed during push

rohit-rp (approved):

PR #3363, opened 2024-08-06
TE-18.1 gRIBI MPLS in UDP encap and decap

dplore:
Test specification for gRIBI encap/decap with an input-policy which classifies and polices incoming packets using a one-rate, two color marker.

Several TODO items exist to acknowledge planned additions to the test.

## Highlights
* introduced new OC subtree (/qos/input-policies) 
  * The operational use case is to configure schedulers (policers) on input subinterfaces
  * Today OC requires definition of queues for schedulers but in practice, input policing typically does not involve queues
  * A new `input-policies` subtree is proposed here, relating classifiers and schedulers to each other.
  * `input-policies` are then applied to an interface with a new OC leaf-list (qos/interfaces/interface/input/policies)

Related PR's:
* #3375 defines scaling requirements for this functionality.
* #3383 defines a policer related to this test
*

PR #3383, opened 2024-08-19
DP-2.2 QoS policer matching on next-hop-group

dplore:
Note: This test is related to TE-18.1 and TE-18.3


LimeHat:
Thanks Darren,
(to continue the conversation from #3363)
As an alternative to the proposed policer definition, I would suggest implementing something like `/qos/policer-policies/` (similar to `/qos/scheduler-policies`), where the ingress policers can be defined.

And interface association can be defined via `/qos/interfaces/interface[interface-id=*]/input/policer-policy` 

Note that there's already an existing classifier-to-interface association via `/qos/interfaces/interface/input`, so the newly proposed classifier association to a policy `/qos/input-policies/input-policy/config/classifier` creates ambiguity. 

Additionally, w.r.t. 
```
  # qos interface config
  #/qos/interfaces/interface/subinterface/input/config/policies:   # TODO:  new OC leaf-list (/qos/interfaces/interface/input/config/policies)
```

`/qos/interfaces/interface` component can be used to reference interfaces or subinterfaces depending on the interface-ref config, so there's no need for the additional `subinterface` context, it will be implied by the interface-ref (where `subinterface` is present). This is, for instance, how classifiers operate already 


Unrelatedly, `matching on next-hop-group` is unusual, especially for the ingress policer. 
Most implementations that I know of operate based on packet/frame fields; and knowledge of the destination NHG requires the implementation to perform a FIB lookup first, before performing the policing action. I can't comment on feasibility, but I have some concerns.

dplore:
@LimeHat 
> > introduced new OC subtree (/qos/input-policies)
> > The operational use case is to configure schedulers (policers) on input subinterfaces
> > Today OC requires definition of queues for schedulers but in practice, input policing typically does not involve queues
> > A new input-policies subtree is proposed here, relating classifiers and schedulers to each other.
> > input-policies are then applied to an interface with a new OC leaf-list (qos/interfaces/interface/input/policies)
> 
> Is there an open PR for this already? (I wasn't able to find it)
> 
> I would argue that this may not be an optimal solution. Overloading the scheduler-policies, which are fundamentally responsible for queue management with additional, unrelated functionality (definition/instantiation of ingress policers on subinterfaces) is questionable.

There is no open PR for the proposed OC model in this README yet.

How should ingress policers be configured on a VOQ style device?   The current [QoS model](https://openconfig.net/docs/models/qos/) does not directly describe how to implement ingress policers on a VOQ style device. 

OC currently combines the notion of policer and queue servicing into one entity.  This isn't an overload based on what is defined today in OC.  (although maybe it should be changed?)

I think the only option for an input policer in the current OC model is to assume there is an input queue as well.  In a device which only (abstractly) implements an output queue, does it make sense to go ahead and use the existing qos model and have two queues and consider the input queue doesn't really exist but can be configured with a scheduler that only supports policing?  

Ingress policing only on input and queuing only on output is a very common scenario.  It seems we should have a direct way to model it.    See the [drawing in this gist ](https://gist.github.com/dplore/17574ee060f78dad380e3bd9bbb70a9d) for some options.  This  README currently proposes option "A".

vishnureddybadveli:
/qos/policer-policies/ concept is looking better Darren Loher.

vishnureddybadveli (approved):
Changes look good to me.

dplore:
I made a major revision of the scheduler configuration to align with the current, unmodified OC /qos model.  Note that the input-queue is a _dummy_ queue required to satisfy the required relationships in the OC model.

Here is what the relationships in the OC qos model look like.  I plan to add this diagram and example configuration to the qos documentation, but thought I would share it here first as this test serves as a canoncial use case for ingress policing on a device without input queues.

oc_qos_overview-Page-3 drawio

I'd like to ask for feedback on these options
1. Using a dummy queue is simple enough and aligns with the current node relationships. (which if so, meets use cases with and without input queuing and scheduling).

2. adding yet another relationship in the schema to define a new `input-type` enum of "FORWARDING_GROUP".  This has a benefit of avoiding a "dummy queue", but slightly increases the model complexity by adding another relationship to an already complex schema.  

3. Some other suggestion?

dplore:
@LimeHat @nandanarista @earies @rgwilton for comment on the QoS model for ingress policing example here.

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+## Test setup
+
+Use TE-18.1 test environment setup.
+
+## Procedure
+
+### TE-18.2.1 Generate and push configuration
+
+* Generate config for 2 classifiers which match on next-hop-group.
+* Generate config for 2 forwarding-groups mapped to "dummy" input queues
```
jaingaurav12, line 23: Darren,, Please elaborate what does this mean.

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+          conditions:
+            next-hop-group:
+                config:
+                    name: "nhg_B2"     # new OC path needed, string related to /afts/next-hop-groups/next-hop-group/state/next-hop-group-id
+          actions:
+            config:
+              target-group: "input_dest_B"
+
+
+  forwarding-groups:
```
jaingaurav12, line 86: Kindly add more context. I do not understand what is this section delivering. Thank you

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+  interfaces:                  # this is repeated per subinterface (vlan)
+    - interface: "PortChannel1.100"
+        config:
+          interface-id: "PortChannel1.100"
+        input:
+          classifers:
+            - classifier:
+              config:
+                name: "dest_A"
+                type: "IPV4"
```
jaingaurav12, line 161: I think type is not needed as policy will have terms for both IPv6 and v6.

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+            }
+          }
+          encap-headers {
+            encap-header {
+              index: 2
+              src_ip: "outer_ipv6_src"
+              dst_ip: "outer_ipv6_dst_A"
+              dst_udp_port: "outer_dst_udp_port"
+              ip_ttl: "outer_ip-ttl"
+              dscp: "outer_dscp"
```
jaingaurav12, line 240: Darren,, Not now, but just wanted to make sure this is the section that will be updated with IPsec SPI detail. correct ?

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore: This is a summary in English for the [OpenConfig ingress policing syntax defined in the readme](https://github.com/openconfig/featureprofiles/pull/3383/files#diff-f219a5b1de6d0179fdad45ecad55c22e7758fc4bd4c729e0e0c37f6104490179R32-R182)

[Additional description is in this comment](https://github.com/openconfig/featureprofiles/pull/3383#issuecomment-2364695288)

I plan to create more extensive documentation on this at openconfig.net in the near future.

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore: The OC  classifiers model is address familiy specific , so one classifier is needed for each address family.

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore: IPSec is covered by a separate test at [TE-18.6](https://github.com/openconfig/featureprofiles/pull/3435)

We wanted to scope the tests separately so we can show 
TE-18.1: encap/decap works
TE-18.2: encap/decap + policer works
TE-18.6: encap/decap + policer +ipsec works

If we put all of these into one test, then it would be hard to tell if the different important parts are working without debugging.

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore: I have updated the OC qos overview documentation to include an in depth explanation of ingress policing.  Please see https://github.com/openconfig/public/pull/1190/files?short_path=11f0b86#diff-11f0b8695aa64acdd535b0d47141c0a373e01f63099a423a21f61a542eda0052

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+    }
+  }
+}
+```
+
+
+### TE-18.2.3 Test traffic
+
+* Send traffic
+  * Send traffic from ATE port 1 to DUT for dest_A and is conforming to cir.
```
robshakir, line 307: What does `is conforming to cir` mean? Can we define what the CIR is, and what the rate that we want is?

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+  * Send traffic from ATE port 1 to DUT for to dest_B and is conforming to
+    cir.
+  * Validate packets are received by ATE port 2.
+    * Validate qos interface scheduler counters
```
robshakir, line 311: For the specification of this test, we probably need to be more concrete as to what "validate the counters" means. What are the required counters that need to be checked?

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+    * Validate afts next hop counters
```
robshakir, line 312: Similarly to above, suggest we are more specific w.r.t definition of "validate".

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+  * Validate outer packet ipv6 flow label assignment
+    * When the outer packet is IPv6, the flow-label should be inspected on the ATE.
+      * If the inner packet is IPv4, the outer IPv6 flow label should be computed based on the IPv4 5 tuple src,dst address and ports, plus protocol
```
robshakir, line 315: Does validate here mean specifically that the same 5-tuple results in the same flow label? Are there other correctness validations needed?

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+      * If the inner packet is IPv6, the inner flow label should be copied to the outer packet.
+  * Increase traffic on flow to dest_B to 2Gbps
```
robshakir, line 317: Is 2Gbps greater than the CIR? I didn't see these rates defined anywhere other than the example config.

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+  /qos/scheduler-policies/scheduler-policy/schedulers/scheduler/one-rate-two-color/config/bc:
+  /qos/scheduler-policies/scheduler-policy/schedulers/scheduler/one-rate-two-color/config/queuing-behavior:
+  /qos/scheduler-policies/scheduler-policy/schedulers/scheduler/one-rate-two-color/exceed-action/config/drop:
+
+  # qos classifier config
+  /qos/classifiers/classifier/config/name:
+  /qos/classifiers/classifier/terms/term/config/id:
+  #/qos/classifiers/classifier/terms/term/conditions/next-hop-group/config/name: # TODO: new OC leaf to be added
+
+  # qos policer config - TODO: a new OC subtree (/qos/policer-policies, essentially copying/moving policer action from schedulers)
```
robshakir, line 337: I think this is outdated.

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+    gNMI.Set:
+      union_replace: true
+      replace: true
+    gNMI.Subscribe:
+      on_change: true
+```
+
+## Required DUT platform
+
+* FFF
```
robshakir, line 362: Can this not be tested on a virtualised platform?

vishnureddybadveli (approved):
Changes look good to me.

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore: specified expected packet loss for each flow

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore: it was decided aft next hop counters are no longer needed for this scenario, so removed this.

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore: Yes, it's only in the example config.  I updated this to be more explicit about why this is done.

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore: removed

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore: Since this test includes dataplane requirements , it must also be testing on the real target platform.  
If only run on a virtual platform, that would certify the virtualized platform is compliant.  But it would not certify a platform implementing the policing on an ASIC is compliant.  In our use case, we want to know the hardware is working.  (when developing the code for this test, I would absolutely try and use a virtualized platform until I got the test perfected, then run on hardware).

TE-18.1 focuses only on gRIBI aft manipulation and verifies via packet forwarding.  A control-plane only version of that test could be written (which doesn't test packet forwarding), and run on a virtualized platform.  But a test with packet forwarding in the dataplane test needs to run on the real target platform.

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore: Good idea, I revised this with a scalable and  more explicit validation method

dplore (commented):
@robshakir ready for your review.

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,363 @@
...
+* Generate config for 2 classifiers which match on next-hop-group.
+* Generate config for 2 forwarding-groups mapped to "dummy" input queues
+  * Note that the DUT is not required to have an input queue, the dummy queue
+    satisfies the OC schema which requires defining nodes mapping
+    classfier->fwd_group->queue->scheduler
+* Generate config for 2 schedulder-policies to police traffic
+* Generate config to apply classifer and scheduler to DUT subinterface with vlan.
+* Use gnmi.Replace to push the config to the DUT.
+
+```yaml
```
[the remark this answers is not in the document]
dplore: I have filed the request for this!  Given that this canonical config section is neither yaml or json verified in these readmes at the moment, do you want me revise this to json?   (It's a 1:1 translation, so I can easily paste it in a translator)

(fwiw, the flattened oc paths section at the end in yaml format is verified to be valid OC schema)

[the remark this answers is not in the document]
dplore: As a little exercise, I made a [little yaml to json convertor](https://gist.github.com/dplore/301722299fb4c6baa5f33d1d31cc5a6e).   This yaml in this readme is syntax validated and output as JSON.  

Next steps will be to validate the OC schema in the JSON  and expand our [github CI check readme validator](https://github.com/openconfig/featureprofiles/tree/main/tools/validate_readme_spec) to look for some special heading in the markdown and do the same.  (maybe a little hacky, but I think this approach will not be too high effort).

testregistry.textproto
```diff
@@ -1001,6 +1001,10 @@ test {
   description: "MPLS in UDP Encapsulation with QoS scheduler"
   readme: ""
 }
+test: {
```
nachikethas, line 1004: For my information, why is this test being tagged under the "TE-*" bucket? Is "TE" basically being overloaded in FNT for "anything programmed through gRIBI"? Note that this is _not_ how "TE" is interpreted internally.

Would it make sense to more explicitly and clearly demarcate the tests based on the project(s) and/or initiatives that own the corresponding feature?

nachikethas (dismissed):

testregistry.textproto
dplore: Thanks for the feedback, you're correct ,I assigned "TE" (traffic engineering) because this was a gRIBI focused test.  But there is no traffic engineering in this test and while there is a strong gRIBI dependency in this test, its focus is application of QoS with gRIBI.   

I'll move this one to the "DP" (data plane) test id series since it's primary focus is on QoS.

dplore:
@vishnureddybadveli this is ready to merge, noting that there are a few TODO items.  You can take up updates for future revisions.

feature/qos/otg_tests/ingress_police_nhg/README.md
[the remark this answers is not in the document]
dplore: Converted to JSON for now.  I decided to include the comments, even though comments in JSON are invalid. 
 If/when yaml validation is implemented, we can convert it.

vishnureddybadveli (approved):
Changes look good to me.

dplore:
Thanks for all the comments.  CLosing this and tracking development here on the [dp-2.2 branch](https://github.com/openconfig/featureprofiles/blob/dp-2.2/feature/qos/otg_tests/ingress_police_nhg/README.md).  When it's closer to ready to be merged, a new PR will be created.

PR #3384, opened 2024-08-19
fixes issues introduced in PR#2795

self-maurya:


rohit-rp (approved):

cprabha:
@self-maurya  pull 2795 only removes interface configuration associating with network instance but not deleting any interface configuration.

Here in this pull we are adding interface config again .  but this is not required right ?

LimeHat:
I'd agree with @cprabha , it feels like this should be a deviation for the vendor that needs it.

cc @dplore

self-maurya:
> @self-maurya pull 2795 only removes interface configuration associating with network instance but not deleting any interface configuration.
> 
> Here in this pull we are adding interface config again . but this is not required ?

This current setup seems to be working for Juniper as well.
And if I don't remove the vrf-1 then it works for other vendors except Juniper.

cprabha:
> be working for Juniper as we

Sure,  It will be helpful if this code can be moved under deviation.

self-maurya:
> > be working for Juniper as we
> 
> Sure, It will be helpful if this code can be moved under deviation.

I had internal discussion with @dplore and few other folks.
As per Darren, the interfaces should be removed from the NI since the delete of the /network-instances/network-instance/ explicitly is a parent of the NI to interface mapping, so the vendor should be removing the interface configs on NI deletion.

@cprabha based on this, could you please raise a PR to introduce a deviation for Juniper.
Sample code
```
if !deviation {
    // reconfigure the interfaces
}
```

cc: @dplore

PR #3375, opened 2024-08-13
TE-18.3 gRIBI and Scheduler mpls-in-udp scale test

dplore:
Note, this test depends on 
* https://github.com/openconfig/featureprofiles/pull/3363 
* https://github.com/openconfig/featureprofiles/pull/3383

PR #3385, opened 2024-08-19
DP-2.4 Police traffic on input interface matching all packets using 1 rate, 2 color marker.

vishnureddybadveli:
Adding test case for one rate, two color policier matching on input interface/sub-interface.


.DS_Store
dplore, line 1: remove this file

feature/.DS_Store
dplore, line 1: remove this file

feature/qos/.DS_Store
dplore, line 1: remove this file

testregistry.textproto
```diff
@@ -945,6 +945,10 @@ test: {
   id: "TE-17.1"
   readme: "https://github.com/openconfig/featureprofiles/blob/main/feature/experimental/gribi/otg_tests/vrf_policy_driven_te/README.md"
 }
+test: {
+  id: "TE-18.2"
```
dplore, line 949: To prevent conflict with other works in progress at #3383 #3363 #3375 

```suggestion
  id: "TE-18.4"
```

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore, line 1: Hi, please remove the duplicate procedure / content from cut and paste of #3383?     I think you'd leave the procedure heading # TE-18.3 Control maximum rate of traffic transmitted or received on an interface.

I think the goal for this instance is to create a test which configures a policer classifying on an input aggregate (lag) interface vlan.    There needs to be a classifier.  I recommend classifying by the input mac of DUT port 1 in order to match all inbound L2 traffic which I think meets the intent?  

Here's the path: 
[/qos/classifiers/classifier/terms/term/conditions/l2/config/destination-mac](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-qos.html#qos-classifiers-classifier-terms-term-conditions-l2-config-destination-mac)

dplore (changes requested):

feature/qos/otg_tests/ingress_police_nhg/README.md
vishnureddybadveli: Sure,

I had added a classifier by input mac in order to match all inbound L2 traffic and drop traffic over limits.

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,216 @@
+# TE-18.2 QoS scheduler with 1 rate 2 color policer, classifying on next-hop group
+
+## Summary
+
+Use the gRIBI applied ip entries from TE-18.1 gRIBI. Configure an ingress scheduler
+to police traffic using a 1 rate, 2 color policer. Configure a classifier to match
+traffic on a next-hop-group.  Apply the configuration to a VLAN on an aggregate
+interface.  Send traffic to validate the policer.
```
dplore, line 8: Can you update this text to reflect the goal of your test?

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,216 @@
...
+## Test setup
+
+Use TE-18.1 test environment setup.
+
+## Procedure
+
+### TE-18.3.1 Generate and push configuration
+
+* Generate config for 2 scheduler polices with an input rate limit.  
+* Generate config for 2 classifiers which match on next-hop-group.
```
dplore, line 23: this needs an update too

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,216 @@
...
+      config:
+        name: “dest_A”
+      terms:
+        - term:   # repeated for address in destination A
+          config:
+            id: "match_1_dest_A"
+          conditions:
+            next-hop-group:
+                config:
+                    name: "nhg_A"     # new OC path needed, string related to /afts/next-hop-groups/next-hop-group/state/next-hop-group-id
```
dplore, line 77: Here and below, this is matching on next-hop-group.  It should be changed to match on destination MAC.

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+### TE-18.3.1 Test traffic
+
+* Send traffic
+  * Send traffic from ATE port 1 to DUT for dest_A and is conforming to cir.
+  * Send traffic from ATE port 1 to DUT for to dest_B and is conforming to
+    cir.
+  * Validate packets are received by ATE port 2.
+    * Validate qos interface scheduler counters
+    * Validate afts next hop counters
```
dplore, line 133: Since we are not matching on next-hop groups, we can remove the requirement to check next-hop-group counters.

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,216 @@
...
+  * Validate outer packet ipv6 flow label assignment
+    * When the outer packet is IPv6, the flow-label should be inspected on the ATE.
+      * If the inner packet is IPv4, the outer IPv6 flow label should be computed based on the IPv4 5 tuple src,dst address and ports, plus protocol
+      * If the inner packet is IPv6, the inner flow label should be copied to the outer packet.
+  * Increase traffic on flow to dest_B to 2Gbps
+    * Validate that flow dest_B experiences ~50% packet loss (+/- 1%)
+
+# TE-18.3 Control maximum rate of traffic transmitted or received on an interface.
```
dplore, line 141: Oh, I see you cut and paste the 18.2 test and added an 18.3 here.

Each readme is independent.  The 18.2 test above should be removed from this file.  

Additionally, this should be numbered `TE-18.4`   Also be more specific about the name of the test as there are many ways to rate limit traffic.  I suggest something like:

`TE-18.4 - Classify traffic on input using destination MAC and police using 1 rate, 2 color marker`

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,216 @@
...
+Use TE-18.1 test environment setup.
+
+## Procedure
+
+### TE-18.4.1 Generate and push configuration
+
+* Use default class (commonly known as the class-default class).
+* Configures a policer classifying on an input aggregate (lag) interface VLAN.
+* Add a classifier by input mac of DUT port 1 in order to match all inbound L2 traffic.
+* Add a limit after which traffic is dropped.
```
dplore, line 164: Consider removing the references to 'class' and 'class-default' which are undefined concepts in OC QoS.

In OC, there are classifiers and schedulers.  Classifers are the match rule,  Schedulers can be configured to police traffic and also optionally shape traffic and optionally perform different de-queueing algorithms.  In this case, we are defining a scheduler with only a policer.

dplore (changes requested):

feature/qos/otg_tests/ingress_police_nhg/README.md
```diff
@@ -0,0 +1,216 @@
+# TE-18.2 QoS scheduler with 1 rate 2 color policer, classifying on next-hop group
```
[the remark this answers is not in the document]
dplore: Oh, I see this is cut and paste from another test.  This whole TE-18.2 test should be removed.  Each readme file should contain only one Test and one or more subtests.

The structure should be something like:

* Test 18.4 - classify on dest mac and police
* Test 18.4.1 - generate and push config
* Test 18.4.2 - send traffic and validate receive traffic on OTG

feature/qos/otg_tests/ingress_police_nhg/README.md
dplore: since this is cut/paste from 18.2, this should just be deleted.

feature/qos/otg_tests/ingress_police_mac_vlan/README.md
```diff
@@ -0,0 +1,181 @@
+# TE-18.4 Classify traffic on input using destination mac and police using 1 rate, 2 color marker
```
LimeHat, line 1: I have a basic question: what is the purpose of destination-mac matching if this is a layer 3 service (this is not explicitly specified, but i'm assuming it is L3 based on the traffic flows and other TE-x.x test case descriptions).

dplore: It's true this is an L3 interface use case.  The intent is to match all packs received on the interface.  There's no explicit classifier condition for matching all packets on an interface today in OC, so I suggested mac address.

We could introduce some new container for this in `/qos/classifiers/classifier/terms/term/conditions` if we think this is better.

feature/qos/otg_tests/ingress_police_mac_vlan/README.md
```diff
@@ -0,0 +1,181 @@
...
+            destination-mac
+    - classifer: “dest_B”
+      config:
+        name: “dest_B”
+      terms:
+        - term:
+          config:
+            id: "match_1_dest_B"
+          conditions:
+            destination-mac
```
dplore, line 84:
```suggestion
            destination-mac: dut_port1_mac_address    # use the mac address from dut port1    
```

feature/qos/otg_tests/ingress_police_mac_vlan/README.md
vishnureddybadveli: May I kindly request Sergey Fomin,
+1 that this is a L3 interface use case. 
Currently vendors like cisco support default policiers on sub-interface with can match on destination mac-address. In order to support similar construct seems this pull request would be helpful for the community.

LimeHat: IMO using the interface/routing mac for wildcard match all is not quite right.

mac-criteria is often used with l2 services where it has very specific meaning/implementation, and mixing it with something completely different is suboptimal. 

Another question is,
>There's no explicit classifier condition for matching all packets on an interface today in OC, so I suggested mac address.

Does it have to be explicit? What if we assume implicit "match all" when no conditions are specified?

vishnureddybadveli: Thanks Sergey Fomin, I agree mac-criteria is often used with l2 services.
Thinking aloud on how to match all packets with lack of classifier condition.

```
Does it have to be explicit? What if we assume implicit "match all" when no conditions are specified?
```

Great point let me check in this direction, thanks for your kind direction.

dplore: > > There's no explicit classifier condition for matching all packets on an interface today in OC, so I suggested mac address.
> 
> Does it have to be explicit? What if we assume implicit "match all" when no conditions are specified?

That sounds good enough for this pull request.  @vishnureddybadveli can you update this PR remove the match critiera and add text to the readme  that we expect the lack of match conditions to cause all packets to be matched.  

Personally I like to avoid implicit rules like this, but we can take it up in a separate PR for the OC models and see where that goes.

vishnureddybadveli: Thanks Darren

Added the required text.

dplore:
Please refresh the canonical OC for policer configuration to match the style in [DP-2.2](https://github.com/openconfig/featureprofiles/blob/dp-2.2/feature/qos/otg_tests/ingress_police_nhg/README.md)

(you can of course remove the classifier terms, but please do use the forwarding-groups and queue configuration)

vishnureddybadveli:
> Please refresh the canonical OC for policer configuration to match the style in [DP-2.2](https://github.com/openconfig/featureprofiles/blob/dp-2.2/feature/qos/otg_tests/ingress_police_nhg/README.md)
> 
> (you can of course remove the classifier terms, but please do use the forwarding-groups and queue configuration)

Yes refreshed with the canonical OC to match the style in DP-2.2 and removed classifier terms, in order to avoid any confusion as DP-2.2 is still been in the works and to decouple with DP-2.2 I had refrained from using any queue configurations, forwarding-groups.

feature/qos/otg_tests/ingress_police_default/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+### DP-2.4.2 Test traffic
+
+* Send traffic
+  * Send traffic from ATE port 1 to DUT for dest_A and is conforming to cir.
+  * Send traffic from ATE port 1 to DUT for to dest_B and is conforming to
+    cir.
+  * Validate qos counters per DUT.
+  * Validate qos counters by ATE port.
+  * Validate packets are received by ATE port 2.
```
dplore, line 147: specify what the validation criteria is,  see DP-2.2 for example

feature/qos/otg_tests/ingress_police_default/README.md
```diff
@@ -0,0 +1,191 @@
...
+  * Validate outer packet ipv6 flow label assignment
+    * When the outer packet is IPv6, the flow-label should be inspected on the ATE.
+      * If the inner packet is IPv4, the outer IPv6 flow label should be computed based on the IPv4 5 tuple src,dst address and ports, plus protocol
+      * If the inner packet is IPv6, the inner flow label should be copied to the outer packet.
```
dplore, line 151: see DP-2.2 for specific recommended way to validate this

testregistry.textproto
```diff
@@ -945,6 +945,10 @@ test: {
   id: "TE-17.1"
   readme: "https://github.com/openconfig/featureprofiles/blob/main/feature/experimental/gribi/otg_tests/vrf_policy_driven_te/README.md"
 }
+test: {
```
dplore, line 948: Please sort this list (move to the other DP test entries)

dplore (dismissed):

.DS_Store
dplore, line 1: You'll need to remove this file from the pull request and then the CI checks will pass.

dplore (approved):

PR #3386, opened 2024-08-19
fixed the failure that was introduced by PR-3376

bkreddy143:
1. PR- 3376 changes changed the signal type to KILL instead of TERM  that resulted to non graceful application restart , fixed by adding that as an parameter to the  KillProces() function.
2. Also, added parameter option for restart after kill to have more flexibility  

"This code is a Contribution to the OpenConfig Feature Profiles project ("Work") made under the Google Software Grant and Corporate Contributor License Agreement ("CLA") and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia’s intellectual property are granted for any other purpose. This code is provided on an "as is" basis without any warranties of any kind."


nsadhasivam (changes requested):
Please resolve the static analysis failure

bkreddy143:
> Please resolve the static analysis failure don't use ALL_CAPS in Go names; use CamelCase

updated Variable to  CamelCase

nsadhasivam (approved):
LGTM!

PR #3389, opened 2024-08-21
Accountz-1.1

dipchauh:


feature/security/gnsi/acctz/tests/record_subscribe_full/record_subscribe_full_test.go
```diff
@@ -0,0 +1,828 @@
...
+		if readyCounter >= 10 {
+			t.Fatal("grpc connection never reached ready state")
+		}
+
+		time.Sleep(time.Second)
+	}
+
+	return conn, addrObj
+}
+
```
morrowc, line 183: there's a bunch of (I think) setup happening here, would it be better to pull the setup into distinct files outside of this file? keeping this file to just the tests you want to run?

At the least it might be nice to at least have some comments for each of these functions:
  // Setup test RPCs to be used in the acctz client tests

dipchauh: Moved grpc/ssh helper functions to `internal/security/acctz/acctz.go` & fixed comments.

feature/security/gnsi/acctz/tests/record_subscribe_full/record_subscribe_full_test.go
```diff
@@ -0,0 +1,254 @@
...
+	var recordIdx int
+	var lastTimestampUnixMillis int64
+
+	for {
+		if recordIdx >= len(records) {
+			t.Log("Out of records to process...")
+			break
+		}
+
+		r := make(chan recordRequestResult)
```
morrowc, line 97: this could benefit from a comment or three...
I suspect this pattern is dangerous, since we re-create this channel each time.
we go-func the reading from the recv a single record into that channel.

evaluation of the record happens...

Why use this channel at all?

internal/security/acctz/acctz.go
```diff
@@ -0,0 +1,917 @@
...
+	shellCommand         = "uname -a"
+	gnmiCapabilitiesPath = "/gnmi.gNMI/Capabilities"
+	gnoiPingPath         = "/gnoi.system.System/Ping"
+	gnsiGetPath          = "/gnsi.authz.v1.Authz/Get"
+	defaultSSHPort       = 22
+)
+
+var gRPCClientAddr net.Addr
+
+// Record represents the structure for an acctz record.
```
morrowc, line 67: I would have thought just using the proto struct directly would work here, why doesn't that work?

internal/security/acctz/acctz.go
```diff
@@ -0,0 +1,917 @@
...
+	})
+	if err != nil {
+		t.Fatalf("Failed sending credentialz rotate account credentials finalize request, error: %s", err)
+	}
+
+	// Brief sleep for finalize to get processed.
+	time.Sleep(time.Second)
+}
+
+func nokiaFailCliRole(t *testing.T) *gnmi.SetRequest {
```
morrowc, line 134: is nokia somehow special? (why call this 'nokiaFail....' ?)

internal/security/acctz/acctz.go
```diff
@@ -0,0 +1,917 @@
...
+		t.Log("DUT does not support `Service` function, will attempt to resolve dut name field.")
+
+		// Suppose ssh could be not 22 in some cases but don't think this is exposed by introspect.
+		dialTarget := fmt.Sprintf("%s:%d", dut.Name(), defaultSSHPort)
+		resolvedTarget, err := net.ResolveTCPAddr("tcp", dialTarget)
+		if err != nil {
+			t.Fatalf("Failed resolving ssh target %s", dialTarget)
+		}
+		target = resolvedTarget.String()
+	} else {
```
morrowc, line 234: chances are this else isn't necessary? (or could be made to be not necessary)

internal/security/acctz/acctz.go
```diff
@@ -0,0 +1,917 @@
...
+	}
+
+	var target string
+	err := binding.DUTAs(dut.RawAPIs().BindingDUT(), &serviceDUT)
+	if err != nil {
+		t.Log("DUT does not support `Service` function, will attempt to resolve dut name field.")
+
+		// Suppose ssh could be not 22 in some cases but don't think this is exposed by introspect.
+		dialTarget := fmt.Sprintf("%s:%d", dut.Name(), defaultSSHPort)
+		resolvedTarget, err := net.ResolveTCPAddr("tcp", dialTarget)
```
morrowc, line 229: is working dns in the test environment expected to be reliable/etc?

feature/security/gnsi/acctz/tests/record_subscribe_full/record_subscribe_full_test.go
dipchauh: Moved channel creation outside the loop & added comments.
Since response stream is expected to be open indefinitely, channels can provide a way to timeout & exit the loop if no records were received for some time (since the `Recv` is a blocking call).
Another option could be to use `context.WithTimeout()`

internal/security/acctz/acctz.go
dipchauh: Removed record struct & using pb directly

internal/security/acctz/acctz.go
dipchauh: This is for cli tests where user needs to send some cli commands which are denied. Currently it seems there are no corresponding OC paths to define denied commands for users. So each vendor would need to create a failure cli role (roles with certain commands not allowed) in their native model.

internal/security/acctz/acctz.go
dipchauh: Yes, this can be made to be not necessary by returning 2 times from the func. I was looking to log the target string once before returning hence the else.

internal/security/acctz/acctz.go
dipchauh: Yes, this would be required in case we are using dut.Name() as the target address then we can check if it's resolvable before dialing. 
Also required for comparing address bits in acctz record.

morrowc (approved):

PR #3401, opened 2024-08-27
Update to the MTU test readme to cover for Tunnel interfaces.

sachendras:
- Separate the test to 2 sub tests
- While the original test covered MTU configuration on physical/Bundle  interfaces only, added subtest covers for Tunnel interface MTU configuration and corresponding emphasis on zero fragmentation plus packet drops.

Updated write-up also covers for missed config and state paths.


feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/README.md
```diff
@@ -7,31 +7,90 @@ IPv4 and IPv6 packet sizes are sent over them.
...
+  * Run traffic flows of the following size over IPv4 and IPv6 between ATE ports. 
+    * 1500 Bytes
+    * 2000 Bytes
+    * 4000 Bytes
+    * 9202 Bytes
+  * Assert ATE reports packets sent and received count are the same, indicating no fragmentation, and 
+    successful transit.
+
+* MTU-1.3.2: Test w/ tunnel interfaces and forwarding plane via the physical and bundled interfaces
+  * Configure DUT with a GRE tunnel interface.
```
dplore, line 26: Something we should clarify by providing the canonical OC is:

From the paths included at the end of the readme, I am guessing the intent is for the stateful interface style configuration.  This should be the target if one is going to run a control protocol over it (such as bgp)?  

The other style would be a policy-forwarding configuration which uses match and encap rules.   Or perhaps you have use cases for both styles? (which should have two readmes if that is true)

The test should define 'canonical OC' used for the test intent.  The follow format is [pending standardization for README](https://github.com/openconfig/featureprofiles/pull/3336).  I suggest using it here:

### OC tunnel interface configuration
```yaml
interfaces:
  - interface: 'tunnel1_if_name'
    config:
      name: 'tunnel1_if_name'
    tunnel:     # configures the tunnel outer packet 
      config:
       src: 'tunnel1_outer_ip_src'
       dst: 'tunnel1_outer_ip_dst'
       ttl: 'tunnel1_outer_ttl'
       gre-key: 'tunnel1_outer_gre_key'
      ipv4:
        config:
          mtu: 'tunnel1_inner_mtu'
        addresses:
          - address:
            config:
              ip: 'tunnel1_inner_ipv4'
              prefix-length: 'tunnel1_inner_ipv4_prefixlen'
      ipv6:
        config:
          mtu: 'tunnel1_inner_mtu'        addresses:
        addresses:
          - address:
            config:
              ip: 'tunnel1_inner_ipv6'
              prefix-length: 'tunnel1_inner_ipv6_prefixlen'
```

sachendras: 1. Do we have a use case for both scenarios?
**Response:**

Just to elaborate on the topics in the question so we set the context for the requirements.

a. Filter based encap: This approach uses a filter on the incoming direction that identifies, L3/L4 or other parameters in the TOS byte to set a forwarding action as encap to tunnel destination address which is reachable via the FIB entry

b. Interface based encap: In this case the router has virtual tunnel interfaces. The DUT will have routes for the destination address of the received traffic pointing at the tunnel interface. The tunnel interface has required encap parameters and the tunnel destination configured. The tunnel destination will be reachable via LPM lookup

While "a" provides us mature filter options to tunnel the desired flow, "b" provides a simple L3 matched tunelling that is independent of ACLs. This is important because, large filters consume memory resources and we should avoid them where possible.

So the answer is, yes we need both cases. We currently have Tun-1 and Tun-2  tests that cover both scenarios. What is missing is something that covers a hybrid of both. We will definitely add the latter use case.

feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/README.md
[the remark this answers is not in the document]
sachendras: I think it depends on how you look at it. If you look from the MTU perspective, we want to confirm NOS MTU functionality for both unencaped as well as encaped traffic. That is why this test has both scenarios and we test for different MTU values. 

If the focus of my test is validating tunnel behavior, then I wouldn't include physical interface scenario. Also, I wouldnt test different MTU values.

feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/README.md
```diff
@@ -7,31 +7,90 @@ IPv4 and IPv6 packet sizes are sent over them.
...
+* MTU-1.3.1: Test with Physical and Bundle interfaces
+  * Run traffic flows of the following size over IPv4 and IPv6 between ATE ports. 
+    * 1500 Bytes
+    * 2000 Bytes
+    * 4000 Bytes
+    * 9202 Bytes
+  * Assert ATE reports packets sent and received count are the same, indicating no fragmentation, and 
+    successful transit.
+
+* MTU-1.3.2: Test w/ tunnel interfaces and forwarding plane via the physical and bundled interfaces
```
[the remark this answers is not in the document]
sachendras: https://github.com/orgs/openconfig/projects/10/views/1?filterQuery=mtu&pane=issue&itemId=78122119 issue raised for the work

feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/README.md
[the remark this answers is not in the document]
dplore: Agreed, there is a not a right or wrong answer here. We can stay with your current interpretation, thanks for considering.

feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/README.md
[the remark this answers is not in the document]
dplore: Thanks, please embed that issue link into the README as a todo.  

(Good example of a TODO link https://github.com/openconfig/featureprofiles/blob/971d8dec564b8302e613295fce2cb0d907cf8935/feature/interface/singleton/otg_tests/singleton_test/README.md?plain=1#L52)

feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/README.md
dplore: Ok, I think you are saying we use both and why.  Thanks!

I am not sure we need to have both together in one test, unless it is a special scaling test.  In my experience, unless you're looking at interop or performance of a very small number of features,  a config scale/IC resources limit test is best added to an integration test (with a full device config push) and probably doesn't fit into FNT.  (because config scale requires configuring a lot of features at the same time).

dplore (dismissed):
LGTM.  Please do add the TODO link into the README to help identify the new part of this README.

feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/README.md
```diff
@@ -7,31 +7,107 @@ IPv4 and IPv6 packet sizes are sent over them.
...
+                  ip: 'tunnel1_interface_ipv4'     # For tunnel decap destination and/or route next-hop
+                  prefix-length: 'tunnel1_interface_ipv4_prefixlen'
+          ipv6:
+            config:
+              mtu: 'tunnel1_inner_mtu'
+            addresses:
+              - address:
+                  config:
+                    ip: 'tunnel1_interface_ipv6'     # For tunnel decap destination and/or route next-hop
+                    prefix-length: 'tunnel1_interface_ipv6_prefixlen'
```
dplore, line 75: This portion of paths needs to be moved out of the heading `## OpenConfig Path and RPC Coverage
`.  THis is because our CI requires only a flat list of OC paths to be validating.

I suggest moving this canonical config to the test setup section.  (The test procedure can reference "push the config defined in the setup to the DUT" for example).

dplore: Moving this will fix the CI check error the PR currently has

sachendras: Done!

feature/mtu/largeippacket/otg_tests/large_ip_packet_transmission/README.md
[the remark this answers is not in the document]
sachendras: Done!

masood-shah (approved):

sezhang2 (approved):
The code has been reviewed, so I just approved it for code submission

PR #3409, opened 2024-08-28
TE-18.5 gRIBI encap mpls in UDP or GRE with inner TTL manipulation

vishnureddybadveli:


vishnureddybadveli:
Not pursuing this instead use policy forwarding rule

PR #3417, opened 2024-09-05
OTG: Update RT-5.6 to change link down on port1 instead of all ports

ANISH-GOTTAPU:
Fix for the issue reported https://github.com/open-traffic-generator/fp-testbed-juniper/issues/77


ANISH-GOTTAPU:
will open another PR to address the issue as experimental folder is removed and all tests are under feature/

PR #3426, opened 2024-09-10
Add `attestz-[123]` to the testregistry.

robshakir:
```
 * (M) testregistry.textproto
  - Add missing attestz test IDs.
```

### DIFF
--- a/testregistry.textproto
+++ b/testregistry.textproto
@@ -73,6 +73,21 @@ test: {
   exec: " "
 }
 test: {
+  id: "attestz-1"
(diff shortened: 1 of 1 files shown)


dplore (approved):
This is a helpful improvement, so LGTM.  But the existing README should be split into 3 README's in separate folders if you feel like doing that in this PR.

robshakir:
I'd rather not take on editing these -- I can if it's a strict requirement, but it seems like we should be enforcing this at original PR review time. Are there clear standards being set? (It's not clear to me that there are.)

PR #3434, opened 2024-09-17
RT-7.8 bgp_comm_match_action_test.go

cprabha:
Hi,

Both ATE Peers should be in different AS otherwise routes will not be forwarded by dut.
Also added deviation.


Swetha-haridasula (approved):
lgtm

ram-mac (approved):
The changes looks good and as per the requirements like have different AS numbers between DUT and ATE; This was even done in another test Nested policies long back.

So approving it.

rohit-rp (approved):

PR #3438, opened 2024-09-18
cfgplugins optical changes + Cisco deviations

karthikeya-remilla:
this PR builds on top of code changes proposed by @hattikals over multiple open PRs.

Changes and brief description for the following files:

1. zr_cd_test: use cfgplugins code + enable operational-mode flag
2. zr_input_output_test: enable operational-mode flag
3. cfgplugins/interface.go: update the common function InterfaceConfig for tests to consume


ahsaanyousaf (approved):

feature/platform/transceiver/tests/zr_cd_test/zr_cd_test.go
```diff
@@ -32,6 +32,8 @@ const (
 var (
 	frequencies        = []uint64{191400000, 196100000} // 400ZR OIF wavelength range
 	targetOutputPowers = []float64{-13, -9}             // 400ZR OIF Tx power range
+	opModeFlag         = flag.Int("operational_mode", 1, "vendor-specific operational-mode for the channel")
```
ejbrever, line 35: just to make it explicit, can you please call this "operationalModeFlag" in all files?

internal/cfgplugins/interface.go
```diff
@@ -33,28 +33,29 @@ const (
 	targetFrequencyToleranceMHz   = 100000
 )
 
-// opticalChannelComponentFromPort returns the name of the optical channel component for the given
+// OpticalChannelComponentFromPort returns the name of the optical channel component for the given
 // port.
-func opticalChannelComponentFromPort(t *testing.T, dut *ondatra.DUTDevice, p *ondatra.Port) string {
+func OpticalChannelComponentFromPort(t *testing.T, dut *ondatra.DUTDevice, p *ondatra.Port) string {
```
ejbrever, line 38: Oh interesting....we actually just went through this in a different place for this PR:
https://github.com/openconfig/featureprofiles/pull/2891

https://github.com/hattikals/featureprofiles/blob/3a34ffd7466037956b25ddea4ffbe34aae09e7ee/internal/components/components.go#L203

I don't know why we have this same code in two places :/

We should probably merge this somehow and only have one. Maybe just to start in this PR, can you just copy the logic from the other file?

feature/platform/transceiver/tests/zr_temperature_test/zr_temperature_test.go
```diff
@@ -92,15 +92,18 @@ func TestZRTemperatureState(t *testing.T) {
...
-	for _, s := range subcomponents {
-		subc, ok := s.Val()
-		if ok {
-			sensorComponent := gnmi.Get[*oc.Component](t, dut1, gnmi.OC().Component(subc.GetName()).State())
-			if sensorComponent.GetType() == sensorType {
-				scomponent := gnmi.OC().Component(sensorComponent.GetName())
-				if scomponent != nil {
-					component1 = scomponent
+	// subcomponents check not required for CISCO
+	if dut1.Vendor() != ondatra.CISCO {
```
ejbrever, line 96: can we rewrite this a little for readability:

var compWithTemperature *oc.Component
switch dut1.Vendor() {
case ondatra.CISCO:
  compWithTemperature = gnmi.OC().Component(transceiverName)
default:
  <...add the for loop logic here...>
}

ejbrever (commented):
I see your note about once getting conditional approval you plan to change more files. Please do not do this.

It really helps to keep changes as small as possible for quick reviews. In the future if you can break the PRs up into small pieces that will be very helpful.

feature/platform/transceiver/tests/zr_temperature_test/zr_temperature_test.go
dplore: Also note, the canonical OC to get temperature is through a subcomponent sensor.  If cisco does not support this, then a deviation should be defined and used instead of `if dut1.Vendor() != ondatra.CISCO `

internal/cfgplugins/interface.go
dplore: I recommended this change offline.  Yes, the goal should be to migrate to using `internal/cfgplugins` for common configuration functions.  For ease of review, all the migration does not need to happen in this PR.

ejbrever: Agree, we should definitely break that up in multiple PRs.

karthikeya-remilla:
> I see your note about once getting conditional approval you plan to change more files. Please do not do this.
> 
> It really helps to keep changes as small as possible for quick reviews. In the future if you can break the PRs up into small pieces that will be very helpful.

@ejbrever @dplore I agree about keeping the PR small for ease of review. I assumed that all files must be migrated in this PR to not break regression runs, since we are changing the common utility function. 

Will split this PR by reducing the changed files. Thanks!

feature/platform/transceiver/tests/zr_temperature_test/zr_temperature_test.go
karthikeya-remilla: Ack - I'll push these changes for zr_temperature_test in my next PR.

internal/cfgplugins/interface.go
karthikeya-remilla: Ack - will revert to the tests using components.OpticalChannelComponentFromPort approved as part of #2891

ejbrever (approved):
Thanks for the changes!

ahsaanyousaf (approved):

PR #3452, opened 2024-09-24
TE-16.3 - fix for transport and subtest-8

singh-prem:
1. Added changes to add loopback address as transport address for vendor=cisco

2. Deleted code from TC-8 which is not correct as per readme. Added code to delete tunnel prefixes from niTEVRF222 such that traffic falls to backup path that leads to decap and lookup.


gggsmith:
@singh-prem could you please resolve the conflicts? thank you!

karthikeya-remilla:
This PR is obsoleted by later ones.

PR #3456, opened 2024-09-25
RT-5.8: IPv6 Link Local : Changes to await for interface up check after replace config for OTG

priyaranjannanda:
RT-5.8: IPv6 Link Local : Changes to await for interface up check after replace config for OTG configuration to avoid OTG failures.


ram-mac (commented):
Changes look good; Need to validate it once.

ram-mac (approved):
Validated the changes and test is passing

self-maurya (approved):

PR #3462, opened 2024-09-26
Transceiver-9: improving test flow

karthikeya-remilla:
Currently in zr_laser_bias_test, we are streaming individual laser-current leaves in sample mode. namely - instant/avg/min/max values. 

During the verification process, since these are pulled at different times, the avg value is sometimes not between the min and max values causing false test failure. 

"The average is not between the maximum and minimum values Avg:232.50 Min:232.60 Max:232.60"

To better suit the test intent, we need to change the test to stream at the container level i.e 
p1Stream := samplestream.New(t, dut1, component.OpticalChannel().LaserBiasCurrent().State(), 10*time.Second)

and then check the condition: min <= avg <= max for all the values "within" the container streamed at the same timestamp. 

Test is passing with these changes.


feature/platform/transceiver/tests/zr_laser_bias_current_test/zr_laser_bias_current_test.go
```diff
@@ -143,26 +117,21 @@ func TestZRLaserBiasCurrentStateInterface_Flap(t *testing.T) {
 	i := d.GetOrCreateInterface(dp1.Name())
 	i.Enabled = ygot.Bool(false)
 	i.Type = oc.IETFInterfaces_InterfaceType_ethernetCsmacd
+	gnmi.Replace(t, dut1, gnmi.OC().Interface(dp1.Name()).Name().Config(), dp1.Name())
```
ejbrever, line 120: can you please share more why this is needed as I think the next line would cover this?

karthikeya-remilla: unintentional - removed the line now!

ejbrever (approved):

ahsaanyousaf (approved):

PR #3463, opened 2024-09-26
add gnmirpc to oc paths

ElodinLaarz:
The intent is to be able to allow both _requirements_ to be made clearer by attaching the required RPC level of support (get, set, subscribe), as well as more detailed information about the current level of _support_.
Just because a path is supported under `gnmi.Get` doesn't mean it's supported `ON_CHANGE` in a `gnmi.Subscribe`!

The messages can be used to either list requirements or, if describing a particular target, it can be used to describe the level of support a target provides.

For read paths, we allow specification of desired (or implemented) support level: streaming, subscribe once, or polling.
Additionally, for paths requiring streaming support, we add specifications for the subscription type: target defined, on change, or sampled.
Finally, we add a `sample_interval_nanoseconds`, which can be used as a 'maximum desired cadence', if specified as a requirement or the 'minimum recommended cadence' if describing a target's capabilities.

For a target, a cadence of `0` (when the `SubscriptionMode` includes `ON_CHANGE` as supported) describes `ON_CHANGE` as the recommended frequency of subscription.


proto/ocpaths.proto
```diff
@@ -76,3 +80,39 @@ message OCPathConstraint {
     string platform_type = 1;
   }
 }
+
+// GNMIRpc describes expected (or supported) behavior for a particular
+// Openconfig path.
+message GNMIRpc {
+  bool get = 1;
```
dplore, line 87: I have heard requests to have some explicit negation of a required oc path.  ie: `bool not_supported = 4;` (or not_used, not_required, etc...)

The use case is to explicitly mark some path is not required which may be useful when:
1.  There is a history where the path was previously supported, but no longer is. 
2. There is uncertainty if the path in question was accidentally omitted from the list.

dplore (approved):

proto/ocpaths.proto
ElodinLaarz: That seems reasonable, overall. I just don't like having negative parameters. (_flashback to `disable no_zero_suppression false`_)

Perhaps we can lump get/set/subscribe into an enum?

```
  // Support allows specification for whether a path is required / supported for each of Get/Set/Subscribe.
  enum Support {
    UNSPECIFIED_SUPPORT = 0;
    UNSUPPORTED = 1;  // This path explicitly is not required / the target has no support for this path.
    GET = 2;  // Get requests must / do succeed for this path.
    SET = 3;  // Set requests must / do succeed for this path.
    SUBSCRIBE = 4;  // Certain subscribe requests must / do succeed for this path, as outlined below.
  }
  repeated Support support = 1;
```

WDYT?

dplore: Ok, turns out the idea for a 'not supported' case is no longer a use case. 

Let's proceed with what you have here in this PR.   No need to add in the enum.

PR #3465, opened 2024-09-26
Update testregistry.textproto

vvlakshmanamurthy:
Updating test registry with the Policy-based VRF selection decap /encap/TZ bit marking TCs.

### DIFF
--- a/testregistry.textproto
+++ b/testregistry.textproto
@@ -397,6 +397,24 @@ test: {
   exec: " "
 }
 test: {
+  id: "PF-1.6"
+  description: "GUE Decapsulation & Policy-based VRF selection"
+  readme: " "
+  exec: " "
+}
+test: {
+  id: "PF-1.7"
+  description: "Policy-based VRF selection"
+  readme: " "
+  exec: " "
+}
+test: {
+  id: "PF-1.8"
+  description: "Policy-based VRF selection , encap & DSCP marking"
+  readme: " "
+  exec: " "
+}
+test: {
   id: "PLT-1.1"
   description: "Interface breakout Test"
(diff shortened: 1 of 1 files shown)


testregistry.textproto
```diff
@@ -396,6 +396,24 @@ test: {
...
 }
+test: {
+  id: "PF-1.6"
+  description: "GUE Decapsulation & Policy-based VRF selection"
+  readme: " "
+  exec: " "
+}
+test: {
+  id: "PF-1.7"
+  description: "Policy-based VRF selection"
```
robshakir, line 407: Did you review the tests in https://github.com/openconfig/featureprofiles/tree/main/feature/experimental/policy/policy_vrf_selection/otg_tests -- there are some tests that cover the same features here.

robshakir (changes requested):
Please sign the CLA or if you are a Googler register your GitHub account.

testregistry.textproto
sachendras: @vvlakshmanamurthy .. imo it must be as follows

test: {
  id: "PF-1.6"
  description: "Policy-based VRF selection and Traffic migration between VRFs by Traffic class"
  readme: " "
  exec: " "
}

We must delete PF-1.7

@robshakir, the existing tests arent similar. Existing test uses VLAN subinterfaces per VRF and also relies on IP-IP inner and outer header details for VRF selection in the policy. May be we must updated the original one to be more specific to its use case. More like in https://github.com/openconfig/featureprofiles/pull/3475/files for TC based VRF selection.

vvlakshmanamurthy: Thanks Sach , will update the test cases.

PR #3466, opened 2024-09-26
Fixing policyforwarding rule using the /afts tree, instead should be …

vishnureddybadveli:
…in the /policy-forwarding tree.

### DIFF
--- a/feature/gribi/otg_tests/mpls_in_udp/README.md
+++ b/feature/gribi/otg_tests/mpls_in_udp/README.md
@@ -248,7 +248,6 @@ decapsulate MPLS in UDP.  # TODO: Move to dedicated README
 openconfig-network-instance:
   network-instances:
     - network-instance: "DEFAULT"
-      afts:
         policy-forwarding:
           policies:
             policy: "default decap rule"


feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -248,7 +248,6 @@ decapsulate MPLS in UDP.  # TODO: Move to dedicated README
 openconfig-network-instance:
   network-instances:
     - network-instance: "DEFAULT"
-      afts:
```
dplore, line 251: This is actually ok.  the path is:


/network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/action/config/decapsulate-mpls-in-udp
--


<br class="Apple-interchange-newline">[/network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/action/config/decapsulate-mpls-in-udp](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-policy-forwarding-policies-policy-rules-rule-action-config-decapsulate-mpls-in-udp)

dplore (changes requested):
The suggested change is not needed, but the gribi proto could be fixed / synced to what is in TE-18.1

PR #3467, opened 2024-09-27
OTG: RT 1.26 

ANISH-GOTTAPU:
With the existing code, we are hitting with tracking bits limitation on ixia-c SW with KNE. Updating the test to make it work on all OTG platforms seamlessly.


Swetha-haridasula (approved):
lgtm

dplore (approved):

PR #3477, opened 2024-09-30
Implementation of remaining containerz tests.

alshabib:
This commit introduces the following tests:

CNTR-1.2
CNTR-1.3
CNTR-1.4
CNTR-1.5
CNTR-1.6


feature/container/containerz/tests/container_lifecycle/README.md
```diff
@@ -98,6 +100,16 @@ Using the container started as part of CNTR-1.2, validate that the container can
...
 
+## CNTR-1.5: Create a volume on the DUT.
+
+Validate the the DUT is capable of creating a volume, reading it back
+and removing it. 
+
+## CNTR-1.6: Upgrade a container on the DUT.
+
+Using the same container started as part of CNTR-1.1, validate that the container
+can upgraded.
```
robshakir, line 111:
```suggestion
can be upgraded. 
```

robshakir, line 111: Consider adding something that defines what "upgraded" means?

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -45,18 +56,35 @@ func TestDeployAndStartContainer(t *testing.T) {
...
 	}
 
-	ret, err := cli.StartContainer(ctx, "cntrsrv", "latest", "./cntrsrv", "test-instance", client.WithPorts([]string{"60061:60061"}))
+	ret, err := cli.StartContainer(ctx, "cntrsrv", "latest", "./cntrsrv", instanceName, client.WithPorts([]string{"60061:60061"}))
 	if err != nil {
 		t.Fatalf("unable to start container: %v", err)
 	}
+
+	// wait for cntr container to come up.
+	time.Sleep(5 * time.Second)
```
robshakir, line 65: I presume there's nothing fancier we can do to determine whether this container is alive? Is this the one that implements `ping` such that the service can be checked -- or should we consider this a wider integration test?

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -65,5 +93,155 @@ func TestDeployAndStartContainer(t *testing.T) {
...
+
+	var logs []string
+	for msg := range logCh {
+		logs = append(logs, msg.Msg)
+		if msg.Error != nil {
+			t.Errorf("logs returned an error: %v", err)
+		}
+	}
+
+	if len(logs) == 0 {
```
robshakir, line 118: I presume that the container here is specified to create some logs so there's no chance that this is a flake?

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -65,5 +93,155 @@ func TestDeployAndStartContainer(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("unable to list containers: %v", err)
+	}
 
+	for cntr := range listCh {
+		t.Errorf("StopContainer did not stop the container: %v", cntr)
+	}
+}
+
+func TestVolumes(t *testing.T) {
```
robshakir, line 165: Can you add a comment here? I think it'd be worth just noting that the test here ensures that a volume can be created and removed from an API perspective -- it doesn't check that it's possible to actually _use_ that volume.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -65,5 +93,155 @@ func TestDeployAndStartContainer(t *testing.T) {
...
+		case prog.Error != nil:
+			t.Fatalf("failed to push image: %v", err)
+		case prog.Finished:
+			t.Logf("Pushed %s/%s\n", prog.Image, prog.Tag)
+		default:
+			t.Logf(" %d bytes pushed", prog.BytesReceived)
+		}
+	}
+
+	_, err = cli.UpdateContainer(ctx, "cntrsrv", "upgrade", "./cntrsrv", instanceName, false, client.WithPorts([]string{"60061:60061"}))
```
robshakir, line 225:
```suggestion
	if _, err = cli.UpdateContainer(ctx, "cntrsrv", "upgrade", "./cntrsrv", instanceName, false, client.WithPorts([]string{"60061:60061"})); err != nil { ... }
```

robshakir (commented):
LGTM generally, some comments in-line.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
alshabib: Actually I set this to retry 5 times.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
alshabib: Yes this is correct.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -65,5 +110,156 @@ func TestDeployAndStartContainer(t *testing.T) {
...
+	volCh, err := cli.ListVolume(ctx, nil)
+	if err != nil {
+		t.Errorf("unable to list volumes: %v", err)
+	}
+
+	wantVolumes := []*client.VolumeInfo{
+		{
+			Name:    "my-vol",
+			Driver:  "local",
+			Options: map[string]string{"device": "", "o": "", "type": "none"},
```
robshakir, line 210: Can we document what these `Options` here are please? Are these default labels that are added by Docker?

robshakir (approved):
Thanks for adding these!

robshakir (approved):

PR #3486, opened 2024-10-01
Create README.md

sachendras:


feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+
+
+**PF-1.6.1: [Baseline]**
+   * Situation:
+       * BE1-AF4 traffic from ATE1:Port1 to ATE2:Port1 while there are no IBGP advertisements from ATE2:Port2 to DUT:Port3
+       * Unencaped traffic received from ATE2 to ATE1:Port1 is routed via the IBGP session between DUT:Port2 and ATE2:Port1<br>
+
+   * Test Steps:
+       * Please run all the health checks suggested above prior to starting the test as a baseline.
+       * Configure DUT as suggested above.
```
robshakir, line 177:
```suggestion
       * Configure DUT as described above.
```

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+           * Make sure the IBGP peering between DUT:Port3 and ATE2:Port2 are down for this subtest.
+       * Start above mentioned flows
+           * Dont start the above mentioned encaped flows from ATE2:Port3<br>
+           
+   * Expectations:
+	    * ATE1:Port1 to ATE2:Port1 flows are expected to be 100% successful and routing over the IBGP peering between DUT:Port2 and ATE2:Port1
```
robshakir, line 183: Can we define 100% successful? I assume 0% packet loss is your pass/fail criteria?

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+	    * ATE2 to ATE1:Port1 Unencaped flows must be 100% successful. The flows are expected to route through the IBGP peering between ATE2:Port1 --> DUT:Port2 --> ATE
+	    * Run post test Healthchecks and compare. Ensure there are no drops, coredumps etc.<br><br><br>
+
+
+**PF-1.6.2: Migrate routing of BE1 flows from ATE1:Port1 to ATE2:Port1 to ATE2:Port3 instead**
+   * Situation:
```
robshakir, line 189: In this scenario -- I think it has become too complex defining the prefixes to selectively advertise in each test case. I'd suggest that we break this down in terms of the prefixes that we have in all scenarios, and then the deltas that are being created per test case.

The alternative is to define these prefixes in terms of sets (and the sessions with a name) and say, for example:

> ### Definitions:
> * `FLOW-SET-1`: ATE port 1 to ATE port 2
> * `BGP-SESSION-1`; ATE port 1 to ATE port 2, iBGP with AS100, policy setting local pref to 42 for all routes.
>
> ### Test Case
> Enable FLOW-SET-1 and BGP-SESSION-1.

I think also defining this in terms of the positive (i.e., enable flow 1) rather than the negative (i.e., disable flow 2) is clearer, because the reader can then just understand that they enable the list of flows that you specified. I can imagine writing this like:

```golang

flows := map[string]*someFlowThingy{
  "flow1": {
     src: "ATE-1",
     dst: "ATE-2",
     // some spec of the flow
  },
  "flow2": {
    src: "ATE-3",
    dst: "ATE-4",
  },  
}

func TestSomething(t *testing.T) {
   enableFlows(t, "flow1", "flow2", ...)
}
```

which is pretty easy to read.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+       * The Traffic that gets encaped, Ensure the TTL value of the inner header is decremented and copied to the outer header. Similarly, TOS byte in the inner header is copied to the Outer header.<br>
+         
+   * Expectations:
+       * Routes to Prefixes IPv4Prefix6/24, IPv6Prefix6/64 learnt from ATE2:Port2 will have a Local-Preference of 200 and hence placed in the FIB. Routes to other prefixes from ATE2 will be learnt from the IBGP peering with ATE2:Port1, will continue to have the default local-Preference of 100 and will be placed in the DUT's FIB  
+       * Flows to destination prefixes IPv4Prefix6/24, IPv6Prefix6/64 will be GUE encaped to tunnel destination IPv4Prefix12/28. the flows must be successful.
+           * Ensure that the TTL on the outer heder is decremented and lower than that of the inner header.
+           * TOS byte in the outer header is the same as the TOS value in the inner header.
+           * Ensure the DUT is streaming accurate data on the number of packets/bytes being Encaped.
+	   * ATE2 to ATE1:Port1 Unencaped flows must be 100% successful. The flows are expected to route through the IBGP peering between ATE2:Port1 --> DUT:Port2 --> ATE
```
robshakir, line 206: This is helpful as a clarification -- since it gives you a clear idea of what to do code w.r.t correctness criteria. 

If you don't agree w.r.t specifying the flows per subtest, then I suggest that we just define the flows in terms of src/dst IP and then define the correctness criteria like you have here.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+	   * Run post test Healthchecks and compare. Ensure there are no drops, coredumps etc.<br><br><br>
+
+
+**PF-1.6.3: Migrate routing of AF1-AF4 flows from ATE1:Port1 to ATE2:Port1 to ATE2:Port3 instead**
+   * Follow the same steps as above in PF-1.6.2 and gradually move one Traffic class at a time in the following order
+       * PF-1.6.3: Migrate routing of AF1 flows from ATE1:Port1 to ATE2:Port1 to ATE2:Port3 instead
```
robshakir, line 212: Can we clarify the implementation here? It seems like we are doing this by advertising a particular prefix one at a time to on the session @ ATE2:port3, right?

We should clarify what steps in 1.6.2 we are covering here -- since migrating per AFI isn't defined it's not really clear what to do here (I think you're adding to the iBGP session advertisements, but it's not clearly specified).

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+       * Same as above in PF-1.6.2 for each traffic class migrated.<br><br><br>
+
+**PF-1.6.7: Encap all traffic flows from ATE2 to ATE1:Port1**
+   * Situation:
+       * Test starts from the situation as in PF-1.6.3 above. Therefore all traffic from ATE1 to ATE2 is being encaped in the DUT and routed towards ATE2:Port3
+
+   * Test Steps:
+       * Please run all the health checks suggested above prior to starting the test as a baseline.
+       * Start all flows above:
+           * Dont start the above mentioned Unencaped flows from ATE2:Port1<br>
```
robshakir, line 231: Again, let's define the positive case by having some way to group the flows.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+    
+   * Expectations:
+       * Flows from ATE1 to ATE2 will be GUE encaped to tunnel destination IPv4Prefix12/28 and routed out to ATE2:Port3. The flows must be successful. Ensure that the TTL on the outer heder is decremented and lower than that of the inner header. TOS byte value in the outer header is the same as the TOS value in the inner header.
+       * Ensure the DUT is streaming accurate data on the number of packets/bytes being Encaped.
+       * ATE2:Port3 to ATE1:Port1 Encaped flows with tunnel destination as IPv4Prefix11/28 must be 100% successful. The flows are expected to route through the IBGP peering between ATE2:Port1 --> DUT:Port2 --> ATE
+       * Ensure the DUT is streaming accurate data on the number of packets/bytes being Decaped.
```
robshakir, line 237: Can we define the counter that we expect to use for this?

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+       * Run post test Healthchecks and compare. Ensure there are no drops, coredumps etc.<br><br><br>
+
+
+**PF-1.6.8: [Negative scenario] EBGP route to the tunnel destination "IPv4Prefix12/28" is removed inflight**
+   * Situation:
+       * Test start from the end state of "PF-1.6.7" whereby the DUT is encaping all traffic from ATE1 to ATE2 to tunnel destination address "IPv4Prefix12/28". Also, all traffic in the direction from ATE2 to ATE1 is being encaped to tunnel destination address "IPv4Prefix11/2". ATE2 is not sending any unencaped flows.
+   * Test Steps:
+       * Please run all the health checks suggested above prior to starting the test as a baseline.
+       * Start all flows above:
+           * Dont start the above mentioned Unencaped flows from ATE2:Port1
```
robshakir, line 247: Same comment as above re: negative vs. positive definitions. Will not comment on the remainder, but I'd suggest we make these changes throughout.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+       * Configure the EBGP on ATE2:Port3 to stop advertising "IPv4Prefix12/28" to DUT:Port4. When this happens, following is expected.
+           * The Static route that points the VPNH to the tunnel destination (NHG or tunnel interface) must become invalid. Therefore, all the routes that are programmed by ATE2:Port2 with a VPNH of "IPv4Prefix17" in its IBGP session with DUT:Port3 must become invalid.
+           * Because of the above, all IBGP routes recevied by the DUT:Port2 from ATE2:Port1 with the default local-preference MUST become valid and hence there must not be any traffic loss.<br><br><br>
```
robshakir, line 250:
```suggestion
           * Because of the above, all IBGP routes received by the DUT:Port2 from ATE2:Port1 with the default local-preference MUST become valid and hence there must not be any traffic loss.<br><br><br>
```

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+           * IPv4Prefix8/24 IPv6Prefix8/64
+           * IPv4Prefix9/24 IPv6Prefix9/64
+           * IPv4Prefix10/24 IPv6Prefix10/64
+           * IPv4Prefix15/32, IPv6Prefix15/128   <<--- ATE2:Port2 IBGP address 
+       * DUT:Port2 --> ATE2:Port1
+           * All Intternet Prefixes learnt from ATE1:Port1
+			
+
+   * ATE2:Port2 <-IBGP-> DUT:Port3 over IPv4-unicast and IPv6-Unicast
+       * ATE2:Port2 -> DUT:Port3. Following prefixes for different services advertised with local-preference 200 and a Virtual Protocol Next-Hop (VPNH) of IPv4Prefix17 for IPv4 destination prefixes and a VPNH of IPv6Prefix17 for IPv6 destination prefixes. [Please note: Following advertisements are made gradually as suggested in the subtests below]:
```
robshakir, line 65: For `IPv4Prefix17` here -- is this a /32, or is this actually a prefix?

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+
+**PF-1.6.8: [Negative scenario] EBGP route to the tunnel destination "IPv4Prefix12/28" is removed inflight**
+   * Situation:
+       * Test start from the end state of "PF-1.6.7" whereby the DUT is encaping all traffic from ATE1 to ATE2 to tunnel destination address "IPv4Prefix12/28". Also, all traffic in the direction from ATE2 to ATE1 is being encaped to tunnel destination address "IPv4Prefix11/2". ATE2 is not sending any unencaped flows.
+   * Test Steps:
+       * Please run all the health checks suggested above prior to starting the test as a baseline.
+       * Start all flows above:
+           * Dont start the above mentioned Unencaped flows from ATE2:Port1
+       * Configure the EBGP on ATE2:Port3 to stop advertising "IPv4Prefix12/28" to DUT:Port4. When this happens, following is expected.
+           * The Static route that points the VPNH to the tunnel destination (NHG or tunnel interface) must become invalid. Therefore, all the routes that are programmed by ATE2:Port2 with a VPNH of "IPv4Prefix17" in its IBGP session with DUT:Port3 must become invalid.
```
robshakir, line 249: It'd be useful in these test cases to define the configuration -- the static route here is specifically:

> static
>   dst: `IPv4Prefix17` 
>   nhop: `IPv4Prefix12`   # what address is actually the NH? this is a /28

so `IPv4Prefix12/28` is withdrawn, and `IPv4Prefix17` (which I assume is a /32) which is the VPNH becomes invalid, and hence at that point prefixes for whom the next-hop is set to `IPv4Prefix17` become invalid.

Given this scenario, you use the alternate routing.

It seems worth clarifying this because it's very easy to get lost in what the scenario here is IMHO.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+           * Ensure the IBGP sessions says established during the migration.
+           * Ensure the DUT is streaming accurate data on the number of packets/bytes being Encaped and this is the same as the packets being sent from ATE1.<br><br><br>
+
+
+
+## OpenConfig Path and RPC Coverage
+```yaml
+paths:
+  # Policy for route leaking between VRFs
+  /network-instances/network-instance/inter-instance-policies/apply-policy
```
robshakir, line 273: I think this isn't needed, there are no VRFs mentioned?

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+  # Following applied in Default VRF
+  /network-instances/network-instance/inter-instance-policies/apply-policy/config/export-policy
+  # Following applied in Non-default VRF
+  /network-instances/network-instance/inter-instance-policies/apply-policy/config/import-policy
+
+  # Config for VRF selection
```
robshakir, line 279: No VRF selection policies mentioned in this README AFAICS?

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+  # Action that sets the VRF for matching conditions.
+  /network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/action/config/network-instance:
+
+  # Creating the Prefix-Set required for the above VRF selection policy
+  /routing-policy/defined-sets/prefix-sets/prefix-set/name:
+  /routing-policy/defined-sets/prefix-sets/prefix-set/config/name:
+  /routing-policy/defined-sets/prefix-sets/prefix-set/prefixes/prefix/config/ip-prefix:
+  /routing-policy/defined-sets/prefix-sets/prefix-set/prefixes/prefix/config/masklength-range:
+
+  # Apply VRF selection policy as per the Test procedure above
```
robshakir, line 297: Assume this whole section around VRF selection is outdated.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+  gnmi:
+    gNMI.Set:
+      union_replace: true
+    gNMI.Subscribe:
+      on_change: true
+```
+
+## Required DUT platform
+
+* Specify the minimum DUT-type:
```
robshakir, line 380: Again, I think this is not the minimum -- must be MFF since MFF > FFF if you need MFF, but I'm not sure why this can't be executed against FFF.

robshakir (changes requested):
I would suggest that this test case needs a diagram that explains where the routing is attached -- this can be added as an image that is then embedded in the README. 

I think it needs to show:

 * each group of prefixes that are advertised
 * where the BGP sessions are between -- interface vs. loopback addresses
 * where the IS-IS adjacencies are

It's quite hard to follow what is happening in each stage without a clearer topological idea of the scenario. 

I think for these combinatorial tests we have to be much clearer about the way the test setup should look.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
+# PF-1.6 Static GUE Encap and BGP path selection
+
+## Summary
```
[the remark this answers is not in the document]
sachendras: ack, thanks!

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
+# PF-1.6 Static GUE Encap and BGP path selection
+
```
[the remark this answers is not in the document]
sachendras: Done, thanks

dplore:
@sachendras please respond to and resolve open comments. Thanks!

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+           * IPv4Prefix3/24 IPv6Prefix3/64
+           * IPv4Prefix4/24 IPv6Prefix4/64
+           * IPv4Prefix5/24 IPv6Prefix5/64
+
+       * DUT:Port1 --> ATE1:Port1
+           * All prefixes for different services learnt from either ATE2:Port1 or ATE2:Port2 depending on the one that is placed in the FIB. 
+	
+   * ATE2:Port1 <-IS-IS(L2)-> DUT:Port2
+       * ATE2:Port1 --> DUT:Port2
+           * IPv4Prefix13, IPv6Prefix13 <<--- ATE2:Port1 IBGP address
```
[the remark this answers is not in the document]
sachendras: ack on that will address it

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+
+   * ATE2:Port2 <-IBGP-> DUT:Port3 over IPv4-unicast and IPv6-Unicast
+       * ATE2:Port2 -> DUT:Port3. Following prefixes for different services advertised with local-preference 200 and a Virtual Protocol Next-Hop (VPNH) of IPv4Prefix17 for IPv4 destination prefixes and a VPNH of IPv6Prefix17 for IPv6 destination prefixes. [Please note: Following advertisements are made gradually as suggested in the subtests below]:
+           * IPv4Prefix6/24 IPv6Prefix6/64 <--- Advertised in PF-1.6.2 to PF-1.6.9
+           *  IPv4Prefix7/24 IPv6Prefix7/64 <--- Advertised in PF-1.6.3 to PF-1.6.9
+           * IPv4Prefix8/24 IPv6Prefix8/64 <--- Advertised in PF-1.6.4 to PF-1.6.9
+           * IPv4Prefix9/24 IPv6Prefix9/64 <--- Advertised in PF-1.6.5 to PF-1.6.9
+           * IPv4Prefix10/24 IPv6Prefix10/64 <--- Advertised in PF-1.6.6 to PF-1.6.9
+       * DUT:Port3 --> ATE2:Port2
+           * All Intternet Prefixes learnt from ATE1:Port1
```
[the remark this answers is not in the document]
sachendras: ack

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+   * ATE2:Port3 <-EBGP-> DUT:Port4 over IPv4-unicast and IPv6-Unicast
+       * ATE2:Port3 --> DUT:Port4
+           * IPv4Prefix12/28 <<-- Edge Encap destination
+       * DUT:Port4 --> ATE2:Port3
+           * IPv4Prefix11/28
+
+### DUT configuration:
+   * EBGP peering between DUT:Port1 and ATE1:Port1
+   * IS-IS adjancency between DUT:Port2 and ATE:Port1
+       * Loopback interface is passive.
```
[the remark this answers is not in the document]
sachendras: ack

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+   * IBGP peering between DUT:Por2 and ATE2:Port1
+   * IBGP peering between DUT:Port3 and ATE2:Port2
+   * EBGP peering between DUT:Port4 and ATE2:Port3
+   * Import and export route policies to match the advertisements for each BGP peering.
```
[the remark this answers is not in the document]
sachendras: I will add some clarification

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: I think that was an error, will correct it

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,584 @@
...
+            *   The DUT must have a static route pointing `"$ATE2_PPNH1.v6/128"`
+                to the NHG created above. The IBGP peer $ATE2_C.IBGP.v6/128 is
+                expected to advertise both IPv4 and IPv6 prefixes with the
+                Next-Hop as `"$ATE2_PPNH1.v6/128"`
+
+*   **GUE Decapsulation:**
+
+    *   For a GUE decapsulation node, configure the following:
+        *   UDP port 6080 must be used for decapsulating IPv4 payload, and UDP
+            port 6615 for IPv6 payload
```
masood-shah, line 170: @sachendras  I think (today) the decision was/is to use a single, designated UDP destination port for gIRO's IP/UDP encap/decap, accommodating both inner IPv4 and IPv6 payloads. This same UDP port indicates the inner payload is IP; and the first nibble of that inner payload will then be used to differentiate between IPv4 and IPv6. The decapsulated packet will subsequently be forwarded based on its inner IP. Should we proceed with updating the documentation to reflect this?

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,584 @@
...
+*   **BGP:**
+
+    *   Define import and export route policies to match the advertisements for
+        each BGP peering.
+
+*   **Static GUE Encapsulation:**
+
+    *   Configure static GUE encapsulation as follows:
+        *   Define UDP ports to be used for IPv4oUDP and IPv6oUDP as 6080 and
+            6615 respectively.
```
masood-shah, line 149: Ditto as above

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,584 @@
...
+.
+
+## Summary
+
+The goal of this test is to:
+
+1.  Test the implementation of Static GUE encap where the Tunnel endpoint is
+    resolved over EBGP, while the Payload's destination is learned over IBGP.
+
+2.  Verify that before GUE encapsulation, the LPM lookup on the payload
```
masood-shah, line 12: We can omit 'LPM' here and simplify the text to focus on the intended result...For example: "Confirm that prior to encapsulation, the DUT selects the correct path to the payload destination by prioritizing the IBGP route with the highest Local Preference, and successfully utilizes backup IBGP routes if the primary path fails."

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,584 @@
...
+    destination performs route selection between different IBGP-learned routes
+    and chooses those with a higher Local Preference, falling back to backup
+    routes if necessary.
+
+3.  Validate that encapsulated traffic has its TOS bits copied from the inner
+    header to the outer header, and that depending on the tunnel destination IP,
+    the implementation marks the TZ bits (LSB bit \#1 and 2 of the DSCP header)
+    as either TZ=11 or TZ=10.
+
+4.  Ensure that the implementation allows configuration of the TTL value for the
```
masood-shah, line 22: Refine it slightly to enhance clarity...such as..."Confirm that the TTL value for the outer IP header created during GUE encapsulation can be explicitly configured."

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: Corrected, ptal

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: Updated, ptal

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: Yes, updated, ptal

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: I can correct that. This says the old intent since I wasnt clear on support

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+       * IPv6Prefix7/64 to IPv6Prefix2/64 [DSCP:AF1]
+       * IPv4Prefix8/24 to IPv4Prefix3/24 [DSCP:AF2]
+       * IPv6Prefix8/64 to IPv6Prefix3/64 [DSCP:AF2]
+       * IPv4Prefix9/24 to IPv4Prefix4/24 [DSCP:AF3]
+       * IPv6Prefix9/64 to IPv6Prefix4/64 [DSCP:AF3]
+       * IPv4Prefix10/24 to IPv4Prefix5/24 [DSCP:AF4]
+       * IPv6Prefix10/64 to IPv6Prefix5/64 [DSCP:AF4]
+<br><br><br>
+
+### Following Health checks to be run Pre and Post every sub test
```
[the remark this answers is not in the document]
sachendras: This is a placeholder for the list of different pre and post tests. In each of the tests below, I have called this placeholder pre and post every test to ensure the test didnt result in any of the errors/drops. Let me know if my understanding is incorrect here.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
masood-shah: Since we've decided this is the way forward, let's change it and start working with vendors to get it codified

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,382 @@
...
+
+A1[ATE1:Port1] <-- IBGP(ASN100) --> B1; 
+B2 <-- IBGP(ASN100) --> C1; 
+B3 <-- IBGP(ASN100) --> C2;
+B4 <-- EBGP(ASN100:ASN200) --> C3;
+```
+
+### Advertisements:
+
+   * ATE1 <-EBGP-> DUT:Port1 over IPv4-unicast and IPv6-Unicast
```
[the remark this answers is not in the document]
sachendras: ok, I will address these

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,616 @@
...
+.
+
+## Summary
+
+The goal of this test is to:
+
+1.  Test the implementation of Static GUE encap where the Tunnel endpoint is
+    resolved over EBGP, while the Payload's destination is learned over IBGP.
+
+2.  Verify that before GUE encapsulation, the LPM lookup on the payload
```
masood-shah, line 12: @sachendras suggesting a slight rewrite, removing LPM, something like this..."Confirm that before GUE encapsulation, the device correctly selects the path for the payload destination from multiple available IBGP routes. Path selection needs to follow the BGP best path algorithm rules, like giving preference to routes with a higher BGP Local Preference. Furthermore, it must successfully switch over to backup/alternative IBGP paths when the preferred path fails"

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: I have addressed both the requirements. Ptal

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: Done

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,627 @@
...
+| `$ATE2_INTERNAL_TE10.v4/32`                                                                                | IPoUDP tunnel destination address on the DUT. This IP MUST receive traffic with with TE bits marked as 10                                                                                                                                                                                                                                 |
+| <br>`$ATE2_INTERNAL6.v4/24`<br>`$ATE2_INTERNAL6.v6/64`<br>`$ATE2_INTERNAL7.v4/24`<br>`$ATE2_INTERNAL7.v6/24`<br>`$ATE2_INTERNAL8.v4/24`<br>`$ATE2_INTERNAL8.v6/64`<br>`$ATE2_INTERNAL9.v4/24`<br>`$ATE2_INTERNAL9.v6/24`<br>`$ATE2_INTERNAL10.v4/24`<br>`$ATE2_INTERNAL10.v6/64` | <br>- Internal Public prefixes<br>- Advertised to the DUT over the IBGP peering between `ATE2_PORT1.IBGP.v[46]`<>`$DUT_lo0.v[46]` and `$ATE2_PORT2.IBGP.v6`<>`$DUT_lo0.v6`<br>- Advertised further to `$ATE1_IBGP.v[46]` over the IBGP peering between `$ATE1_IBGP.v[46]`<>`DUT_lo0.v[46]` over their respective AFI peering |
```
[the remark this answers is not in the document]
sachendras: That was stale info, corrected now. Thanks for spotting

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
dplore, line 1: IMO, this test could be broken up into several, smaller, separate tests:

1. static route to BGP NH with IP in GUE encapsulation 
1. static route to BGP NH with IP in GUE decapsulation
1. qos for static route to BGP NH with IP in GUE encap. (depends on 1.)
1. qos for static route to BGP NH with IP in GUE decap (depends on 2.)
1. scale qos for static route to BGP NH with IP in GUE encap (depends on 3.)
2. 1. scale qos for static route to BGP NH with IP in GUE decap (depends on 4.)

We used this pattern of building up a base encap test and then layering on qos and scale  for the policy-forwarding tests for MPLS in GUE encap/decap with success

sachendras: Is the ask to have separate FNTs for each?
If so, there are FNTs already in review that are covering smaller portions that you mentioned. However here I want multidimensional test as the title suggests.

nupkanoi (approved):

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,635 @@
...
+    header to the outer header, and that depending on the tunnel destination IP,
+    the implementation marks the TZ bits (LSB bit \#1 and 2 of the DSCP header)
+    as either TZ=11 or TZ=10.
+
+4.  Confirm that the TTL value for the outer IP header created during GUE
+    encapsulation can be explicitly configured.
+
+5.  Confirm that the DUT performs GUEv1 Decap of traffic received in the reverse
+    direction, and that during decapsulation, the outer DSCP and TTL bits are
+    not copied to the inner header.
```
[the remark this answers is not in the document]
sachendras: Why is this being called out now? You did the review of the PR last year too https://github.com/openconfig/featureprofiles/pull/3486#issuecomment-2557882672

Also, the requirement here is to have Multidimensional test. So I prefer to have the test multidimensional. Single dimension tests are expected to be covered in other Pulls.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,635 @@
...
+| | $ATE2_INTERNAL9.v6/64 --> $ATE1_PORT1_INTERNET4.v6/64 | AF3 | |
+| | $ATE2_INTERNAL10.v4/24 -->  $ATE1_PORT1_INTERNET5.v4/24 | AF4 | |
+| | $ATE2_INTERNAL10.v6/64 -->  $ATE1_PORT1_INTERNET5.v6/64 | AF4 | |
+
+
+### DUT Configuration:
+
+In addition to the adjacencies and peering configurations described in the
+tables above, the DUT requires the following configurations:
+
```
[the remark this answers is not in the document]
dplore: here's another example where one test refers to a second test to "import"  or re-use the configuration.  

https://github.com/openconfig/featureprofiles/blob/fccde73b7301161320faf1afd431d26b8976b531/feature/policy_forwarding/otg_tests/mpls_gre_ipv4_encap_scale_test/README.md?plain=1#L29

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,635 @@
...
+        routed over the connection between `DUT_Port2` <> `ATE2_Port1`. Since no
+        encapsulation is performed on the DUT in this scenario, the TZ bits are
+        expected to be `00`.
+    *   Flow-Set #5 must also achieve 100% success, with flows routing through
+        the connection between `ATE2_Port1` --> `DUT_Port2` --> `ATE1_Port1`.
+    *   Execute post-test health checks and compare the results with the
+        baseline. Verify that there are no drops, core dumps, or other issues.
+
+
+**RT-3.52.2: BE1 Traffic Migrated from being routed over the DUT_Port2 -->
```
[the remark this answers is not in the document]
sachendras: Sure, but this one I wanted a multidimensional test covering functionality of a certain set of features in tandem.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,635 @@
...
+*   **Expectations:**
+
+    *   Same as RT-3.52.2 for each traffic class migrated, except for tunnel
+        destination `$ATE2_INTERNAL_TE10.v4/32`, where TZ bits will be marked
+        as 10.
+
+
+        
+
+**RT-3.52.7: DUT as a GUE Decap Node**
```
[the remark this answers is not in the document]
sachendras: I have several smaller tests for each feature in isolation. As mentioned earlier, I want to keep this one as a multidimensional test.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,635 @@
...
+| ASN2                                 |                      | `$DUT_Port2`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |
+
+**Different Flows used throughout the test:**
+You are correct, the previous markdown table was inaccurate. Here's the rebuilt markdown version:
```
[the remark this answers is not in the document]
sachendras: yep, corrected

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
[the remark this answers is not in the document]
sachendras: The canonical JSON is added here so I dont see a need for reference. Resolving this comment.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
masood-shah: ty!

masood-shah (commented):
LGTM

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,635 @@
...
+        `$ATE2_INTERNAL_TE11.v4/32` as the Next-hop and the latter will point at
+        `$ATE2_INTERNAL_TE10.v4/32` as the Next-hop.
+    *   Initiate Flow-Set #1 and Flow-Set #2
+    *   Initiate Flow-Set #5
+
+*   **Expectations:**
+
+    *   Flow-Set #1 and Flow-Set #2 are expected to achieve 100% success and be
+        routed over the connection between `DUT_Port2` <> `ATE2_Port1`. Since no
+        encapsulation is performed on the DUT in this scenario, the TZ bits are
```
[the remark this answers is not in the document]
dplore: You'll need to define tz bits as this is not standard

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,635 @@
...
+        established.
+    *   No prefixes are exchanged over the IBGP peering between
+        `$ATE2_C.IBGP.v6` and `$DUT_lo0.v6`.
+
+*   **Test Steps:**
+
+    *   Execute all health checks mentioned above prior to running the tests.
+    *   Configure the DUT as described previously for IS-IS adjacency, IBGP and
+        EBGP peering and the corresponding routing policies, static routes for
+        GUE encapsulation, and policies to selectively apply Traffic Class (TZ)
```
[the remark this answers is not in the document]
dplore: Thanks!

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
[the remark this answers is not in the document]
dplore: Oh I see an explanation was added

sachendras:
ack, added a diagram as you suggested and also embedded it in to the readme. While I have all the protocols, adjacencies and the IP addresses used for the peering detailed in the diagram, I have also added multiple tables to the test explaining what addresses are advertised by the different peers, Ptal

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: I have revamped the test significantly to improve readability. This comment may be stale at this point so resolving. Please feel free to re-open.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: Address this concern. Ptal at the defined tables. Hopefully that addresses the concern here.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: I added flow-sets so resolving the comment

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: lot of the comments may be stale at this point given the significant changes to the Readme so far.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: Marking this resolved since the flow-sets are defined now.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,706 @@
...
+|                      ATE2 (ASN 100 & 200)           |  |          |
+|-----------------------------------------------------+  |          |
+| Port1: (to DUT)                                     <--+          |
+|  - $ATE2PORT1_DUTPORT2_1.v4/31                      |             |
+|  - $ATE2PORT1_DUTPORT2_1.v6/127                     |             |
+|-----------------------------------------------------+             |
+| Port3: (to DUT)                                     <-------------+
+|  - `$ATE2PORT3_DUTPORT4_1.v4/31`                    |
+|  - `$ATE2PORT3_DUTPORT4_1.v6/127`                   |
+|-----------------------------------------------------+
```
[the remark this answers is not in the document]
sachendras: I added a statement that ATE1 and ATE2 can be the same device.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,709 @@
...
+        Flow-Set #1, Flow-Set #2, and Flow-Set #5
+    *   The DUT receives `$ATE2_INTERNAL_TE11.v4/30` and
+        `$ATE2_INTERNAL_TE10.v4/30` from `$ATE2_Port3`, while advertising
+        `$DUT_TE11.v4/32`, `$DUT_TE10.v4/32`, `$DUT_TE11.v4/30` and
+        `$DUT_TE10.v4/30` to `$ATE2_Port3` via EBGP
+
+*   **Test Steps:**
+
+    *   Execute the health checks described previously
+    *   Configure the DUT according to the prior instructions
```
dplore, line 298: Has the configuration changed?   The DUT config defined in the setup was pushed in 3.52.1 already.  Do you want to change the DUT config in some way and push it again?

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,709 @@
...
+        (Zero loss), routing via the IBGP peering between `$ATE2_IBGP.v[46]` and
+        `$DUT_lo0.v[46]`
+    *   Post-test health checks should be performed and compared against the
+        baseline. Verify the absence of drops or core dumps. If any, the test
+        Must fail
+
+**RT-3.52.3: AF1-AF4 Traffic Migrated to DUT --> ATE2_Port3**
+
+Follow the same steps as in RT-3.52.2 and gradually move one Traffic class at a
+time in the following order. Note changes in RT-3.52.5 and RT-3.52.6
```
dplore, line 338: specify the config needed to move traffic

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,709 @@
...
+    *   Stop Flow-Set #5 and start Flow-Set #4, resulting in active flows for
+        Flow-Set #1 through Flow-Set #4.
+
+*   **Expectations:**
+
+    *   Traffic from ATE1 to ATE2 should be GUE encapsulated with tunnel
+        destinations `$ATE2_INTERNAL_TE11.v4/32` and `$ATE2_INTERNAL_TE10.v4/32`
+        and routed out `$DUT_Port4<>$ATE2_Port3`.
+
+    *   The DUT should provide accurate streaming data for the number of
```
dplore, line 393: specify OC paths needed for this (here and in the OC Paths at the yaml stanza at end of this README)

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,709 @@
...
+    *   `ATE2_Port3` sends encapsulated flows (Flow-Set #3 and Flow-Set #4) to
+        `ATE1_Port1` through the DUT.
+
+        BE1 to AF2 flows are expected to have a tunnel destination of
+        `$DUT_TE11.v4/32`, while AF3 and AF4 flows should have `$DUT_TE10.v4/32`
+        as their tunnel destination.
+
+        Traffic should reach the destination successfully.
+
+    *   The DUT should accurately stream data on decapsulated packets/bytes.
```
dplore, line 405: specify OC paths needed for this (here and in the OC Paths at the yaml stanza at end of this README)

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,709 @@
...
+        `DUT_Port2` --> `ATE1_Port1`.
+    *   Execute post-test health checks and compare the results with the
+        baseline. Verify that there are no drops, core dumps, or other issues.
+
+**RT-3.52.2: BE1 Traffic Migrated from being routed over the DUT_Port2 -->
+ATE2_Port1 path to DUT_Port4 --> ATE2_Port3 path**
+
+*   **Situation:**
+
+    *   The baseline test (RT-3.52.1) is running, with flows active in
```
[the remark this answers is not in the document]
sachendras: Addressed, asked to restart the flows.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: no, I updated.. it is the same as the baseline config,

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,709 @@
...
+        `$ATE2_Port3` and `$DUT_Port4`, and these flows must be 100% successful
+        (zero loss). Inshort, change in path due to higher local preference on
+        the received routes from `$ATE2_C.IBGP.v6/128` should cause zero packet
+        loss
+    *   The outer header TTL should be 127 upon arrival at `ATE2_Port1` (before
+        decapsulation)
+    *   The outer header DSCP bits should be the same as the inner header DSCP
+        bits when received at `ATE2_Port1` (before decapsulation).
+    *   The DUT should accurately stream data regarding the number of
+        packets/bytes encapsulated
```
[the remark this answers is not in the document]
sachendras: Added the paths and also clarified which parts to be validated in ATE v/s which parts in OC

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: Done

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,743 @@
...
+| IBGP Peer:                                |
+|  - $ATE1_IBGP.v4/32                       |
+|  - $ATE1_IBGP.v6/128                      |
+|                                           |
+| Prefixes Advertised:                      |
+|  - $ATE1_PORT1_INTERNET[1-5].v4/24        |
+|  - $ATE1_PORT1_INTERNET[1-5].v6/64        |
+|                                           |
+| ATE1_Port1 <-> DUT_Port1:                 |
+|  - $ATE1PORT1_DUTPORT1_1.v4/31            |
```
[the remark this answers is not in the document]
sachendras: Well, intention is to have 
$ATE1PORT1_DUTPORT1/31 as the prefix in which

$ATE1PORT1_DUTPORT1_1/32 is the ATE IP and $ATE1PORT1_DUTPORT1_2/32 is the DUT IP.

Which is why in the diagram, the port config on the ATE and the DUT sides carry their respective IPs from the same subnet.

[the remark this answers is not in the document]
ram-mac: ACK

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,743 @@
...
+| Flow-Set#4 from ATE2_Port3 --> ATE1_Port1 are GUE encaped with Tunnel destination as $DUT_TE10.v4/32 | `$ATE2_INTERNAL9.v4/24 --> $ATE1_PORT1_INTERNET4.v4/24`  | AF3  | `$DUT_TE10.v4/30`           |
+| Flow-Set#4 from ATE2_Port3 --> ATE1_Port1 are GUE encaped with Tunnel destination as $DUT_TE10.v4/32 | `$ATE2_INTERNAL9.v6/64 --> $ATE1_PORT1_INTERNET4.v6/64`  | AF3  | `$DUT_TE10.v4/30`           |
+| Flow-Set#4 from ATE2_Port3 --> ATE1_Port1 are GUE encaped with Tunnel destination as $DUT_TE10.v4/32 | `$ATE2_INTERNAL10.v4/24 --> $ATE1_PORT1_INTERNET5.v4/24` | AF4  | `$DUT_TE10.v4/30`           |
+| Flow-Set#4 from ATE2_Port3 --> ATE1_Port1 are GUE encaped with Tunnel destination as $DUT_TE10.v4/32 | `$ATE2_INTERNAL10.v6/64 --> $ATE1_PORT1_INTERNET5.v6/64` | AF4  | `$DUT_TE10.v4/30`           |
+| Flow-Set#5 from ATE2:Port1 to ATE1:Port1 are sent Unencaped                                          | `$ATE2_INTERNAL6.v4/24 --> $ATE1_PORT1_INTERNET1.v4/24`  | BE1  | N/A                         |
```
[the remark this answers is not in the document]
sachendras: ack, updated all the internet prefixes with user prefixes. Hopefully that helps?

So, the following,
- $ATE1_PORT1_INTERNET[1-5].v4/24
- $ATE1_PORT1_INTERNET[1-5].v6/64 

will change to:
- $ATE1_PORT1_user[1-5].v4/24
- $ATE1_PORT1_user[1-5].v6/64

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,743 @@
...
+`$ATE2_Port3<>$DUT_Port4`                    | EBGP                 | <br>- `ATE2_Port3` advertises `$ATE2_INTERNAL_TE10.v4/30` and `$ATE2_INTERNAL_TE11.v4/30` to `DUT_Port4`<br>- `DUT_Port4` advertises `$DUT_TE10.v4/32`, `$DUT_TE11.v4/32`, `$DUT_TE10.v4/30` and `$DUT_TE11.v4/30` to `ATE2_Port3`
+Local ASN as ASN100 for following interfaces |                      | `$ATE1_IBGP.v[46]`, `$DUT_lo0.v[46]`, `$ATE2_IBGP.v[46]`, `$DUT_lo0.v[46]`, `$ATE2_C.IBGP.v6`, `$ATE2_M.IBGP.v[46]`, `$DUT_lo0.v[46]`, `$DUT_Port2`
+Local ASN as ASN200 for following interfaces |                      | `$ATE2_Port3`
+
+**Different Flows used throughout the test:**
+
+| Src_destination of flows | From_IP --> To_IP | DSCP | Tunnel endpoint used |
+| :------------------------------------------------------------------------------------------------- | :------------------------------------------------------ | :--- | :-------------------------- |
+| Flow-Set#1 from ATE1_Port1 --> ATE2_[Either Port1 or Port3 depending on the FIB entries of the DUT]  | `$ATE1_PORT1_INTERNET1.v4/24 --> $ATE2_INTERNAL6.v4/24`  | BE1  | `ATE2_INTERNAL_TE11.v4/32`  |
```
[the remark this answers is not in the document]
sachendras: Not sure I follow, below is from a snippet of the `All IP addresses for the test:`  table. Also, I have the same captured under the "Topology" section plus the IBGP table for advertisements. Are you referring to certain section of the doc?

<br>`$ATE2_INTERNAL6.v4/24`<br>`$ATE2_INTERNAL6.v6/64`<br>`$ATE2_INTERNAL7.v4/24`<br>`$ATE2_INTERNAL7.v6/64`<br>`$ATE2_INTERNAL8.v4/24`<br>`$ATE2_INTERNAL8.v6/64`<br>`$ATE2_INTERNAL9.v4/24`<br>`$ATE2_INTERNAL9.v6/64`<br>`$ATE2_INTERNAL10.v4/24`<br>`$ATE2_INTERNAL10.v6/64`                                                           | <br>- Internal Public prefixes<br>- Advertised to the DUT over the IBGP peering between `ATE2_PORT1.IBGP.v[46]`<>`$DUT_lo0.v[46]` and `$ATE2_C.IBGP.v6`<>`$DUT_lo0.v6`<br>- Advertised further to `$ATE1_IBGP.v[46]` over the IBGP peering between `$ATE1_IBGP.v[46]`<>`DUT_lo0.v[46]` over their respective AFI peering

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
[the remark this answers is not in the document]
sachendras: They aren't, but 2 different IP address of the same /31 prefix. _1 represents ATE side of the IP whereas _2 is DUT side of the IP. Following from the All_IPaddresses table above.

<br>`$ATE1PORT1_DUTPORT1_1.v4/31`<br>`$ATE1PORT1_DUTPORT1_1.v6/127`                                                                                                                                                                                                                                                                        | Configured on the ATE1PORT1 port of ATE1_Port1<>DUT_Port1 point-to-point connection and used for IS-IS adjacency.
<br>`$ATE1PORT1_DUTPORT1_2.v4/31`<br>`$ATE1PORT1_DUTPORT1_2.v6/127`                                                                                                                                                                                                                                                                        | Configured on the DUTPORT1 port of ATE1_Port1<>DUT_Port1 point-to-point connection and used for IS-IS adjacency.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,743 @@
...
+|  - $ATE1PORT1_DUTPORT1_2.v6/127           |
+|-------------------------------------------|
+|                DUT (ASN-100)              |
+|-------------------------------------------|
+| Loopback:                                 |
+|  - $DUT_lo0.v4                            |
+|  - $DUT_lo0.v6                            |
+|-------------------------------------------|
+| Port2: (to ATE2_Port1)                    <------------+ (IS-IS/IBGP)
+|  - $ATE2PORT1_DUTPORT2_2.v4/31            |            |
```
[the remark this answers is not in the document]
sachendras: It is not OR, I corrected that to "and". Hopefully it helps now? Basically IS-IS provides reachability for IBGP peering.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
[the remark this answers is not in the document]
sachendras: I have addressed this requirement and added tables too for all the IPs used in the test

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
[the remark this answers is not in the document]
sachendras: Updated this in the test

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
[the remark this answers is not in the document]
sachendras: The  ambiguity doesnt exist anymore as the test is rewritten with added details and table references. The new "BGP Peering Details" table explicitly defines which prefixes are advertised from the ATE side in each session

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: This comment is no longer valid, so resolving. The test is rewritten since the last review

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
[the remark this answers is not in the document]
ram-mac: Thanks; This helps

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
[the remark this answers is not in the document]
ram-mac: Yes, i thought the Internal prefixes are reference from other documents internally. But now it is clear that it is referred to as internally to the README itself.

ram-mac (changes requested):
@sachendras - There are some issues with OC paths in the README which is why it is failing the ci/cd checks. Can you fix them?

ram-mac (approved):
LGTM

robshakir (approved):
Approving based on @rohit-rp and @ram—mac’s reviews.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,745 @@
...
+*   Migrate routing of AF4 flows from `DUT_Port2` --> `ATE2_Port1` to
+`DUT_Port4` --> `ATE2_Port3`.
+
+*   `$ATE2_C.IBGP.v6` will advertise `$ATE2_INTERNAL10.v4/24` and
+    `$ATE2_INTERNAL10.v6/64` with next-hop as `$ATE2_PPNH2.v6/128`
+    Traffic towards `$ATE2_INTERNAL10.v[46]` will have tunnel
+    destination `$ATE2_INTERNAL_TE10.v4/32`.
+*   BE1-AF4 are now migrated
+*   Complete validation same as RT-3.52.2 above
+```
```
dplore, line 366: Need to remove this to get the README to validate

```suggestion
```

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,745 @@
...
+
+## RT-3.52.1: Baseline
+
+*   **Test Steps:**
+
+    *   Generate DUT Configuration as specified in the test setup
+    *   Use gnmi to push the configuration to the DUT
+    *   Ensure no prefixes are exchanged over the IBGP peering between
+        `$ATE2_C.IBGP.v6` and `$DUT_lo0.v6`. Validate this using OC
+        ```
```
dplore, line 262: The use of the triple backtick has to be on the 1st column, otherwise you get unpredictable results.  Notice how much of the remainder (but not all) is treated as static text.  

Convert all the triple back ticks to start and end on the 1st column and check the mdformat results to make sure this looks as intended.   

Right now major sections of the doc are getting excluded as static /codeblock text which is causing validation to fail for ocpaths and canonical_oc

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,745 @@
...
+        /network-instances/network-instance/afts/ipv4-unicast/ipv4-entry/state/prefix
+
+        /network-instances/network-instance/afts/ipv6-unicast/ipv4-entry/state/counters/octets-forwarded
+        /network-instances/network-instance/afts/ipv6-unicast/ipv4-entry/state/counters/packets-forwarded
+        /network-instances/network-instance/afts/ipv6-unicast/ipv4-entry/state/next-hop-group
+        /network-instances/network-instance/afts/ipv6-unicast/ipv4-entry/state/origin-protocol
+        /network-instances/network-instance/afts/ipv6-unicast/ipv4-entry/state/prefix
+        ```
+
+## Canonical OpenConfig for GUEv1 Encapsulation configuration
```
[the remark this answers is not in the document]
sachendras: Done

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
sachendras: Done

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,743 @@
...
+    *   Routes to prefixes `$ATE2_INTERNAL6.v4/24` and `$ATE2_INTERNAL6.v6/64`,
+        learned from `$ATE2_C.IBGP.v6/128`, should be placed in the FIB. Other
+        prefixes from ATE2 will continue to be learnt via the IBGP peering
+        between `$ATE2_PORT1.IBGP.v[46]` and `$DUT_lo0.v[46]` and hence will be
+        in the DUT's FIB.
+    *   Flows destined for `$ATE2_INTERNAL6.v4/24` and `$ATE2_INTERNAL6.v6/64`
+        should be GUE-encapsulated with tunnel destination
+        `$ATE2_INTERNAL_TE11.v4` and routed over the EBGP peering between
+        `$ATE2_Port3` and `$DUT_Port4`, and these flows must be 100% successful
+        (zero loss).
```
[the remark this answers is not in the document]
sachendras: done

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,743 @@
...
+
+Follow the same steps as in RT-3.52.2 and gradually move one Traffic class at a
+time in the following order. Note changes in RT-3.52.5 and RT-3.52.6
+
+### RT-3.52.3: Migrate AF1 Flows
+
+```
+*   Migrate routing of AF1 flows from `DUT_Port2` --> `ATE2_Port1` to
+`DUT_Port4` --> `ATE2_Port3`.
+*   BE1 and AF1 are now migrated
```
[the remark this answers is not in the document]
sachendras: This is already provided in context when you read through the test in its entirety on how to do the migration.

feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md
```diff
@@ -0,0 +1,743 @@
...
+        as their tunnel destination. Traffic should reach the destination
+        successfully with zero loss.
+
+    *   Post-test health checks should be performed and compared against the
+        baseline. Verify the absence of drops or core dumps. If any, the test
+        Must fail
+
+### RT-3.52.8: Negative Scenario - EBGP Route for remote tunnel endpoints Removed
+
+Inflight
```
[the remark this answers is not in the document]
sachendras: it is the continuation of the heading. Looks like got cut-off due to the 80 character limit. I corrected. Thanks!

singhavnish2516 (approved):

PR #3492, opened 2024-10-02
Update testregistry.textproto

sachendras:
New test PF-1.6 for Multidimensional Static GUE encap/Decap w/o any VRFs

### DIFF
--- a/testregistry.textproto
+++ b/testregistry.textproto
@@ -409,6 +409,12 @@ test: {
   exec: " "
 }
 test: {
+  id: "PF-1.6"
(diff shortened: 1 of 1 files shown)


testregistry.textproto
```diff
@@ -408,6 +408,12 @@ test: {
   readme: " "
```
dplore, line 408: Probably PF-1.5 should be a TUN-*.* ID test because it sounds like from the description that it is based on /interfaces/interface/tunnel configuration

dplore (approved):
consider updating ID's as suggested

PR #3494, opened 2024-10-02
Update testregistry.textproto

vvlakshmanamurthy:
Adding RT-3.33 DSCP based traffic steering from Non-default VRF to Default VRF using Policy based VRF selection


testregistry.textproto
```diff
@@ -759,6 +759,12 @@ test: {
   readme: ""
   exec: " "
 }
+test: {
+  id: "RT-3.33"
```
dplore, line 763: Recommend PF-1.6. (if this test is to use policy-forwarding (pbr, filter based forwarding)

vvlakshmanamurthy (commented):
updated

testregistry.textproto
vvlakshmanamurthy: Updated the same.

dplore (approved):

PR #3504, opened 2024-10-09
Readme for RT-5.11: Interface Load-Interval for Statistics Sampling

ihebboubaker:
This PR is related to Interface Load-Interval for Statistics Sampling


feature/interface/loadinterval/otg_tests/README.md
```diff
@@ -0,0 +1,60 @@
...
+## OpenConfig Path and RPC Coverage
+
+```yaml
+paths:
+  ## Config paths
+  /interfaces/interface/rates/config/load-interval:
+  ## State paths
+  /interfaces/interface/rates/state/load-interval:
+  /interfaces/interface/state/counters/in-octets:
+  /interfaces/interface/state/counters/out-octets:
```
dplore, line 49: note, these paths are an incrementing, total count of packets.  There are no OC paths for "packet or octet rate" (ie: there is no gauge like snmp has).    It's expected that a management system will stream the counters and compute a rate based on the incrementing counters.

dplore (changes requested):
The load-interval is expected to affect the CLI / any user interface on the device.  So it can be configured, but there is no direct way to validate this in OC.

So this test can configure the load-interval, but the only validation would be to get `/interfaces/interface/rates/state/load-interval` to confirm that it was set.

feature/interface/loadinterval/otg_tests/README.md
ihebboubaker: Yes, I agrre. So the objective is to monitor the in/out octet counters over a defined timeframe while transmitting traffic at a fixed rate. This allows us to compare the collected data with the anticipated values and confirm the counters' accuracy.

rohit-rp:
RT-5.11 is IPv6 ND suppression, the ID needs to be changed.
@ihebboubaker can you resolve it and update the PR

dplore:
@ram-mac to review

PR #3508, opened 2024-10-10
Update testregistry.textproto

vvlakshmanamurthy:
Adding a Testcase for ARP policers.


testregistry.textproto
```diff
@@ -930,6 +930,12 @@ test: {
   readme: "https://github.com/openconfig/featureprofiles/blob/main/feature/interface/staticarp/ate_tests/static_arp_test/README.md"
   exec: " "
 }
+test: {
+  id: "TE-1.2"
```
dplore, line 934: Add to DP list of tests.  (dataplane).

TE is for traffic engineering features.

vvlakshmanamurthy: done

PR #3516, opened 2024-10-14
Fix description for TE-9.1 in testregistry

dplore:


xw-g (changes requested):
Thanks for catching this. 

Maybe we should pick a different ID? https://github.com/openconfig/featureprofiles/blob/main/feature/gribi/otg_tests/mpls_compliance/README.md?plain=1

nsadhasivam (approved):
LGTM!

dplore:
> Thanks for catching this.
> 
> Maybe we should pick a different ID? https://github.com/openconfig/featureprofiles/blob/main/feature/gribi/otg_tests/mpls_compliance/README.md?plain=1

done

PR #3517, opened 2024-10-16
fix prefix_set_test

ram-mac:
The Cisco OCAGENT name was not defined inside gnoi.go file causing the test to fail.
Now this is been added and the test is now passing.


nsadhasivam (approved):
LGTM!

PR #3521, opened 2024-10-17
Add a test to retain inner packet TTL if inner packet TTL=1

vishnureddybadveli:
Added the missed functionality of retaining inner TTL.

### DIFF
--- a/feature/gribi/otg_tests/mpls_in_udp/README.md
+++ b/feature/gribi/otg_tests/mpls_in_udp/README.md
@@ -197,7 +197,7 @@ openconfig-network-instance:
                             destination-ip: "outer_ipv6_dst_def"
                             ip-ttl: outer_ip-ttl
                             dscp: outer_dscp
-                            inner-ttl-min: 2
+                            inner_ip_ttl_min: 1
 ```
 
 * Generate the policy forwarding configuration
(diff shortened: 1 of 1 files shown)


vishnureddybadveli:
As per vendors there is support to retain inner packet TTL, test case does not describe how exactly it is done byvendor implementation. Test case mentions that match should check for inner packet TTL == 1 but retaining inner TTL is vendor implementation.

feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -278,6 +278,133 @@ openconfig-network-instance:
...
+          encap-headers {
+            encap-header {
+              index: 2
+              type: OPENCONFIG_AFT_TYPES:UDP
+              udp {
+                src_ip: "outer_ipv6_src"
+                dst_ip: "outer_ipv6_dst_B"
+                dst_udp_port: "outer_dst_udp_port"
+                ip_ttl: "outer_ip-ttl"
+                inner_ip_ttl_min: 1
```
dplore, line 389: Can you also add your validation criteria to  [TE-18.1.2 above](https://github.com/openconfig/featureprofiles/pull/3521/files#diff-a6b35105c1accdb1889d9c529f4a13d6844d4a9844a90f3d3f736eba728502c5R200).   

Also, I think you need to define what the TTL should be on some IP traffic flow and what the action should be for each scenario.  (Define flow with TTL is 1 and another with TTL 2 and describe validation for both)

vishnureddybadveli: Thx,

Added different procedure for TTL 1 and another one for common case.

Added TTL min to TE-18.1.2.

vishnureddybadveli:
Plan is to use policy forwarding rule instead.

PR #3524, opened 2024-10-18
add deviation for leaf constraint

ram-mac:
power_admin_down_up_test is failing due to below error;
"Leaf 'name' inside the 'config' container must also be set to value '0/FC0' to satisfy the leafref constraint on the list key"

This error is for Leaf-ref validation for list keys which is enforced in XR, and this is fixed by setting the leaf in config container as well. While all other vendors work without this explicit configuration.

So there is deviation to track the behavior separately.


feature/platform/tests/power_admin_down_up_test/metadata.textproto
```diff
@@ -30,3 +30,12 @@ platform_exceptions: {
     skip_controller_card_power_admin: true
   }
 }
+platform_exceptions: {
+  platform: {
+    vendor: CISCO
+  }
+  deviations: {
+    leaf_ref_validation: true
```
nsadhasivam, line 38: leaf_ref_validation this is a generic name, Can the deviation name be specific to this oc config.?

nsadhasivam (changes requested):

feature/platform/tests/power_admin_down_up_test/metadata.textproto
ram-mac: actually i kept it generic on purpose; because there are some other tests where we need to do a leaf reference; Like some of the bgp config path specifically. Otherwise we keep it specific to power related on all components;

nsadhasivam: Can you please share the details of the oc leaves

cc: @dplore please share your inputs

ram-mac: Have change the deviaion name to power_enable_disable_leaf_ref_validation; The changes recommended is now fixed;

nsadhasivam (approved):
LGTM!

PR #3528, opened 2024-10-20
Update testregistry.textproto

vvlakshmanamurthy:
Flow matching using ACL and to port mirror or redirect.


testregistry.textproto
```diff
@@ -230,11 +230,17 @@ test: {
   exec: " "
 }
 test: {
-  id: "DP-1.17"
+  id: "CPT-1.1"
```
dplore, line 233: please alphabetize the list by id

dplore (commented):
please alphabetize, otherwise LGTM

dplore (approved):

PR #3531, opened 2024-10-21
traffic related and tolerance related changes

ram-mac:
1. This test uses random port assignments. While there are a lot of flows (64k) which should minimize variation between test runs, the tolerance in the test for load balancing deviation is extremely low (0.01%) so a small deviation could cause a failure.
2. This test uses a non-deterministic number of packets. While this probably has a small effect, there will be some variation in the number of packets sent in every test.

So the changes here are as follows
1. modify tolerance to 0.1% instead of 0.01%
2. traffic transmission is now set to 10k pps and 200k packets in total.


bkreddy143:
Can we consider to add 5 tuple variance ? currently it is on src and dst ports has variance . May be we can add src,dst ip as well ? at least adding src_ip variance for both inner and outer header doesn't need  change in dut config.

ram-mac:
> Can we consider to add 5 tuple variance ? currently it is on src and dst ports has variance . May be we can add src,dst ip as well ? at least adding src_ip variance for both inner and outer header doesn't need change in dut config.

the src/dst ip is limited i guess in comparison to the src/dst ports to create the flows. So IMO it may not have that much of an impact

nsadhasivam (approved):
LGTM!

feature/interface/aggregate/otg_tests/balancing_test/balancing_test.go
```diff
@@ -557,7 +561,7 @@ func (tc *testCase) testFlow(t *testing.T, l3header string) {
 	beforeTrafficCounters := tc.getCounters(t, "before")
 
 	tc.ate.OTG().StartTraffic(t)
-	time.Sleep(15 * time.Second)
+	time.Sleep(20 * time.Second)
```
dplore, line 564: You might want to add a little more time.  The traffic rate you set will take 20 seconds, and you are only sleeping for 20 seconds.  If the OTG takes some variable time which is greater than 0 seconds to ramp up to the Pps specified, it will take a little bit longer than 20 seconds to complete.  So this test could easily be flaky unless you allow a bit more time for the traffic to complete.  

Better than sleep though, can you research the OTG api and see if there is a call you can use to block on the traffic flow completing?  This would guarantee the flow is complete versus guessing on a timeout/sleep interval.

PR #2805, opened 2024-03-13
[gNMI-1.11, RT-5.1/3] - add InterfaceCountersUpdateDelayed deviation for CISCO

kjahed:
CISCO interface counters telemetry does not reflect the latest value. This PR adds InterfaceCountersUpdateDelayed and modify gNMI-1.10, gNMI-1.11, RT-5.1, and RT-5.3 to wait until the counters are updated (for a max timeout) when activated.

In addition, for gNMI-1.10, modify `verifyChassisIsAncestor` to ignore ancestors with no type. This is similar to PR #2808 for gNMI-1.4.

PR #3536, opened 2024-10-22
[gNMI-1.10] add InterfaceCountersUpdateDelayed deviation

kjahed:
1- Add InterfaceCountersUpdateDelayed deviation for CISCO (see PR #2805)
2- Use lookup instead of get for component type. Some of the parent component may not map to any of identity types defined under OPENCONFIG_HARDWARE_COMPONENT, for those components type leaf would not be returned (see PR #2808).


dplore (changes requested):
Thanks for this.  Your new function is better than the existing.  Our intent is to test using gnmi.Subscribe with SAMPLE mode.  I suggest you replace  the existing `fetchInAndOutPkts` with your new function.

ram-mac:
@kjahed - Can you please take care of the comments requested by Darren?

kjahed:
> @kjahed - Can you please take care of the comments requested by Darren?

@ram-mac  done.

dplore (approved):
Ran this in our environment and it is passing as expected, thanks.

ram-mac (approved):
LGTM

PR #3537, opened 2024-10-22
DP-1.2: Fixing validation of non strict priority scheduler case for Juniper

charantejag504:
DP-1.2: Fixing validation of non strict priority scheduler case for Juniper

Fixing scheduler priority state validation for non strict schedulers as per vendor support


sezhang2 (approved):

sezhang2:
Hi @Swetha-haridasula,

Looks like the PR needs one more approval. Please check.

thanks!

PR #3540, opened 2024-10-24
Updating code for kne platform

priyaranjannanda:
Hi ,

"WaitForArp" needs way to invoke arp request when invoked this api.

As of now "WaitForArp " Ixia API call does not send explicit arp request , which is very much needed when tests have interface change events in containerized environment. (Virtual platforms).

As of now ARP is resolved only during Pushconfig API and not when we invoke "WaitForARP".

Thanks,
Priya Ranjan


feature/interface/ip/ipv6_link_local/otg_tests/ipv6_link_local_test/ipv6_link_local_test.go
```diff
@@ -85,6 +87,14 @@ func TestIPv6LinkLocal(t *testing.T) {
 	otgSrcToDstFlow(t, top, ateSrc.IPv6, ateDst.IPv6, linkLocalFlowName)
 
 	ate.OTG().PushConfig(t, top)
+	// https://github.com/openconfig/featureprofiles/issues/3410
+	// Below code will be removed once ixia issue is fixed.
+	if slices.Contains(kneDeviceModelList, dut.Model()) {
+		ate.OTG().StartTraffic(t)
+		time.Sleep(15 * time.Second)
```
ram-mac, line 94: Couple of comments here
1) Can we add a code to check if the traffic is started on Ixia and then follow step 2.
2) Instead of Sleeping for 15 seconds of time; We can make changes in the traffic profile to send fixed traffic with a duration of 15 or 20seconds and we would not have to stop the traffic in ln #9

ram-mac:
Adding more reviewers for this PR

priyaranjannanda:
Hi All,

Fix needs to be in kne to resolve Arp. Detailed analysis is here https://github.com/open-traffic-generator/fp-testbed-juniper/issues/88

I am closing this pull request.

--
Thanks & Regards,
Priya Ranjan

PR #3542, opened 2024-10-26
TE-16.2/TE-17.1: Removed disabling autonegotiation and explicitly setting port speed a…

mohanasm:
Removed explicit disable of auto-negotiation and setting port-speed,duplex for specific PMD type. It is not a requirement per README.


Swetha-haridasula (approved):
lgtm

PR #3543, opened 2024-10-26
fixing cisco fib values and add sleep

ram-mac:
Cisco was failing because of BGP neighbors not formed. So have fixed following

1. No BGP neighbor was resolved by adding 30secs delay in the test before verification of the BGP; 
2. After fixing the BGP issue, the test stills failed with an error "Please provide BGP route count for vendor to maxout FIB CISCO in var vendorSpecRoutecount"; So this was fixed by adding 250k value to the Cisco as below.

	vendorSpecRoutecount = map[ondatra.Vendor]uint32{
	}

These fixes has now helped in fixing the test.


nsadhasivam (approved):
LGTM!

feature/gribi/otg_tests/fib_failed_due_to_hw_res_exhaust_test/fib_failed_due_to_hw_res_exhaust_test.go
```diff
@@ -404,7 +405,7 @@ func verifyBgpTelemetry(t *testing.T, dut *ondatra.DUTDevice) {
 	var nbrIP = []string{atePort1.IPv6}
 	t.Logf("Verifying BGP state.")
 	bgpPath := gnmi.OC().NetworkInstance(deviations.DefaultNetworkInstance(dut)).Protocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP").Bgp()
-
+	time.Sleep(30 * time.Second)
```
dplore, line 408: why add another sleep inside the function when it's also been added outside the function call at 

https://github.com/openconfig/featureprofiles/pull/3543/files#diff-95f47bbdec2fb8f97a024715f0396c01a34b2c03a06f581cf4126e68fb9e1336R294

Wouldn't it be more clear to just increase the sleep times that are already there?

LimeHat: If anything, I don't think either of the sleep timers makes sense to solve the stated issue. 
https://github.com/openconfig/featureprofiles/blob/b03beb684baa8e0b89b76679c8a2915d8a37e2b0/feature/gribi/otg_tests/fib_failed_due_to_hw_res_exhaust_test/fib_failed_due_to_hw_res_exhaust_test.go#L413-L417

gnmi.watch.await already blocks for a minute, and this timer can be increased if needed.

ram-mac: ACK @dplore @LimeHat 
Will fix it, by increasing the timer in gnmi.Watch function.

PR #3544, opened 2024-10-28
fix bgp_session_mode_configuration

ram-mac:
Fixed the error where pushing the config for bgp sessions fails for following reasons
1. default address family being not enabled in peer-group
2. transport address is not defined for the peer-group level or neighbor level
3. OTG nbrLvlPassive, peerLvlActive mode were not configured; So this had to be configured.


dplore:
Thanks for fixing this @ram-mac and for your review @Swetha-haridasula .  

This should be refactored to use the bgp cfgplugin at https://github.com/openconfig/featureprofiles/blob/main/internal/cfgplugins/bgp.go

This contains common functions for creating bgp sessions that can be reused across any test using bgp.  This way we are not repeating code in each test and when we discover changes are needed, they can be inherited by all tests.

PR #3545, opened 2024-10-28
Update ocpaths.proto with releases and yang version

arista-allan:
Add a richer definition of a release identifier via new message, which has the original timestamp,
but adds an optional version number, which is typically how they are formally identified, plus an
optional moniker, which is how they are sometimes informally identified.

Add to ocpath three optional release identifiers: one for origination (the release when a path became
available from a vendor), one for deprecation (the release when a path is deprecated by a vendor),
and one for removal (when a path is no longer supported by a vendor).

Add to ocpath an optional yangModelVersion, which is typically the public YANG model version when
a path appears in a model, typically an OpenConfig consortium model - this is vendor independent.


proto/ocpaths.proto
```diff
@@ -62,6 +75,24 @@ message OCPath {
...
+  // paths. origination must be defined (not enforced). not mutually 
+  // exclusive with origination and removal.
+  optional Release deprecation = 7;
+
+  // the date/time when this path was removed from support. not present
+  // for active paths or if unknown. origination must be defined (not enforced).
+  // not mutually exclusive with deprecation.
+  optional Release removal = 8;
+
+  // the YANG model version that first defined this path. this is typically
```
LimeHat, line 93: is there any value in tracking this for each vendor/sw/hw release? looks like vendors will need to maintain a database with such information in order export it..?

or is the expectation that the data will be provided by OC community and not used by vendors?

proto/ocpaths.proto
```diff
@@ -62,6 +75,24 @@ message OCPath {
   // GNMIRpc describes expected (or supported) behavior for a particular
   // Openconfig path.
   GNMIRpc gnmi_rpc = 5;
+
+  // the date/time when this path became supported. not present if 
+  // unknown.
+  optional Release origination = 6;
+
+  // the date/time when this path became deprecated. not present for active
```
LimeHat, line 83: how "active" is defined in this context?

aren't deprecated path still usable?

proto/ocpaths.proto
```diff
@@ -26,6 +26,19 @@ package openconfig.ocpaths;
 
 option go_package = "github.com/openconfig/featureprofiles/proto/ocpaths_go_proto;ocpaths";
 
+// Release is at minimum a timestamp for associated with an artifact.
+// By convention, itt is expected to uniquely identify and artifact
```
LimeHat, line 30: a few typos on this line (`itt` `and artifact`)

proto/ocpaths.proto
```diff
@@ -26,6 +26,19 @@ package openconfig.ocpaths;
...
+message Release {
+  // the date/time of the relase
```
LimeHat, line 32: typo: `relase`

proto/ocpaths.proto
```diff
@@ -26,6 +26,19 @@ package openconfig.ocpaths;
...
+  google.protobuf.Timestamp date = 1;
+
+  // the optional version of the release, typically numbers and dots.
+  optional string version = 2;
+
+  // the optional moniker (nickname) for the release
+  optional string name = 3;
+}
```
LimeHat, line 40: https://github.com/openconfig/featureprofiles/blob/13c822568b5abf9a3efbf5dd78f43a3e826feb5f/proto/nosimage.proto#L39-L43

there are no nicknames for releases in nosimage.proto

proto/ocpaths.proto
```diff
@@ -26,6 +26,19 @@ package openconfig.ocpaths;
...
 option go_package = "github.com/openconfig/featureprofiles/proto/ocpaths_go_proto;ocpaths";
 
+// Release is at minimum a timestamp for associated with an artifact.
+// By convention, itt is expected to uniquely identify and artifact
+message Release {
+  // the date/time of the relase
+  google.protobuf.Timestamp date = 1;
+
+  // the optional version of the release, typically numbers and dots.
+  optional string version = 2;
```
LimeHat, line 36: in proto3 syntax there's no need to use `optional` syntax unless you need an explicit presence indicator for the field (any field can be omitted)

LimeHat (changes requested):
In general, it would be good to hear more on the motivation for these changes.

A concern I have is that the concept of nosimage.proto moves from a snapshot of supported features in a given release/platform combo to an abstract, database-like mechanism that keeps track of history of changes, some of which are not even related to vendor changes.

proto/ocpaths.proto
arista-allan: Not the OC community - this is vendor-specific. I'll update the comments to make this clear.
That database would be the NOS images from previous releases. External logic would walk those
NOS images and calculate any origination/deprecation/removal dates to populate the current NOS
image. As a vendor, it is fairly common for our customers to ask for this information. But it is marked
as optional (I saw your later comment on this - I'll comment there) as a reminder that a vendor can
release a NOS image without this information.

proto/ocpaths.proto
arista-allan: ah, I needed a word for not-deprecated and made one up (without definition). 
I'll clear this comment up.

proto/ocpaths.proto
arista-allan: who knew that i get paid to type. :face-palm:

proto/ocpaths.proto
arista-allan: Good point. The concept of the release triplet was not spread around to the rest of the
schema. You could rightly make the point that it should be. I have some reviewers coming
from G that I'll ask about that. Thanks for raising.

proto/ocpaths.proto
LimeHat: >Not the OC community - this is vendor-specific.
>  // the YANG model version that first defined this path.

In that case perhaps I don't understand how the "yang model version" is defined. We, as a vendor, don't have any additional versioning of OC models. Or perhaps you're talking about your PR in general but not about the specific field?
That question is related to the field only (unfortunately I quoted the first line of the comment which didn't produce a nice preview). As for the overall PR reasoning, I left a comment above.

arista-allan: that yangModelVersion typically identifies the OpenConfig consortium model that defines this leaf.

proto/ocpaths.proto
arista-allan: per comments, this does have a constraint on the value to match /system/state/software-version

proto/ocpaths.proto
LimeHat: >that yangModelVersion typically identifies the OpenConfig consortium model that defines this leaf.

That brings me back to the original question: what is the value of tracking it per vendor/platform? :-)

proto/ocpaths.proto
LimeHat: Usually, comments (or a separate README/API contract document) are used to indicate mandatory fields when there's a need. I think you already did that in the parent comment (above the `message Release`). I don't think there's a way to enforce it in the syntax.

Although I wouldn't necessarily agree that specifying a release date without a release version is a great idea :)

>and mutual exclusion.

this can be done via `oneof`

proto/ocpaths.proto
arista-allan: tracking this was a specific requirement handed to us by G

LimeHat: @dplore would you be able to provide context?

proto/ocpaths.proto
dplore: Use a comment above the field to indicate mandatory

proto/ocpaths.proto
dplore: [NOSImageProfile](https://github.com/openconfig/featureprofiles/blob/main/proto/nosimage.proto) should be used to describe the contents of an operating system's support.

A series of NOSImageProfile messages (textproto files) can be used to learn when a path is introduced and removed.   

Any fields related to the  NOS support should be present in the NOSImageProfile proto.

dplore (changes requested):
[NOSImageProfile](https://github.com/openconfig/featureprofiles/blob/main/proto/nosimage.proto) should be used to describe the contents of an operating system's support.

A series of NOSImageProfile messages (textproto files) can be used to learn when a path is introduced and removed.

It seems all the new fields being proposed in ocpaths.proto really belong in in the NOSImageProfile proto.

proto/ocpaths.proto
dplore: I believe the version of the NOS should remain in the NOSImageProfile and is not needed here.

proto/ocpaths.proto
dplore: If this field is intended to represent the OC yang model version, then when used inside of a NOSImageProfile, this field can assert the OC yang model version the NOS supports.  I am not sure it has to be the 'earliest'.  It can merely be whatever version the NOS was built to support.

proto/ocpaths.proto
dplore: We need to be clear if we are talking about the OC release (https://github.com/openconfig/public/releases)  or the NOS release (vendor operating system version)?   

Because this release is in the OCPath.proto, it should in my mind be the OC release.  Is that your intent @arista-allan ?

arista-allan:
Thanks for the feedback. I'll have cycles to put on this shortly. And will suggest improvements for approval.

arista-allan:
Not sure why, but i'm not getting any notifications when comments are made here. Sigh. Maybe 2025 will be better.

dplore:
> Not sure why, but i'm not getting any notifications when comments are made here. Sigh. Maybe 2025 will be better.

Not sure if your email is filtering, but there are various ways you can subscribe to github notifications:
https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github

proto/ocpaths.proto
LimeHat: >It can merely be whatever version the NOS was built to support.

This doesn't require per-path granularity though. Today there's already a field that indicates the supported OC release https://github.com/openconfig/featureprofiles/blob/1d9ac4a2986756f560e9635aea36ad88afa07482/proto/ocpaths.proto#L50-L54 - I remember we had a conversation about this some time ago, and the intent was for vendors to follow the OC release numbering (instead of individual model numbering)

arista-allan: i finally have some free cycles to circle back on this.

PR #3546, opened 2024-10-29
added breakout config test

mastarkey:
Test is for breakout config for the following modes 

- 4x100
- 2x100
- 4x10


feature/platform/tests/breakout_configuration/breakout_test.go
```diff
@@ -0,0 +1,350 @@
...
+				configContainer := &oc.Component_Port_BreakoutMode_Group{
+					Index:         ygot.Uint8(0),
+					NumBreakouts:  ygot.Uint8(tc.numbreakouts),
+					BreakoutSpeed: oc.E_IfEthernet_ETHERNET_SPEED(tc.breakoutspeed),
+				}
+				groupContainer := &oc.Component_Port_BreakoutMode{Group: map[uint8]*oc.Component_Port_BreakoutMode_Group{1: configContainer}}
+				breakoutContainer := &oc.Component_Port{BreakoutMode: groupContainer}
+				portContainer := &oc.Component{Port: breakoutContainer, Name: ygot.String(componentName)}
+
+				if deviations.VerifyExpectedBreakoutSupportedConfig(dut) {
```
nraghavendran, line 203: not sure why this deviation is needed. if breakout config is not supported. then the test should fail right?

nraghavendran: if this applies only for 2x100G, pls add a comment to call it out explicitly/indicate in the deviation title

LimeHat: If this is about the optic modules, IMO that check (if its needed) should rely on OpenConfig-defined telemetry, e.g. `ethernet-pmd`, and be implemented for all vendors. 

Vendor-proprietary checks make sense only if this related to limitations of a specific platform.

feature/platform/tests/breakout_configuration/breakout_test.go
```diff
@@ -0,0 +1,359 @@
...
+
+						gnmi.Update(t, dut, gnmi.OC().Interface(portName).Config(), i)
+					}
+
+					t.Log("Configuring the ATE")
+					configureOTG(t, ate, &tc.breakoutspeed, Ateipv4Subnets, Dutipv4Subnets,
+						int(tc.numbreakouts))
+				})
+
+				t.Run(fmt.Sprintf("Ping ATE from DUT via Breakout Interface %v %v",
```
LimeHat, line 283: I don't think there should be dependency on gnoi ping. At the very least its not in the readme. 

IMO a simple traffic test (between the two ATE ports) makes more sense.

feature/platform/tests/breakout_configuration/metadata.textproto
```diff
@@ -0,0 +1,16 @@
+# proto-file: github.com/openconfig/featureprofiles/proto/metadata.proto
+# proto-message: Metadata
+plan_id: "PLT-1.1"
+uuid: "6ac4eed0-9e04-46f5-85d8-ca88ee5fec7d"
+description: "Interface breakout Test"
+testbed: TESTBED_DUT_ATE_9LINKS_LAG
```
LimeHat, line 6: Doesn't seem that you need 9LINKS_LAG here.

feature/platform/tests/breakout_configuration/breakout_test.go
```diff
@@ -0,0 +1,359 @@
...
+		numbreakouts  uint8
+		breakoutspeed oc.E_IfEthernet_ETHERNET_SPEED
+		portPrefix    string
+		dutIntfIp     string
+		ateIntfIp     string
+	}{
+		{
+			numbreakouts:  4,
+			breakoutspeed: oc.IfEthernet_ETHERNET_SPEED_SPEED_100GB,
+			portPrefix:    "HundredGigE",
```
LimeHat, line 159: some vendor-specific hardcoded prefixes.. can they be eliminated?

LimeHat (changes requested):

nraghavendran (commented):
+1 to address # DUT links

feature/platform/tests/breakout_configuration/breakout_test.go
mastarkey: not sure what path your referring to with `ethernet-pmd` however when I check the [Openconfig Schema](https://openconfig.net/projects/models/schemadocs/jstree/openconfig-platform.html) for platform it takes to components/component and there are leafs that will give this cli output as its use is to know the supported modes of a given optic and prevent us from creating a false positive failure result. So as far as I know there is no baked in OC way of getting this data.

feature/platform/tests/breakout_configuration/breakout_test.go
mastarkey: its in the readme to use ping in the procedure check the bottom bullet it mentions `Verify that DUT responds to ARP/ICMP on all tested interfaces`

feature/platform/tests/breakout_configuration/metadata.textproto
mastarkey: Thanks resolved

feature/platform/tests/breakout_configuration/breakout_test.go
LimeHat: In that case the ping should be initiated from OTG ("DUT responds")

feature/platform/tests/breakout_configuration/breakout_test.go
LimeHat: ethernet-pmd: https://github.com/openconfig/public/blob/ea290e31e8e9fbad5135ac28eb74a718fe502ec8/release/models/platform/openconfig-platform-transceiver.yang#L533-L543

If you think this is not enough to determine the breakout capability, can you please elaborate on what else is needed?  If there's a property of the transceiver that is not exposed in OC models that is needed to determine that, let's fix that in OC.

feature/platform/tests/breakout_configuration/breakout_test.go
LimeHat: I don't understand why is this marked as resolved, since it is not..

LimeHat (changes requested):
Looks like my comments were marked as "resolved" instead of being actually addressed :-( please take another look at the three items that were marked as "resolved"

feature/platform/tests/breakout_configuration/breakout_base_test.go
```diff
@@ -0,0 +1,232 @@
...
+		// Handle potential errors from Atoi conversion
+		if errI != nil || errJ != nil {
+			return false
+		}
+		return numberI < numberJ
+	})
+}
+
+// Function to get compName and set up port name based on provided interface and port prefix
+func getCompName(dut *ondatra.DUTDevice, string, portPrefix string, t *testing.T) (string, string, bool) {
```
LimeHat, line 192: this whole function is vendor-specific.
if it is impossible to eliminate vendor specific logic and dependency on prefixes, then at least vendor-specific items should be guarded with switch/case statements and/or maps to select the correct choices for each vendor.

dplore: It is expected that component/interface names are vendor specific.  Agreed, they should be wrapped/guarded by switch statement like switch dut.Vendor()

LimeHat: >It is expected that component/interface names are vendor specific.

I agree. My comment is more generic: in this function, and throughout the test (I left a similar comment above, which was left unanswered), there are a lot of dependencies on hardcoded names/prefixes/etc, and I suspect that some of this logic can (and should) be replaced with dynamic discovery or other methods. 

For example, not all vendors use prefixes to distinguish interfaces with different port speeds; however the current logic seem to require that:

https://github.com/openconfig/featureprofiles/blob/676ca31db676d47dd127c1e8a5f75b7ddd1f0686/feature/platform/tests/breakout_configuration/breakout_test.go#L156-L189

Thus this logic needs to be revisited.

mastarkey: these area's you have called out as vendor specify have been wrapped in if `dut.Vendor() == ondatra.CISCO {}` so only cisco devices will access that code block.

feature/platform/tests/breakout_configuration/breakout_test.go
```diff
@@ -0,0 +1,369 @@
...
+
+						gnmi.Update(t, dut, gnmi.OC().Interface(portName).Config(), i)
+					}
+
+					t.Log("Configuring the ATE")
+					configureOTG(t, ate, &tc.breakoutspeed, Ateipv4Subnets, Dutipv4Subnets,
+						int(tc.numbreakouts))
+				})
+
+				t.Run(fmt.Sprintf("Ping ATE from DUT via Breakout Interface %v %v",
```
LimeHat, line 293: The previous comments haven't been addressed yet.

feature/platform/tests/breakout_configuration/metadata.textproto
```diff
@@ -0,0 +1,16 @@
+# proto-file: github.com/openconfig/featureprofiles/proto/metadata.proto
+# proto-message: Metadata
+plan_id: "PLT-1.1"
+uuid: "6ac4eed0-9e04-46f5-85d8-ca88ee5fec7d"
+description: "Interface breakout Test"
+testbed: TESTBED_DUT_ATE_9LINKS
```
LimeHat, line 6: What is the logic behind selecting this topology? It doesn't appear that you're using 9 links in the test code.

feature/platform/tests/breakout_configuration/breakout_test.go
```diff
@@ -0,0 +1,369 @@
...
+	sort.Slice(ports, func(i, j int) bool {
+		return ports[i].ID() < ports[j].ID()
+	})
+
+	if *breakoutspeed == oc.IfEthernet_ETHERNET_SPEED_SPEED_100GB {
+		t.Logf("Speed is %v", *breakoutspeed)
+	} else if *breakoutspeed == oc.IfEthernet_ETHERNET_SPEED_SPEED_10GB {
+		t.Logf("Speed is needed to start port assignment on port5 as that is "+
+			"where 10G ports are in setup %v", *breakoutspeed)
+		ports = ports[4:] // Assuming ports 5+ are 10G
```
LimeHat, line 91: Is there a basis for this assumption?

feature/platform/tests/breakout_configuration/breakout_test.go
```diff
@@ -0,0 +1,369 @@
...
+	}{
+		{
+			numbreakouts:  4,
+			breakoutspeed: oc.IfEthernet_ETHERNET_SPEED_SPEED_100GB,
+			portPrefix:    "HundredGigE",
+			dutIntfIp:     dutPort1.IPv4,
+			ateIntfIp:     atePort1.IPv4,
+		},
+		{
+			portPrefix:    "HundredGigE",
```
LimeHat, line 167: I think this was supposed to be removed (and the other 2 instances as well)?

LimeHat (changes requested):

feature/platform/tests/breakout_configuration/breakout_test.go
```diff
@@ -0,0 +1,369 @@
...
+			dutIntfIp:     dutPort2.IPv4,
+			ateIntfIp:     atePort2.IPv4,
+		},
+	}
+
+	gnoiClient := dut.RawAPIs().GNOI(t)
+	ate := ondatra.ATE(t, "ate")
+
+	for _, tc := range cases {
+		tc := tc // Capture range variable
```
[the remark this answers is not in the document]
mastarkey: these issues have been discussed with Google and they are ok with code as is.

feature/platform/tests/breakout_configuration/breakout_base_test.go
```diff
@@ -0,0 +1,232 @@
...
+		ips[i] = fmt.Sprintf("%s/%d", incrementedIP.String(), 24) // Format as a string with a /24 subnet.
+	}
+
+	return ips, nil
+}
+
+// findNewPortNames fetches the parent breakout port interface states and the new port convention.
+// Example being 4x100 parent port would be FourHundredGigE0/0/0/10 this will find and return
+// the newly broken out ports of OneHundredGigE0/0/0/0/10/0-4
+func findNewPortNames(dut *ondatra.DUTDevice, t *testing.T, originalPortName string, numBreakouts uint8) ([]string, error) {
```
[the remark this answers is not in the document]
mastarkey: these issues have been discussed with Google and they are ok with code as is.

feature/platform/tests/breakout_configuration/breakout_test.go
mastarkey: these issues have been discussed with Google and they are ok with code as is.

nraghavendran (approved):

PR #3548, opened 2024-10-29
moved flow helpers under otg config helpers , as flow is also configu…

Surajrawal-eqa:
…ration in otg


arukumani-eqa (approved):

internal/otg_helpers/otg_config_helpers/otgflowhelpers.go
dplore, line 1: The file names stutter.  Recommend naming like:L

otg_helpers/config/flow/

PR #3549, opened 2024-10-29
deviation for cisco

ram-mac:
Cisco session state is set to Idle when both the BGP neighbors are set to passive mode (on DUT and on OTG)
And the script is looking for an Active state and hence the test is failing

So, added a deviation to modify the expected state of BGP in such case.


nsadhasivam (dismissed):
LGTM!

rohit-rp (approved):

PR #3553, opened 2024-10-30
deviation for cisco 1.51 multipath only under global/afi/safi/

ram-mac:
cisco deviation for multipath.
 Currently cisco enables multipath and max-paths at the global/afi/safi level and not under global/ path.
 So added a deviation to take care of this test.


Swetha-haridasula (approved):
lgtm

PR #3557, opened 2024-11-05
added required approvals workflow

self-maurya:


self-maurya:
Raised another PR - https://github.com/openconfig/featureprofiles/pull/3563

PR #3558, opened 2024-11-05
fix leaf constraint error and address family initialisation error

ram-mac:
This test is failing with the below errors
1) Leaf 'name' inside the 'config' container must also be set to value 'DEFAULT' to satisfy the leafref constraint on the list key"
2) "'BGP' detected the 'warning' condition 'The address family has not been initialized'"

The changes here would fix the above two errors.


nsadhasivam (approved):
LGTM!

PR #3561, opened 2024-11-05
Added defer function to cleanup fib table

bkreddy143:
currently after test is complete OTG retains its config and due to which following tests are failing as fib is full and not able to install any more routes.  added defer function to stop BGP protocol that will clear all routes injecected .


nsadhasivam (approved):
LGTM!

PR #3573, opened 2024-11-08
[gNMI-1.14] prune device config before push

kjahed:
Similar to gNMI-1.15, device config must be pruned before it can be pushed again. This PR: 
- moves `GetDeviceConfig` from gNMI-1.15 to `/internal/fptest/config.go` so that it can be reused by gNMI-1.14
- modifies gNMI-1.14 metadata verification to ignore any notification with a timestamp newer than that of the configuration `SetResponse`.


ram-mac:
hi,
can the readme be fixed?

ram-mac (approved):
LGTM

PR #3575, opened 2024-11-12
RT_7.8 changes and deviation

arvbaska1:

- Change AS# for ATE Port 2 , since routes will not be advertised with same AS on both sides to avoid loop.
- SetSendCommunity under deviation SkipBgpSendCommunityType
- Under deviation BGPConditionsMatchCommunitySetUnsupported for cisco, reconfigured community-set
- Added prefix length with the subnet for the communityIndex verification func


ram-mac:
@arvbaska1 - Can you resolve the conflicts and push the commit. Will validate it once done.

ram-mac (approved):
Changes looks good.
Validated the test on Cisco and it is passing now.

rohit-rp (approved):

PR #3577, opened 2024-11-12
fixing interface level config and timer

ram-mac:
Fixed the adjacency issue by enabling the level 2 at the interface level.
Also added a wait time after the pushing config to the ixia and dut


Swetha-haridasula (approved):
lgtm

feature/isis/otg_tests/isis_metric_style_wide_enabled_test/isis_metric_style_wide_enabled_test.go
```diff
@@ -191,13 +193,15 @@ func TestISISWideMetricEnabled(t *testing.T) {
...
 	ts.PushAndStart(t)
+	time.Sleep(time.Minute * 2)
 
 	statePath := isissession.ISISPath(ts.DUT)
 	intfName := ts.DUTPort1.Name()
 	if deviations.ExplicitInterfaceInDefaultVRF(ts.DUT) {
 		intfName += ".0"
 	}
 	t.Run("ISIS telemetry", func(t *testing.T) {
+		time.Sleep(time.Minute * 2)
```
nsadhasivam, line 204: Is the delay for ISIS to come up?

Can we use gnmi.Await instead?

PR #3581, opened 2024-11-13
Rename Packages in features/security.

morrowc:
The package name config for each golang test package could have been more specified.

Supposedly this might fix the copy action into google's internal  repository?


dplore (approved):
LGTM.  This may not be the only issue, but it will help resolve some build errors/expectations

PR #3585, opened 2024-11-13
AcctzPB fix - there were 2 definitions for 'acctz', that's bad.

morrowc:
Set the PB versions of the import to use an alias: acctzpb
Leave the featureprofile version of acctz alone.


dplore (changes requested):
also needed in record_history_truncation_test.go

dplore (dismissed):

PR #3589, opened 2024-11-14
fix linkbw_any regex

dplore:
earlier regex matches only single char for bandwidth.  Now it will match any number of chars in the bandwidth portion of the community


nsadhasivam (approved):
LGTM!

PR #3590, opened 2024-11-14
code base for slim GCI control plane test infra

gopalakrishv:
Developed a code for bring up the control plane test infra with lemmings in KNE topology to work on lemming development activities.

Created a DUT with two IXIA ports connected to the DUT and configured the interfaces, configured the route entry through gRIBI programming.

Same has been validated with the FIB program success acknowledgement.


dplore:
Fails on lemming kne in this PR?  See [Log](https://storage.cloud.google.com/featureprofiles-ci-logs-openconfig/log-81d207ff-8c84-4ba3-9feb-9f195573e720.txt)

Also, can you add a peer team member to review?

gopalakrishv:
> Fails on lemming kne in this PR? See [Log](https://storage.cloud.google.com/featureprofiles-ci-logs-openconfig/log-81d207ff-8c84-4ba3-9feb-9f195573e720.txt)
> 
> Also, can you add a peer team member to review?

This test case is passed in Lemmings, please find the pass log for your reference.

https://fusion2.corp.google.com/invocations/ecfb03f4-df7d-49bd-b15b-cd0c66bed690/targets/%2F%2Fthird_party%2Fopenconfig%2Ffeatureprofiles%2Ffeature%2Fgribi%2Fotg_tests%2Fgribi_static_route_on_lemmings_test:gribi_static_route_on_lemmings_test_kne_openconfig_lemming/tests

feature/gribi/otg_tests/gribi_static_route_on_lemmings_test/BUILD
```diff
@@ -0,0 +1,36 @@
+load("//ops/netops/lab/wbb/testbeds:testbeds.bzl", "featureprofiles_test_suite")
```
Surajrawal-eqa, line 1: @gopalakrishv We should not push Build file in Git. It will be created internally once the PR is merged

feature/gribi/otg_tests/gribi_static_route_on_lemmings_test/gribi_static_route_on_lemmings_test.go
```diff
@@ -0,0 +1,188 @@
+package basic_static_route_support_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"google3/third_party/open_traffic_generator/gosnappi/gosnappi"
```
Surajrawal-eqa, line 9: Import packages from github.com, google3 is for internal usage. Same applies for all the imports used in this file. Refer internal/otg_helpers/otg_config_helpers/otgconfighelpers.go for reference

Surajrawal-eqa (changes requested):

feature/gribi/otg_tests/gribi_static_route_on_lemmings_test/gribi_static_route_on_lemmings_test.go
```diff
@@ -0,0 +1,188 @@
+package basic_static_route_support_test
```
arukumani-eqa, line 1: I suspect this is a already running gribi test on lemmings ....

Lemmings SKU defined here -- http://shortn/_qSGuvLKhMn

Also the gribi calls are same as file which does similar test -- http://shortn/_ghVJjmX5gu

might be we just need to validate and confirm that the existing test works. 

@dplore any thoughts here?

feature/gribi/otg_tests/gribi_static_route_on_lemmings_test/README.md
```diff
@@ -0,0 +1,48 @@
+# GRIBI static route on lemmings test
```
arukumani-eqa, line 1: Looks like this has more coverage as well...

http://shortn/_FcU1Zv7i6k

If Lemmings SKU is already covered in existing tests , then we can use the same i guess.

arukumani-eqa (changes requested):

feature/gribi/otg_tests/gribi_static_route_on_lemmings_test/README.md
gopalakrishv: Purpose of this test case is to give minimum environment for developing the control plane by using the lemmings

feature/gribi/otg_tests/gribi_static_route_on_lemmings_test/BUILD
gopalakrishv: I have removed the BUILD file

feature/gribi/otg_tests/gribi_static_route_on_lemmings_test/gribi_static_route_on_lemmings_test.go
gopalakrishv: import file paths are updated to gitub

feature/gribi/otg_tests/gribi_static_route_on_lemmings_test/gribi_static_route_on_lemmings_test.go
```diff
@@ -0,0 +1,185 @@
+package gribi_static_route_on_lemmings_test
```
arukumani-eqa, line 1: probably we want to exclude lemming from the namings here ... 

the test seems to be agnostic of platform and the lemmings option is passed as a flag to the test in wbb bind -- use_kne_topo ...

I agree for the need of a simple test in case you want to demo.

arukumani-eqa (changes requested):

feature/gribi/otg_tests/gribi_static_route_on_lemmings_test/gribi_static_route_on_lemmings_test.go
gopalakrishv: If I remove the lemmings from the name, it will look like a generic one and will be duplicated due to other test cases, which was covering all the aspects, also this coverage is required to have it for Lemming development. We can execute the test case only on Lemming by executing the below command 

blaze test --notest_loasd --test_output=streamed --cache_test_results=no -- //third_party/openconfig/featureprofiles/feature/gribi/otg_tests/gribi_static_route_on_lemmings_test:gribi_static_route_on_lemmings_test_kne_openconfig_lemming

arukumani-eqa: I suspect if we can include platform aspects into the test files... the tests are really platform agnostic....

Also i believe this is not lemmings only test ... the same test should be able to work on other vendors as well... Also, i do not think the other tests cover this simple case right? 

+ @dplore for any thoughts

arukumani-eqa (changes requested):

feature/gribi/otg_tests/gribi_static_route_on_lemmings_test/gribi_static_route_on_lemmings_test.go
arukumani-eqa: Approving as this is not a blocker to run tests... 

@dplore when you have a chance .... please let us know on your thoughts so that we can do it in a seperate change if any is needed.

arukumani-eqa (approved):

Surajrawal-eqa (approved):

PR #3598, opened 2024-11-15
Bug fix validate.go

dplore:
Current code only works if a string is passed as the version.  If no string is provided, the a clone is attempted on a branch name of "v", which won't exist in the public OC repo


ElodinLaarz (dismissed):
LGTM.

ElodinLaarz (approved):

PR #3602, opened 2024-11-15
Fixed issue where MTU not getting set for some MTUs

mastarkey:
there is an if condition that was set for setting the expected mtu for 1500, this excluded the other two MTU sizes of 5000 and 9236 so added those two types to the if condition as well as they were failing with the got value of nil and want of expected mtu size


ram-mac:
The change looks good;

ram-mac (approved):
LGTM

self-maurya (approved):

nikki-blr (approved):

nsadhasivam (approved):
LGTM!

PR #3603, opened 2024-11-20
fixing bgp_afi_safi_defaults_test with deviations

ram-mac:
1) BGP Neighbor Configuration: The script currently configures both IPv4 and IPv6 address families as "true" for both v4 and v6 BGP peers. The Readme specifies that only the corresponding address family should be set to "true" for each peer type. This is fixed now

2) BGP Capabilities Verification: The script currently retrieves all capabilities for a neighbor and iterates through them. We should instead only retrieve and check the capabilities for the relevant AFI/SAFI family (IPv4 or IPv6). I have implemented a fix for this as well.
   Cisco does not support GetAll/Wildcard capability for AFI/SAFI and hence added deviation

3) Test Case: Validate AFI-SAFI OC Defaults at Peer Group Level:
The argument for BGPv4 peers is incorrectly setting the "isV4Only" parameter to "false." 
This is corrected now.

4) Validate AFI-SAFI Not Enabled: New subsection added to verify only the BGP session and keep validating BGP capabilities as another subtest.

5) Enabling the extNh.ExtendedNextHopEncoding 
This feature is enabled by default for Cisco and we do not have to explicitly enable it via OC path. 
 Currently added a deviation to set this leaf.


dplore (commented):
Thanks for the changes.  Couple of suggestions on intent/description update.

The deviation `BgpExtendedNextHopEncodingLeafUnsupported` should not be added unless we can confirm why it is there.  See inline comments.

dplore (approved):

dplore (approved):

PR #3614, opened 2024-11-29
RT-5.8 monitor interface stats before proceeding with test

karthikeya-remilla:
using gnmi.Watch.Await instead of time.sleep


feature/interface/ip/ipv6_link_local/otg_tests/ipv6_link_local_test/ipv6_link_local_test.go
```diff
@@ -255,9 +255,17 @@ func verifyLinkLocalTraffic(t *testing.T, dut *ondatra.DUTDevice, ate *ondatra.A
 	p1 := dut.Port(t, "port1")
 	beforeInPkts := gnmi.Get(t, dut, gnmi.OC().Interface(p1.Name()).Counters().InPkts().State())
 	ate.OTG().StartTraffic(t)
-	time.Sleep(15 * time.Second)
+	if deviations.InterfaceCountersUpdateDelayed(dut) {
```
dplore, line 258: A better pattern is to use gnmi.Watch.Await,  for example:

https://github.com/openconfig/featureprofiles/blob/be17a31486f158e46f857fbe6e9822becb84c8bf/feature/interface/singleton/otg_tests/singleton_test/singleton_test.go#L448-L461

See also the ondatra guidance for how to use Await to reach a goal state:

https://github.com/openconfig/ondatra/blob/b07a95699a4eed8aa606762baadc0a96074c4394/gnmi/gnmi.go#L42-L86

dplore (changes requested):
Please use gnmi.Watch.Await (with the generous 30 second timeout you have specified)  so the test will not hang indefinitely) instead of sleep.

dplore:
@ram-mac handing over to you to finish the review.

feature/interface/ip/ipv6_link_local/otg_tests/ipv6_link_local_test/ipv6_link_local_test.go
karthikeya-remilla: done - please check my change now.

karthikeya-remilla:
@dplore could you please shed some light on why gnmi.Await was replaced by time.Sleep in this line?

https://github.com/openconfig/featureprofiles/blob/cb3f696e54174fa86001245adf002bc888dd380f/feature/interface/ip/ipv6_link_local/otg_tests/ipv6_link_local_test/ipv6_link_local_test.go#L131

This was changed as part of #3443

ram-mac:
> @dplore could you please shed some light on why gnmi.Await was replaced by time.Sleep in this line?
> 
> https://github.com/openconfig/featureprofiles/blob/cb3f696e54174fa86001245adf002bc888dd380f/feature/interface/ip/ipv6_link_local/otg_tests/ipv6_link_local_test/ipv6_link_local_test.go#L131
> 
> This was changed as part of #3443

@karthikeya-remilla i added this static sleep; With gnmi.Await i see the value is not fetched, we can add it back and see why the values are not fetched with gnmi.Await

feature/interface/ip/ipv6_link_local/otg_tests/ipv6_link_local_test/ipv6_link_local_test.go
ram-mac: Changes looks good now.

ram-mac (commented):
@karthikeya-remilla can you add back the gnmi.Await instead of static sleep and then i can approve this PR

karthikeya-remilla:
> @karthikeya-remilla can you add back the gnmi.Await instead of static sleep and then i can approve this PR

Done. Please check now.

ram-mac:
> @ram-mac handing over to you to finish the review.

@dplore - my review is done and i have approved the changes

dplore (approved):

dplore:
@self-maurya any idea why PR approval is not working for this PR?  There are two approvals?

Note, we recently updated codeowners (#3655) to use featureprofiles-approvers as the repo-wide codeowner. Could that change be related to the issue?

self-maurya (approved):

PR #3620, opened 2024-12-04
gNOI-3.1: Changes to complete_chassis_reboot_test

hattikals:
Making changes to script as the gnmi.Get request was failing as it was reaching dut after 120s. Added extra check to make sure , that time elapsed is less than 120s when failing.


Swetha-haridasula (approved):
lgtm

self-maurya (approved):

PR #3624, opened 2024-12-05
DP-1.14: Adding deviation for no zero suppression

charantejag504:
Adding deviation for no zero suppression

1. Sample mode introduced in DP-1.14 requires even zero values to be exported.
2. Removing default zero suppression to support the sample mode validations.


sezhang2 (commented):
Should the PR tile change to **no zero suppression** ?

BTW, the deviation is used for the workaround, but **no zero suppression** is required (not a workaround).

sezhang2:
Please also fix the code:

_Error: feature/qos/otg_tests/qos_basic_test/qos_basic_test.go:1793:9: unnecessary use of fmt.Sprintf (S1039)_

charantejag504:
@sezhang2, updated the PR title.

charantejag504:
Addressed all the comments.

PR #3626, opened 2024-12-06
RT-1.54 bgp_override_as_path_split_horizon_test.go

cprabha:
Hi,

Fixing issue with route policy, it was commented out in previous git pull.
If we dont add condition install protocol eq to BGP, all direct routes will be installed and sent prefixes does not match.

Thanks,
Prabha


dplore (approved):
LGTM if checks pass

PR #3628, opened 2024-12-06
RT-1.51: Add Await when checking for ECMP next hop

yini101:
BGP may take time to propagate the ECMP routes, add Await to wait for the stable state.


dplore (approved):
LGTM if all checks pass

PR #3629, opened 2024-12-06
updated the aft route summary check as the await method was returning…

mastarkey:
the code change from `gnmi.get` to `gnmi.await` when we run with gnmi.await the prefix return value is always 0 so changed to gnmi.watch which is allowing us to gather the actual route information.


sudhinj:
please address the comment updated by sudhinj

karthikeya-remilla:
Obsolete PR. Closing this.

PR #3646, opened 2024-12-13
Avoid adopting potentially unreleased changes in third-party action.

robshakir:
Adopt security recommendation that fixed SHAs are used in GitHub workflows to avoid code changing implementation under our feet.


dplore (approved):

rohit-rp (approved):

dplore:
FYI: @bstoll

PR #3656, opened 2024-12-17
Update sub test case to retain TTL=1 for incoming packets with TTL = 1

vishnureddybadveli:
Update sub test case to retain TTL=1 for incoming packets with TTL = 1


feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -253,6 +253,132 @@ openconfig-network-instance:
...
+Create AFT entries using gRIBI to match on IP and TTL = 1 to
+encapsulate the matching packets in MPLS in UDP with inner ip ttl as 1.
+
+#### gRIBI RPC content
+
+The gRIBI client should send this proto message to the DUT to create AFT
+entries.  See [OC AFT Encap PR in progress](https://github.com/openconfig/public/pull/1153)
+for the new OC AFT model nodes needed for this.  
+
+```proto
```
dplore, line 266: Please revise the format to use what is in TE-18.1.1 (proto specific format)

ref: https://github.com/openconfig/featureprofiles/pull/3656/files#diff-a6b35105c1accdb1889d9c529f4a13d6844d4a9844a90f3d3f736eba728502c5R48-R78

feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -253,6 +253,132 @@ openconfig-network-instance:
...
+            }
+          }
+          encap-headers {
+            encap-header {
+              index: 2
+              src_ip: "outer_ipv6_src"
+              dst_ip: "outer_ipv6_dst_A"
+              dst_udp_port: "outer_dst_udp_port"
+              ip_ttl: "outer_ip-ttl"
+              inner_ip_ttl_min: 1
```
dplore, line 311: Please add a TODO comment as this field does not exist in the OC AFT model (yet) and is also not yet present in gRIBI proto

feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -253,6 +253,132 @@ openconfig-network-instance:
...
+        }
+      }
+    }
+  }
+}
+```
+
+#### Procedure - A
+* Send traffic from ATE port 1 to DUT port 1 with inner packet TTL as 1.
+* Validate afts next hop counters
```
dplore, line 377: For this subtest, we are only checking the IP TTL, so this next hop counter validation can be removed.

feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -253,6 +253,132 @@ openconfig-network-instance:
...
+* Using OTG, validate ATE port 2 receives MPLS-IN-UDP packets
+  * Validate destination IPs are outer_ipv6_dst_A and outer_ipv6_dst_B
+  * Validate MPLS label is set
+  * Validate inner packet ttl as 1.
```
dplore, line 381: Add a validation for the outer ttl as well (it should also be the value specified in "outer_ip-ttl")

dplore (dismissed):

dplore:
@robshakir for 2nd review+approval since other gribi feature owners are unavailable or not connected to this use case.

feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -253,6 +253,57 @@ openconfig-network-instance:
...
+                "policy": "retain ttl",
+                "rules": [
+                  {
+                    "config": {
+                      "sequence-id": 1,
+                    },
+                    "ipv6": {
+                      "config": {
+                        "destination-address": "router_ip"
+                        "ip-ttl": 1
```
dplore, line 282: This will force the outer packet ttl to be set to 1 no matter what the inner/payload packet's TTL is.  See:

https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-afts-next-hops-next-hop-encap-headers-encap-header-udp-v4-state-ip-ttl

There currently isn't any OC config path or gRIBI proto defined to implement the description of this new subtest TE-18.1.5.  You can propose a new field

feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -253,6 +253,57 @@ openconfig-network-instance:
 * Generate traffic from ATE port 1
 * Validate ATE port 2 receives the innermost IPv4 traffic with correct VLAN and inner_decap_ipv6
 
+### TE-18.1.5 Rewrite the ingress innner packet TTL = 1, if the incoming TTL = 1.
+Canonical OpenConfig for policy forwarding, matching IP prefix and TTL = 1 with action
+redirect and inner packet TTL = 1.
```
dplore, line 258: I think this is intended to be programmed using gRIBI.modify, correct?  The JSON below looks like what would be used in a gnmi.Set.  You'll need some proto encoded structure like in TE-18.1.1.  You should be able to copy TE-18.1.1 and add a new (proposed) field for this special TTL behavior.  Perhaps named something like  `preserve_ttl_1` or `ttl_1_exception`.

feature/gribi/otg_tests/mpls_in_udp/README.md
dplore: See comment above regarding adding a TODO for a new gRIBI field like `preserve_ttl_1`

dplore (changes requested):

feature/gribi/otg_tests/mpls_in_udp/README.md
vishnureddybadveli: hmm... I think this will be gnmi set since traffic policy will be used

feature/gribi/otg_tests/mpls_in_udp/README.md
vishnureddybadveli: I agree but we go by traffic policy so do we need to propose a new gRIBI field?

feature/gribi/otg_tests/mpls_in_udp/README.md
arkaab: this will be via vendor CLI.

feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -253,6 +253,57 @@ openconfig-network-instance:
...
+                      "sequence-id": 1,
+                    },
+                    "ipv6": {
+                      "config": {
+                        "destination-address": "router_ip"
+                        "ip-ttl": 1
+                      }
+                    },
+                    "action": {
+                      "next-hop-group": "nhg" #TODO: Need to add to model as nhg will be doing encap and setting TTL.
```
arkaab, line 286: Is this already a field in NHG ? 
Is the proposal here to have a separate NHG just for cloud router traffic ?

arkaab: Modeling this as an encap parameter is fine.

feature/gribi/otg_tests/mpls_in_udp/README.md
vishnureddybadveli: I agree its not going to be gRIBI per se, although immediate usage is vendor CLI.

feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -253,6 +253,57 @@ openconfig-network-instance:
...
+                },
+                "policy": "retain ttl",
+                "rules": [
+                  {
+                    "config": {
+                      "sequence-id": 1,
+                    },
+                    "ipv6": {
+                      "config": {
+                        "destination-address": "router_ip"
```
jaingaurav12, line 281: Can we please make this a generic ? any packets w/ ttl 1 ...... why Router IP ?

feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -253,6 +253,57 @@ openconfig-network-instance:
...
+Canonical OpenConfig for policy forwarding, matching IP prefix and TTL = 1 with action
+redirect and inner packet TTL = 1.
+
+```json
+{
+  "openconfig-network-instance": {
+    "network-instances": [
+      {
+        "afts": {
+          "policy-forwarding": {
```
jaingaurav12, line 266: What about inner IPv4 packets

feature/gribi/otg_tests/mpls_in_udp/README.md
vishnureddybadveli: We agreed upon match with (Router IP + TTL=1) as we thought there could be use case packets(for example trace route, customer packets) having TTL = 1 not destined to Router.

feature/gribi/otg_tests/mpls_in_udp/README.md
vishnureddybadveli: Thank you,

There was an effort from partner team to model a nhg as an encap parameter.(https://github.com/openconfig/public/compare/master...danameme:public:temp1-static-gue)

I was planning to piggyback on it by adding the support for setting TTL in nhg.

feature/gribi/otg_tests/mpls_in_udp/README.md
vishnureddybadveli: Thank you,

This is complicated currently I do not see a way to do this in a single instruction, my proposal is there would be an additional separate config that would have been pre-provisioned for the HW pipeline to look into inner IPv4 packets. I do not have a way to represent this as of now.

feature/gribi/otg_tests/mpls_in_udp/README.md
vishnureddybadveli: Thanks Darren

I do not see any reason we will be using gRIBI.modify.

feature/gribi/otg_tests/mpls_in_udp/README.md
vishnureddybadveli: Thank you,
The proposal says match on (Router IP + outer IP TTL=1) but specifies action as set inner packet TTL=1.

feature/gribi/otg_tests/mpls_in_udp/README.md
vishnureddybadveli: It was decided to drop redirect action instead use action as set ttl = 1

vishnureddybadveli:
Darren and jaingaurav12,

May I request your approval and merge of the changes.

feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -183,6 +183,57 @@ encapsulate in GRE.
...
+                "policy": "retain ttl",
+                "rules": [
+                  {
+                    "config": {
+                      "sequence-id": 1,
+                    },
+                    "ipv6": {
+                      "config": {
+                        "destination-address": "router_ip"
+                        "ip-ttl": 1
```
trlongth, line 212: There seems to be an existing field `hop-limit` that match the description with what we are looking here.
Can we use that insetad:
https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-policy-forwarding-policies-policy-rules-rule-ipv6-config-hop-limit

feature/gribi/otg_tests/mpls_in_udp/README.md
dplore: Ah, I see. Got it.  The action of `set-ip-ttl` sounds good.  I don't see a TODO in the README to add that to the OC models though.

feature/gribi/otg_tests/mpls_in_udp/README.md
dplore: Thanks @trlongth.  

The rule should include the existing [hop-limit](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-policy-forwarding-policies-policy-rules-rule-ipv6-config-hop-limit) field.

The action should include a proposed `set-ip-ttl` field.

trlongth: So for my understanding, we should:
- Replace the "ip-ttl" in the rule with "hop-limit".
- Propose a new "set-ip-ttl" in the action.

feature/gribi/otg_tests/mpls_in_udp/README.md
vishnureddybadveli: Thanks Darren

I agree currently I did not add a OC yet. 
I will create a PR and add path.

trlongth (approved):

dplore (approved):

nandanarista (commented):
the policy lgtm

feature/gribi/otg_tests/mpls_in_udp/README.md
```diff
@@ -183,6 +183,57 @@ encapsulate in GRE.
 * Validate ATE port 2 receives GRE traffic with correct inner and outer IPs
 
 
+### TE-18.1.5 Rewrite the ingress innner packet TTL = 1, if the incoming TTL = 1.
```
xw-g, line 186: Is this test already implemented in code? If not, let's add a [TODO] here.

xw-g (dismissed):

dplore (approved):

PR #3662, opened 2024-12-19
Modified TC5 final port up check as timing was causing flaky results

mastarkey:
This PR increases the final verification wait time in the TC5 "short down" scenario from 10 seconds to 30 seconds. The README does not mandate strict upper limits for this step, only that the interface should be stable after the short down event. Extending the wait time provides a sufficient buffer to confirm that the DUT has fully settled in the "UP" state, reducing test flakiness without violating any documented requirements.


self-maurya (dismissed):

ram-mac (dismissed):
LGTM

self-maurya (approved):

ram-mac (approved):

PR #3664, opened 2024-12-19
Remove RT-1.26.3 static route preference

dplore:
The functionality in the README is not correct and the feature as stated is not required.


Swetha-haridasula:
Why is the functionality not correct? Could you eleborate?
 The feature is required as we utilize preference/AD for HA of management links

dplore:
This is an incorrect use of preference.  Preference is defined as selecting the desired route between different protocols. (for example: static vs. BGP vs. ISIS).   Preference is not defined (in OC) as selection criteria between two routes in the same protocol.   

Reference: 
https://github.com/openconfig/public/blob/850e572ae45613d8f29f25ac8d87a4d51db2f8fc/release/models/local-routing/openconfig-local-routing.yang#L174-L185

Swetha-haridasula:
@dplore  In case management HA CPN topology, we have default route being received by BGP and we have a backup static default route as well. The design is to use the iBGP default route as more preferred route over the static route. Hence the AD of static route is modified. Please refer to b/373412048 for recent CPN HA design change as well as 'https://docs.google.com/document/d/1l9KsbhgzLb3lXgxsledrqktYIam5TBoMqvy5em3m5hU/edit?tab=t.0#heading=h.jzghbleb96qw ' where option#1 was finalized.

dplore:
We should use a next hop with a higher metric to get the backup static
route behavior.

@Rafal Szarecki ***@***.***> for any additional comments

Swetha-haridasula:
@dplore but AD is the first things that gets looked into for route selection (When the prefix is received from multiple protocols).
The default route is available via static and iBGP. If we did not modify the AD then static route is choosen even if the metric of the route is modified here.
Requested mellowd@google.com also inputs in bug as he knows the CPN design here

rszarecki:
If I understand correctly we have one or more default BGP routes AND static
default route with single next-hop. The intention is that static route
serves as backup - so-stalled floating static.

In this case preference of static should be modified to be > then BGP.
Probably 220 will just work.

We use above in MVC management HA.

--------------
Rafal Szarecki

rszarecki:
No.
First think is to select best route in each protocol independently, using
protocols specific attributes  - locPref, as-path Len, med for BGP. SPF
path metric for Isis, next-hop metric for static.
Then AD is the only metric used to select between protocols.

IMHO the OpenConfig model for static is ambiguous, because it allows 2
attributes - metric and preference per each next-hop.
W/o standardization which one is more important.
And inconsistent with other protocols, which do not allow for distinct AD
per path/next-hop or even per prefix.
IMHO preference/AD should be attribute of prefix, not of next-hop. This
will solve ambiguity, while reducing flexibility a bit. Still more flexible
than other protocols.

Or maybe we need better describe use and behavior of static’preference as.
- not used to select best among static route next-hops
- carried by selected best static route to cross-protocol comparison of
per-protocol best routes which selects system-best-and-active route.





--------------
Rafal Szarecki

Swetha-haridasula:
@rszarecki . Yes I agree of your opinion _of better describe use and behavior of static’preference as' carried by selected best static route to cross-protocol comparison of
per-protocol best routes which selects system-best-and-active route'_ . 
I do agree it is confusing as the path here '/network-instances/network-instance/protocols/protocol/static-routes/static/next-hops/next-hop/config/preference' might be mis-judging due to the term next-hop ( though i believe it may be because only for static route we have the privilege/mandate of mentioning next-hop unlike for any other igp/egp's).

But the usage of preference/AD for management HA is necessary. @dplore , please let me know of your thoughts

Swetha-haridasula:
closing as this feature is currently used in CPN management HA. kindly feel free to re-open in case of discrepency

PR #3665, opened 2024-12-20
updating header's and adding flow parameters

rahulsy26:


sancheetaroy:
Please correct all static analysis errors in the file.

rahulsy26:
Looks like there is another PR raised which have same and new changes with test. So closing this bug which is duplicated to the below.

https://github.com/openconfig/featureprofiles/pull/4147

PR #3667, opened 2024-12-20
System-1: System testing : Added Deviation

mnp-juniper:
Resolved conflicts for #3416


Swetha-haridasula:
needs changes to align with other vendors and general use usecase. 
1. Juniper must first add new deviation GNMIGetOnRootUnsupported as in PR#2605
2. Then add 3 Subtests. First for state, second for config, 3rd for operational
3. For first subtest add another deviation GNMIGetOnStateUnsupported and add the alternative for it which is using gnmi subscribe I believe
3. For second subtest deviation is not needed for Juniper
4. For 3rd subtest add a deviation and skip it if there is no alternative available

PR #3672, opened 2024-12-27
Bug fixes: 1) time.Duration function wrong unit; 2) transceiver disable OC path

jianchen-g:
bug fixes: 1) in zr_laser_bias_current_test file, the time.Duration function returns value in nanosecond by default, needs to convert it to second; 2) in zr_low_power_mode_test file, to use the transceiver disable OC path, the input needs to be transceiver name instead of interface name.


Ankur19 (approved):
LGTM

dplore (approved):

PR #2855, opened 2024-03-27
System-1: System testing: Updates to g_protocol_test.go

cakhil45:
1.  Added encoding type in GNMI get request 
2. Authz getpolicy is used without adding any policy. This leads to RPC error.  Added code to handle and validate the RPC error.

PR #2605, opened 2024-01-31
System-1: added p4rt capabilities deviation and fixed gnmi call

kjahed:
This PR modifies System-1:
- In TestP4RTClient: add P4RT capabilities unsupported deviation for CISCO and use a Read request instead.
- In TestGNMIClient: make a get request for a well-supported leaf.

PR #3674, opened 2025-01-02
system_base_test: run TestGNMIClient for defined path

brianneville:
In PR #2855 the commit
https://github.com/openconfig/featureprofiles/commit/87f3397bc322ec35e34a470cd793b070db882e63 made a change to the path being requested when the  connection is being tested. 
Previously an empty GetRequest was being sent, which meant that the GetResponse would also be empty. 
This was changed so that now the test is requesting the root node.

However, its possible that the GetResponse for the root node is larger than the maximum response size for which the client is configured (i.e. the gRPC  default max response size of 4MiB)

To get around this we should run the request to a defined path so that we are guaranteed a small response. The response is ignored anyway, so it doesnt matter what path we request here.

Use the `gnmi_get_on_root_unsupported` introduced in https://github.com/openconfig/featureprofiles/pull/2605 to avoid performing a get on the root node

### DIFF
--- a/feature/system/system_base_test/tests/system_g_protocol_test/metadata.textproto
+++ b/feature/system/system_base_test/tests/system_g_protocol_test/metadata.textproto
@@ -22,3 +22,11 @@ platform_exceptions: {
     gnmi_get_on_root_unsupported: true
   }
 }
+platform_exceptions: {
+  platform: {
+    vendor: ARISTA
+  }
+  deviations: {
+    gnmi_get_on_root_unsupported: true
+  }
+}


self-maurya:
@brianneville could you please resolve the conflicts

brianneville:
> @brianneville could you please resolve the conflicts

Done in the latest patch version.
I'm now getting github actions failing with unrelated errors:
"""
Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: 704facf57e6136b1bc63b828d79edcd491f0ee84`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down
"""

self-maurya (approved):

PR #3679, opened 2025-01-06
RT-1.54: Remove add path configuration

yini101:
Disable add path on ixia since the test does not rely on add path.


dplore:
@yini101 , note this does not pass on cEOS:

https://gist.github.com/dplore/034940b29d45d5b67846269f90109fdf

dplore:
Please address the issue in the attached gist.  Thanks!

yini101:
@dplore Thank you for the review! 
```failed while setting /routing-policy: failed while unmarshaling Container routing-policy: {"openconfig-routing-policy:policy-definitions":{"policy-definition":[{"config":{"name":"ALLOW"},"name":"ALLOW","statements":{"statement":[{"actions":{"config":{"policy-result":"ACCEPT_ROUTE"}},"conditions":{"config":{"install-protocol-eq":"openconfig-policy-types:BGP"}},"config":{"name":"ALLOW"},"name":"ALLOW"}]}}]}}: error decoding JSON value for node routing-policy/policy-definitions/policy-definition[name='ALLOW']/statements/statement[name='ALLOW']/conditions/config: An unexpected element "install-protocol-eq" is present``` It was failing because of `install-protocol-eq` configuration.

Arista doesn't support or need `install-protocol-eq` leaf to suppress other routes from being accepted. Could you please consider adding deviation for it because other vendors may need it as per #3626.

ram-mac (approved):
The changes looks good and it is inline with the README

self-maurya (approved):

PR #3683, opened 2025-01-07
fix counters for interfaces

ram-mac:
Below changes are to address gaps in verifying counters on the device
1. Counters in-broadcast-pkts, in-discards, in-errors, in-fcs-errors, in-multicast-pkts, out-broadcast-pkts, out-discards, out-errors, out-fcs-errors, out-multicast-pkts are added;
2. Counter 'InUnicastPkts' was repeated in ln-129 and ln-117; So the counters was supposed to be InUnicastPkts in ln-129 and this is now fixed
3. Some of the subinterfaces counters were skipped in the test. Now these are fixed


feature/gnmi/otg_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
```diff
@@ -548,3 +603,39 @@ func waitOTGARPEntry(t *testing.T) {
 	}).Await(t)
 
 }
+
+func (tc *testCase) setupAggregateAtomically(t *testing.T) {
```
dplore, line 607: this should be moved to  internal/cfgplugins/interface.go and represent the generate config part of the work.  

In the Test function, a gnmi.Replace can use the output of the cfgplugin

dplore (changes requested):

feature/gnmi/otg_tests/telemetry_interface_packet_counters_test/telemetry_interface_packet_counters_test.go
ram-mac: Darren, actually these functions are not used in this test, will remove it from here.

ram-mac: @dplore - can you please approve this PR? i have made the relevant changes.

dplore (dismissed):

dplore (approved):

singhavnish2516:
Hi @ram-mac / @dplore ,
We have merged these changes as part of https://github.com/openconfig/featureprofiles/pull/4110. I think we can close this PR.

PR #3723, opened 2025-01-28
RT_1.54 bgp_override_as_path_split_horizon_test.go

cprabha:
Hi,

updating test per discussion, removing install protocol bgp in configure route policy.
If we dont configure install protocol bgp then all direct, interface routes will be advertised and verify bgp prefix might not be consistent. Hence removed the code for the same.


feature/bgp/otg_tests/bgp_override_as_path_split_horizon_test/bgp_override_as_path_split_horizon_test.go
```diff
@@ -215,21 +215,6 @@ func advBGPRouteFromOTG(t *testing.T, args *otgTestArgs, asSeg []uint32) {
 	time.Sleep(30 * time.Second)
 }
 
-// verifyPrefixesTelemetry confirms that the dut shows the correct numbers of installed,
-// sent and received IPv4 prefixes.
-func verifyPrefixesTelemetry(t *testing.T, dut *ondatra.DUTDevice, nbr string, wantInstalled, wantSent uint32) {
-	t.Helper()
```
ram-mac, line 221: @cprabha - why are we removing the verifyPrefixesTelemetry; this i believe is going to validate whether prefixes received/installed on the device.

cprabha: @ram-mac  If  we dont enable install protocol BGP , It will install direct/interface routes into bgp table.  And Sent prefixes count does not as expected. This is the reason i have enabled install protocol BGP, this will install only bgp routes and avoid installing direct routes

cprabha: @ram-mac plz check now , i have added vendor check for route policy.  please let me know if this looks good

feature/bgp/otg_tests/bgp_override_as_path_split_horizon_test/bgp_override_as_path_split_horizon_test.go
```diff
@@ -240,6 +240,9 @@ func configureRoutePolicy(t *testing.T, dut *ondatra.DUTDevice, name string, pr
 		t.Fatalf("AppendNewStatement(%s) failed: %v", name, err)
 	}
 	stmt.GetOrCreateActions().PolicyResult = pr
+	if dut.Vendor() == ondatra.JUNIPER {
```
ram-mac, line 243: Hi,
 can you add this as part of a deviation for Juniper?
 So we can track it separately on this one.

ram-mac (changes requested):

feature/bgp/otg_tests/bgp_override_as_path_split_horizon_test/bgp_override_as_path_split_horizon_test.go
cprabha: @ram-mac modified it to prefix set, its there in other scripts . hope this works . please check

ram-mac: @cprabha - this way of handling is not acceptable. particularly "if dut.Vendor() == ondatra.JUNIPER" part, the code should be vendor agnostic and if there is some issue with the OC path then it has to be handled differently by adding deviation. In this case i do not see any changes.

ram-mac (changes requested):

feature/bgp/otg_tests/bgp_override_as_path_split_horizon_test/bgp_override_as_path_split_horizon_test.go
cprabha: @ram-mac please refresh and check , I have added prefix set instead of install protocol.

ram-mac (approved):
LGTM, Validated against vendors and the tests are passing and no regression introduced due to this PR.

nikki-blr (approved):

rohit-rp (approved):

PR #3725, opened 2025-01-28
AnCX tests correction

AmrNJ:
bgp_override_as_path_split_horizon_test - Removed install-protocol-eq which is not needed
isis_interface_hello_padding_enable_test - Increasing timeout to avoid flakiness


ram-mac (approved):
Looks good

self-maurya (approved):

PR #3744, opened 2025-01-31
Remove deviation ttl_copy_unsupported

xw-g:
This doesn't seems like a good usage of deviation (bypass one of the focus of the test).

### DIFF
--- a/feature/gribi/otg_tests/basic_encap_test/metadata.textproto
+++ b/feature/gribi/otg_tests/basic_encap_test/metadata.textproto
@@ -22,7 +22,6 @@ platform_exceptions:  {
     vendor: JUNIPER
   }
   deviations:  {
-    ttl_copy_unsupported: true
   }
 }
 platform_exceptions:  {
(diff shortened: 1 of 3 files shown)


xw-g:
Internal b/392061912.

feature/gribi/otg_tests/basic_encap_test/metadata.textproto
```diff
@@ -22,7 +22,6 @@ platform_exceptions:  {
     vendor: JUNIPER
   }
   deviations:  {
-    ttl_copy_unsupported: true
```
robshakir, line 25: Should we also remove this from `metadata.proto` and also https://github.com/openconfig/featureprofiles/blob/ddc6255868db1fbb7eb917e48771800f119d0963/internal/deviations/deviations.go#L739 ?

xw-g: I was thinking about separate CL to clean up the code. This being clean up the metadata.

robshakir: SGTM.

robshakir (approved):

xw-g:
b/392061912 is already conclusive, hence I'm going to force the merge.

PR #3755, opened 2025-02-07
MTU-1.5: Path MTU handing

nsadhasivam:
MTU-1.5: Path MTU handing
- Tests ensures that DUT generates ICMP "Fragmentation Needed and Don't Fragment was Set" for packets exceeding egress interface MTU.

MTU-1.5.1 IPv4 Path MTU (for 2000, 4000, 9000 bytes) verify
- ATE Port-1 receives ICMP type-3, code-4
- packet-processing-aggregate report fragment packet discards
- fragment-total-drops increments as per the amount of packets sent
- low CPU (<20%) utilization on controllers

Similarly for MTU-1.5.2 IPv6 Path MTU


proto/metadata.proto
```diff
@@ -700,6 +700,12 @@ message Metadata {
     // Cisco: b/385298159
     bool gnmi_get_on_root_unsupported = 251;
 
+    // Device does not support packet processing aggregate drops.
```
ram-mac, line 703: @nsadhasivam - can you create bug for these deviations.

feature/mtu/otg_tests/pmtu_handing/metadata.textproto
```diff
@@ -0,0 +1,18 @@
...
+plan_id: "MTU-1.5"
+description: "Path MTU handing"
+testbed: TESTBED_DUT_ATE_2LINKS
+platform_exceptions: {
+  platform: {
+    vendor: NOKIA
+  }
+  deviations: {
+    explicit_port_speed: true
+    explicit_interface_in_default_vrf: true
```
ram-mac, line 14: @nsadhasivam - i do not see packet_processing_aggregate_drops_unsupported marked in this metadata.textproto. Is this available now?

nsadhasivam: packet_processing_aggregate_drops_unsupported is not needed for Nokia, but it would be used for Arista.

proto/metadata.proto
nsadhasivam: Sure, after pr merge with share tracking bug to vendor for oc path support.

feature/mtu/otg_tests/pmtu_handing/metadata.textproto
ram-mac: Can this deviation be added to the Vendor Arista in this case, if it is not added the deviations will be false by default. Iam not sure how it is handled in the code.

proto/metadata.proto
ram-mac: @nsadhasivam the bug need to be added in the metadata.proto so it the reference is there. Create a placeholder  bug and you can add the details later.

ram-mac (changes requested):

feature/mtu/otg_tests/pmtu_handing/metadata.textproto
nsadhasivam: When the script is enabled for cisco the textproto file can be updated.

proto/metadata.proto
nsadhasivam: updated tracking bug

ram-mac (approved):

ram-mac (approved):

rohit-rp (dismissed):

Swetha-haridasula (dismissed):
lgtm

nikki-blr (approved):

PR #3756, opened 2025-02-07
adding deviation for cisco delimiter, and skipping empty string 

fmolinar:
Adding the deviation for the banner test as Cisco devices expect a delimiter to be sent, and skipping the empty string banner test since Cisco currently doesnt support empty string send as banner.


feature/system/tests/system_base_test/banner_test.go
```diff
@@ -30,31 +31,32 @@ import (
...
 			gnmi.Replace(t, dut, config.Config(), testCase.banner)
 
 			t.Run("Get MOTD Config", func(t *testing.T) {
-				if testCase.banner == "" {
-					if gnmi.LookupConfig(t, dut, config.Config()).IsPresent() {
-						t.Errorf("MOTD Banner not empty")
-					} else {
-						t.Logf("No response for the path is expected as the config is empty")
+
+				if testCase.description == "Empty String" {
```
robshakir, line 53: We should not be matching on the description of a test case. This should be based on the content if we need to.

feature/system/tests/system_base_test/banner_test.go
```diff
@@ -66,15 +68,18 @@ func TestMotdBanner(t *testing.T) {
 			})
 
 			t.Run("Get MOTD Telemetry", func(t *testing.T) {
-				if testCase.banner == "" {
-					if gnmi.LookupConfig(t, dut, config.Config()).IsPresent() {
-						t.Errorf("MOTD Telemetry Banner not empty")
-					} else {
-						t.Logf("No response for the path is expected as the config is empty")
+				if testCase.description == "Empty String" {
```
robshakir, line 71: Ditto the above.

feature/system/tests/system_base_test/banner_test.go
```diff
@@ -66,15 +68,18 @@ func TestMotdBanner(t *testing.T) {
...
+					if dut.Vendor() != ondatra.CISCO {
```
robshakir, line 72: Please do not use vendor matches, use deviations to indicate non-compliance.

feature/system/tests/system_base_test/banner_test.go
```diff
@@ -97,31 +102,30 @@ func TestMotdBanner(t *testing.T) {
...
 			gnmi.Replace(t, dut, config.Config(), testCase.banner)
 
 			t.Run("Get Login Banner Config", func(t *testing.T) {
-				if testCase.banner == "" {
-					if gnmi.LookupConfig(t, dut, config.Config()).IsPresent() {
-						t.Errorf("Config Login Banner not empty")
-					} else {
-						t.Logf("No response for the path expected is expected as the config is empty")
+				if testCase.description == "Empty String" {
+					if dut.Vendor() != ondatra.CISCO {
```
robshakir, line 123: Same comments as above, do not use vendor matches or matches on test case descriptions. This should all be gated by deviations.

robshakir (changes requested):

feature/system/tests/system_base_test/banner_test.go
fmolinar: I made changes in the new PR

feature/system/tests/system_base_test/banner_test.go
fmolinar: I created a deviation in the new PR

mastarkey:
Hi @robshakir can you trigger the CI/CD checks here so we have an idea if those are passing.

fmolinar:
outdated

PR #3759, opened 2025-02-11
RT-5.8: IPv6 Link Local: Added Sleep of 15sec after start traffic to have the got packets flow in DUT.

priyaranjannanda:
- Added Sleep of 15sec after start traffic to have the packets flow in DUT.


feature/interface/ip/ipv6_link_local/otg_tests/ipv6_link_local_test/ipv6_link_local_test.go
```diff
@@ -256,6 +256,7 @@ func verifyLinkLocalTraffic(t *testing.T, dut *ondatra.DUTDevice, ate *ondatra.A
 	p1 := dut.Port(t, "port1")
 	beforeInPkts := gnmi.Get(t, dut, gnmi.OC().Interface(p1.Name()).Counters().InPkts().State())
 	ate.OTG().StartTraffic(t)
+	time.Sleep(15 * time.Second)
```
LimeHat, line 259: there's already gnmi.Watch with the 30-second timer on the very next line. If that 30s is not sufficient, you can just increase that value; no need to add a blocking sleep

LimeHat (changes requested):

rohit-rp (approved):

PR #3799, opened 2025-02-11
Updated as per #3401

nsadhasivam:
Update Subtest1 to cover MTU-1.3.1: Test with Physical and Bundle interfaces


self-maurya (approved):

Swetha-haridasula (approved):
lgtm

PR #3801, opened 2025-02-12
gNMI-1.10-Add support for GetAll gnmi requests

AmrNJ:
Support added to check the following paths

 - /interfaces/interface[name=*]/state/
 - /interfaces/interface[name=*]/ethernet/
 - /components/component[name=*]/transceiver/


Swetha-haridasula (approved):
lgtm

PR #3812, opened 2025-02-13
Added route reflector scale test

self-maurya:


rohit-rp (dismissed):

feature/bgp/routereflector/otg_tests/route_reflector_scale/route_reflector_scale_test.go
```diff
@@ -0,0 +1,532 @@
...
+			for k := 0; k < 25; k++ {
+				ips = append(ips, netip.MustParsePrefix(fmt.Sprintf("172.%d.%d.%d/32", i, j, k)))
+			}
+		}
+	}
+
+	return ips
+}
+
+func createInternetPrefixesV6(t *testing.T) []netip.Prefix {
```
ram-mac, line 493: There is one issue with the Internet prefixes, the prefixes generated in the script is contiguous ones it does not represent the Internet routes both ipv4 and ipv6
Now the question is how to add these routes. There are 3 ways we can do it
1) The entire internet table can be downloade from the potaro.net and it the routes can be read from that file which will include all the different as path attributes;
2) Have a internet peer in the lab to which we could establish a connection with and then it will have all the Internet routes
3) Use Ixia to advertise the Internet routes to DUT and then readvertise it accordingly.

feature/bgp/routereflector/otg_tests/route_reflector_scale/route_reflector_scale_test.go
```diff
@@ -0,0 +1,532 @@
...
+}
+
+func createInternetPrefixesV4(t *testing.T) []netip.Prefix {
+	t.Helper()
+	var ips []netip.Prefix
+	// Create 1500000 /32s - internet prefixes
+	for i := 50; i < 74; i++ {
+		for j := 0; j < 250; j++ {
+			for k := 0; k < 25; k++ {
+				ips = append(ips, netip.MustParsePrefix(fmt.Sprintf("182.%d.%d.%d/32", i, j, k)))
```
ram-mac, line 476: @self-maurya - Can you modify this to add a /24 prefix set because /32 is host routes

feature/bgp/routereflector/otg_tests/route_reflector_scale/route_reflector_scale_test.go
```diff
@@ -0,0 +1,532 @@
...
+			}
+		}
+	}
+
+	// Create 500000 /32s - internal prefixes
+	for i := 16; i < 24; i++ {
```
ram-mac, line 482: @self-maurya - Can you modify these to advertise routes as /30 prefixes

feature/bgp/routereflector/otg_tests/route_reflector_scale/route_reflector_scale_test.go
```diff
@@ -0,0 +1,532 @@
...
+		for i := 0; i < 1000; i++ {
+			ip := netip.MustParsePrefix(fmt.Sprintf("2001:db8:%d:%d::/64", i, j))
+			ips = append(ips, ip)
+		}
+	}
+
+	// Create 400000 /64s - internal prefixes
+	for j := 10; j < 50; j++ {
+		for i := 0; i < 1000; i++ {
+			ip := netip.MustParsePrefix(fmt.Sprintf("fc00:abcd:%d:%d::/64", i, j))
```
ram-mac, line 507: @self-maurya - Please modify these to /126 prefixes as they would consume higher memory footprint on the device.

ram-mac (changes requested):

feature/bgp/routereflector/otg_tests/route_reflector_scale/route_reflector_scale_test.go
self-maurya: As discussed internally, I've have changed the prefixes to /30 for IPv4 and /126 for IPv6.

feature/bgp/routereflector/otg_tests/route_reflector_scale/route_reflector_scale_test.go
self-maurya: fixed

feature/bgp/routereflector/otg_tests/route_reflector_scale/route_reflector_scale_test.go
self-maurya: fixed

rohit-rp (approved):

PR #3816, opened 2025-02-14
g_protocol_test- added type config 

Sagrohi:
Script was missing type config because of that it was failing, so added type config.

### DIFF
--- a/feature/system/system_base_test/tests/system_g_protocol_test/system_g_protocol_test.go
+++ b/feature/system/system_base_test/tests/system_g_protocol_test/system_g_protocol_test.go
@@ -76,7 +76,7 @@ func TestGNMIClient(t *testing.T) {
 			Encoding: gpb.Encoding_JSON_IETF,
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


feature/system/system_base_test/tests/system_g_protocol_test/system_g_protocol_test.go
```diff
@@ -76,7 +76,7 @@ func TestGNMIClient(t *testing.T) {
 			Encoding: gpb.Encoding_JSON_IETF,
 		}
 	} else {
-		req = &gpb.GetRequest{Encoding: gpb.Encoding_JSON_IETF, Path: []*gpb.Path{{Elem: []*gpb.PathElem{}}}}
+		req = &gpb.GetRequest{Encoding: gpb.Encoding_JSON_IETF, Path: []*gpb.Path{{Elem: []*gpb.PathElem{}}}, Type: gpb.GetRequest_CONFIG}
```
LimeHat, line 79: This should be a deviation

LimeHat (changes requested):
This was already discussed in #3265 with a colleague of yours. gNMI specification does not impose such restrictions, and this should be gated behind a deviation.

Swetha-haridasula:
This is not needed as we are getting it on parent path

PR #3827, opened 2025-02-18
DP-1.17: Add deviation and test configuration

yini101:
Add deviation interfcae_enabled and interface IPv4 configuration


feature/qos/ecn/otg_tests/dscp_transparency_test/dscp_transparency_test.go
```diff
@@ -363,6 +363,12 @@ func configureDUTPort(
 ) {
 	gnmiOCRoot := gnmi.OC()
 
+	intf := portAttrs.NewOCInterface(port.Name(), dut)
```
self-maurya, line 366: where we are using this intf and s variable

self-maurya (changes requested):

feature/qos/ecn/otg_tests/dscp_transparency_test/dscp_transparency_test.go
self-maurya: @yini101 please look into this, check if added code snippet is required or not.

yini101: I have updated line 376 to use intf. Thanks for pointing it out.

self-maurya (approved):

PR #3831, opened 2025-02-19
increase the timeout for addressing flakiness

ram-mac:
Many times the interfaces are not coming up after the power enable/disable and the test is failing since the Interfaces are not available after a reboot.
So, increase the timeout to 8minutes vs 5minutes


Swetha-haridasula (approved):
lgtm

self-maurya (approved):

PR #3834, opened 2025-02-20
fixed README.md related to global AFI-SAFI level behavior and also added the neighbor AFI-SAFI under global

ram-mac:
1)README.md is modified as per the Google requirement
2)Added the neighbor AFI/SAFI under global.

Have validated the changes against all vendors and they are all working as intended with no failures due to these changes.


rohit-rp (approved):

nsadhasivam (approved):
LGTM!

PR #3836, opened 2025-02-20
fixing ipv4_entry_test

ram-mac:
Fixing ipv4_entry_test
1. replaced the instance creation using utility fptest
2. few changes in staticARPWithMagicUniversalIP function, gnmi.replace was changed to gnmi.update


Swetha-haridasula (approved):
lgtm

self-maurya (approved):

PR #3848, opened 2025-02-25
P4RT Table entries fixed to only send TTL1 

ArpitJoshi18:
P4RT entries are modified to only send trace-route(V4+V6) entries with TTL=1. 
Both modified scripts are now installing P4RT entries for GDP,LLDP and trace-route (V4+V6) with TTL1. 

P4RT-performance_test already takes care of it. 

May need to open it for P4RT-traceroute_packetin_with_vrf_selection_test as well since it also installs TTL0.


Swetha-haridasula (approved):
LGTM

gggsmith (approved):
LGTM

PR #3850, opened 2025-02-27
Add deviations back

yini101:
Add the following deviations back:
- unsupported_static_route_next_hop_recurse: true
- skip_setting_disable_metric_propagation: true
- skip_bgp_send_community_type: true
- default_import_export_policy_unsupported: true 
as all of them are needed by Arista.


feature/bgp/policybase/otg_tests/chained_policies_test/metadata.textproto
```diff
@@ -15,6 +15,7 @@ platform_exceptions: {
     default_network_instance: "default"
     missing_value_for_defaults: true
     skip_setting_statement_for_policy: true
+    default_import_export_policy_unsupported: false
```
rohit-rp, line 18: Shouldn't this be true

yini101: Right, thanks for catching this. I've updated it.

rohit-rp (approved):

self-maurya (approved):

PR #3855, opened 2025-02-28
Modifications to isisRT2.14-isis_drain_test

divyarema:
Added aft checks before checking traffic


rohit-rp:
@divyarema If we need to enhance an Automation we should also make relevant additions to the README for consistency. Please can you add AFT check to the readme. Also I want to know how did get this requirement?

divyarema:
> @divyarema If we need to enhance an Automation we should also make relevant additions to the README for consistency. Please can you add AFT check to the readme. Also I want to know how did get this requirement?

@rohit-rp  Script validates traffic after route advertisement /modifications of metric without route check.To avoid intermittent traffic loss seen during test, added AFT check prior traffic verification.

PR #3857, opened 2025-02-28
Rt 1.23

fmolinar:
Reverted the deviation: bgp_global_extended_next_hop_encoding_unsupported
updated the get afi safi state with lookup, because of the case of enabled: false. By using lookup instead of get, ondatra is able to handle the case where the state returns null when the afi safi is disabled.

### DIFF
--- a/feature/bgp/otg_tests/bgp_afi_safi_defaults/bgp_afi_safi_defaults_test.go
+++ b/feature/bgp/otg_tests/bgp_afi_safi_defaults/bgp_afi_safi_defaults_test.go
@@ -355,8 +355,14 @@ func verifyBgpCapabilities(t *testing.T, dut *ondatra.DUTDevice, afiSafiLevel st
 				if !nbr.isV4 && afiSafi == oc.BgpTypes_AFI_SAFI_TYPE_IPV4_UNICAST {
 					continue
 				}
-				t.Logf("AFI-SAFI state: %v", gnmi.Get(t, dut, statePath.Neighbor(nbr.neighborip).AfiSafi(afiSafi).State()))
-				afiSafiList = append(afiSafiList, gnmi.Get(t, dut, statePath.Neighbor(nbr.neighborip).AfiSafi(afiSafi).State()))
+				t.Logf("AFI-SAFI state: %v", gnmi.Lookup(t, dut, statePath.Neighbor(nbr.neighborip).AfiSafi(afiSafi).State()))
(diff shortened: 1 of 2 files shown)


feature/bgp/otg_tests/bgp_afi_safi_defaults/bgp_afi_safi_defaults_test.go
```diff
@@ -355,8 +355,14 @@ func verifyBgpCapabilities(t *testing.T, dut *ondatra.DUTDevice, afiSafiLevel st
 				if !nbr.isV4 && afiSafi == oc.BgpTypes_AFI_SAFI_TYPE_IPV4_UNICAST {
 					continue
 				}
-				t.Logf("AFI-SAFI state: %v", gnmi.Get(t, dut, statePath.Neighbor(nbr.neighborip).AfiSafi(afiSafi).State()))
-				afiSafiList = append(afiSafiList, gnmi.Get(t, dut, statePath.Neighbor(nbr.neighborip).AfiSafi(afiSafi).State()))
+				t.Logf("AFI-SAFI state: %v", gnmi.Lookup(t, dut, statePath.Neighbor(nbr.neighborip).AfiSafi(afiSafi).State()))
+				res := gnmi.Lookup(t, dut, statePath.Neighbor(nbr.neighborip).AfiSafi(afiSafi).State())
```
ram-mac, line 359: I see gnmi.Lookup replacing gnmi.Get. Although both the functions would do a one time subscription of the desired OC path as long we get a value for the OC path we should be good. 
@dplore and @robshakir

ram-mac (changes requested):

feature/bgp/otg_tests/bgp_afi_safi_defaults/bgp_afi_safi_defaults_test.go
ram-mac: @fmolinar - ACK 
Changes LGTM

ram-mac (approved):

feature/bgp/otg_tests/bgp_afi_safi_defaults/metadata.textproto
```diff
@@ -12,7 +12,9 @@ platform_exceptions: {
   deviations: {
     ipv4_missing_enabled: true
     bgp_extended_next_hop_encoding_leaf_unsupported: true
+    bgp_global_extended_next_hop_encoding_unsupported: true
```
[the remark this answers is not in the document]
ram-mac: Resolving this conversation as the changes are taken care

nsadhasivam (approved):
LGTM!

PR #3860, opened 2025-03-03
Retry gRIBI connection after trigger

s-nandu:
When the process restarts or the routing engine switches over, for a brief period of time the device might not accept gRIBI connection until it is ready to service the request.
Adding a retry mechanism to handle that behavior


nflath (approved):

Swetha-haridasula (approved):
lgtm

self-maurya (approved):

PR #3862, opened 2025-03-04
Large Metadata of size 100k

Swetha-haridasula:
This is to add a addition check if config push with large metadata size of 100KB is supported on config push


feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
```diff
@@ -275,4 +287,18 @@ func TestLargeSetConsistency(t *testing.T) {
 	wg.Wait()
 	time.Sleep(5 * time.Second)
 	checkMetadata2(t, gnmiClient, dut)
+	b := make([]byte, 100*1024)
```
robshakir, line 290: Can we make this a separate subtest such that it's clear when the large metadata subtest is failing vs. the logic above? It seems like it could even be its own `Test`?

feature/system/gnmi/metadata/tests/large_set_consistency_test/README.md
```diff
@@ -29,6 +29,8 @@ dut.testbed
     * stop sending REQUEST requests
     * wait 5 seconds
     * SubscribeRequest once for configuration root-level metadata and verify that it is identical corresponding root-level metadata of 2nd_LARGE_CONFIGURATION.
+* Push a Large metadata of size 100KB and wait for SetResponse that confirm push was sucesfull.
```
robshakir, line 32: The specification for metadata states ([here](https://github.com/openconfig/public/blob/master/release/models/extensions/openconfig-metadata.yang#L36)) states that it can be up to 1MiB in size. It seems prudent for us to determine whether that contract is being met, can we make a table-driven test that covers some additional sizes?

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
```diff
@@ -275,4 +287,18 @@ func TestLargeSetConsistency(t *testing.T) {
...
+		t.Fatalf("Error reading bytes: %v", err)
+	}
+	largeMetadata := strconv.Itoa(largeMetadataInt)
+	// send 3rd large metadata update request in one goroutine
+	gpbSetRequest = buildGNMISetRequest(t, largeMetadata, baselineConfig)
+	t.Log("gnmiClient Set large metadataconfig request")
+	if _, err := gnmiClient.Set(context.Background(), gpbSetRequest); err != nil {
+		t.Fatalf("gnmi.Set unexpected error: %v", err)
+	}
+	checkLargeMetadata(t, gnmiClient, dut, largeMetadata, done)
```
robshakir, line 302: Here's where I suggest that this becomes table-driven:

```suggestion
  for _, tt := range []struct{
     name string
     size int
     wantErr bool
  }{{
     name: "100KiB",
     size: 100*1024,
  }, {
     name: "1MiB",
     size: 1*1024*1024,
  }, {
     name: "2MiB",
     size: 2*1024*1024,
     wantErr: true
  }}{
	largeMetadataInt, err := rand.Read(tt.size)
	if err != nil && err != io.EOF {
		t.Fatalf("Error reading bytes: %v", err)
	}
	largeMetadata := strconv.Itoa(largeMetadataInt)
	// send 3rd large metadata update request in one goroutine
	gpbSetRequest = buildGNMISetRequest(t, largeMetadata, baselineConfig)
	t.Log("gnmiClient Set large metadataconfig request")
	_, err := gnmiClient.Set(context.Background(), gpbSetRequest);
	if (err != nil) != tt.wantErr {
		t.Fatalf("gnmi.Set unexpected error: %v, wantErr? %v", err, tt.wantErr)
	}
	if err == nil {
	    // probably only need to check value if there wasn't an error
     	checkLargeMetadata(t, gnmiClient, dut, largeMetadata, done)
    }
 }
```

By doing this, we are going to test more than just the 100KiB limit. If we find a limitation then we know that we have a device that isn't compliant with the API expectations.

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
Swetha-haridasula: I think it is valid to create a separate Sub test to check for various sizes of metadata. Will change the code to include it

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
Swetha-haridasula: Will add this logic to a make it a sub-test

Swetha-haridasula: @robshakir , my question is Should the test fail if the config config push with metadata of size>1MiB is successful. The proto says the metadata size does not exceed 1MiB but not necessary that device must reject if it is greater than 1MiB

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
Swetha-haridasula: Done

self-maurya (dismissed):

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
```diff
@@ -208,6 +210,34 @@ func checkMetadata2(t *testing.T, gnmiClient gpb.GNMIClient, dut *ondatra.DUTDev
...
+	want := largeMetadata
+	if got != want && getRespTimeStamp < done.Load() {
+		t.Errorf("extractMetadataAnnotation: got %v, want %v", got, want)
+	}
+}
+
+func testLargeMetadata(t *testing.T, gnmiClient gpb.GNMIClient, dut *ondatra.DUTDevice, baselineConfig *oc.Root, size int, done *atomic.Int64) {
+	randomBytes := make([]byte, size)
+	_, err := io.ReadFull(rand.Reader, randomBytes)
+	if err != nil {
```
robshakir, line 225:
```suggestion
	if _, err := io.ReadFull(rand.Reader, randomBytes); err != nil {
```

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
```diff
@@ -208,6 +210,34 @@ func checkMetadata2(t *testing.T, gnmiClient gpb.GNMIClient, dut *ondatra.DUTDev
...
+		t.Fatalf("failed to generate random bytes: %v", err)
+	}
+	// Encode the bytes to a base64 string.
+	largeMetadata := base64.StdEncoding.EncodeToString(randomBytes)
+	largeMetadata = largeMetadata[:size]
+	// send large metadata update request in one goroutine
+	gpbSetRequest := buildGNMISetRequest(t, largeMetadata, baselineConfig)
+	t.Log("gnmiClient Set large metadataconfig request")
+	_, err = gnmiClient.Set(context.Background(), gpbSetRequest)
+	if err != nil {
```
robshakir, line 235:
```suggestion
	if _, err = gnmiClient.Set(context.Background(), gpbSetRequest); err != nil {
```

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
```diff
@@ -234,7 +264,9 @@ func TestLargeSetConsistency(t *testing.T) {
 	if _, err := gnmiClient.Set(context.Background(), gpbSetRequest); err != nil {
 		t.Fatalf("gnmi.Set unexpected error: %v", err)
 	}
-	checkMetadata1(t, gnmiClient, dut, done)
+	t.Run("check Metadata1", func(t *testing.T) {
```
robshakir, line 267: Suggest to make this a more meaningful string than `Metadata1`. Can we call this something that describes what the function does?

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
```diff
@@ -274,5 +306,31 @@ func TestLargeSetConsistency(t *testing.T) {
 
 	wg.Wait()
 	time.Sleep(5 * time.Second)
-	checkMetadata2(t, gnmiClient, dut)
+	t.Run("check Metadata2", func(t *testing.T) {
```
robshakir, line 309: Same comment as above.

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
```diff
@@ -274,5 +306,31 @@ func TestLargeSetConsistency(t *testing.T) {
...
+		checkMetadata2(t, gnmiClient, dut)
+	})
+
+	// Large metadata Test cases.
```
robshakir, line 313: Suggest to make this a different `Test` -- so just make this be:

```golang
func TestLargeMetadata(t *testing.T) {
...
}
```

since we are testing something other than the base metadata that is described above.

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
```diff
@@ -274,5 +306,31 @@ func TestLargeSetConsistency(t *testing.T) {
...
+		size int
+	}
+	testCases := []testCase{
+		{
+			name: "Metadata with Size 100KiB",
+			size: 100 * 1024,
+		},
+		{
+			name: "Metadata with Size 1MiB",
+			size: 1 * 1024 * 1024,
```
MarcCharlebois, line 325: https://github.com/openconfig/public/blob/dcbd009789135270a9c9297991b90c7ca063a08c/release/models/extensions/openconfig-metadata.yang#L32
OC model is described as Max Len 1048000, and 1024*1024 = 1048576, which goes above the max length, should be changed to 1000*1024 (Mebibyte to Megabyte) @robshakir

feature/system/gnmi/metadata/tests/large_set_consistency_test/metadata.textproto
```diff
@@ -4,7 +4,7 @@
 uuid: "e9cdd09c-14b9-4e3c-b38c-584ef9611c3f"
 plan_id: "gNMI-1.14"
 description: "OpenConfig metadata consistency during large config push"
-testbed: TESTBED_DUT_ATE_2LINKS
+testbed: TESTBED_DUT
```
MarcCharlebois, line 7: https://github.com/openconfig/featureprofiles/blob/29c99d6b98c21187f3a023fc0f9fc38d3d52d70f/feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go#L33 - Test still uses port - https://github.com/openconfig/featureprofiles/blob/main/topologies/dut.testbed#L7 does not include any ports, this will result in TB failures.

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
Swetha-haridasula: Ok. Got your point which is valid. Fixed the test to reflect same

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
Swetha-haridasula: changed it to shortStringMetadata1

feature/system/gnmi/metadata/tests/large_set_consistency_test/large_set_consistency_test.go
robshakir: Good catch. :-)

Swetha-haridasula:
@robshakir kindly review. Resolved your comments

robshakir (dismissed):
Thanks for the changes.

self-maurya (approved):

AmrNJ (approved):
LGTM.

PR #3863, opened 2025-03-04
IPv6 ND RA Suppress test

rohit-rp:
https://github.com/openconfig/featureprofiles/issues/3321

### DIFF
--- a/feature/interface/ip/ipv6_ND/otg_tests/suppress_ipv6_nd_ra_test/README.md
+++ b/feature/interface/ip/ipv6_ND/otg_tests/suppress_ipv6_nd_ra_test/README.md
@@ -1,4 +1,4 @@
-# RT-5.11: Suppress IPv6 ND Router Advertisement [Depreciated]
+# RT-5.12: Suppress IPv6 ND Router Advertisement [Depreciated]
 
 ## Summary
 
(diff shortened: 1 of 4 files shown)


LimeHat (changes requested):
How is that different from rt-5.9?
https://github.com/openconfig/featureprofiles/tree/main/feature/interface/ip/ipv6_ND/otg_tests/disable_ipv6_nd_ra_test

rohit-rp:
> How is that different from rt-5.9? https://github.com/openconfig/featureprofiles/tree/main/feature/interface/ip/ipv6_ND/otg_tests/disable_ipv6_nd_ra_test

RT-5.9 should test only disable `/router-advertisement/config/enable`
vs
RT-5.11 is to test only suppression `/router-advertisement/config/suppress`

LimeHat:
> RT-5.11 is to test only suppression /router-advertisement/config/suppress

`Suppress` is a deprecated leaf (and currently used in RT-5.9 as a deviation by vendors who don't support new OC).

https://github.com/openconfig/public/blob/b0d7b808f02c99a5307368604a638d52b98dd593/release/models/interfaces/openconfig-if-ip.yang#L1079-L1087

If we are writing a test to explicitly validate legacy implementations, can this be indicated in the README (and the fact that it is not required for vendors who support newer OC revisions)?  
cc @dplore @cfernanz @xw-g

rohit-rp:
> > RT-5.11 is to test only suppression /router-advertisement/config/suppress
> 
> `Suppress` is a deprecated leaf (and currently used in RT-5.9 as a deviation by vendors who don't support new OC).
> 
> https://github.com/openconfig/public/blob/b0d7b808f02c99a5307368604a638d52b98dd593/release/models/interfaces/openconfig-if-ip.yang#L1079-L1087
> 
> If we are writing a test to explicitly validate legacy implementations, can this be indicated in the README (and the fact that it is not required for vendors who support newer OC revisions)? cc @dplore @cfernanz @xw-g

Yes I have indicated this in the readme. Eventually this test will be removed.

ram-mac (approved):

self-maurya (approved):

PR #3894, opened 2025-03-06
Gnoi2.1 plq test

AmrNJ:
The plq test is modified to,

Run on a single DUT connected with 2 400G links and 2 100g links.
DUT port 1 <--> DUT port 2
DUT port 3 <--> DUT port 4
Runs on both single and member link interfaces


feature/gnoi/packet_link_qualification/tests/packet_link_qualification_test/packet_link_qualification_test.go
```diff
@@ -337,23 +362,70 @@ func testLinkQualification(t *testing.T, dut1 *ondatra.DUTDevice, dut2 *ondatra.
...
-		generatorsetupDuration:    30 * time.Second,
-		reflectorsetupDuration:    60 * time.Second,
+		generatorsetupDuration:    time.Duration(math.Max(30, genPblqMinSetup)) * time.Second,
+		reflectorsetupDuration:    time.Duration(math.Max(60, refPblqMinSetup)) * time.Second,
 		testDuration:              120 * time.Second,
 		generatorPostSyncDuration: 5 * time.Second,
 		reflectorPostSyncDuration: 10 * time.Second,
-		tearDownDuration:          30 * time.Second,
+		generatorTeardownDuration: time.Duration(math.Max(30, genPblqMinTearDown)) * time.Second,
+		reflectorTeardownDuration: time.Duration(math.Max(30, refPblqMinTearDown)) * time.Second,
```
dplore, line 418: Thanks for doing this, the logic looks good.  Please remove the comments and move this code into a function call (such as `func setPlqTimers` so the code is easier to read.

dplore (commented):
Thanks for all the improvements here and adding the testbeds.  One style comment to resolve.  Also, please provide link to passing test results

dplore (dismissed):

feature/gnoi/packet_link_qualification/tests/packet_link_qualification_test/metadata.textproto
```diff
@@ -3,8 +3,8 @@
 
 uuid: "e98e61ec-03cd-4bff-8094-2fc69491962a"
 plan_id: "gNOI-2.1"
-description: "Packet-based Link Qualification"
-testbed: TESTBED_DUT_DUT_4LINKS
+description: "Packet-based Link Qualification on 100G and 400G links"
+testbed: TESTBED_DUT_400ZR_100G_4LINKS
```
LimeHat, line 7: Can we split ZR into a separate test case so the results can be obtained on non-ZR testbeds? I think there's a number of testbeds (including testbeds @ Google labs) that don't have ZR optics yet.

LimeHat (changes requested):

feature/gnoi/packet_link_qualification/tests/packet_link_qualification_test/metadata.textproto
LimeHat: ZR/ZR+ interfaces typically require special config (things like optical channels; `transceiver-x` test cases demonstrate that) so those ports are not really interchangeable with others. This config is probably missing from this PR btw.

Therefore ZR and non-ZR interfaces are not interchangeable with each other, you can't simply "downgrade" to another 400G testbed if ZR is not available.

I understand that you don't want to replicate the core test logic in different files; maybe it would be better to create an internal helper lib that can be shared between the two tests?

AmrNJ: We only wanted the test to run on 400g speed links irrespective of optics as plq is transparent to optic pmd.

LimeHat: >as plq is transparent to optic pmd.

PLQ itself is transparent, but the preliminary configuration is not. 400ZR interfaces require additional config compared to non-ZR.

ZR test cases (TRANSCEIVER-x) use internal/cfgplugins helper for that, e.g. 
https://github.com/openconfig/featureprofiles/blob/c80b9f3ef6d58aa39dd4f3cf972dcb515f4f7581/feature/platform/transceiver/inventory/tests/zr_inventory_test/zr_inventory_test.go#L55-L62
https://github.com/openconfig/featureprofiles/blob/c80b9f3ef6d58aa39dd4f3cf972dcb515f4f7581/internal/cfgplugins/interface.go#L54-L73

If you want to handle any 400G optic, you need to include additional if/else or switch/case statements to differentiate interface config based on PMD.

AmrNJ: What about 400g FR? Does it need the additional config as well?

rohit-rp: > ZR/ZR+ interfaces typically require special config (things like optical channels; `transceiver-x` test cases demonstrate that) so those ports are not really interchangeable with others. This config is probably missing from this PR btw.

The configs are not missing from the PR. We have testbed with necessary config already applied and we do not interchange any ports. So this along with rewriting the testbed assignments during run works for us

> Since ZR and non-ZR interfaces are not interchangeable with each other, you can't simply "downgrade" to another 400G testbed if ZR is not available.

In our test env we do not need to do this.

> I understand that you don't want to replicate the core test logic in different files; maybe it would be better to create an internal helper lib that can be shared between the two tests?

I will have Amritha merge the code and I understand it may not work for you right now. But I have created [an issue](https://github.com/openconfig/featureprofiles/issues/3970) to split the test and ensure it works for external test environments also. We will take it up in April

LimeHat: If you need an internal reference w.r.t ZR-related configs paths you can take a look at b/297268414#comment10 which was provided by the optical team.

rohit-rp: > > We have testbed with necessary config already applied and we do not interchange any ports.
>
> Typically FNT tests are responsible for e2e configuration of interfaces used in test, which ensures both OC compliance and validation of production workflows.

I get you, but we have FNT's that do e2e configuration of interfaces, this specific test does not intend to do that. We are not using this test to validate interface config rather gnoi link qual feature only. gnoi link qual is optics/speeds/port agnostic hence this feature and interface config do not have any interdependencies.

> If you're ok not to use OC to configure ZR with specific vendors that's cool, but I just don't see why this pattern should be normalized or standardized.

The intent of this test is not to configure ZR, just test gnoi link qual.

Let us merge this one as it works for us and helps us run the test. We can split the test into two later and tracked under https://github.com/openconfig/featureprofiles/issues/3970

rohit-rp (approved):

dplore:
Ppre-configuration should be limited to the minimum configuration required to make the device accessible to gnmi for config/monitoring.   The featureprofiles tests should do all the configuration needed for the feature  inside the test.  So in this case, it sounds like we may need need to move some of our interface configuration into featureprofiles code.  (probably in the internal/interfaces folder which is intended for shared interface configurations.

PR #3957, opened 2025-03-07
PLT-1.1 Fixed Port Issues

mastarkey:
Test is for breakout config for the following modes

4x100
2x100
4x10


AmrNJ (approved):
LGTM

feature/platform/tests/breakout_configuration/breakout_test.go
```diff
@@ -0,0 +1,436 @@
...
+				configContainer := &oc.Component_Port_BreakoutMode_Group{
+					Index:         ygot.Uint8(0),
+					NumBreakouts:  ygot.Uint8(tc.numbreakouts),
+					BreakoutSpeed: oc.E_IfEthernet_ETHERNET_SPEED(tc.breakoutspeed),
+				}
+				groupContainer := &oc.Component_Port_BreakoutMode{Group: map[uint8]*oc.Component_Port_BreakoutMode_Group{1: configContainer}}
+				breakoutContainer := &oc.Component_Port{BreakoutMode: groupContainer}
+				portContainer := &oc.Component{Port: breakoutContainer, Name: ygot.String(componentName)}
+
+				if deviations.VerifyExpectedBreakoutSupportedConfig(dut) {
```
rohit-rp, line 278: @mastarkey I understand the need for this but it will also lead to cosmetic pass if the test runs on unsupported device because we are skipping a test case. I think we are trying to solve a problem this test is not meant to solve. Lets remove this check if the test fails because it runs on the wrong device/port, thats an infrastructure issue for the testing team to fix.

rohit-rp (changes requested):

feature/platform/tests/breakout_configuration/breakout_test.go
AmrNJ: The plan is to create a testbed.proto file that selects only the breakout ports. We will remove this deviation once we figure out the testbeds that will be used.

AmrNJ (approved):
LGTM

rohit-rp (approved):

PR #3965, opened 2025-03-10
Add grpc-services constraint to ocpaths.textproto

dplore:
This allows specifying a constraint to assert OC Paths which are restricted to a specific GRPC service. (ie: the use of gnmi, gnsi, gnpsi, gribi in the /system/grpc-servers/grpc-server subtree)

The relevant identity-ref is at `/system/grpc-servers/grpc-server/config/services`

### DIFF
--- a/proto/ocpaths.proto
+++ b/proto/ocpaths.proto
@@ -77,8 +77,11 @@ message OCPath {
 // fields for each type of constraint
 message OCPathConstraint {
   oneof constraint {
-    string platform_type = 1;  // derived from OC module oc-platform-types
-    string grpc_service = 2;   // derived from OC module oc-sys-grpc
+    // Specifies a OPENCONFIG_HARDWARE_COMPONENT or OPENCONFIG_SOFTWARE_COMPONENT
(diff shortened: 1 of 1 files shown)


proto/ocpaths.proto
```diff
@@ -77,7 +77,8 @@ message OCPath {
 // fields for each type of constraint
 message OCPathConstraint {
   oneof constraint {
-    string platform_type = 1;
+    string platform_type = 1;  // derived from OC module oc-platform-types
+    string grpc_service = 2;   // derived from OC module oc-sys-grpc
```
robshakir, line 81: Do you want to list the identity that is actually being used to generate these values? (I assume that they are called `PLATFORM_TYPE` and `GRPC_SERVICE` but it might be clearer to explicitly state it?

robshakir (dismissed):

proto/ocpaths.proto
dplore: Good idea, I'll add

dplore: Done

Ankur19 (approved):

PR #3967, opened 2025-03-10
Wikidoc: Fix bad metadata handling

bstoll:
Remove the file feature/interface/ip/ipv6_link_local/metadata.textproto; it is a duplicate of feature/interface/ip/ipv6_link_local/otg_tests/ipv6_link_local_test/metadata.textproto.  wikidoc will also now ignore test cases with a metadata file but no README.

We also now need to be explicit that we need write access to update the wiki; there was a recent org wide security change. All workflows used to have write permission by default.


dplore (approved):

ram-mac (approved):
LGTM

PR #3968, opened 2025-03-11
RT-7.1 & RT-1.33-Code and deviation changes

AmrNJ:
RT-7.1
1. The test adds bgp_default_policy_unsupported deviation in Arista
2. The initial prefix verification only checks if all the prefixes advertised by otg port 1 is received and advertised to otg port 2 in the absence of import/export routing policies. 

RT-1.33
1. Remove initial prefix validation as it is not mentioned in the README. The test would eventually fail if there is a miss in otg prefix advertisement. 
2. Validate prefixes only on ebgp neighbor 1 after applying import policy on ebgp neighbor 1. Neigbour 2 will anyways be validated after applying the export policy.


ram-mac (approved):
Based on the discussion on the BGP routes verification this PR changes are looking good. Hence approving it.

PR #3973, opened 2025-03-12
RT-2.1 changes

arvbaska1:
- TestISISHelloTimer : Set level enabled leaf to true since OC default is false for this leaf.

- In TestAuthentication : enable is-is authentication under isis/interface/authentication container since for P2P links config attributes for one specific level is blocked.


feature/isis/otg_tests/base_adjacencies_test/base_adjacencies_test.go
```diff
@@ -455,12 +455,23 @@ func TestAuthentication(t *testing.T) {
 				auth.AuthType = oc.KeychainTypes_AUTH_TYPE_SIMPLE_KEY
 				auth.AuthPassword = ygot.String(password)
 				for _, intf := range isis.Interface {
-					intf.GetOrCreateLevel(2).GetOrCreateHelloAuthentication().Enabled = ygot.Bool(tc.enabled)
-					if tc.enabled {
-						intf.GetLevel(2).GetHelloAuthentication().AuthPassword = ygot.String("google")
-						intf.GetLevel(2).GetHelloAuthentication().AuthMode = tc.mode
-						intf.GetLevel(2).GetHelloAuthentication().AuthType = oc.KeychainTypes_AUTH_TYPE_SIMPLE_KEY
+					if deviations.IsisAuthenticationInterfaceLevelUnsupported(ts.DUT) {
+						intf.GetOrCreateAuthentication().Enabled = ygot.Bool(tc.enabled)
```
ram-mac, line 459: So Cisco is not supporting authentication at interface level is it? So, this deviation will do authentication in the global level only? Which means we are skipping the interface level2 based configuration.

@ihebboubaker - To check this part from a requirement perspective. If this is a requirement, then this deviation is not acceptable as we are skipping exercising the path.

internal/deviations/deviations.go
```diff
@@ -1325,3 +1325,8 @@ func OperStatusForIcUnsupported(dut *ondatra.DUTDevice) bool {
 func ExplicitDcoConfig(dut *ondatra.DUTDevice) bool {
 	return lookupDUTDeviations(dut).GetExplicitDcoConfig()
 }
+
+// IsisAuthenticationInterfaceLevelUnsupported returns true if Isis Authentication is not supported under OC ISIS interface/levels but supported under IS-IS Interface/Authentication.
```
ram-mac, line 1329: @arvbaska1 - Please raise a FR or bug for this before adding it.

ram-mac (changes requested):

feature/isis/otg_tests/base_adjacencies_test/base_adjacencies_test.go
arvbaska1: It supported, but for P2P links the adjacencies are shared by both levels, so configuring some attribute for only one level (level2 in this case) will be rejected.
This is not global authentication this is authentication container under the ISIS Interface as well https://github.com/openconfig/public/blob/master/release/models/isis/openconfig-isis.yang#L1385

arvbaska1: i will rename the deviation more appropriately

ksgireesha:
@self-maurya  and @rohit-rp , Please review this.

feature/isis/otg_tests/base_adjacencies_test/base_adjacencies_test.go
ram-mac: @arvbaska1 - can u rename it and repush the commit

Please also created a bug for this deviation as mentioned in the other comment

@Tavneet-Sidhu13 - FYI

ram-mac (changes requested):

feature/isis/otg_tests/base_adjacencies_test/base_adjacencies_test.go
arvbaska1: updated deviation, added buganizer and resolved conflicts

ram-mac (approved):
The changes look good and it is being validated.

rohit-rp (approved):

PR #3977, opened 2025-03-12
Update README.md for PF-1.2

sugrimov:
Add a condition to verify inner packet QoS marking after encapsulation.


robshakir (approved):
LGTM, thanks for the change.

PR #3978, opened 2025-03-12
Fix for SFLOW-1 to work with non-default sampling rate

singh-prem:
This PR adds a deviation to allow a different ingress sampling rate than 1/1,000,000 for platforms.
No change is needed for a platform that supports default ingress sampling rate.

### DIFF
--- a/feature/afts/otg_tests/afts_base/afts_base_test.go
+++ b/feature/afts/otg_tests/afts_base/afts_base_test.go
@@ -187,7 +187,14 @@ func (tc *testCase) configureDUT(t *testing.T) error {
 		}
 	})
 	ts.ATEIntf1.Isis().Advanced().SetEnableHelloPadding(false)
-	ts.PushAndStart(t)
+
+	if err := ts.PushAndStart(t); err != nil {
+		return err
+	}
+
+	if _, err = ts.AwaitAdjacency(); err != nil {
+		return fmt.Errorf("no IS-IS adjacency formed: %v", err)
+	}
 	return nil
 }
 
--- a/feature/afts/otg_tests/afts_base/metadata.textproto
+++ b/feature/afts/otg_tests/afts_base/metadata.textproto
@@ -20,7 +20,9 @@ platform_exceptions: {
   }
   deviations: {
     default_network_instance: "default"
-    isis_single_topology_required: true
+    isis_instance_enabled_required: true
+    isis_interface_afi_unsupported: true
     low_scale_aft: true
+    interface_enabled: true
   }
 }


masood-shah (commented):
this implements the use of the deviation but doesnt explain why the default might be unsupported or where the value 262144 comes from. could you please link the specific technical reference (eg, Cisco doc ID, bug) that documents this platform's min supported ingress sFlow sampling rate and the technical justification for its deviation from the default within this test suite?

dplore:
> this implements the use of the deviation but doesnt explain why the default might be unsupported or where the value 262144 comes from. could you please link the specific technical reference (eg, Cisco doc ID, bug) that documents this platform's min supported ingress sFlow sampling rate and the technical justification for its deviation from the default within this test suite?

We need to decide if the deviated sample rate is acceptable for this test?  The 1/1,000,000 rate was our operational standard.   If we want to allow different rates or a range of rates, this test is the place to specify it.  Can you help figure out our requirement @masood-shah ?

masood-shah:
> > this implements the use of the deviation but doesnt explain why the default might be unsupported or where the value 262144 comes from. could you please link the specific technical reference (eg, Cisco doc ID, bug) that documents this platform's min supported ingress sFlow sampling rate and the technical justification for its deviation from the default within this test suite?
> 
> We need to decide if the deviated sample rate is acceptable for this test? The 1/1,000,000 rate was our operational standard. If we want to allow different rates or a range of rates, this test is the place to specify it. Can you help figure out our requirement @masood-shah ?



> > this implements the use of the deviation but doesnt explain why the default might be unsupported or where the value 262144 comes from. could you please link the specific technical reference (eg, Cisco doc ID, bug) that documents this platform's min supported ingress sFlow sampling rate and the technical justification for its deviation from the default within this test suite?
> 
> We need to decide if the deviated sample rate is acceptable for this test? The 1/1,000,000 rate was our operational standard. If we want to allow different rates or a range of rates, this test is the place to specify it. Can you help figure out our requirement @masood-shah ?

Happy to help, but first, could someone explain how the values 1/1,000,000 and 262,144 were derived or determined?

arvbaska1:
> this implements the use of the deviation but doesnt explain why the default might be unsupported or where the value 262144 comes from. could you please link the specific technical reference (eg, Cisco doc ID, bug) that documents this platform's min supported ingress sFlow sampling rate and the technical justification for its deviation from the default within this test suite?

the min. and max. supported sample rate can differ based on platforms, for 8000 it can sample from 1:1 to 1:262144
sample rate. This documented below.
https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/netflow/configuration/b-netflow-configuration-ios-xr-8000/netflow-sflow-key-concepts.html#flow-sampler

internal/cfgplugins/sflow.go
```diff
@@ -36,7 +36,8 @@ func NewSFlowGlobalCfg(t *testing.T, batch *gnmi.SetBatch, newcfg *oc.Sampling_S
 	if newcfg == nil {
 		c.Enabled = ygot.Bool(true)
 		c.SampleSize = ygot.Uint16(256)
-		c.IngressSamplingRate = ygot.Uint32(1000000)
+		// override ingress sampling rate if default value of 1000000 is not supported
+		c.SetIngressSamplingRate(deviations.SflowIngressMinSamplingRate(d))
```
dplore, line 40: This needs to be wrapped with a conditional (if the deviation exists, then use it).
for example:
```
		if deviations.SflowIngressMinSamplingRate(d) {
			switch d.Vendor() {
			case ondatra.CISCO:
             		        c.SetIngressSamplingRate(deviations.SflowIngressMinSamplingRate(d))
			}
		} else {
		        c.IngressSamplingRate = ygot.Uint32(1000000)
		}
```


It is still up to @masood-shah to decide to approve

arvbaska1: sure, ack.

arvbaska1: @dplore  Could you help review the PR (run the checks) , the requested changes have been made

PR #4001, opened 2025-03-13
Update README.md - remove ASIC-based forwarding engine requirement

gggsmith:
The Ondatra testing framework doesn't support this requirement

### DIFF
--- a/feature/qos/ecn/ate_tests/ecn_enabled_traffic_test/README.md
+++ b/feature/qos/ecn/ate_tests/ecn_enabled_traffic_test/README.md
@@ -37,13 +37,6 @@ minimum threshold.
         priority.
         *   AF3, AF2, AF1, BE1 and BE0 will use WRR queues.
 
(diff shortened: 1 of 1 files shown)


feature/qos/ecn/ate_tests/ecn_enabled_traffic_test/README.md
```diff
@@ -37,13 +37,6 @@ minimum threshold.
         priority.
         *   AF3, AF2, AF1, BE1 and BE0 will use WRR queues.
 
-*   Test results should be independent of the location of interfaces. For
-    example, 2 input interfaces and output interface could be located on
-
-    *   Same ASIC-based forwarding engine
-    *   Different ASIC-based forwarding engine on same line card
-    *   Different ASIC-based forwarding engine on different line cards
-
```
sezhang2, line 46: It is not a requirement to test all the cases with different locations of interfaces. It describes the test result should not be impacted by  the location of interfaces.

PR #4026, opened 2025-03-17
initial commit for plt1.2

rohit-rp:


feature/platform/tests/parent_component_validation/parent_component_validation.go
```diff
@@ -0,0 +1,76 @@
...
+	}{
+		{
+			desc:    "Port1",
+			port:    "port1",
+			pattern: "^(SwitchChip|NP|xe-0/0/0|0/0/CPU0)$",
+		},
+		{
+			desc:    "Port2",
+			port:    "port2",
+			pattern: "^(SwitchChip|NP|xe-0/0/0|0/0/CPU0)$",
```
AmrNJ, line 62: Does the pattern match all vendor scenarios?

rohit-rp: No this is incorrect for Cisco, updated to the correct pattern

AmrNJ (approved):
LGTM.

dplore (commented):
Hi, this overlaps / duplicates this test that also does parent component validation:

https://github.com/openconfig/featureprofiles/blob/8c4806b67aed4d00a3cf1c51ec7571564fe7602c/feature/platform/tests/telemetry_inventory_test/telemetry_inventory_test.go#L99 (gNMI-1.4: Telemetry: Inventory)

It would be okay to move the parent component validation into a separate test if you like.  

Also, I believe "gNMI-1.4: Telemetry: Inventory" implementation is better because it is more vendor neutral and utilizes more of the intended OC.  It properly checks for component type `INTEGRATED_CIRCUIT` instead of hardcoding vendor specific regex.  

Also this test was merged with static analysis failures.  Please fix them.  Now, every PR will have these failures until someone fixes them.

PR #4028, opened 2025-03-18
TRANSCEIVER-11: remove stale workaround for optical channel format

LiamWalsh98:
Cisco devices now return the correct format, meaning this workaround would cause a failure

Old e.g: `OpticalChannel0_0_0_18`

instead of : `Ex: OpticalChannel0/0/0/18`


yiwenhu-g (approved):
LGTM

amrindrr (approved):

PR #4031, opened 2025-03-20
changing the tests to handle gnmi.Get() response gracefully

abhijit-dhar:
Added check - for some cases gnmi.Get() query when LLDP is disabled; should be handled carefully; such that value not found exception can be handled gracefully instead of test cases ended abruptly.

### DIFF
--- a/feature/lldp/otg_tests/core_lldp_tlv_population_test/core_lldp_tlv_population_test.go
+++ b/feature/lldp/otg_tests/core_lldp_tlv_population_test/core_lldp_tlv_population_test.go
@@ -26,20 +26,20 @@ import (
 	"github.com/openconfig/featureprofiles/internal/deviations"
 	"github.com/openconfig/featureprofiles/internal/fptest"
 	"github.com/openconfig/featureprofiles/internal/otgutils"
+	gpb "github.com/openconfig/gnmi/proto/gnmi"
 	"github.com/openconfig/ondatra"
 	"github.com/openconfig/ondatra/gnmi"
 	"github.com/openconfig/ondatra/gnmi/oc"
 	otgtelemetry "github.com/openconfig/ondatra/gnmi/otg"
 	"github.com/openconfig/ondatra/otg"
 	"github.com/openconfig/ygnmi/ygnmi"
-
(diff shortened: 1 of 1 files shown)


feature/lldp/otg_tests/core_lldp_tlv_population_test/core_lldp_tlv_population_test.go
```diff
@@ -152,7 +153,12 @@ func configureDUT(t *testing.T, name string, lldpEnabled bool) (*ondatra.DUTDevi
 		gnmi.Replace(t, node, gnmi.OC().Interface(p.Name()).Enabled().Config(), true)
 	}
 
-	return node, gnmi.Get(t, node, gnmi.OC().Lldp().Config())
+	tsState := gnmi.Lookup(t, node, gnmi.OC().Lldp().State())
+	_, isPresent := tsState.Val()
```
self-maurya, line 157: can we utilize the received value;
```
  lldpState, isPresent := tsState.Val()
  if isPresent {
    return node, lldpState
  }
```

feature/lldp/otg_tests/core_lldp_tlv_population_test/core_lldp_tlv_population_test.go
```diff
@@ -257,43 +266,52 @@ func checkOTGLLDPNeighbor(t *testing.T, otg *otg.OTG, c gosnappi.Config, expLldp
...
-
-	// Get the LLDP port neighbor ID and state of the node.
-	if _, ok := gnmi.Watch(t, dut, interfacePath.State(), time.Minute, func(val *ygnmi.Value[*oc.Lldp_Interface]) bool {
-		intf, present := val.Val()
-		return present && len(intf.Neighbor) > 0
-	}).Await(t); !ok {
-		t.Error("Number of neighbors: got 0, want > 0.")
-		return
-	}
+	if conf != nil {
```
self-maurya, line 281: instead of this return if conf is nil
```
if conf == nil {
  return
}
```

self-maurya (changes requested):

self-maurya (approved):

PR #4032, opened 2025-03-20
added support for keysight port time stamp

abhijit-dhar:
Test case holddown_timers_test.go is now updated with new API to read the timestamp when the last link state change (say up or down). The API looks as "gnmi.Get(t, ate.OTG(), gnmi.OTG().Port(p1.ID()).LastChange().State())". This is not supported in KNE platform. In KNE platform this API will return 0. The kne situation is handled in the test cases itself. In the KNE test cases will follow old path.

But in hardware platform the test case requires ports to time-synced with the DUT - this is the basic requirement of the test. If DUT is not time-synced with the port the test will fail. CODEOWNERS Please comment if a deviation would be appropriate to handle this situation as per guidelines [https://github.com/openconfig/featureprofiles/tree/main/internal/deviations#when-to-use-deviations](url)

### DIFF
--- a/feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
+++ b/feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
@@ -308,9 +308,9 @@ func flapOTGInterface(t *testing.T,
 	}
 
 	tsState := gnmi.Lookup(t, ate.OTG(), gnmi.OTG().Port(p1.ID()).LastChange().State())
-	_, isPresent := tsState.Val()
+	lastChange , isPresent := tsState.Val()
 	if isPresent {
-		otgStateChangeTsUint64 = gnmi.Get(t, ate.OTG(), gnmi.OTG().Port(p1.ID()).LastChange().State())
+		otgStateChangeTsUint64 = lastChange
 	}
 
 	// convert string type change to time.time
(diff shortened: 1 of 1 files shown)


self-maurya (approved):

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -305,6 +307,12 @@ func flapOTGInterface(t *testing.T,
 		t.Logf("Interface %s status got %v, want %v", aggID, DutLastChangeOper2, expectedStatus.String())
 	}
 
+	tsState := gnmi.Lookup(t, ate.OTG(), gnmi.OTG().Port(p1.ID()).LastChange().State())
+	_, isPresent := tsState.Val()
+	if isPresent {
+		otgStateChangeTsUint64 = gnmi.Get(t, ate.OTG(), gnmi.OTG().Port(p1.ID()).LastChange().State())
```
self-maurya, line 313: use existing value;

```
lastChange, isPresent := tsState.Val()
if isPresent {
  otgStateChangeTsUint64 = lastChange
}
```

self-maurya (changes requested):

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
```diff
@@ -643,6 +655,9 @@ func TestTC5ShortDOWN(t *testing.T) {
 		}
 	})
 
+	// WAIT TILL DUT INTERFACES ARE UP
+	time.Sleep(time.Second * 5)
```
AmrNJ, line 659: Please use gnmi.Await instead of time.sleep.

AmrNJ (changes requested):

feature/interface/holdtime/otg_tests/holdtime_test/holddown_timers_test.go
Swetha-haridasula: @abhijit-dhar please update to gnmi.Await

PR #4035, opened 2025-03-20
RT-6.1: LLDP testcase changes for cisco

karthikeya-remilla:
PR to incorporate LLDP behavior change from ios-xr version 25.1.1: interface needs to be added to OC datastore explicitly before enabling LLDP on it

Proposed changes:
leafref fix + interface operation
remove the deviation "lldp_interface_config_override_global" which is not applicable anymore

Note: go test flag "lldpInterfaceConfigOverrideGlobal" has been added for backward compatibility - if consuming IOS-XR 2441 images for test, set this flag to true


rohit-rp:
@karthikeya-remilla Please resolve conflicts

karthikeya-remilla:
@rohit-rp is the ATE version of LLDP not used anymore? I think it is deleted? 

If so, we can close this PR, as it is not applicable anymore as the changes are applicable only to ATE

rohit-rp:
Yes right it was deleted by me in /pull/4094

PR #4043, opened 2025-03-22
Fixing chained_policies_test for the export policy changes PR 3366

ram-mac:
PR 3366 added export policy to satisfy the default behavior of BGP.
For Cisco this is not required as the default behavior of BGP is to reject all routes when there is no policy for the BGP peer. 

Hence removing the changes that is not required.


nsadhasivam (approved):
LGTM!

PR #4048, opened 2025-03-24
fix security issues

self-maurya:


internal/security/credz/credz.go
```diff
@@ -72,7 +72,8 @@ func SetupUser(t *testing.T, dut *ondatra.DUTDevice, username string) {
 // - Must use 4 of the 5 character classes ([a-z], [A-Z], [0-9], [!@#$%^&*(){}[]\|:;'"], [ ]).
 func GeneratePassword() string {
 	// Create random length between 24-32 characters long.
-	length := minPasswordLength + rand.Intn(maxPasswordLength-minPasswordLength+1)
+	delta := maxPasswordLength - minPasswordLength + 1
+	length := minPasswordLength + rand.Intn(delta)
```
singhavnish2516, line 76: Please check if this is not marked as security concern although it is a false positive. If it is marked, we can use crypto rand with no performance issues.

singhavnish2516:
LGTM

internal/security/credz/credz.go
self-maurya: If this doesn't solve this false positive, then we can force close https://github.com/openconfig/featureprofiles/security/code-scanning/2 directly as this is false positive.

singhavnish2516 (approved):

singhavnish2516 (approved):
LGTM

PR #1027, opened 2023-01-27
TE-13.2 route_removal_during_failover_test

cprabha:
Hi,

Please review automation for TE-13.2 testcase.
solves https://github.com/openconfig/featureprofiles/issues/608

Thanks,
Prabha

PR #4054, opened 2025-03-24
RT-7.5: Update link_bandwidth_test

yini101:
Add `extCommunitySetArista` and set the per vendor `extCommunitySet` when being used.

### DIFF
--- a/feature/bgp/policybase/otg_tests/link_bandwidth_test/link_bandwidth_test.go
+++ b/feature/bgp/policybase/otg_tests/link_bandwidth_test/link_bandwidth_test.go
@@ -131,6 +131,12 @@ var (
 		"linkbw_any": "^link-bandwidth:.*",
 	}
 
+	extCommunitySetArista = map[string]string{
+		"linkbw_1M":  "link-bandwidth:23456:1M",
+		"linkbw_2G":  "link-bandwidth:23456:2G",
+		"linkbw_any": "^lbw:.*:.*$",
+	}
+
 	communitySet = map[string]string{
 		"regex_match_comm100": "^100:.*$",
 	}
@@ -665,10 +671,19 @@ func configureExtCommunityRoutingPolicy(t *testing.T, dut *ondatra.DUTDevice) {
 	root := &oc.Root{}
 	var communitySetCLIConfig string
 	var extCommunitySetCLIConfig string
+
+	switch dut.Vendor() {
+	case ondatra.CISCO:
+		extCommunitySet = extCommunitySetCisco
+	case ondatra.JUNIPER:
+		extCommunitySet = extCommunitySetJuniper
+	case ondatra.ARISTA:
+		extCommunitySet = extCommunitySetArista
+	}
+
 	if deviations.BgpExtendedCommunitySetUnsupported(dut) {
 		switch dut.Vendor() {
 		case ondatra.CISCO:
-			extCommunitySet = extCommunitySetCisco
 			for name, community := range extCommunitySet {
 				if name == "linkbw_any" && deviations.CommunityMemberRegexUnsupported(dut) {
 					communitySetCLIConfig = fmt.Sprintf("community-set %v \n dfa-regex '%v' \n end-set", name, community)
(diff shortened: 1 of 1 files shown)


AmrNJ:
@yini101 Link-bandwidth is the standard identifier for Extended Community. Any reason why regex alone has a different string as the same string does work for "link-bandwidth:23456:1M and link-bandwidth:23456:2G any reason why regex has a different string? We could use the deviation for interim but it will be better if we can standardize this.

yini101:
> @yini101 Link-bandwidth is the standard identifier for Extended Community. Any reason why regex alone has a different string as the same string does work for "link-bandwidth:23456:1M and link-bandwidth:23456:2G any reason why regex has a different string? We could use the deviation for interim but it will be better if we can standardize this.

The `lbw:` prefix is the existing link bandwidth matching behavior in EOS for a long time now. It has been the behavior prior to the https://github.com/openconfig/public/pull/1027 merging into OpenConfig. The other issue is that `bgp-ext-community-type` is for pushing in link bandwidth as a restricted string that can be parsed, thus that part could be enforced to use `link-bandwidth:` from the operator's input. Internally, it's easy to parse the input `link-bandwidth:` for the set ext-community side; however, for match ext-community reg-exp, it's not a good idea to change the reg-exp string that's defined by the operator as that could be very problematic.

I also could not find any clear standard on what to match against in the BGP RFC.

feature/bgp/policybase/otg_tests/link_bandwidth_test/link_bandwidth_test.go
```diff
@@ -114,7 +114,7 @@ var (
 	extCommunitySet = map[string]string{
 		"linkbw_1M":  "link-bandwidth:23456:1M",
 		"linkbw_2G":  "link-bandwidth:23456:2G",
-		"linkbw_any": "^link-bandwidth:.*:.*$",
```
nsadhasivam, line 117: Please retain the linkbw_any as "^link-bandwidth:.*:.*$", as per [#1027](https://github.com/openconfig/public/pull/1027/files) and [#README](https://github.com/openconfig/featureprofiles/blob/9b05130d1035d5400e6bfb3ef59c14a80017d97a/feature/bgp/policybase/otg_tests/link_bandwidth_test/README.md?plain=1#L49)

To handle the behaviour on EOS you may use MatchLinkBandwidthAnyRegex to return "^lbw:.*:.*$" when the deviation is used, and let L117 "linkbw_any": "^link-bandwidth:.*:.*$", remain the default as per the oc std and README

nsadhasivam (changes requested):

feature/bgp/policybase/otg_tests/link_bandwidth_test/link_bandwidth_test.go
yini101: I've updated the test to retain the original `linkbw_any`.

AmrNJ:
Hi @yini101 I think it will be better if we create a extendedCommunitySet for Arista with the right regex like how we have for other vendors.

yini101:
> Hi @yini101 I think it will be better if we create a extendedCommunitySet for Arista with the right regex like how we have for other vendors.

I changed it as per suggested.

feature/bgp/policybase/otg_tests/link_bandwidth_test/link_bandwidth_test.go
```diff
@@ -672,6 +678,9 @@ func configureExtCommunityRoutingPolicy(t *testing.T, dut *ondatra.DUTDevice) {
 			t.Fatalf("Unsupported vendor %s for native command support for deviation 'BgpExtendedCommunitySetUnsupported'", dut.Vendor())
 		}
 	} else {
+		if deviations.MatchLinkBandwidthRegexDiffPrefix(dut) {
```
AmrNJ, line 681: Can you remove this deviation and just have a switch case here instead as the communities are anyways vendor specific.

yini101: I've updated the code as you suggested.

feature/bgp/policybase/otg_tests/link_bandwidth_test/metadata.textproto
```diff
@@ -14,6 +14,7 @@ platform_exceptions:  {
     missing_value_for_defaults:  true
     interface_enabled:  true
     default_network_instance:  "default"
+    match_link_bandwidth_regex_diff_prefix: true
```
AmrNJ, line 17: Why do we still need this deviation?

yini101: I removed the deviation now.

yini101:
This PR has absorbed #4222 for it to pass.

yini101:
@AmrNJ Is there anything else that needs to be done on my end for this to merge? I think I resolved all of the issues. The `PR Approval Workflow / check-approvals (pull_request_target)` check is still failing.

nsadhasivam (approved):
LGTM!

rohit-rp (approved):

PR #4071, opened 2025-03-26
add test 1.16 - bgp add-path scale with policy

ram-mac:
Adding test case 1.16 for BGP ADDPATH SCALE with POLICY


sugrimov (commented):
This is pretty much a repeat of RT-1.15.
I think we should just have a test to match selected routes with community and do add-path for them. IMO, there is point to repeat everything we test in 1.15.

dplore:
> This is pretty much a repeat of RT-1.15. I think we should just have a test to match selected routes with community and do add-path for them. IMO, there is point to repeat everything we test in 1.15.

If you like and want to see these results separately, you can have two subtests in RT-1.15.  ie:

RT-1.15.1 - test addpath scale
RT-1.15.2 - test addpath scale with policy

However, if there is no reason to see test results without policy (because our use case is to always use policy with add path), then I agree, you can just have one test which includes policy.

ram-mac:
> This is pretty much a repeat of RT-1.15. I think we should just have a test to match selected routes with community and do add-path for them. IMO, there is point to repeat everything we test in 1.15.

OK; It makes sense to have single test with a policy configuration in place. Will merge RT-1.15 and RT-1.16 into single test.

ram-mac:
@sugrimov @dplore - Please review and approve; i have merged RT-1.15 and RT-1.16; So we will have just one case covering both. 

I think now it is upto mark.

dplore (changes requested):
Only thing missing now is the canonical OC for bgp add-paths configuration as required by the [featureprofiles README template](https://github.com/openconfig/featureprofiles/blob/main/doc/test-requirements-template.md).    

I think it's best if it's added to RT-1.3 and this test (RT-1.15) should link to it.

ram-mac:
> Only thing missing now is the canonical OC for bgp add-paths configuration as required by the [featureprofiles README template](https://github.com/openconfig/featureprofiles/blob/main/doc/test-requirements-template.md).
> 
> I think it's best if it's added to RT-1.3 and this test (RT-1.15) should link to it.

Added now and we have the Json format for the ADD-path in details. Will attach a link from this README into RT-1.3

self-maurya (approved):

nikki-blr (approved):

PR #4073, opened 2025-03-26
Modified the test and corresponding function calls to set the MTU values correctly on the NTE ports

abhijit-dhar:
To set the BGP TCP Segment size to desired value, no separate Api is needed in the BGP, rather setting the MTU in the Ixia port achieves the desired purpose. Hence modified the test and corresponding function signatures and calls to set the MTU values correctly on the IXIA ports.


self-maurya (approved):

nsadhasivam (approved):
LGTM!

nsadhasivam:
@nikki-blr Please help merge the changes

PR #3395, opened 2024-08-22
RT-1.27 : static_route_bgp_redistribution_test.go 

cprabha:
Hi,

Adding deviations and also fixing BGP config and corresponding OTG configuration.

PR #4075, opened 2025-03-27
RT-1.27: Fix test configurations and add deviations

yini101:
* Add deviation default_no_igp_metric_propagation for default behavior of not propagating IGP metric by default.
  * Add Arista specific CLI for propagating IGP metric
* Add deviation skip_bgp_peer_group_send_community_type for skipping setting send community type per bgp peer-groups
* Fix up test usage of `BgpCommunitySetRefsUnsupported` to use `SetCommunitySetRefs` instead of the deprecated `SetCommunitySetRef`.
* Use `send-community-type`
* Fix 1.27.11 and 1.27.22 test case to point to the peer address instead of self.


internal/cfgplugins/bgp_policy.go
```diff
@@ -78,3 +78,20 @@ func DeviationJuniperRoutingPolicyBGPActionSetCommunity(t *testing.T, dut *ondat
...
+// DeviationAristaRoutingPolicyBGPActionSetMedPropagate is used as an alternative to
+// /routing-policy/policy-definitions/policy-definition/statements/statement/actions/bgp-actions/config/set-med.
+// This deviation implements CLI to perform the equivalent function.
+func DeviationAristaRoutingPolicyBGPActionSetMedPropagate(t *testing.T, dut *ondatra.DUTDevice, policyName string) {
+	// route-map route-policy-v4 statement propagate-metric permit 5
+	//   set metric igp-metric
+	//   continue
+	// exit
+
+	cliConfig := fmt.Sprintf("route-map %s statement propagate-metric permit 5\n", policyName)
```
nsadhasivam, line 91: @ihebboubaker Please confirm if the CLI is used in production?

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
```diff
@@ -450,6 +451,13 @@ func configureTableConnection(t *testing.T, dut *ondatra.DUTDevice, isV4, mPropa
 		addressFamily = oc.Types_ADDRESS_FAMILY_IPV6
 	}
 
+	if importPolicy == "" && deviations.DefaultNoIgpMetricPropagation(dut) {
```
nsadhasivam, line 454: Lets not combine deviations with other conditions.	
Instead we could use:
	
	if importPolicy == "" {
                if deviations.DefaultNoIgpMetricPropagation(dut) {
		             importPolicy = redistributeStaticPolicyNameV4
		             if !isV4 {
			                  importPolicy = redistributeStaticPolicyNameV6
		             }
	    }

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
```diff
@@ -502,11 +510,18 @@ func configureTableConnection(t *testing.T, dut *ondatra.DUTDevice, isV4, mPropa
 			cfgplugins.DeviationCiscoTableConnectionsStatictoBGPMetricPropagation(t, dut, isV4, med, importPolicy)
 		}
 	}
+
+	if mPropagation && deviations.DefaultNoIgpMetricPropagation(dut) {
```
nsadhasivam, line 514: Avoid using deviations combined with other condition

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
```diff
@@ -1511,13 +1531,13 @@ func validateRedistributeRouteWithTagSet(t *testing.T, dut *ondatra.DUTDevice, a
 	}
 
 	if isV4 {
-		if deviations.SkipSettingDisableMetricPropagation(dut) {
+		if deviations.SkipSettingDisableMetricPropagation(dut) && !deviations.DefaultNoIgpMetricPropagation(dut) {
```
nsadhasivam, line 1534: Avoid combining 2 deviations in the same if

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
```diff
@@ -1698,7 +1725,7 @@ func TestBGPStaticRouteRedistribution(t *testing.T) {
 			name:  "1.27.5 redistribute-ipv4-route-policy-as-prepend",
 			setup: func() { redistributeStaticRoutePolicyWithASN(t, dut, isV4) },
 			validate: func() {
-				validatePrefixASN(t, ate, isV4, atePort1.Name+".BGP4.peer", "192.168.10.0", []uint32{65499, 65499, 65499, 64512})
+				validatePrefixASN(t, ate, isV4, atePort1.Name+".BGP4.peer", "192.168.10.0", []uint32{64512, 65499, 65499, 65499})
```
nsadhasivam, line 1728: Why is the order of ASN changes?

nsadhasivam (changes requested):

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
yini101: I think that would make the code much harder to follow because this is an && expression. So it would look like:
```
if ipv4 {
  if deviation1 {
    if !deviation2 {
      validate(ipv4)
    } else {
      validate(zero)
    }
  } else {
    validate(zero)
  }
} else {
  if deviation1 {
    if !deviation2 {
      validate(ipv6)
    } else {
      validate(zero)
    }
  } else {
    validate(zero)
  }
}
```

Do you think I should still change it?

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
yini101: @AmrNJ do you happen to remember why this was changed?

internal/cfgplugins/bgp_policy.go
yini101: I would like to avoid using the CLI if possible... Arista has a OpenConfig support to configure the metric propagation, but it's behind a deviation which changes the type to add an enum type. Is it possible to use that instead? I am unsure on how to configure this in featureprofiles, are there any examples on that?

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
AmrNJ: This was changed with PR #3395. The correct order as per the README is below. 
validatePrefixASN(t, ate, isV4, atePort1.Name+".BGP4.peer", "192.168.10.0", []uint32{65499, 65499, 65499, 64512})

nsadhasivam: @yini101 can you please revert the change as per the above comment.

internal/cfgplugins/bgp_policy.go
nsadhasivam: @yini101 

Can you please share the openconfig and the deviation details for the below.?

Arista has a OpenConfig support to configure the metric propagation, but it's behind a deviation which changes the type to add an enum type. Is it possible to use that instead?

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
self-maurya: how about we do something like below to reduce a bit of cognitive complexity:

```
func validateLearnedPrefix(t *testing.T, ate *ondatra.ATEDevice, bgpPeerName, subnet string, expectedMED uint32, shouldBePresent, ipv4 bool) {
  if ipv4 {
    validateLearnedIPv4Prefix(t, ate, bgpPeerName, subnet, expectedMED, shouldBePresent)
  } else {
    validateLearnedIPv6Prefix(t, ate, bgpPeerName, subnet, expectedMED, shouldBePresent)
  }
}

if deviation1 {
  if deviation2 {
    validate(zero)
  } else {
    validateLearnedPrefix(t, ate, bgpPeerName, subnet, expectedMED, shouldBePresent, ipv4)
  }
} else {
  validate(zero)
}
```

Also, doing this way would make it easier to cleanup the deviations later.

self-maurya (changes requested):

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
nsadhasivam: @yini101 Can you please modify as above.

internal/cfgplugins/bgp_policy.go
yini101: https://github.com/aristanetworks/yang/blob/master/EOS-4.34.0F/release/openconfig/models/policy/arista-rpol-deviations.yang#L211
```
  typedef arista-bgp-set-med-type {
    type union {
      type uint32;
      type enumeration {
        enum IGP {
          description
            "set the MED value to the IGP cost toward the
             next hop for the route, only applies for BGP learned routes";
        }
        enum METRIC {
          description
            "set the MED value to the IGP metric of a redistributed route";
        }
      }
    }
    description
      "Type definition for specifying how the BGP MED can be set in BGP
      policy actions.  The MED can be specified as a direct integer
      value or setting it to the IGP or METRIC cost.  To be used in conjunction
      with `bgp-set-med-action` in order to specify the appropriate
      action to take on this value.";
  }
```


https://github.com/aristanetworks/yang/blob/master/EOS-4.34.0F/release/openconfig/models/policy/arista-rpol-deviations.yang#L763
```
  deviation "/oc-rpol:routing-policy/oc-rpol:policy-definitions/oc-rpol:policy-definition/oc-rpol:statements/oc-rpol:statement/oc-rpol:actions/oc-bgp-pol:bgp-actions/oc-bgp-pol:config/oc-bgp-pol:set-med-action" {
    description
      "EOS supports adding IGP and METRIC to set-med value";
    deviate replace {
      type arista-rpol-deviations:arista-bgp-set-med-action;
    }
  }
```

I believe that `METRIC` would be the enum value to use here.

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
yini101: Looking into the discussion, in the README itself might be incorrect. The RFC4271 as per Section 6.3 states that the AS_PATH should start with the local AS on the left most side as the receiver if the session is an external peer. This means that the correct order is `[64512, 65499, 65499, 65499]`.

[RFC4271 as per Section 6.3](https://datatracker.ietf.org/doc/html/rfc4271#section-6.3):
```
   If the UPDATE message is received from an external peer, the local
   system MAY check whether the leftmost (with respect to the position
   of octets in the protocol message) AS in the AS_PATH attribute is
   equal to the autonomous system number of the peer that sent the
   message.  If the check determines this is not the case, the Error
   Subcode MUST be set to Malformed AS_PATH.
```

The reference for the original PR discussion #3395 is [RFC4271 Section 5.1.2](https://www.rfc-editor.org/rfc/rfc4271.html#section-5.1.2):
```
         1) if the first path segment of the AS_PATH is of type
            AS_SEQUENCE, the local system prepends its own AS number as
            the last element of the sequence (put it in the leftmost
            position with respect to the position of octets in the
            protocol message).  If the act of prepending will cause an
            overflow in the AS_PATH segment (i.e., more than 255 ASes),
            it SHOULD prepend a new segment of type AS_SEQUENCE and
            prepend its own AS number to this new segment.
```
AS-path prepend feature is a local policy decision and to my knowledge it is not dictated by the RFC4271. That being said, this section of the RFC is about how to manipulate the AS_PATH attribute when prepending the local AS during advertisement which also reinforces that the local AS that appears should be on the left most side.

cprabha: README is correct as per RFC also as per bgp yang .

As per BGP yang. order should be []uint32{65499, 65499, 65499, 64512}). Please check.  
 revision "2024-11-13" {
    description
      "Add use-last-as leaf to set-as-path-prepend to prepend last AS.";
    reference "8.1.0";
  }

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
yini101: Done

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
yini101: There might be a misunderstanding here. The `set-as-path-prepend` is not a feature that is standardized by any RFC and does not define how this feature to behave. The `use-last-as` is an option as part of `set-as-path-prepend` and it takes the last ASN in the _current_ AS Path and prepends it. This is also not dictated as part of any RFC as well. This should be all done before the local AS is prepended.

What is part of the standard is prepending the local AS in the AS path as per RFC4271. It states to prepend the local AS number as the last element with a clarification of it being on the left most position. This means it should be `{64512, 65499, 65499, 65499}`. Am I missing something here?

AmrNJ: The right order of ASN is indeed 64512 64599 64599 64599 64512. The readme needs to be updated. The most recent commit already has the right order.

internal/cfgplugins/bgp_policy.go
yini101: Hello @nsadhasivam and @ihebboubaker,

Is this acceptable? How can I update the test to push the `METRIC` enum value which is the Arista specific augment?

feature/bgp/static_route_bgp_redistribution/otg_tests/static_route_bgp_redistribution_test/static_route_bgp_redistribution_test.go
rszarecki: - The policy action `set-as-path-prepend` should insert ASN BEFORE eBGP mandatory AS path manipulation (adding own ASN on path). This is because on-the-wire last as MUST be == ASN of peer. It's value can be used for sanity/safety check against ANS adverised in OPEN messages.
- the value of ASN to be prepended in policy action can be: a) explicit value OR b) `use-last-as`.
Because of a) policy-inserted ASN must not be "left"/more-last vs. eBGP inserted own AS.
The b) is bit ambigous if this "last-as" is last AS on as-path before or after eBGP-mandatory insertion. My understanding will last ASN on path as existing in Adj-RIB-in (import policy) or Local-RIB (export policy). Bottom line, the last as on-the-wire send by eBGP MUST be local ASN of speaker.

internal/cfgplugins/bgp_policy.go
yini101: The latest patch utilizes the `RawAPIs` idea, thanks Nisha!

nsadhasivam (approved):
LGTM!

self-maurya (approved):

PR #4077, opened 2025-03-27
Added VRF scale

vishnureddybadveli:
MPLS in GRE and MPLS in UDP Scale Test


feature/gribi/otg_tests/mpls_in_udp_scale/README.md
```diff
@@ -50,9 +50,24 @@ outer_ip-ttl =        "64"
   * Update policer-policies in a batch of `sched_q` = 1,000
   * Policer-policies changes should take effect within `sched_r` / 2 time
 
-#### Scale profile A - many vlans
+#### Scale profile A - many vlans 1 NH per NHG with same MPLS label
 
 * 20 ip destinations * 1,000 vlans = 20,000 'flows'
+* 1000 VRF have 1000 policy forwarding NHG with 8 NH
+* 1 default route in each VRF pointing to policy forwarding NHG.
+* Create 20K gRIBI NHG with 1 NH per NHG with same MPLS label.
```
dplore, line 58: The configuration should also  require generating configuration of static LSP's for the MPLS label to VLAN mapping, right?

feature/gribi/otg_tests/mpls_in_udp_scale/README.md
```diff
@@ -50,9 +50,24 @@ outer_ip-ttl =        "64"
...
+* 100K longest exact match route entries with 5 pointing to 1 gRIBI NHG.
```
dplore, line 59: Clarify that these are gRIBI programmed prefixes.  (ie: they are not static routes, right)?  So it should read tsomething like:

Create 5 unique gRIBI prefixes which point to 1 NHG and repeat this 20k times, resulting in 100K total prefixes, pointing to 20k NHG.

Is that your intent?

feature/gribi/otg_tests/mpls_in_udp_scale/README.md
```diff
@@ -50,9 +50,24 @@ outer_ip-ttl =        "64"
   * Update policer-policies in a batch of `sched_q` = 1,000
   * Policer-policies changes should take effect within `sched_r` / 2 time
 
-#### Scale profile A - many vlans
+#### Scale profile A - many vlans 1 NH per NHG with same MPLS label
 
 * 20 ip destinations * 1,000 vlans = 20,000 'flows'
+* 1000 VRF have 1000 policy forwarding NHG with 8 NH
```
dplore, line 56:
```suggestion
* 1000 network-instances have each have 1 static route pointing a static  NHG with 8 NH.  The encap-headers for the NH should indicate MPLS in GRE encapsulation as per the canonical OC defined in test TE-18.1
```

dplore: NOte the canonical OC needs to be updated to TE-18.1 to use the static route and static NHG,NH paths defined in https://github.com/openconfig/public/pull/1234

feature/gribi/otg_tests/mpls_in_udp_scale/README.md
```diff
@@ -50,9 +50,24 @@ outer_ip-ttl =        "64"
...
+* 1 default route in each VRF pointing to policy forwarding NHG.
+* Create 20K gRIBI NHG with 1 NH per NHG with same MPLS label.
+* 100K longest exact match route entries with 5 pointing to 1 gRIBI NHG.
+* Each ingress vlan has 20 policer-policies = 10,000 'token buckets'
+* The 20 ip destinations are split evenly between the 20 policers
```
dplore, line 61:
```suggestion
* The QoS classifier should contain rules to match each unique IP destination to a unique scheduler with a 2 color, 1 rate policer as per the configuration in TE-18.1.  The result should be 20,000 policers.
```

dplore (changes requested):

vishnureddybadveli:
@dplore may I kindly request to please take a look at the changes.

feature/gribi/otg_tests/mpls_in_udp_scale/README.md
```diff
@@ -34,56 +34,47 @@ outer_ip-ttl =        "64"
...
-    `flow_q` flows to use different NH,NHG
-
-#### Procedure - Policer + Flow Scale
-
-* For each scale profile, create the following subsets TE-18.1.6.n
-  * Program all 20,000 flows
-  * Every `sched_r` interval use gnmi.Set to replace `sched_q` scheduler policies
-  * Verify packet loss changes for all flows within `sched_r` / 2 time
+  * 1,024 vlans
+  * 1024 network instances
```
dplore, line 40: nit:  network-instance is not part of a flow, but rather the DUT configuration.  (this section is for flows)

feature/gribi/otg_tests/mpls_in_udp_scale/README.md
```diff
@@ -34,56 +34,47 @@ outer_ip-ttl =        "64"
...
+  * gRIBI client update rate `flow_r` = 100 updates (Modify requests) per second, 60 operations per update/
```
dplore, line 41: what are the 60 "operations" per update?  Do you mean each modify request should contain 60 ip prefixes+ NHG + NH?

feature/gribi/otg_tests/mpls_in_udp_scale/README.md
```diff
@@ -34,56 +34,47 @@ outer_ip-ttl =        "64"
...
+  * DUT packet forwarding updated within 1 second after adding entries (after FIB_PROGRAMMED response)
```
dplore, line 42: This is not part of the flow definition or scale profile, but rather a performance verification.  It should be moved into a subtest where some method needs to be specified for how this verification can be done.

feature/gribi/otg_tests/mpls_in_udp_scale/README.md
dplore: Could this be simplified to be client requests that happen as fast as possible, and measure when packet loss becomes zero?    I think some specific steps need to go into how to measure packet loss.  The earlier text that is deleted in this PR was:

```
  * Validate RIB_AND_FIB_ACK with FIB_PROGRAMMED is received from DUT within
    1 second
  * Measure packet loss.  Target packet loss <= 50%.
  * Repeat adding 200 flows until 20,000 flows have been added
  * Once reaching 20,000 flows, perform 1 iteration of modifying the first
    `flow_q` flows to use different NH,NHG
```

dplore (changes requested):

PR #4091, opened 2025-03-31
Update testregistry.textproto

sachendras:
Define new test for "Multidimensional test for Static GUE Encap/Decap based on BGP path selection and selective DSCP marking (Original Pull/3486).


testregistry.textproto
```diff
@@ -896,6 +896,12 @@ test: {
   readme: ""
   exec: " "
 }
+test: {
+  id: "RT-3.52"
+  description: "Multidimensional test for Static GUE Encap/Decap based on BGP path selection and selective DSCP marking based on tunnel destination."
+  readme: ""
```
self-maurya, line 902: Can you also add the readme path: https://github.com/openconfig/featureprofiles/blob/main/feature/policy_forwarding/encapsulation/otg_tests/staticgueencap_and_bgp_path_selection/README.md

self-maurya (changes requested):

PR #4098, opened 2025-04-01
RT-5.6: Replace loopback mode to TERMINAL 

dipchauh:
Per the conversation in b/404876484, removing loopback deviation and setting loopback mode to TERMINAL according to new definitions in OpenConfig [PR1155](https://github.com/openconfig/public/pull/1155)


feature/interface/interface_loopback_aggregate/otg_tests/interface_loopback_aggregate/interface_loopback_aggregate_test.go
```diff
@@ -228,17 +228,18 @@ func TestInterfaceLoopbackMode(t *testing.T) {
...
 	})
 
-	t.Run("Configure interface loopback mode FACILITY on DUT AE interface", func(t *testing.T) {
-		if deviations.InterfaceLoopbackModeRawGnmi(dut) {
-
-			gnmi.Update(t, dut, gnmi.OC().Interface(dutPort1.Name()).LoopbackMode().Config(), oc.Interfaces_LoopbackModeType_TERMINAL)
-
+	t.Run("Configure interface loopback mode on DUT AE interface", func(t *testing.T) {
+		var loopbackMode oc.E_Interfaces_LoopbackModeType
+		switch dut.Vendor() {
```
self-maurya, line 233: Vendor based check are discouraged as this in someway hides the deviation.

If intention of this change is only to replace deviation with switch case, in I'd suggest to drop these changes and close the PR.

self-maurya (changes requested):

feature/interface/interface_loopback_aggregate/otg_tests/interface_loopback_aggregate/interface_loopback_aggregate_test.go
trathod1: Based on discussion here - [b/404876484#comment3](https://partnerissuetracker.corp.google.com/issues/404876484#comment3), switch-case was used to support both vendors using old and new definition. 

If we intent to use deviation without switch-case, we can update to script to use `TERMINAL` (new definition) by default and have old definition (`FACILITY`) under deviation.

AmrNJ: @trathod1 The use of both definitions are approved as per comment. But it is preferred to have this as a deviation in order to better track delivery.

AmrNJ (changes requested):

feature/interface/interface_loopback_aggregate/otg_tests/interface_loopback_aggregate/interface_loopback_aggregate_test.go
trathod1: @AmrNJ Sounds good. We will make the expected `TERMINAL` as the default and also update the README for the same based on [b/404876484#comment3](https://partnerissuetracker.corp.google.com/issues/404876484#comment3). And if some vendor supporting only the old definition and uses `FACILITY`, they can introduce the deviation to use that.

dipchauh:
Similar PR [4413](https://github.com/openconfig/featureprofiles/pull/4413) already merged so closing this PR.

PR #4100, opened 2025-04-02
Bandwidth computation for forwarding-unviable links

AmrNJ:
Ensure that the total bandwidth computation for the bundle interface does not include links where "forwarding viable" is set to false.

### DIFF
--- a/feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
+++ b/feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
@@ -961,7 +961,6 @@ func checkBundleViableLinksBW(t *testing.T, dut *ondatra.DUTDevice, dutPorts []*
 		if got := gnmi.Get(t, dut, gnmi.OC().Interface(port.Name()).ForwardingViable().State()); got != false {
 			portSpeed := ethernetPortSpeedToMbps(gnmi.Get(t, dut, gnmi.OC().Interface(port.Name()).Ethernet().PortSpeed().State()))
 			lagSpeed += portSpeed
-			t.Logf("Lag speed is %v", lagSpeed)
 		}
 	}
 	return lagSpeed, nil
--- a/feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
+++ b/feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
@@ -202,18 +202,18 @@ func TestAggregateAllNotForwardingViable(t *testing.T) {
 
 	t.Logf("ISIS cost of LAG_2 lower then ISIS cost of LAG_3 Test-01")
 	t.Run("RT-5.7.1.1: Setting Forwarding-Viable to False on Lag2 all ports except port 2", func(t *testing.T) {
-		bundleBW := checkBundleViableLinksBW(t, dut, dutPortList[1:agg2.ateLagCount+1])
-		if got, _ := gnmi.Lookup(t, dut, ocpath.Root().Interface(aggIDs[1]).Aggregation().LagSpeed().State()).Val(); got != bundleBW {
-			t.Errorf("Forwarding unviable links counted as part of LAG bandwidth, got %v, want %v", got, bundleBW)
+		expectedbundleBW, _ := checkBundleViableLinksBW(t, dut, dutPortList[1:agg2.ateLagCount+1])
+		if got, _ := gnmi.Lookup(t, dut, ocpath.Root().Interface(aggIDs[1]).Aggregation().LagSpeed().State()).Val(); got != expectedbundleBW {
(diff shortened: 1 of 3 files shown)


feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -202,6 +202,9 @@ func TestAggregateAllNotForwardingViable(t *testing.T) {
 
 	t.Logf("ISIS cost of LAG_2 lower then ISIS cost of LAG_3 Test-01")
 	t.Run("RT-5.7.1.1: Setting Forwarding-Viable to False on Lag2 all ports except port 2", func(t *testing.T) {
+		if got, _ := gnmi.Lookup(t, dut, ocpath.Root().Interface(aggIDs[1]).Aggregation().LagSpeed().State()).Val(); got != 500000 {
```
ram-mac, line 205: @AmrNJ - Can this value 500000 be derived based on the number of member links for the aggregate interface? or Can this be used as a constant instead of direct value?

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
```diff
@@ -210,6 +213,9 @@ func TestAggregateAllNotForwardingViable(t *testing.T) {
 		if err := confirmNonViableForwardingTraffic(t, dut, ate, atePortList[2:agg2.ateLagCount+1], dutPortList[2:agg2.ateLagCount+1]); err != nil {
 			t.Fatal(err)
 		}
+		if got, _ := gnmi.Lookup(t, dut, ocpath.Root().Interface(aggIDs[1]).Aggregation().LagSpeed().State()).Val(); got != 100000 {
```
ram-mac, line 216: Can this value 100000 be derived based on the number of member links for the aggregate interface? or Can this be used as a constant instead of direct value?

ram-mac (changes requested):
Please see if the suggested changes can be made.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
LimeHat, line 205: I don't think that you can assume BW of an interface; it is a generic 8-link testbed which can use links with any BW (100G or 400G in most cases)
https://github.com/openconfig/featureprofiles/blob/eb86322a511be5e468d92a0647a8110afc641d05/feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/metadata.textproto#L7

LimeHat (changes requested):
Is this aligned with the current OpenConfig definitions?

https://github.com/openconfig/public/blob/50a6b4075923a2cf8b46b86f1f3cd7ecb17ea0f8/release/models/interfaces/openconfig-if-aggregate.yang#L142-L165

The current definition says that the speed is `based on speed of active member interfaces`, and, AFAIK, non-viable interfaces are not removed from the active list.

On top of that, I'd like to note that a non-forwarding viable interface still can receive traffic in rx direction. With the proposed changes, it means that a LAG can receive more bandwidth in RX direction than reported by this telemetry leaf

@dplore @AmrNJ can we please start with an OpenConfig PR that clarifies that?

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
AmrNJ: Added checkBundleViableLinksBW to compute the BW of forwarding viable links.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
AmrNJ: Added checkBundleViableLinksBW to dynamically calculate the bundle BW. 
I agree the LAG will be able to receive more BW that what is reported. We will raise a PR to update OC definition.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
ram-mac: Thanks for taking care of the comments.

feature/interface/aggregate/otg_tests/aggregate_all_not_viable_test/aggregate_all_not_forwarding_viable_test.go
ram-mac: Thanks for taking care of this request

ram-mac (approved):
Changes look good to me

self-maurya (approved):

PR #4103, opened 2025-04-03
Rt 5.5 increase holdtime down tolerance

AmrNJ:
Increasing the holdtime down tolerance to 700ms. This is to account for processing delays after issuing RF request to OTG.


ram-mac (approved):
LGTM

PR #4113, opened 2025-04-08
AUTHZ: add OC counter and policy test

nkrishk:
This PR introduces enhancements to the [`authz.go`] file, focusing on improving the handling of gRPC counters and policy verification logic. 

It modifies the 
1. `Verify` function to ensure counter updates 
2. `Rotate` function to validate Authz policy


internal/security/authz/authz.go
```diff
@@ -69,6 +73,44 @@ type Rule struct {
...
+
+	policy.GrpcAuthzPolicyCreatedOn = createdOn
+	policy.GrpcAuthzPolicyVersion = version
+
+	return policy, nil
+}
+
+func verifyPolicySame(t *testing.T, policy1, policy2 GrpcAuthzPolicy) {
+	if policy1.GrpcAuthzPolicyCreatedOn != policy2.GrpcAuthzPolicyCreatedOn {
+		t.Fatalf("Policy Created On mismatch: policy1=%d, policy2=%d", policy1.GrpcAuthzPolicyCreatedOn, policy2.GrpcAuthzPolicyCreatedOn)
```
dipchauh, line 106: The OC Path `grpc-authz-policy-created-on` is stored as nanoseconds since epoch [(link)](https://github.com/openconfig/public/blob/50a6b4075923a2cf8b46b86f1f3cd7ecb17ea0f8/release/models/gnsi/openconfig-gnsi-authz.yang#L69-L77) while the `created_on` field in Authz UploadRequest is passed as seconds since epoch [(link)](https://github.com/openconfig/gnsi/blob/bfdfaf995f62789565ae384ae94e31cf77951291/authz/authz.proto#L210-L212). Kindly make the appropriate conversions if you want to compare these fields.

internal/security/authz/authz.go
```diff
@@ -233,6 +384,31 @@ func Verify(t testing.TB, dut *ondatra.DUTDevice, spiffe *Spiffe, rpc *gnxi.RPC,
...
 	}
+	counterBefore := GrpcCounter{}
+	_, tempPolicy := Get(t, dut)
+	if !strings.Contains(spiffe.ID, "deny-all") {
+		if len(tempPolicy.DenyRules) > 0 && tempPolicy.DenyRules[0].Name != "no-one-can-gnmi" {
+			if rpc == gnxi.RPCs.GnsiAuthzProbe {
+				t.Logf("Probe")
+			}
+			data := gnmi.LookupAll(t, dut, gnmi.OC().System().GrpcServerAny().State())
+			sysGrpcServerName, pres := data[0].Val()
```
dipchauh, line 395: There can be multiple gRPC server instances on the dut serving different gRPC services. How do we know which server to pick to verify the counters from since 0th entry might not always be the correct one?

nkrishk: @dipchauh currently CISCO supports only one gRPC server instances so this logic should work for now

internal/security/authz/authz.go
nkrishk: made appropriate conversion to convert into nano second.. Done.

internal/security/authz/authz.go
LimeHat: the test shouldn't be based on what Cisco supports; it is supposed to be vendor-neutral in accordance with what is defined by OpenConfig

nkrishk: I understand.  
Assuming we will use the DEFAULT gRPC server details in our binding file, I will hard-code the gRPC server as "DEFAULT."  
If you have any alternative suggestions, please let me know.

internal/security/authz/authz.go
```diff
@@ -69,6 +73,44 @@ type Rule struct {
...
+func getPolicyFromOcPath(t *testing.T, dut *ondatra.DUTDevice) (GrpcAuthzPolicy, error) {
+
+	policy := GrpcAuthzPolicy{
+		GrpcAuthzPolicyCreatedOn: 0,
+		GrpcAuthzPolicyVersion:   "",
+	}
+
+	// Perform a gNMI Get or Subscribe to retrieve the policy data
+	createdOn := gnmi.Get(t, dut, gnmi.OC().System().Aaa().Authorization().GrpcAuthzPolicyCreatedOn().State())
+	// if err != nil {
```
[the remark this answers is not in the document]
MarcCharlebois: I think maybe there's a bit of confusion. I think having the error checking here is beneficial, just should not have been commented. If the policy Data gnmi.Get is not successful, correctly reporting the error should be done.

internal/security/authz/authz.go
```diff
@@ -233,6 +378,31 @@ func Verify(t testing.TB, dut *ondatra.DUTDevice, spiffe *Spiffe, rpc *gnxi.RPC,
...
 	}
+	counterBefore := GrpcCounter{}
+	_, tempPolicy := Get(t, dut)
+	if !strings.Contains(spiffe.ID, "deny-all") {
+		if len(tempPolicy.DenyRules) > 0 && tempPolicy.DenyRules[0].Name != "no-one-can-gnmi" {
+			if rpc == gnxi.RPCs.GnsiAuthzProbe {
+				t.Logf("Probe")
+			}
+			data := gnmi.LookupAll(t, dut, gnmi.OC().System().GrpcServerAny().State())
+			sysGrpcServerName, pres := data[0].Val()
```
morrowc, line 389: is 'pres' here 'ok' ?

I suspect if you reach into a 'is not there' slice element you're going to have a 'panic' bad day.
Would it make sense to FIRST check:
  "if len(data) != 1 ....."
  
  regardless of the discussion (which is 100% valid) about the name/why '0' may work here, reaching into a thing that maybe doesn't exist is always a bad plan.

internal/security/authz/authz.go
```diff
@@ -233,6 +378,31 @@ func Verify(t testing.TB, dut *ondatra.DUTDevice, spiffe *Spiffe, rpc *gnxi.RPC,
...
+			if !pres {
+				t.Fatalf("Got nil system grpc server data")
+			}
+
+			if tConcrete, ok := t.(*testing.T); ok {
```
morrowc, line 394: it's odd that you need to cast t to *testing.T ?

internal/security/authz/authz.go
```diff
@@ -216,9 +273,97 @@ type HardVerify struct {
...
+
+	// Ensure LastAccessAccept and LastAccessReject in counter2 are greater than or equal to counter1
+	if counter2.LastAccessAccept < counter1.LastAccessAccept || counter2.LastAccessReject < counter1.LastAccessReject {
+		return actualDiff, false
+	}
+
+	return actualDiff, true
+}
+
 // Verify uses prob to validate if the user access for a certain rpc is expected.
```
morrowc, line 363: probe?

internal/security/authz/authz.go
```diff
@@ -216,9 +273,97 @@ type HardVerify struct {
...
 // It also execute the rpc when HardVerif is passed and verifies if it matches the expectation.
```
morrowc, line 364: HardVerify ?

internal/security/authz/authz.go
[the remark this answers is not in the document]
nkrishk: The `gnmi.Get` function will not returning an error. Instead, the error is being managed at the ygnmi level, which causes a test fatal failure being triggered. - https://github.com/openconfig/ondatra/blob/main/gnmi/gnmi.go#L271-L278.

It was assumed that the error checking would be sufficient.

internal/security/authz/authz.go
nkrishk: Since `Verify` accepts `testing.TB` (an interface), and `FetchUpdate` requires `*testing.T` (a concrete type), so  asserted that `t` is of type `*testing.T` before calling `FetchUpdate`.

morrowc (commented):
pending comment answers.

internal/security/authz/authz.go
LimeHat: there's no naming of grpc-servers in the binding, and it is not correct to assume the default name.

to select the correct server you can either:
1) retrieve `/system/grpc-servers/grpc-server/state/services` and find the server which has gnmi service enabled, or
2) if you query stats only after they are populated (after RPCs were executed and counters are >0], you can retrieve stats for all grpc-servers select counters that are present from any server

[both solutions assume that there's only 1 gnmi-enabled grpc-server, but that should be mostly ok]

LimeHat: Also since authz is not limited to gnmi you need to do this exercise for each service type. For different services you'd expect to have different grpc-servers, generally speaking (e.g. IANA allocated ports for gnmi, gribi and p4rt are different); even if it may not be true in your particular implementation/configuration

karthikeya-remilla:
Obsolete PR. Closing this.

PR #4114, opened 2025-04-08
Create tunnel_ecmp_hashing README

markkud:


feature/tunnel/otg_tests/tunnel_ecmp_hashing/README.md
dplore, line 1: Each test ie: TUN-2.(8,9,10) should be in a separate folder and README file.

The folder should be feature/policy-forwarding/otg_tests/<test_name_here>/README.md

The reason to use the policy-forwarding folder is these ECMP tests are related to policy-forwarding rules (decap is for MPLS in UDP is: /network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/action/config/decapsulate-mpls-in-udp) 

Looking in the policy-forwarding folder, I see there already is a test for MPLS in UDP decap:
https://github.com/openconfig/featureprofiles/tree/main/feature/policy_forwarding/otg_tests/mpls_gue_ipv4_decap_test

and this test has verification of load balancing in it.  Please review

dplore (changes requested):

feature/tunnel/otg_tests/tunnel_ecmp_hashing/README.md
```diff
@@ -0,0 +1,173 @@
...
+      union_replace: true
+      replace: true
+    gNMI.Subscribe:
+      on_change: true
+
+```
+
+
+
+# TUN-2.9 ECMP hashing based with GUE decap
```
dplore, line 58: I don't see a test for decap of IP in UDP / GUEv1 already here in featureprofiles.   The closest test is this one: https://github.com/openconfig/featureprofiles/tree/main/feature/policy_forwarding/otg_tests/mpls_gue_ipv4_decap_test which does IPv4 in MPLS in GUE decap.  

I suggest you could create a new README and copy that test which covers a very similar configuration and modify it to be IPv4 in GUE decap (just by removing the MPLS stuff), and then also ensure the load balancing/ECMP you want is present.  You'd place it in a folder like:   feature/policy_forwarding/otg_tests/ipv4_udp_decap_test

feature/tunnel/otg_tests/tunnel_ecmp_hashing/README.md
```diff
@@ -0,0 +1,173 @@
...
+    gNMI.Set:
+      union_replace: true
+      replace: true
+    gNMI.Subscribe:
+      on_change: true
+
+```
+
+
+# TUN-2.10 ECMP hashing based with GRE encap
```
dplore, line 112: THere is an existing IP in GRE encap test here: https://github.com/openconfig/featureprofiles/tree/main/feature/policy_forwarding/encapsulation/otg_tests/encap_gre_ipv4

As above, please review and ensure it has the ECMP verification you want.  If not, instead of creating a new test here, propose a modification to the existing test.

feature/tunnel/otg_tests/tunnel_ecmp_hashing/README.md
markkud: Thank you Darren, will exclude this test as its already covered.

feature/tunnel/otg_tests/tunnel_ecmp_hashing/README.md
markkud: Added this test under policy-forwarding as suggested.

feature/tunnel/otg_tests/tunnel_ecmp_hashing/README.md
markkud: Thank you Darren, confirmed that this is already covered.

dplore (commented):
This remaining test now sounds like the same scope as another README in progress at:
https://github.com/openconfig/featureprofiles/pull/4135

Your topology is different, but functionally it seems like the goal is the same.  UDP encap/GUEv1 packets are sent to a DUT, the DUT decaps using policy-forwarding `decapsulate-gue` action and packets are load balanced across a set of output interfaces.   (The other test is a bit more comprehensive and includes L3 and L2/aggregate interface load balancing)

PR #4116, opened 2025-04-08
updating containerz lifecycle tests to run with ONDATRA

alshabib:


feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -2,44 +2,53 @@ package container_lifecycle_test
...
-	}
+func TestMain(m *testing.M) {
+	fptest.RunTests(m)
+}
 
-	return cli
+func containerzClient(ctx context.Context, t *testing.T) *client.Client {
+	dut := ondatra.DUT(t, "dut")
+
+	dut.Config().New().WithAristaText(`
```
robshakir, line 33: This should be gated by a deviation to indicate that this is not supported as OpenConfig currently. You can define a new deviation and then handle this as a switch based on `dut.Vendor`.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -2,44 +2,53 @@ package container_lifecycle_test
...
+
+	t.Logf("Waiting for device to ingest its config.")
+	time.Sleep(time.Minute)
+
+	return client.NewClientFromStub(dut.RawAPIs().GNOI(t).Containerz())
 }
 
 func startContainer(ctx context.Context, t *testing.T) *client.Client {
 	cli := containerzClient(ctx, t)
+	cli.RemoveContainer(ctx, instanceName, true)
```
robshakir, line 51: Suggest to add a comment as to why this is needed (otherwise it's a bit odd that we remove a container when we are calling a fn to start a container).

robshakir, line 51: Can this return an error?

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -98,18 +90,28 @@ func TestDeployAndStartContainer(t *testing.T) {
...
+		})
+		if err != nil {
+			t.Fatalf("unable to list container state for %s", instanceName)
+		}
+
+		for info := range ch {
+			if info.Error != nil {
+				t.Fatalf("unable to list containers: %v", info.Error)
+			}
+			if info.State == "RUNNING" {
```
robshakir, line 105: Should we make the `RUNNING` a const somewhere?

robshakir (changes requested):

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -2,45 +2,73 @@ package container_lifecycle_test
...
-		return grpc.DialContext(ctx, target, grpc.WithTransportCredentials(insecure.NewCredentials()))
-	}
-	cli, err := client.NewClient(ctx, *containerzAddr)
-	if err != nil {
-		t.Fatalf("unable to dial containerz: %v", err)
+
+	dut := ondatra.DUT(t, "dut")
+	switch dut.Vendor() {
+	case ondatra.ARISTA:
+		if deviations.ContainerzOCUnsupported(dut) {
```
robshakir, line 40: Can we file an internal bug to make sure that we have OC modelling for the parameters that we need here (and across vendors)?

robshakir (approved):

robshakir:
Could you fix static analysis please?

robshakir (approved):
LGTM.

PR #4117, opened 2025-04-09
Add OC path for /lacp/interfaces/interface/config/interval

singhavnish2516:
* Add OC path for /lacp/interfaces/interface/config/interval


Swetha-haridasula (approved):
lgtm

PR #4129, opened 2025-04-14
Default COPP test readme

rohit-rp:


feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,184 @@
...
+    *   ATE port-2 IPv4 address ```ap2-v4 = 192.168.1.6/30```
+
+    *   DUT port-1 IPv6 address ```dp1-v6 = 2001:DB8::1/126```
+    *   ATE port-1 IPv6 address ```ap1-v6 = 2001:DB8::2/126```
+
+    *   DUT port-2 IPv6 address ```dp2-v6 = 2001:DB8::5/126```
+    *   ATE port-2 IPv6 address ```ap2-v6 = 2001:DB8::6/126```
+
+
+### SYS-1.1.1 - Test CoppSystemBgp
```
ram-mac, line 41: @rohit-rp - Can we add capability to filter protocol packets and to count in the ACL here so we cover the security part as one of the requirement.

ram-mac (changes requested):

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
rohit-rp: No we cant do that here, we are testing the pre applied copp policy and the test does not modify or configure any existing/new ACL

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
dplore, line 1: This test should be generic, not tied to a specific platform or vendor.  We are simply testing the factory default control-plane-traffic rate limits.

If there are deviations required for the packet loss thresholds, deviations could be proposed.

dplore (dismissed):

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,184 @@
...
+*   Configure BGP between ATE Port-1 and DUT port-1
+*   Generate BGP traffic (update/keepalive) from ATE port-1 within the system defined thresholds
+*   Generate BGP traffic (update/keepalive) from ATE port-1 exceeding the system defined thresholds
+*   Monitor the CoPP counters queue output (show cpu counters queue summary) to verify that:
+    *   Packets exceeding the threshold are dropped for each queue
+    *   Packets within the threshold are not dropped
+    *   BGP Session remains stable throughout the test
+    *   Analyze the CPU utilization to ensure it remains stable under traffic stress
+
+### SYS-1.1.2 - Test CoppSystemL3DstMiss
```
masood-shah, line 56: tename: Test CoPP for L3 Destination Miss Traffic (eg ASIC couldn't find a route for the destination, forcing the packet to be sent to the controller)

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,184 @@
...
+        * shape : 25000 kbps
+        * bandwidth : 1500 kbps
+*   Enable LLDP on DUT port-1
+*   Generate LLDP packets from ATE port-1 to DUT port-1 within and exceeding the system defined thresholds
+*   Monitor the CoPP counters queue output (show cpu counters queue summary) to verify that:
+    *   Packets exceeding the threshold are dropped for each queue
+    *   Packets within the threshold are not dropped
+    *   Analyze the CPU utilization to ensure it remains stable under traffic stress
+
+## Canonical OC Configuration
```
masood-shah, line 154: add a note: when the goal evolves to testing configurable CoPP, this section would ideally contain the (future) standardized OC CoPP config model

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,184 @@
...
+  # LDP:
+  /network-instances/network-instance/mpls/signaling-protocols/ldp/global/config/lsr-id:
+  /network-instances/network-instance/mpls/signaling-protocols/ldp/interface-attributes/interfaces/interface/config/interface-id:
+
+rpcs:
+  gnmi:
+    gNMI.Set:
+```
+
+## Minimum DUT platform requirement
```
masood-shah, line 181: define based on capabilities.... for example "DUT must support CoPP mechanisms for common control plane traffic types (BGP, LDP, LLDP, L2/L3 exceptions etc). verification requires gNMI telemetry interface: ideally standardized OC CoPP state model (future), otherwise OC platform counters (control-plane-traffic), vendor-specific gNMI, or the least desirable CLI access to CoPP counters

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,184 @@
...
+    *   DUT port-1 IPv6 address ```dp1-v6 = 2001:DB8::1/126```
+    *   ATE port-1 IPv6 address ```ap1-v6 = 2001:DB8::2/126```
+
+    *   DUT port-2 IPv6 address ```dp2-v6 = 2001:DB8::5/126```
+    *   ATE port-2 IPv6 address ```ap2-v6 = 2001:DB8::6/126```
+
+
+### SYS-1.1.1 - Test CoppSystemBgp
+
+*   Use the default CoPP with defined thresholds for the queue:
```
Swetha-haridasula, line 43: Instead of checking for default CoPP of a specific vendor, lets define the CoPP policy as part of this test case readme and then validate if the policy is complied

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,184 @@
...
+### SYS-1.1.2 - Test CoppSystemL3DstMiss
+
+*   Use the default CoPP with defined thresholds for the queue:
+    *   CoppSystemL3DstMiss
+        * shape : 2500 kbps
+        * bandwidth : 250 kbps
+    *   CoppSystemL3LpmOver
+        * shape : 1000 kbps
+        * bandwidth : 200 kbps
+*   Note: `CoppSystemL3DstMiss` for destinations missing in FIB and `CoppSystemL3LpmOver` for route lookup in software
```
singhavnish2516, line 65: Please correct queue name to `CoppSystemL3LpmOverflow`

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
dplore: This is a good thought, but testing the factory defaults is a valid test unto itself.   

Adding configuration of copp policy greatly expands the scope of the test beyond the requirements because it  introduces new development needed for OC.   This will time to figure out and get support for because OC's `control-plane-traffic` config  needs some additional `/qos/classifiers/classifier/terms/term/conditions` paths  to cover the matching rules that are commonly needed for copp policies.  (ie: match dest tcp 179 and destination  ip-address of "the system" ).   I suggest this is left to a separate test because this OC feature support is not available from implementations right now.

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
dplore: agreed, doing a control-plane-traffic acl config can be in a separate test for the paths in the tree:

[/system/control-plane-traffic/ingress/acl](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-system.html#system-control-plane-traffic-ingress-acl)

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,184 @@
...
+### SYS-1.1.1 - Test CoppSystemBgp
+
+*   Use the default CoPP with defined thresholds for the queue:
+    *   CoppSystemBgp
+        * shape : 25000 kbps
+        * bandwidth : 1500 kbps
+*   Configure BGP between ATE Port-1 and DUT port-1
+*   Generate BGP traffic (update/keepalive) from ATE port-1 within the system defined thresholds
+*   Generate BGP traffic (update/keepalive) from ATE port-1 exceeding the system defined thresholds
+*   Monitor the CoPP counters queue output (show cpu counters queue summary) to verify that:
```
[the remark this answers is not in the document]
dplore: Need to enumerate the paths needed:

```
/components/component/integrated-circuit/pipeline-counters/control-plane-traffic/vendor
/components/component/integrated-circuit/pipeline-counters/control-plane-traffic/state/dropped-aggregate	
/components/component/integrated-circuit/pipeline-counters/control-plane-traffic/state/dropped-bytes-aggregate
```

I think others have pointed out that these paths might be insufficient to capture what you're looking for.  If so, please propose new OC paths needed to meet your requirements.

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
dplore: An OC path for this is:
/components/component/integrated-circuit/pipeline-counters/drop/state/no-route	

This is not specifically a control-plane-traffic path though, merely a count of packets that didn't have a route.  If you are looking for packets which were dropped due to a rate limit, then we'll need a new path defined, or need to query a vendor extension path from: `/components/component/integrated-circuit/pipeline-counters/control-plane-traffic/vendor`

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
LimeHat: >This test should be generic, not tied to a specific platform or vendor. We are simply testing the factory default control-plane-traffic rate limits.

The way README is written is not aligned with that goal.

There's an assumption of a specific COPP model based on a specific implementation (e.g. individually shaped queues for each protocol, such as BGP, LLDP, etc; with a fixed CIR/PIR). This model simply cannot be assumed for other implementations.

rohit-rp: I have updated the readme to be as generic as possible, omitting vendor specific details and added the note that Vendor implementation may differ and all vendors may not support all the subtests in this readme \

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,184 @@
+# SYS-1.1: Test default COPP policy thresholds for Arista
+
+## Summary
+
+***NOTE:*** This test is a CLI test until necessary YANG/OC is developed and it is applicable to Arista `7280DR3AK-36` only
```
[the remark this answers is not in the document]
rohit-rp: Done

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+*   Ensure that the default CoPP protects the control plane from excessive traffic and prioritizes critical control plane packets
+*   Verify that the default CoPP queues function correctly and handle packet drops as expected
+*   Default policy thresholds have been verified using CLI: `show policy-map type copp copp-system-policy` on `7280DR3AK-36`
```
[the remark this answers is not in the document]
rohit-rp: Done

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,184 @@
...
+A[ATE] <-- Port1(IPv4/6) --> B[DUT];
+B[DUT] <-- Port2(IPv4/6) --> C[ATE];
+```
+
+## Procedure
+
+### Initial setup
+
+*   Connect DUT port-1, 2 to ATE port-1, 2 respectively
+*   Configure IPv4 and IPv6 addresses on DUT and ATE ports as shown below
```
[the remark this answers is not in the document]
rohit-rp: Done

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,184 @@
...
+    *   ATE port-1 IPv6 address ```ap1-v6 = 2001:DB8::2/126```
+
+    *   DUT port-2 IPv6 address ```dp2-v6 = 2001:DB8::5/126```
+    *   ATE port-2 IPv6 address ```ap2-v6 = 2001:DB8::6/126```
+
+
+### SYS-1.1.1 - Test CoppSystemBgp
+
+*   Use the default CoPP with defined thresholds for the queue:
+    *   CoppSystemBgp
```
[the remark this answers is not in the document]
rohit-rp: Ack

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
rohit-rp: Yes we are specifically looking for the queue counter (testing threshold)

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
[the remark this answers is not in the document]
rohit-rp: I wish to decouple the proposal and this readme work. I can work with Vendors and Open FR to have them define the counters/state paths as they see fit for their implementation. Once we are fine with their implementation I will add these paths and use them in the automation. Does that sound reasonable?

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,178 @@
...
+*   Configure BGP between ATE Port-1 and DUT port-1
+*   Generate BGP traffic (update/keepalive) from ATE port-1 within the system defined thresholds
+*   Generate BGP traffic (update/keepalive) from ATE port-1 exceeding the system defined thresholds
+* Monitor CoPP telemetry for the BGP traffic class using the best available method to verify that:
+    *   ***[TODO]*** **gNMI/OC:** Query standardized OC CoPP state model and use OC path via gNMI
+        *   `/components/component/integrated-circuit/pipeline-counters/control-plane-traffic/vendor/..`
+    *   **CLI fallback**: Execute vendor-specific CLI (like show cpu counters queue summary or more specific CoPP counter commands) and parse the output.
+    *   Packets exceeding the threshold are dropped for each queue
+    *   Packets within the threshold are not dropped
+    *   BGP Session remains stable throughout the test
```
dplore, line 56: probably new two sessions on two ports, one which has a good neighbor and a second which has a misbehaving neighbor with too many updates. The good neighbor session should not be affected even when the bad neighbor is rate limited.  The rate limited session may go down, but the good neighbor should stay up.

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,178 @@
...
+    *   Analyze the CPU utilization to ensure it remains stable under traffic stress
```
dplore, line 57: Need to define a CPU utilization threshold here which might be tricky to get something predictable and "correct".  It might be better to stick to the simple verification that the "good neighbor" is not affected?

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+### SYS-1.1.2 - Test L3 destination missing and Longest Prefix Match (LPM) overflow counters
+
+*   Generate traffic from ATE port-1 for destination IP addresses using `172.16.0.0/16` range
+*   The route for these destinations will/should not be present on DUT
+*   Generate traffic within the system defined thresholds and exceeding them
```
dplore, line 63: since we don't know the system thresholds (they are some factory default), instead send a very high amount of traffic (near wire speed) with no ip route.   Then verify bgp sessions stay and and CPU utilization can be fetched and logged.

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
masood-shah: For CoPP validation, CPU monitoring should focus on verifying control-plane stability under duress, rather than adherence to a fixed utilization threshold. Perhaps, initially, establish baseline CPU metrics via OC paths like /system/cpus/cpu/state/* and /system/processes/process/state/cpu-utilization, or vendor CLI commands as a fallback. During traffic injection designed to trigger CoPP, monitor these metrics on both the controller (aka RP, RE, Supervisor) and, critically, on any linecard CPUs if CoPP actions (packet policing) are distributed or offloaded to them. 

Basically, we want to validate CoPP is doing its job right if the device's CPU doesn't freak out and hit 100% all the time when it's super busy. We should still be able to manage the device, and the CPU load should settle down, even if CoPP has to drop some traffic and mess with active connections. This shows the important parts of the system aren't getting swamped :)

dplore: Agreed, so if verifying the "good neighbor" is up by ensuring traffic flows using that neighbor's route do not have packet loss?   (the bad neighbor's traffic flows may be impacted if we are in fact able to get some bgp packets to be dropped by the policer)

You might have to generate/fake a TCP syn flood from the OTG on tcp/179 (bgp) to get the control-plane-traffic policy to start dropping packets.  Would that suffice?  (I am not sure how else to get the OTG device to generate "too many" bgp packets)

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,178 @@
...
+    *   ATE port-2 IPv4 address ```ap2-v4 = 192.168.1.6/30```
+
+    *   DUT port-1 IPv6 address ```dp1-v6 = 2001:DB8::1/126```
+    *   ATE port-1 IPv6 address ```ap1-v6 = 2001:DB8::2/126```
+
+    *   DUT port-2 IPv6 address ```dp2-v6 = 2001:DB8::5/126```
+    *   ATE port-2 IPv6 address ```ap2-v6 = 2001:DB8::6/126```
+
+
+### SYS-1.1.1 - Test CoPP for BGP Traffic
```
Swetha-haridasula, line 45: could we mark this sub-test as todo

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
dplore: I see, the policing could be implemented system wide (such as a kernel level, iptables style rule filtering for all tcp/179 traffic system wide), so the device could start policing all tcp 179 traffic, impacting all neighbors.  So I agree that we can only reliably have a very simple test which sending a lot of tcp/179 traffic and verifies that the qos scheduler counters (copp/policer counters) increment.

So the change needed for this test is to remove the line "BGP Session remains stable throughout the test", as this is not guaranteed. 

```
/system/control-plane-traffic/ingress/qos/scheduler-policy/scheduler-statistics/scheduler/state/violating-pkts	
/system/control-plane-traffic/ingress/qos/scheduler-policy/scheduler-statistics/scheduler/state/violating-octets	
```

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
dplore: So how does one write code to "Analyze the CPU utilization to ensure it remains stable under traffic stress"?   That is a very complicated request for some heuristic algorithm for what it means to be stable.  :)

Instead, we can remove this, or do something really simple like, verify CPU utilization is < 100%.

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
dplore: This line needs to be changed to something like, send 100,000 pps on tcp/179 to the device.   

As it reads now, somehow we have to figure ou what the system defined thresholds are, and we don't know this / don't want to add vendor specific constraints into this test.

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
dplore: SGTM,  please add the reference to the counter here, then you can mark this comment resolved.  Thanks!

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+*   Configure BGP between ATE Port-1 and DUT port-1
+*   Generate BGP traffic (update/keepalive) from ATE port-1 within the system defined thresholds
+*   Generate BGP traffic (update/keepalive) from ATE port-1 exceeding the system defined thresholds
+* Monitor CoPP telemetry for the BGP traffic class using the best available method to verify that:
+    *   ***[TODO]*** **gNMI/OC:** Query standardized OC CoPP state model and use OC path via gNMI
```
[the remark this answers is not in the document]
masood-shah: yeah to maintain vendor-agnosticism as much as possible, the README should reference the generic OC path up to the vendor/ level: /components/component/integrated-circuit/pipeline-counters/control-plane-traffic/vendor/

automation: for specific vendor implementations, the automation should then append the vendor-specific elements (e.g., arista/sand/state) to the path. this allows us to immediately leverage the detailed OC telemetry available from vendors like Arista, while still acknowledging the general OC structure in the documentation

Arista's OC paths for CoPP counters are now confirmed to be a direct, hardware-level equivalent to CLI, removing prior uncertainty. Their OC implementation for CoPP counters provides direct, one-to-one mapping with the CLI and reflects hardware counters. Therefore, for Arista, there's no technical barrier to using OC immediately in our FNT automation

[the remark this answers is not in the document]
rohit-rp: Thanks @dplore and @masood-shah, then no changes are necessary.

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,176 @@
...
+    *   Log the CPU utilization to ensure it remains < 100% under traffic stress
+        *   `/system/cpus/cpu/state/total/instant`
+
+### SYS-1.1.4 - Test System IP Unicast
+
+*   Configure a loopback250 interface on the DUT with IP address `10.10.10.10/32`
+*   Generate IPv4 unicast traffic to the loopback250's IP Address
+*   Generate traffic within and exceeding the system defined thresholds
+* Monitor CoPP telemetry for the system destined IP Unicast traffic class using the best available method to verify that:
+    *   ***[TODO]*** **gNMI/OC:** Query standardized OC CoPP state model and use OC path via gNMI
```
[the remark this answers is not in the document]
rohit-rp: Decided to keep the vendor agnostic path

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,176 @@
...
+    *   Packets within the threshold are not dropped
+    *   Log the CPU utilization to ensure it remains < 100% under traffic stress
+        *   `/system/cpus/cpu/state/total/instant`
+
+### SYS-1.1.7 - Test LDP Thresholds
+
+*   Configure LDP session between ATE port-1 and DUT port-1
+*   Generate LDP traffic within and exceeding the system defined thresholds
+* Monitor CoPP telemetry for LDP traffic using the best available method to verify that:
+    *   ***[TODO]*** **gNMI/OC:** Query standardized OC CoPP state model and use OC path via gNMI
```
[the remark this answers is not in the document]
rohit-rp: Decided to keep the vendor agnostic path

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,176 @@
...
+    *   Packets within the threshold are not dropped
+    *   Log the CPU utilization to ensure it remains < 100% under traffic stress
+        *   `/system/cpus/cpu/state/total/instant`
+
+### SYS-1.1.3 - Test System L2 Unicast thresholds
+
+*   Generate traffic from ATE port-1 destined towards switch’s system mac address
+*   Generate traffic within and exceeding the system defined thresholds
+* Monitor CoPP telemetry for System destined L2 Unicast traffic class using the best available method to verify that:
+    *   ***[TODO]*** **gNMI/OC:** Query standardized OC CoPP state model and use OC path via gNMI
```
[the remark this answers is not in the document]
rohit-rp: Decided to keep the vendor agnostic path

feature/system/control_plane_traffic/otg_tests/default_copp_test/README.md
```diff
@@ -0,0 +1,176 @@
...
+    *   Packets within the threshold are not dropped
+    *   Log the CPU utilization to ensure it remains < 100% under traffic stress
+        *   `/system/cpus/cpu/state/total/instant`
+
+### SYS-1.1.6 - Test LACP thresholds
+
+*   Generate random LACP Data Units (LACPDUs) from ATE port-1 to DUT
+*   Generate traffic within and exceeding the system defined thresholds
+* Monitor CoPP telemetry for LACP traffic using the best available method to verify that:
+    *   ***[TODO]*** **gNMI/OC:** Query standardized OC CoPP state model and use OC path via gNMI
```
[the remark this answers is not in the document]
rohit-rp: Decided to keep the vendor agnostic path

PR #4130, opened 2025-04-15
Modified test cases for AFT FNT phase 1 test cases.

sudhinj:
These test cases are to test the prefix next hop group, next hops and it interfaces, origin.


sudhinj:
The modified test case is uploaded by Suhasini so there is no need for 2 pull request.

PR #4131, opened 2025-04-15
Adding MPLS-2.2: MPLS forwarding via static LSP to BGP next-hop Test

keysightgems:
**Readme Location:** https://github.com/openconfig/featureprofiles/blob/main/feature/mpls/otg_tests/static_bgp_nexthop/README.md

The log is attached here: https://partnerissuetracker.corp.google.com/issues/415458482

Since this is the first time I have raised an issue with deviation, I'm not sure if the metadata pb file should be added or will be generated automatically. Please let me know if the generated pb file should be added as part of the PR.

Added Issue ID's:
1. 409229653 
2. 409240869 
3. 410150319


feature/mpls/otg_tests/static_bgp_nexthop/static_bgp_nexthop_test.go
```diff
@@ -0,0 +1,473 @@
...
+
+	if count >= 2 {
+		t.Logf("Packet loss percentage for flow: got %v, want %v", got, lossTolerance)
+		return false
+	}
+	return true
+}
+
+// Enable MPLS Forwarding
+func configureMplsStaticLsps(t *testing.T, dut *ondatra.DUTDevice) {
```
dplore, line 347: please coordinate with https://github.com/openconfig/featureprofiles/pull/4212/files#diff-b5c7bd119da0491c8203a39ae9c42b48450a258a081d59deada69e0672677469R1347

One of you should create a cfgplugin for creating static LSP's and use the same deviation.  Please avoid duplicating config generation for MPLS static LSP and associated deviations.  Thanks!

keysightgems: Added in the https://github.com/openconfig/featureprofiles/pull/4139. Thanks!

feature/mpls/otg_tests/static_bgp_nexthop/static_bgp_nexthop_test.go
```diff
@@ -0,0 +1,447 @@
...
+	eth := flow.Packet().Add().Ethernet()
+	eth.Src().SetValue(atePort1.MAC)
+	eth.Dst().SetValue(macAddress)
+
+	mpls := flow.Packet().Add().Mpls()
+	mpls.Label().SetValue(mplsLabelV4)
+
+	ip4 := flow.Packet().Add().Ipv4()
+	ip4.Src().SetValue(atePort1.IPv4)
+	ip4.Dst().SetValue(bgpNHv4)
```
sachendras, line 268: Shouldnt the flow destination be `ipv4Dst         = "203.0.113.0/24"` rather? This is a fundamental deviation from the Readme which requires "the traffic flow's destination to be a separate prefix that resolves via that BGP next-hop" and not the BGP next-hop itself.

feature/mpls/otg_tests/static_bgp_nexthop/static_bgp_nexthop_test.go
```diff
@@ -0,0 +1,447 @@
...
+	})
+	t.Run("Verify IPv6 traffic discard when BGP-NH-V6 is not available.", func(t *testing.T) {
+		withdrawBGPRoutes(t, []string{"BGP6.peer2.Route"})
+		verifyIpv6TrafficDiscard(t, ate, topo)
+	})
+}
+
+// configureDUT sets up the DUT interfaces, static LSPs, and BGP neighbors.
+func configureDUT(t *testing.T, dut *ondatra.DUTDevice) {
+	t.Helper()
```
sachendras, line 162: Also, where is the following from the Readme covered?

"Configure static routes on the DUT to discard traffic destined for BGP-NH-V4 and BGP-NH-V6. These routes should point to a Null0 with an administrative distance of 254 to ensure they are less preferred than the BGP routes. This prevents the DUT from using its IGP to reach the BGP next-hops.

feature/mpls/otg_tests/static_bgp_nexthop/metadata.textproto
```diff
@@ -0,0 +1,17 @@
...
+description:  "MPLS forwarding via static LSP to BGP next-hop."
+testbed:  TESTBED_DUT_ATE_4LINKS
+platform_exceptions:  {
+  platform:  {
+    vendor:  ARISTA
+  }
+  deviations:  {
+    interface_enabled:  true
+    default_network_instance:  "default"
+    static_mpls_lsp_oc_unsupported: true
```
sachendras, line 15: What is this deviation expected to do? Configure using vendor CLI? Maybe cfgplugins.NewStaticMPLSLabel addresses it?

feature/mpls/otg_tests/static_bgp_nexthop/static_bgp_nexthop_test.go
keysightgems: updated flow destination. Thanks!

proto/metadata.proto
```diff
@@ -816,6 +816,9 @@ message Metadata {
     // Juniper b/419536104
     bool config_leaf_create_required = 291;
 
+    // Devices does not support static mpls lsp
+    bool static_mpls_lsp_oc_unsupported = 292;
```
dplore, line 820: looks like a duplicate of existing deviation `bool static_mpls_unsupported = 283;` at 
https://github.com/openconfig/featureprofiles/pull/4131/files#diff-4891c0dff9aa45c82082583dfb4f85fa166c328f493edd8adff75f6a84e576aaR791

internal/deviations/deviations.go
```diff
@@ -1462,3 +1462,8 @@ func UseOldOCPathStaticLspNh(dut *ondatra.DUTDevice) bool {
 func ConfigLeafCreateRequired(dut *ondatra.DUTDevice) bool {
 	return lookupDUTDeviations(dut).GetConfigLeafCreateRequired()
 }
+
+// StaticMplsLspOCUnsupported returns true if static mpls lsp parameters are unsupported
+func StaticMplsLspOCUnsupported(dut *ondatra.DUTDevice) bool {
```
dplore, line 1467: as noted elsewhere, I think this is a duplicate of a deviation already defined at `StaticMplsOCUnsupported`

dplore (changes requested):
I think the deviation for static LSP's is already defined, please reuse that, and modify your cfgplugin to return a gnmi.BatchReplace.   Then a gnmi.Set  ([example](https://github.com/openconfig/featureprofiles/blob/2a098804983e8496c9b422e78dca647cace0bb1f/feature/sflow/otg_tests/sflow_base_test/sflow_base_test.go#L253)) to push the config to the DUT in your _test.go code.

keysightgems:
> I think the deviation for static LSP's is already defined, please reuse that, and modify your cfgplugin to return a gnmi.BatchReplace. Then a gnmi.Set ([example](https://github.com/openconfig/featureprofiles/blob/2a098804983e8496c9b422e78dca647cace0bb1f/feature/sflow/otg_tests/sflow_base_test/sflow_base_test.go#L253)) to push the config to the DUT in your _test.go code.

Updated gnmi batch.

proto/metadata.proto
dplore: Ah, thank you for pointing that out, you're correct.  But thanks for changing it here.

internal/cfgplugins/mpls.go
```diff
@@ -53,3 +53,38 @@ func MPLSStaticLSP(t *testing.T, batch *gnmi.SetBatch, dut *ondatra.DUTDevice, l
 		gnmi.BatchReplace(batch, gnmi.OC().NetworkInstance(deviations.DefaultNetworkInstance(dut)).Mpls().Config(), mplsCfg)
 	}
 }
+
+// Configure static MPLS label binding (LBL1) using CLI with deviation, if OC is unsupported on device
+func NewStaticMPLSLabel(t *testing.T, batch *gnmi.SetBatch, dut *ondatra.DUTDevice, lspName string, incomingLabel uint32, intfName string, nextHopIP string, protocolType string) {
```
dplore, line 58: this looks like a near duplicate of `func MPLSStaticLSP`, please combine them.  (It's ok to rename the function to NewMPLSStaticLSP, but if you choose to do that, you'll need to update any dependencies.

dplore (changes requested):
Very close now!

Please also address the gofmt errors.

internal/cfgplugins/mpls.go
keysightgems: Yes, you'r correct but `func MPLSStaticLSP` is already merged into main and uses the duplicate deviation "StaticMplsLspOCUnsupported". If we combine `func NewMPLSStaticLSP`, the already merged code will result in an error. We also need to update the deviation to "StaticMplsUnsupported".

Please suggest what we can do?

dplore: why would it fail?   The only difference in the code is the `if intfName != "" ` check.   Seems like that is compatible.

keysightgems: No, I meant the deviation needs to be replaced from StaticMplsLspOCUnsupported to StaticMplsUnsupported. If I change it, the already merged code that checks for StaticMplsLspOCUnsupported won't find it and will go to the else part.

dplore: Hi, what I think should be done here is instead of introducing `NewStaticMPLSLabel`, instead update the existing `MPLSStaticLSP` and use it in your test.

The only change is the  `if intfName != ""`  check which I think would be backwards compatible with how that function is currently. used.

dplore: If you can successfully make this change, fix the static check issues and resolve conflicts I can approve quickly.

PR #4137, opened 2025-04-16
TE-9.2 StaticLspNextHopsNotSupported deviation added

fmolinar:
Cisco doesn't yet support the latest revision which officially adopted a version of our original augmentation to support explicit path static lsps.  Our augmented model changed to deprecated version.


feature/mpls/otg_tests/static_lsp_test/metadata.textproto
```diff
@@ -11,5 +11,6 @@ platform_exceptions:  {
   }
   deviations:  {
     ipv4_missing_enabled:  true
+    skip_static_lsp_next_hops:  true
```
ram-mac, line 14: @fmolinar - change this name to indicate that this deviation for exercising the old OC path something like
use_old_oc_path_static_lsp_nh

feature/mpls/otg_tests/static_lsp_test/static_lsp_test.go
```diff
@@ -148,8 +148,14 @@ func configureStaticLSP(t *testing.T, dut *ondatra.DUTDevice, lspName string, in
 	mplsCfg := d.GetOrCreateNetworkInstance(deviations.DefaultNetworkInstance(dut)).GetOrCreateMpls()
 	staticMplsCfg := mplsCfg.GetOrCreateLsps().GetOrCreateStaticLsp(lspName)
 	staticMplsCfg.GetOrCreateEgress().SetIncomingLabel(oc.UnionUint32(incomingLabel))
-	staticMplsCfg.GetOrCreateEgress().GetOrCreateLspNextHop(lspNextHopIndex).SetIpAddress(nextHopIP)
-	staticMplsCfg.GetOrCreateEgress().GetOrCreateLspNextHop(lspNextHopIndex).SetPushLabel(oc.UnionUint32(implicitNull))
+	if deviations.StaticLspNextHopsNotSupported(dut) {
```
ram-mac, line 151: Add a comment with description to say that the old path is used as deviation.

proto/metadata.proto
```diff
@@ -737,6 +737,10 @@ message Metadata {
     // Devices that do not support containerz config via OpenConfig.
     bool containerz_oc_unsupported = 266;
 
+    // StaticLspNextHopsNotSupported for devices that do not support static lsp next-hops.
```
ram-mac, line 740: @fmolinar - create a bug ID and add it to the comment here.
The description also need to reflect that the static lsp next hop new oc path (mentioned the path) and this deviation is to run the test with the old oc path.

ram-mac (changes requested):
@fmolinar - Please address the comments and push the commit and i will validate it.

PR #4145, opened 2025-04-21
📙: Add README for gRIBI performance measurement test.

robshakir:
```
 * (A) feature/gribi/otg_tests/gribi_programming_performance/README.md
   - Define a new test for gRIBI programming performance.
```


feature/gribi/otg_tests/gribi_programming_performance/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+ * Utilising the topology described above, configure DUT and ATE.
+ * Establish a gRIBI connection to the DUT, using `FIB_ACK`, `SINGLE_PRIMARY` and `PERSIST` session parameters.
+ * For each of the pool sizes defined above, referred to as N:
+  - Select a set of N routes from the DFZ.
+  - Order the set, and select the 25th, 50th, 75th, 90th, 95th, 100th percentile entries.
+  - Define traffic flows destined to the prefixes selected above.
+  - Begin generating traffic.
+  - Program all entries into the DUT:
+    - Record the programming latency for each prefix for both RIB and FIB ACKs to be received.
```
alshabib, line 37: Should we program before generating traffic?

alshabib (approved):

feature/gribi/otg_tests/gribi_programming_performance/README.md
robshakir: Per the logic here, I'm saying that we have two means of measuring programming latency:

 * the reported operation latency within the test (this can use a time that we send the operation vs. the [`timestamp`](https://github.com/openconfig/gribi/blob/master/v1/proto/service/gribi.proto#L308) that is reported in the `AFTResult` that is returned.
 *  the amount of loss that we observe for the prefixes that we are monitoring above.

We have to start traffic before programming to ensure that we're already sending packets before we start measuring.

Does this make sense? I can include more verbose text if needed.

alshabib: Ah this makes sense, I actually missed that this was for programming latency. I for some reason thought it was about scale.

robshakir: Ah, I think it's both -- it validates that we can hit X scale, as well as then determine the latency of the programming.

robshakir:
@Swetha-haridasula or @nachikethas  -- PTAL, we'd like to start implementing this test.

PR #4185, opened 2025-05-07
Enhance containerz tests to obtain better coverage

goabhinav:
Adding different scenarios for containerz.


feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -86,7 +87,29 @@ func startContainer(ctx context.Context, t *testing.T) *client.Client {
...
+	t.Logf("Verifying image %s:latest exists after push", imageName)
+	imgListCh, err := cli.ListImage(ctx, 0, nil) // List all images
+	if err != nil {
+		t.Fatalf("Failed to list images after push: %v", err)
+	}
+	foundImage := false
+	for img := range imgListCh {
+		if img.Error != nil {
+			t.Errorf("Error received during ListImage iteration: %v", img.Error)
+			continue
```
alshabib, line 100: Should we exit if an error occurs during a ListImage operation?

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +130,798 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
+			t.Errorf("ListImage() did not find the pushed image %s:%s", localImageName, localImageTag)
+		}
+
+		// Cleanup the successfully pushed image for this sub-test.
+		if err := cli.RemoveImage(ctx, localImageName, localImageTag, true); err != nil {
+			t.Logf("Cleanup: Failed to remove image %s:%s: %v", localImageName, localImageTag, err)
+		}
+	})
+
+	// Negative Test: Attempt to push a non-existent tar file.
```
alshabib, line 881: This test is really testing the client and not the containerz server. It is not not useful but I am not sure we want/need this tested here.

goabhinav: Yeah thats right. Removed it now.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
goabhinav: Hmmm my thought process was that listImage would list all the images on the host/device and not just that we just pushed. So the other images that are not of our interest shouldn't bother us giving error. If our image gives error it will be never marked as found and test will fail.

alshabib (dismissed):

feature/container/containerz/tests/container_lifecycle/containerz_test.go
goabhinav: Ack. I've changed it to fail the test now.

Thanks.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +127,689 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
+		}
+	})
+
+	// Negative Test: Attempt to retrieve logs from a stopped container instance.
+	t.Run("LogsFromStoppedInstance", func(t *testing.T) {
+		stoppedInstanceName := "test-instance-for-stopped-logs"
+		localImageName := imageName
+		localImageTag := "latest"
+
+		// 1. Ensure clean state for this specific container instance.
```
robshakir, line 319: For these container setup functions -- can we pull them out into a common function somewhere? It seems like we could re-use a call here across different cases to make the tests a bit more succinct.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +127,689 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
+		foundWantImgAndInstance := false
+		var listedContainersForDebug []string
+
+		for cnt := range listCh {
+			if cnt.Error != nil {
+				t.Errorf("Error received during ListContainer iteration: %v", cnt.Error)
+				continue
+			}
+			listedContainersForDebug = append(listedContainersForDebug, cnt.Name+":"+cnt.ImageName)
+			// Note: Docker prepends "/" to container names. Adjust if your runtime doesn't.
```
robshakir, line 488: Do we know if different container runtimes actually have different behaviours here? (Does podman, for example? I think that most devices are either using podman or docker.)

Generally, I don't think we should be suggesting that people make changes to the tests to make them pass -- either this should be done with a deviation ("I don't support feature X" or "I have different behaviour for feature X" which is considered a bug), or be made common across vendors.

The alternative here is that we make the test strip any leading "/" in all cases.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +127,689 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
+		}
+
+		var foundContainers []string
+		for cntr := range listCh {
+			if cntr.Error != nil {
+				t.Errorf("Error received during ListContainer iteration for %s: %v", instanceName, cntr.Error)
+				continue
+			}
+			// Check if the specific instanceName is still listed.
+			// Note: Docker prepends "/" to container names. Adjust if your runtime doesn't.
```
robshakir, line 530: Same comment as above.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +127,689 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
+			t.Logf("%s %s %s", cntr.Name, cntr.ImageName, cntr.State)
```
robshakir, line 531: Can we make this log message have some context?

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +127,689 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
+	})
+
+	// Negative Test: Upgrade to a non-existent image
+	t.Run("UpgradeToNonExistentImage", func(t *testing.T) {
+		cli := startContainer(ctx, t) // Starts 'instanceName' with 'imageName:latest'
+		defer stopContainer(ctx, t, cli)
+
+		nonExistentImage := "non-existent-image-for-upgrade"
+		_, err := cli.UpdateContainer(ctx, nonExistentImage, "latest", "./cntrsrv", instanceName, false, client.WithPorts([]string{"60061:60061"}))
+		if err == nil {
```
robshakir, line 765:
```suggestion
		if _, err := cli.UpdateContainer(ctx, nonExistentImage, "latest", "./cntrsrv", instanceName, false, client.WithPorts([]string{"60061:60061"})); err == nil {
```

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +127,689 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
+
+	// Negative Test: Upgrade to an existing image but non-existent tag.
+	t.Run("UpgradeToNonExistentTag", func(t *testing.T) {
+		cli := startContainer(ctx, t)
+		defer stopContainer(ctx, t, cli)
+
+		nonExistentTag := "non-existent-tag-for-upgrade"
+		// Ensure the base image 'imageName:latest' exists from startContainer.
+		_, err := cli.UpdateContainer(ctx, imageName, nonExistentTag, "./cntrsrv", instanceName, false, client.WithPorts([]string{"60061:60061"}))
+		if err == nil {
```
robshakir, line 785:
```suggestion
		if _, err := cli.UpdateContainer(ctx, imageName, nonExistentTag, "./cntrsrv", instanceName, false, client.WithPorts([]string{"60061:60061"})); err == nil {
```

robshakir (changes requested):
Can we please try and stick with small changes? @alshabib -- can you sync up with @goabhinav please?

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +127,678 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
+			if prog.Error != nil {
+				t.Fatalf("failed to push image: %v", prog.Error)
+			}
 		}
 
-		for info := range ch {
-			if info.Error != nil {
-				t.Fatalf("unable to list containers: %v", info.Error)
+		_, err = cli.StartContainer(ctx, imageName, "latest", "./cntrsrv", instanceName, client.WithPorts([]string{"60061:60061"}))
+		if err != nil {
```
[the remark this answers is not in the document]
goabhinav: Done

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +127,678 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
+			t.Fatalf("unable to start container: %v", err)
+		}
+		defer stopContainer(ctx, t, cli)
+
+		for i := 0; i < 5; i++ {
```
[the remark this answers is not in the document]
goabhinav: Added suitable comments after refactoring.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +127,678 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
-
-	t.Fatalf("test-instance was not started.")
+		t.Fatalf("Container %s was not started.", instanceName)
+	})
+
+	// Negative Test: Attempt to start container with a non-existent image
+	t.Run("StartWithNonExistentImage", func(t *testing.T) {
+		nonExistentImageName := "non-existent-image"
+		_, err := cli.StartContainer(ctx, nonExistentImageName, "latest", "./cmd", "test-non-existent-img", client.WithPorts([]string{"60061:60061"}))
+		if err == nil {
```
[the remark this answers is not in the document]
goabhinav: Done

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +127,678 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
+	})
+
+	// Negative Test: Attempt to start container with an existing image but non-existent tag
+	t.Run("StartWithNonExistentTag", func(t *testing.T) {
+		// Ensure the base image exists (pushed in the positive test or a previous run)
+		// If not, this test might give a false positive for "image not found" instead of "tag not found".
+		// For simplicity, we assume 'imageName' ("cntrsrv") with 'latest' tag was pushed.
+		nonExistentTag := "non-existent-tag"
+		_, err := cli.StartContainer(ctx, imageName, nonExistentTag, "./cmd", "test-non-existent-tag", client.WithPorts([]string{"60061:60061"}))
+		if err == nil {
```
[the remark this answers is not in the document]
goabhinav: Done

feature/container/containerz/tests/container_lifecycle/containerz_test.go
```diff
@@ -107,182 +127,678 @@ func stopContainer(ctx context.Context, t *testing.T, cli *client.Client) {
...
+	})
+
+	// Negative Test: Attempt to start container with a non-existent image
+	t.Run("StartWithNonExistentImage", func(t *testing.T) {
+		nonExistentImageName := "non-existent-image"
+		_, err := cli.StartContainer(ctx, nonExistentImageName, "latest", "./cmd", "test-non-existent-img", client.WithPorts([]string{"60061:60061"}))
+		if err == nil {
+			t.Errorf("Expected error when starting container with non-existent image %s, but got nil", nonExistentImageName)
+			// Attempt to clean up if it somehow started
+			cli.RemoveContainer(ctx, "test-non-existent-img", true)
```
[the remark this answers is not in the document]
goabhinav: Done

feature/container/containerz/tests/container_lifecycle/containerz_test.go
goabhinav: I'm not sure what would be the result for podman. For docker getting "/" in prefix of container name seems odd to me.
I agree we should not make changes in test to make them pass.
To handle this I'm trimming "/" from the prefix in all cases for now.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
goabhinav: done

feature/container/containerz/tests/container_lifecycle/containerz_test.go
goabhinav: Oops it was added for some debugging while testing. Removed it.

feature/container/containerz/tests/container_lifecycle/containerz_test.go
goabhinav: Done

feature/container/containerz/tests/container_lifecycle/containerz_test.go
goabhinav: Done

robshakir (approved):
Thanks, this looks lots cleaner.

Thanks for making the changes!

alshabib (dismissed):

alshabib (approved):

alshabib (approved):

PR #4186, opened 2025-05-08
Add code to verify trapstats rate after line card reboot in Juniper

singhavnish2516:
* Add code to verify trapstats rate after line card reboot for Juniper devices


Swetha-haridasula (approved):
lgtm

PR #4191, opened 2025-05-11
To Include BGP session Adjacency with Martian IP

Swetha-haridasula:
This Test is to include BGP session check with Martian IP


feature/bgp/otg_tests/base_bgp_session_parameters/base_bgp_session_parameters_test.go
```diff
@@ -526,3 +528,72 @@ func TestParameters(t *testing.T) {
 		})
 	}
 }
+
+// TestBGPMartianIPNeighbor is to verify BGP session establishment with Martian IP.
+// Verification is done through BGP adjacency implicitly.
+func TestBgpMartianIPNeighbor(t *testing.T) {
```
ram-mac, line 534: @Swetha-haridasula - Couple of comments
1. We are specifically adding a martian ip coverage. Should we not be adding in the README too?
2. Can we not use the existing function and add the additional argument in the function to cover this testcase instead of creating a new function?

ram-mac (changes requested):

feature/bgp/otg_tests/base_bgp_session_parameters/base_bgp_session_parameters_test.go
Swetha-haridasula: Yes. Will add it in README
Yes will include change in existing test block

ram-mac (approved):
Looks good to me after changes.

rohit-rp (approved):

PR #4193, opened 2025-05-12
Add coverage for /network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/config/description

singhavnish2516:
* Add coverage for /network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/config/description


ram-mac (commented):
Looks Good; Approving it.

ram-mac (approved):

self-maurya (approved):

PR #4189, opened 2025-05-09
Add UnreferencedAftFibAckUnsupported deviation

mudhireddy:
Add a new deviation to skip fiback verification for unreferenced NH, NHG aft entries as part of testing

PR #4197, opened 2025-05-12
Add UnreferencedAftFibAckUnsupported deviation

mudhireddy:
Add a new deviation to skip fiback verification for unreferenced NH, NHG aft entries as part of testing


mudhireddy:
@nflath, fyi, This one replaces https://github.com/openconfig/featureprofiles/pull/4189

mudhireddy:
@nflath , Can you please provide your comments on this pull request

mudhireddy:
Can someone review this please?

nflath:
What is the rationale for this?

mudhireddy:
> What is the rationale for this?

Arista's EOS will not install entries into hardware if there are no references to the afts by design and will not send FIB_ACK for those. 

NextHopGroup gets installed only when there is an IP aft that uses it. 
NextHop gets installed only when there is at least one installed NextHopGroup that uses it.

This test expects to receive FIB_ACK for NH which are referenced by NHG but there is no IP aft that references NHG. So EOS will not support it and hence this deviation.

gggsmith (dismissed):
Instead of using a deviation for this, we'd prefer a different approach. It's important that we have clear visibility into which tests are affected by this non-compliance, and we can't let a "green" status mask these issues.

With that in mind, please create a separate test for this and move all the failing test cases into it.

Thank you!

gggsmith:
> Thanks for the review. sure I can create a separate test-suite for failing cases.
> 
> Test cases for this feature profile test is coming from a different repo as per README.md file which is https://github.com/openconfig/gribigo/tree/main/compliance
> 
> I need to separate the failed cases in that repo first and create a PR for it and let it merge. After that I can work on featureprofiles repo to split compliance test as suggested in the review.
> 
> Please let me know if this aligns with your expectation and if there are any pitfalls of this approach.
> 
> P.S: Is there any documentation available to use to create new test-suites like TE-15.2 in featureprofiles repo ? Thanks.

Apologies for the late reply. Rather than making the gribigo library aware of different gRIBI support levels, I think we should create a new test (e.g., `gribigo_fiback_compliance`, `TE-15.2`) that uses `Reference` to generate the required test cases.

mudhireddy:
> Rather than making the gribigo library aware of different gRIBI support levels, I think we should create a new test (e.g., `gribigo_fiback_compliance`, `TE-15.2`) that uses `Reference` to generate the required test cases.

No worries and Thank you for your reply. I understood it as, use the Reference field for these non-compliant test cases ("TE-3.1.2", "TE-3.1.3") to skip running them as part of TE-15.1 and add a new test suite `TE-15.2`, `gribigo_fiback_compliance` that only runs those skipped in TE-15.1. am I interpreting it correct ?

Also can you please point me to any document that describes how to add new test suite ?

rohit-rp (approved):

PR #4201, opened 2025-05-14
Update to GNMI v0.14.1

jasdeep-hundal:
`go mod tidy` is updating the gNMI version when I'm looking to update the Ondatra version. This requires a change in `internal/security/gen/generate.go`, separating the patch to update gNMI to clarify intent.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/open-traffic-generator/snappi/gosnappi v1.17.1
 	github.com/openconfig/containerz v0.0.0-20250408205203-029560584812
 	github.com/openconfig/entity-naming v0.0.0-20250108173956-784f643e8b64
-	github.com/openconfig/gnmi v0.11.0
+	github.com/openconfig/gnmi v0.14.1
 	github.com/openconfig/gnoi v0.6.1-0.20250206212518-26d177339690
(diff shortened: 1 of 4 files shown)


dplore:
There's a segfault in the TestDial/gnmi_server test:

```
=== RUN   TestDial/gnmi_server
I0514 00:32:42.803906   15985 cntrsrv.go:177] cntr server listening on [::]:60061
--- FAIL: TestDial (22.18s)
    --- PASS: TestDial/self-dial (1.00s)
    --- PASS: TestDial/timeout (4.55s)
    --- PASS: TestDial/self-dial_and_ping (3.20s)
    --- PASS: TestDial/bad_target_server (4.25s)
    --- PASS: TestDial/gribi_server (7.20s)
    --- FAIL: TestDial/gnmi_server (1.99s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9d3c2c]
```

jasdeep-hundal:
Fixed. I'm not sure when/why the change happened, but the latest generated code for GNMI servers does require the embedding by value change: https://pkg.go.dev/github.com/openconfig/gnmi@v0.14.1/proto/gnmi#UnimplementedGNMIServer

dplore (approved):

PR #4203, opened 2025-05-14
 Added deviations  for config leaf requires create and metadata size adjustment

mnp-juniper:
This deviation addresses the requirement to create a config leaf before a gnmi get is attempted on that leaf.


mnp-juniper:
Closing this pull.
will address these in two different pulls separetely .

PR #4167, opened 2025-04-28
Deviation update

cprabha:
Hi,

I have updated BGP scripts with deviation for setmed action set.

Thanks,
Prabha

PR #4204, opened 2025-05-14
RT-1.27 static_route_bgp_redistribution_test.go

cprabha:
Hi,

1. Added deviation for as path prepend order difference and also set med action set. 

     bgp_set_med_action_unsupported : This deviation is previously approved, added to this test as it is applicable.  
                                                                (Cisco b/414333771)
                                                                gitpull req opened with this deviation for other scripts:  (https://github.com/openconfig/featureprofiles/pull/4167/files) 

    bgp_as_path_prepend_order_mismtach: This is new deviation :  b/425632068
    

2. We need to delete any referenced policy which is imported in BGP/table connection before gnmi replace/delete operation. Hence added code to deleting reference policy in table connection.
    

Thanks,
Prabha


trathod1:
Adding reference to ongoing discussion which suggest correct order should be `64512 65499 65499 65499` - https://github.com/openconfig/featureprofiles/pull/4075/files#r2071158137

ram-mac:
@cprabha few comments
1. the required sequence "64512 64599 64599 64599" and this test is actually now changed to the sequence 
"64599 64599 64599 64512"; We have fixed the test with the right sequence using PR4265
2.The README will be update soon for this test
3.The med related changes are already added via PR4267

Hence, iam closing this PR. Please raise another PR for fixing this test

cprabha:
> Validated this PR and now it is passing for Juniper. Approving it now. But there are some conflicts and failures that needs to be resolved.

I have resolved conflicts, please check

ram-mac:
@cprabha - you need to fix the canonical OC part also; Once that is done i can merge it.

PR #4182, opened 2025-05-05
TRANSCEIVER-9 Fixed with deviations & added cooling-off period.

snaragund:
- Added deviation for skipping 9.4 section as per https://partnerissuetracker.corp.google.com/issues/414842051.
- Using deviation `missing_zr_optical_channel_tunable_parameters_telemetry`
- Added cooling-off period as defined.

"This code is a Contribution to the OpenConfig Feature Profiles project ("Work") made under the Google Software Grant and Corporate Contributor License Agreement ("CLA") and governed by the Apache License 2.0. No other rights or licenses in or to any of Nokia's intellectual property are granted for any other purpose. This code is provided on an "as is" basis without any warranties of any kind."

PR #4209, opened 2025-05-15
Vendor specific ZR optical channel operational mode

AmrNJ:
Update vendor specific operational mode values for ZR optic channels. 
Update: Logic is updated to use vendor specific values when '0' is passed as the OperationalMode variable.

### DIFF
--- a/feature/platform/transceiver/chromatic_dispersion/tests/zr_cd_test/zr_cd_test.go
+++ b/feature/platform/transceiver/chromatic_dispersion/tests/zr_cd_test/zr_cd_test.go
@@ -32,8 +32,7 @@ const (
 var (
 	frequencies                = []uint64{191400000, 196100000} // 400ZR OIF wavelength range
 	targetOutputPowers         = []float64{-13, -9}             // 400ZR OIF Tx power range
-	operationalMode     uint16
-	operationalModeFlag *uint16
+	operationalModeFlag = flag.Int("operational_mode", 0, "vendor-specific operational-mode for the channel")
 )
 
 func TestMain(m *testing.M) {
@@ -99,12 +98,7 @@ func TestCDValue(t *testing.T) {
 
 	dp1 := dut.Port(t, "port1")
 	dp2 := dut.Port(t, "port2")
-	if operationalModeFlag != nil {
-		operationalMode = uint16(*operationalModeFlag)
-	} else {
-		operationalMode = cfgplugins.GetOperationalModeFlag(dut1)
-	}
-	cfgplugins.Initialize(operationalMode)
+	cfgplugins.Initialize(operationalModeFlag, dut)
 	cfgplugins.InterfaceConfig(t, dut, dp1)
 	cfgplugins.InterfaceConfig(t, dut, dp2)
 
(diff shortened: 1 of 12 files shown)


ahsaanyousaf:
Can you please add Juniper and Nokia operational modes related flags as well?

For standard OIF 400ZR Juniper should be 1 and Nokia should be 1083

@rezachit to review and followup

dplore:
Please place the common code into internal/cfgplugins/interface.go or a similar shared configuration library.   Do not cut and paste the same code across more than 1 or 2 tests.

dplore (changes requested):
Please place the common code into internal/cfgplugins/interface.go or a similar shared configuration library. Do not cut and paste the same code across more than 1 or 2 tests.

rezachit:
Echoing dplore's comment, you can add vendor specific mode selection as a function of Vendor in the internal/cfgplugins/interface.go in Initialize function... If the operationalMode was set in the test flag use that value otherwise use a switch/case for vendor (Cisco-->5003, Nokia-->1083, Arista-->1, Juniper-->1). No need to change any test.

rezachit:
In the tests you did not pass t as *testing.T into cfgplugins.Initialize function.

AmrNJ:
Logic is updated to use vendor specific values when '0' is passed as the OperationalMode variable. All common code is now moved into the cfgplugins.Initialize function.

feature/platform/transceiver/chromatic_dispersion/tests/zr_cd_test/zr_cd_test.go
```diff
@@ -99,9 +100,9 @@ func TestCDValue(t *testing.T) {
 
 	dp1 := dut.Port(t, "port1")
 	dp2 := dut.Port(t, "port2")
-
-	operationalMode = uint16(*operationalModeFlag)
-	cfgplugins.Initialize(operationalMode)
+	operationalModeValue = uint16(*operationalModeFlag)
```
rezachit, line 103: Do you mind to simplify the changes in the tests, keep the same naming for operationalMode.. Also you can call the getOpMode() function inside the Initialize(t, dut, mode) functions ... In this case the changes to the test files are minimum.

rezachit (changes requested):

feature/platform/transceiver/chromatic_dispersion/tests/zr_cd_test/zr_cd_test.go
AmrNJ: Sure, this is done now.

snaragund:
@dplore if merge these changes earliest then I will make changes in #4182 according to this commit in zr_laser_bias_current_test.go file also.

internal/cfgplugins/interface.go
```diff
@@ -46,15 +45,39 @@ func init() {
 }
 
 // Initialize assigns OpMode with value received through operationalMode flag.
-func Initialize(operationalMode uint16) {
+func Initialize(t *testing.T, dut *ondatra.DUTDevice, initialOperationalMode uint16) uint16 {
```
dplore, line 48:
```suggestion
func InterfaceInitialize(t *testing.T, dut *ondatra.DUTDevice, initialOperationalMode uint16) uint16 {
```

internal/cfgplugins/interface.go
```diff
@@ -46,15 +45,39 @@ func init() {
...
+			opmode = initialOperationalMode
+			t.Logf("cfgplugins.Initialize: Using provided initialOperationalMode: %d", opmode)
 		}
+		t.Logf("cfgplugins.Initialize: Initialization complete. Final opmode set to: %d", opmode)
 	})
+	return GetOpMode()
+}
+
+// GetOpMode returns the opmode value after the Initialize function has been called
+func GetOpMode() uint16 {
```
dplore, line 79:
```suggestion
func InterfaceGetOpMode() uint16 {
```

dplore (dismissed):
Suggested name change to match style of the test of the interface cfgplugin. Otherwise lgtm

self-maurya (approved):

PR #4220, opened 2025-05-19
Added 5 attempt check for FecUncorrectableBlocks and then check for fec value to zero

priyaranjannanda:
Added logic for 5 attempt and then check whether fed value for zero.


feature/platform/transceiver/fec_uncorrectable_frames/tests/zr_fec_uncorrectable_frames_test/zr_fec_uncorrectable_frames_test.go
```diff
@@ -47,17 +47,29 @@ func TestMain(m *testing.M) {
...
-	}
-	fec, ok := fecStream.Val()
-	if !ok {
-		t.Fatalf("Error capturing streaming Fec value")
-	}
-	if reflect.TypeOf(fec).Kind() != reflect.Uint64 {
-		t.Fatalf("fec value is not type uint64")
+	var fec uint64
+	// Try up to 5 times, sleeping 10s between, if FEC uncorrectable blocks are not zero
+	for attempt := 0; attempt < 5; attempt++ {
```
yiwenhu-g, line 52: Why are 5 attempts needed here? I've tried adding a single time delay of up to three minutes, but the errors persist. Is this problem related to link flapping or data reporting?

priyaranjannanda: > Why are 5 attempts needed here? I've tried adding a single time delay of up to three minutes, but the errors persist. Is this problem related to link flapping or data reporting?

The problem is not related to link flapping. The changes is to give ample time before marking test case as failure. 

We are waiting for 10sec each attempt ( 50s in total) for each stream new values. 

--
Thanks & Regards,
Priya Ranjan

yiwenhu-g: To clarify, I'm unable to pass this test with a single attempt after waiting for up to 3 minutes. This suggests the issue isn't just the time delay. Can you try a similar test to confirm this behavior?

PR #4223, opened 2025-05-20
Explicitly set extended community set name

AmrNJ:
The change is to add coverage to the below path and so extended community set name is set explicitly. 
/routing-policy/defined-sets/bgp-defined-sets/ext-community-sets/ext-community-set/config/ext-community-set-name:


feature/bgp/policybase/otg_tests/link_bandwidth_test/link_bandwidth_test.go
```diff
@@ -688,10 +688,8 @@ func configureExtCommunityRoutingPolicy(t *testing.T, dut *ondatra.DUTDevice) {
 		for name, community := range extCommunitySet {
 			rp := root.GetOrCreateRoutingPolicy()
 			pdef := rp.GetOrCreateDefinedSets().GetOrCreateBgpDefinedSets()
-			stmt, err := pdef.NewExtCommunitySet(name)
```
ram-mac, line 691: @AmrNJ  - Why are we doing away with the error handling from the code?

AmrNJ: GetOrCreateExtCommunitySet function also does the error handling.

ram-mac (approved):
Looks Good

self-maurya (approved):

PR #4232, opened 2025-05-22
RT-5.1 interface unnumbered OC path not supported by Cisco

fmolinar:
RT-5.1 uses CLI to replace the leaf 
/openconfig-interfaces:interfaces/interface/subinterfaces/subinterface/openconfig-if-ip:ipv4/unnumbered/config/enabled


AmrNJ:
This PR is no longer required as the changes needed for IP unnumbered interface OC path is reverted with #4242

PR #4260, opened 2025-06-03
Remove all feature.textproto files and references.

robshakir:
```
 * (D) **/feature.textproto
  - remove feature.textproto since these files are no longer being
    maintained. the README.md for each test case defines the paths
    that are expected to be supported.
 * (M) proto/feature.proto
  - remove all fields other than the feature profile ID that is
    used elsewhere.
 * (D) tools/validate_paths.go
  - remove util that checks contents of feature.textproto.
```

Housekeeping change from your friendly neighbourhood OpenConfig tech lead. This change removes some stale content from the repo, and makes changes to feature.proto to align with the use of this file elsewhere.


masood-shah (approved):
thanks for taking care of this cleanup, Rob! i realized since PR is to delete all feature.textproto files from the repository, the automated check that validates them is no longer necessary and would fail

singhavnish2516 (approved):

ElodinLaarz (approved):

robshakir:
I removed the requirement to have the path validation report successfully, since this merge will remove it.

Thanks for the reviews folks.

PR #4265, opened 2025-06-04
Rt 1.27 update deviation and code

AmrNJ:
1. Update the test to use set_med_action deviation.
2. Update setup function for 1.27.8 subtest. PR #3395 has added gnmi.Delete for removing import policies which is causing regression.


nsadhasivam (approved):
LGTM!

nikki-blr (approved):

PR #4276, opened 2025-06-11
Fix controller_card_redundancy_test

AnilKR123:
In testlet 1: testControllerCardInventory, the following GNMI queries are problematic because they are related to last switchover, and the test has not yet performed a switchover:

- /components/component/state/last-switchover-time

- /components/component/state/last-switchover-reason/trigger

- /components/component/state/last-switchover-reason/details

This will cause the queries to return empty and the test to fail. To fix this issue I moved these queries to testlet 2: testControllerCardSwitchover, where the test performs switchover. By running these queries after a switchover the test is guaranteed to receive valid data in these queries. 

I also found a bug in testlet 2, where after switchover the test is supposed to be verifying that both the active and standby supervisors eventually become switchover ready. The problem is there is a bug in the code such that the test is verifying that the active supervisor is switchover ready 2 times and not checking on the standby at all. I fixed this so that the test verifies both the active and standby are switchover ready. I also bumped the await up to 25 minutes because 20 minutes may not always be enough for the supervisors to become switchover ready.

### DIFF
--- a/feature/platform/controllercard/tests/controller_card_redundancy_test/controller_card_redundancy_test.go
+++ b/feature/platform/controllercard/tests/controller_card_redundancy_test/controller_card_redundancy_test.go
@@ -118,11 +118,33 @@ func testControllerCardSwitchover(t *testing.T, dut *ondatra.DUTDevice, controll
 		t.Errorf("Get rpStandbyAfterSwitch: got %v, want %v", got, want)
 	}
 
+	// Validate controller card last switchover time
+	lastSwitchoverTime := gnmi.OC().Component(rpActiveAfterSwitch).LastSwitchoverTime()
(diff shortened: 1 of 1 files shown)


AnilKR123:
Hi @Swetha-haridasula, could you please approve my PR? Thanks!

Swetha-haridasula (approved):
LGTM

singhavnish2516 (approved):
LGTM. @Swetha-haridasula please confirm that 25 minutes is acceptable for switchover.

singhavnish2516 (changes requested):
Requesting changes as @Swetha-haridasula confirmed that we need the timer to be as per the README.

singhavnish2516 (approved):

PR #4277, opened 2025-06-11
Update testregistry.textproto

ElodinLaarz:
Split out the `SFLOW` and `gNPSI` tests into the numbered values in their READMEs.

Also fix the whitespace of the file.


testregistry.textproto
```diff
@@ -1901,8 +1914,26 @@ test: {
   readme: "https://github.com/openconfig/featureprofiles/blob/main/feature/gnoi/factory_reset/tests/factory_reset_test/README.md"
 }
 test: {
```
dplore, line 1916: There should be a 1:1 mapping between tests and README.md files.

dplore (changes requested):

PR #4280, opened 2025-06-11
PLT-1.1

fmolinar:
adding deviations breakout_group_index_0 and num_physical_channels_unsupported.
b/421356455

1. The path /openconfig-platform:components/component/port/openconfig-platform-port:breakout-mode/groups/group/config/num-physical-channels is currently not supported by Cisco, we are working on raising a feature to be supported.

2. The index used in the test to configure the groups changed from 0 to 1, Cisco index starts at 0 not 1, this is the response given by dev team
"The index needs to be changed to 0 instead of 1, when we did the oc-platform-port uprev we have changed the index to start with 0 instead of 1 based on  the sample config provided in the oc-platform-port mode https://github.com/openconfig/public/blob/1b22027df79f508a15a06a0ec3342ed23c70643b/release/models/platform/openconfig-platform-port.yang#L190"

/components/component[name=Port0/0/0/30]/port/breakout-mode/groups/group[index=1] <----


AmrNJ:
@fmolinar If there are no existing standard values defined for index and can be any vendor dependent value, I think we should go with a switch case instead of a deviation here.

AmrNJ (approved):
LGTM

rohit-rp (approved):

PR #4343, opened 2025-06-26
RT_1.1 base_bgp_session_parameters_test.go

cprabha:
Hi,

We need  to configure cli to enable Martian IP usage. Hence added deviation and configuring native cli to allow martian ip configuration.

Thanks,
Prabha


Swetha-haridasula:
Closed as tracked in b/417395488 and b/387408549

PR #4345, opened 2025-06-27
Fix directory structure

singhavnish2516:
* Update the test paths which do not follow the directory structure: `feature/${feature}/${subfeature}/${testkind}/${testname}`
* Update READMEs to fix the format wherever it is incorrect


Swetha-haridasula (approved):
lgtm

PR #4348, opened 2025-06-27
Code to optimize the PR Approval workflow

singhavnish2516:
* Add code to optimize the PR Approval workflow to run only on manual trigger and approvals. The expectation is to solve the rate limiting issue.
* Ref doc: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_review


ttsuhasini (approved):

ElodinLaarz:
Will this fail if the checks are kicked off in another way (not manual / not a review).

e.g. I just merged the branch to see, and the check doesn't happen? (Perhaps force the check on submit, as well?) 
Just guessing, but perhaps https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request

dplore:
> Hi @dplore, I think we should do the following steps (if not done already) to make sure that this workflow is passing before we allow merge to ensure that no one can merge the PR to main without approvals:
> 
> ```
> 1. Click on Settings (top menu).
> 2. Select Branches from the sidebar.
> 3. Under “Branch protection rules”, click Add rule.
> 4. For Branch name pattern, enter main.
> 5. Check Require status checks to pass before merging.
> 6. Select the workflow(s) you want to make required under Status checks that are required.
> The list will show all recent workflow runs.
> 7. If your workflow hasn’t run yet, trigger it by opening a PR or pushing a commit.
> 8. Click Create or Save changes.
> ```

Yes, this is already in place.  But admin users are allowed to override the branch protection rule

ElodinLaarz:
> > > Hi @dplore, I think we should do the following steps (if not done already) to make sure that this workflow is passing before we allow merge to ensure that no one can merge the PR to main without approvals:
> > > ```
> > > 1. Click on Settings (top menu).
> > > 2. Select Branches from the sidebar.
> > > 3. Under “Branch protection rules”, click Add rule.
> > > 4. For Branch name pattern, enter main.
> > > 5. Check Require status checks to pass before merging.
> > > 6. Select the workflow(s) you want to make required under Status checks that are required.
> > > The list will show all recent workflow runs.
> > > 7. If your workflow hasn’t run yet, trigger it by opening a PR or pushing a commit.
> > > 8. Click Create or Save changes.
> > > ```
> > 
> > 
> > Yes, this is already in place. But admin users are allowed to override the branch protection rule
> 
> Hi @ElodinLaarz / @dplore, Then I think the above changes should be good enough to make sure that none of the PRs go through without at least 1 PR approval until and unless it is overridden by the admin user, which is anyways the case. Do you think we should be good with these changes? If yes, let's try to merge this?
> 
> Thanks

I just want to make sure that "workflow has run and passed" is different than "workflow has run and skipped"? If we think that the former will happen with the current setup-- sounds okay to me.

dplore:
> > > Hi @dplore, I think we should do the following steps (if not done already) to make sure that this workflow is passing before we allow merge to ensure that no one can merge the PR to main without approvals:
> > > ```
> > > 1. Click on Settings (top menu).
> > > 2. Select Branches from the sidebar.
> > > 3. Under “Branch protection rules”, click Add rule.
> > > 4. For Branch name pattern, enter main.
> > > 5. Check Require status checks to pass before merging.
> > > 6. Select the workflow(s) you want to make required under Status checks that are required.
> > > The list will show all recent workflow runs.
> > > 7. If your workflow hasn’t run yet, trigger it by opening a PR or pushing a commit.
> > > 8. Click Create or Save changes.
> > > ```
> > 
> > 
> > Yes, this is already in place. But admin users are allowed to override the branch protection rule
> 
> Hi @ElodinLaarz / @dplore, Then I think the above changes should be good enough to make sure that none of the PRs go through without at least 1 PR approval until and unless it is overridden by the admin user, which is anyways the case. Do you think we should be good with these changes? If yes, let's try to merge this?
> 


The desire is to have at least one approver from ALL code owner groups.  The way the branch protection rule work is if 1 approver from ANY codeowner group approves, then the PR is able to be merged.  

In other words, we want one approver from the `featureprofiles-(.*)-owner` AND one approver from `featureprofiles-approvers`

dplore (approved):

PR #4350, opened 2025-06-27
 P4RT1.3 P4RT behavior when a device/node is down

ArpitJoshi18:
Two changes are made to support P4RT-1.3 :
1. For DisableLinecard function: Added dynamic verification of disabled Linecard by using GNMI.OperStatus to confirm Linecard is INACTIVE. (Similar as EnableLinecard func)
This change is for all vendors. 
2. For Juniper, Added delay through deviation to wait after component down operation is confirmed, to ensure all the processes pertaining to the Linecard is inactive. 
This change is only for Juniper vendor.


feature/p4rt/tests/p4rt_device_down_test/p4rt_device_down_test.go
```diff
@@ -405,7 +435,11 @@ func TestP4rtConnect(t *testing.T) {
 	}
 
 	// Disable line card associated with port2
-	toggleLinecardPower(t, dut, lc2ComponentName, false)
+	disableLinecard(t, dut)
+	if deviations.ComponentPowerdownDelay(dut) {
```
rohit-rp, line 439: Waiting for processes to start and stop seems logical but on the other hand I also expect LC status to be reflected once relevant process have come up. By definition a deviation means that you are noncompliant with the standard or expected behaviour. is that the case and do you plan to fix this deviation?
Also don't you need to do the same after the LC is enabled?

ArpitJoshi18: That is correct. LC status reflect upon any state change. 
This was opened previously under different behavior,  these changes are not required anymore. 
Discarding the pull-request.

PR #4353, opened 2025-06-30
To fix the weight of BE0 queue

Swetha-haridasula:
This is to correct the weight of BE) queue


sezhang2 (approved):

feature/qos/otg_tests/one_sp_queue_traffic_test/README.md
```diff
@@ -188,7 +188,70 @@ paths:
   /qos/scheduler-policies/scheduler-policy/schedulers/scheduler/inputs/input/state/queue:
   /qos/scheduler-policies/scheduler-policy/schedulers/scheduler/inputs/input/state/weight:
 
+## Canonical OpenConfig for keepalive-interval and hold-time at neighbour and peer-group level for BGP
```
singhavnish2516, line 191: The heading should read `## Canonical OC`. Any other thing can be part of the sub-headings starting with "###". It should be part of JSON codeblock. YAML codeblock is now incorrect as you have mistakenly added this in the YAML block.

singhavnish2516 (changes requested):
Please address the comments.

PR #4362, opened 2025-07-01
Waiting for successful reboot of device

priyaranjannanda:
- The earlier logic was coming out from loop on error.
- Added else block to break the loop on reboot success.


priyaranjannanda:
Git pull https://github.com/openconfig/featureprofiles/pull/5558

PR #4364, opened 2025-07-02
Add example code to generate canonical oc

singhavnish2516:
* Add example code to generate canonical oc


doc/canonical_oc_generation_example/main.go
```diff
@@ -0,0 +1,68 @@
+package main
```
dplore, line 1: should have a package comment

PR #4365, opened 2025-07-02
Authz fix

ram-mac:
changes in authz
 - marked hardverify to false as now this cannot be done with the current system. Need some changes 

changes in authz1_4_test
 - have RPC checks to be specific instead of all the RPC's


mihirpitale-googler (approved):
lgtm.

PR #4370, opened 2025-07-03
rt1.12

hmod2001:
This PR fixes a validation failure in the TestAlwaysCompareMED ONDATRA test where BGP routes retained the MED value of 50 even after MED policies were removed from the DUT. The test previously failed because the expected MED value after policy removal was 0, but the prefixes retained the prior configured value due to improper cleanup or re-advertisement delays.

Proposed changes
Explicitly delete the existing SET-MED routing policies from peer groups.

Replace those policies with a permissive ALLOW import policy to accept routes without modifying MED.

Log updated peer group configurations to verify changes.

Introduce a 30-second delay to allow BGP route reprocessing.

test passing with logs added below

[logs (1).log](https://github.com/user-attachments/files/21042318/logs.1.log)

### DIFF
--- a/feature/bgp/otg_tests/bgp_always_compare_med/bgp_always_compare_med_test.go
+++ b/feature/bgp/otg_tests/bgp_always_compare_med/bgp_always_compare_med_test.go
@@ -626,6 +626,8 @@ func TestAlwaysCompareMED(t *testing.T) {
 	t.Run("Remove MED settings on DUT", func(t *testing.T) {
 		t.Log("Disable MED settings on DUT.")
 		dutPolicyConfPath := gnmi.OC().NetworkInstance(deviations.DefaultNetworkInstance(dut)).Protocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP").Bgp()
+
+		// Apply the allow-all policy to remove MED modifications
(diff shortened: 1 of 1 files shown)


feature/bgp/otg_tests/bgp_always_compare_med/bgp_always_compare_med_test.go
```diff
@@ -634,6 +636,61 @@ func TestAlwaysCompareMED(t *testing.T) {
...
+
+		// First, disable all neighbors
+		for _, nbrIP := range nbrIPs {
+			nbrPath := bgpPath.Neighbor(nbrIP)
+			t.Logf("Disabling BGP neighbor %s", nbrIP)
+			gnmi.Update(t, dut, nbrPath.Enabled().Config(), false)
+		}
+
+		// Wait for sessions to go down completely
+		t.Log("Waiting for BGP sessions to go down...")
```
ram-mac: @hmod2001 - The wait time is relative to the timer set on the BGP speaker. We should just use the status check on the device for the period of Keepalive configured on the device., by default it is 10secs within which if there is no bgp keepalives received then it would send out notification message and tear the session.

ram-mac (changes requested):

PR #4371, opened 2025-07-03
RT-5.3 set config container leafs corresponding to list key leafs

fmolinar:
added the function fptest.ConfigureDefaultNetworkInstance(t, dut) to set config container leafs corresponding to list key leafs as well as initialize the static protocol.


ram-mac (approved):
LGTM. This should fix the balance test failure

PR #4405, opened 2025-07-15
Adding PF-1.9 Egress handling of TTL Test

keysightgems:
**Readme Location:** https://github.com/openconfig/featureprofiles/blob/main/feature/ttl/otg_tests/egress/README.md

**The log is attached here:** https://partnerissuetracker.corp.google.com/issues/415458482


proto/metadata.proto
```diff
@@ -852,6 +852,15 @@ message Metadata {
     // Nokia b/430183279
     bool explicit_swap_src_dst_mac_needed_for_loopback_mode = 301;
 
+    // Devices does not support static mpls lsp
+    bool static_mpls_lsp_oc_unsupported = 302;
+
+    // Device doesnot support gre enacapsulation
```
danameme, line 858:
```suggestion
    // Device does not support gre enacapsulation
```

proto/metadata.proto
```diff
@@ -852,6 +852,15 @@ message Metadata {
...
+    bool gre_decapsulation_oc_unsupported = 303;
+
+    // Device doesnot support decapsulation group
```
danameme, line 861:
```suggestion
    // Device does not support decapsulation group
```

danameme (approved):

internal/cfgplugins/mpls.go
```diff
@@ -53,3 +60,30 @@ func MPLSStaticLSP(t *testing.T, batch *gnmi.SetBatch, dut *ondatra.DUTDevice, l
 		gnmi.BatchReplace(batch, gnmi.OC().NetworkInstance(deviations.DefaultNetworkInstance(dut)).Mpls().Config(), mplsCfg)
 	}
 }
+
+func MPLSStaticLSPByPass(t *testing.T, batch *gnmi.SetBatch, dut *ondatra.DUTDevice, lspName string, incomingLabel uint32, nextHopIP string, protocolType string, byPass bool) {
```
dplore, line 64: Hi, it is not clear to me what this bypass feature is?  Can you add some description?  I don't see bypass or access-list mentioned in the README as a requirement?

Since the OC is the same for this function as the `MPLSStaticLSP`, I am wondering if we are missing some required functionality in OC to cover this?

@danameme  for more comments

keysightgems: @dplore , @vvlakshmanamurthy 

Yes, you're right — this is not mentioned in the README. However, during implementation of test case **PF-1.9.11**, we encountered a requirement to validate **TTL behavior for inner/outer IP and MPLS layers**. To achieve correct forwarding and TTL preservation in that scenario, we found that a static bypass LSP was needed. Without it, the expected behavior could not be validated.

Initially, I considered extending the existing `MPLSStaticLSP` function, but that would require introducing an additional input parameter. Since this function is already used in merged scripts, changing its signature would break backward compatibility.

To avoid disruption, I introduced a new function `MPLSStaticLSPByPass` specifically for scenarios that require bypass behavior. Please let me know how you'd like to proceed — either keep the separate function or refactor the original one with conditional support for bypass LSPs.

**I took reference from the existing code :** https://github.com/openconfig/featureprofiles/blob/main/feature/policy_forwarding/otg_tests/mpls_gre_ipv4_decap_test/mpls_gre_ipv4_decap_test.go#L327  (here used `MPLSStaticLSPConfig` function, here values are hardcoded so introduced a new function ``MPLSStaticLSPByPass`)
https://github.com/openconfig/featureprofiles/blob/main/internal/cfgplugins/policyforwarding.go#L218

keysightgems:
> (also note, gofmt error here: https://github.com/openconfig/featureprofiles/actions/runs/16669491033/job/47357562654?pr=4405)

Fixed!

internal/cfgplugins/mpls.go
dplore: @vvlakshmanamurthy this is one for you or your team to decide if you want to approve.

vvlakshmanamurthy: I dont think this is the right way to do i.e. adding bypass. 

@keysightgems can you please share a bug or conversation from Arista ? If not i can create one, please let me know.

keysightgems: > I dont think this is the right way to do i.e. adding bypass.
> 
> @keysightgems can you please share a bug or conversation from Arista ? If not i can create one, please let me know.

@vvlakshmanamurthy There is no bug or conversation with Arista, this was used based on the existing script. Please create an issue. Thanks!

ANISH-GOTTAPU:
> Bypass config should be clarified with Arista team and the same needs to be removed from the code once clarified and please do not submit this until the "bypass" knob is clarified.

Hi @vvlakshmanamurthy : Will you be checking this with Arista or you want us to take it. We don't have permission to raise issue with Arista, we can just ask them in Vendor chat. Please let us know who will take this forward.

keysightgems:
> Bypass config should be clarified with Arista team and the same needs to be removed from the code once clarified and please do not submit this until the "bypass" knob is clarified.

Clarified with Arista team, they accepted with "access-list bypass". Thanks!

PR #4407, opened 2025-07-15
Removed ATE Interface Status Check for Port Validation

priyaranjannanda:
The ATE port status check is not required for this test case.

However, when attempting to perform the ATE port check via a gNMI call, the operation fails with the following error:

<error>
interface_loopback_aggregate_test.go:268: Failed to get client: error dialing IxNetwork: error requesting API key: error on HTTP request: 
POST https://10.53.241.176:40051/api/v1/auth/session
tls: first record does not look like a TLS handshake
</error>

This suggests a potential TLS misconfiguration or an incorrect endpoint being used for the gNMI call.


priyaranjannanda:
Closing the pull request and will raise a new one.

PR #4413, opened 2025-07-17
RT-5.6 Use loopback mode TERMINAL ref:PR1155

nsadhasivam:
RT-5.6 Use loopback mode TERMINAL ref:https://github.com/openconfig/public/pull/1155
As per b/404876484


Swetha-haridasula (changes requested):
Please change the run log to reflect Terminal, also justify the reason to change the await to 5 minuts for link to be Up

PR #4416, opened 2025-07-18
Fixed the issues in Gnmi19

mnp-juniper:
Fixed the  few issues in the code : (No deviation required)
-------------------------------
Removed  the hardcoded IP addresses 
Fixed the wrong ISIS protocol type 
Used array to generate ae names in the loop instead of netutils method
Added a 60s sleep after the switchover


mnp-juniper:
raised fresh pull 
https://github.com/openconfig/featureprofiles/pull/4544

PR #4433, opened 2025-07-22
Fix expected TTL in outer header for RT-3.53.5 and RT-3.53.7

danameme:


vvlakshmanamurthy (approved):

feature/static/otg_tests/next_hop_encap_udp/README.md
```diff
@@ -225,16 +225,20 @@ Verify:
           "next-hop-groups": {
             "next-hop-group": [
               {
-                "name": "ENCAPNHG1",
```
dplore: you need to keep this key

dplore (approved):
couple of comments that I think you should address before merging, but otherwise, LGTM

PR #4437, opened 2025-07-24
AFT-1.1: Update Arista deviations

yini101:
Update Arista deviations for AFT-1.1 to enable interfaces and correctly configure ISIS. Also add a check for ISIS when configuring and establishing an adjacency.

### DIFF
--- a/feature/afts/otg_tests/afts_base/afts_base_test.go
+++ b/feature/afts/otg_tests/afts_base/afts_base_test.go
@@ -187,7 +187,14 @@ func (tc *testCase) configureDUT(t *testing.T) error {
 		}
 	})
 	ts.ATEIntf1.Isis().Advanced().SetEnableHelloPadding(false)
-	ts.PushAndStart(t)
+
+	if err := ts.PushAndStart(t); err != nil {
+		return err
+	}
+
+	if _, err = ts.AwaitAdjacency(); err != nil {
+		return fmt.Errorf("no IS-IS adjacency formed: %v", err)
+	}
 	return nil
 }
 
--- a/feature/afts/otg_tests/afts_base/metadata.textproto
+++ b/feature/afts/otg_tests/afts_base/metadata.textproto
@@ -20,7 +20,9 @@ platform_exceptions: {
   }
   deviations: {
     default_network_instance: "default"
-    isis_single_topology_required: true
+    isis_instance_enabled_required: true
+    isis_interface_afi_unsupported: true
     low_scale_aft: true
+    interface_enabled: true
   }
 }


feature/afts/otg_tests/afts_base/metadata.textproto
```diff
@@ -20,7 +20,9 @@ platform_exceptions: {
   }
   deviations: {
     default_network_instance: "default"
-    isis_single_topology_required: true
+    isis_instance_enabled_required: true
+    isis_interface_afi_unsupported: true
     low_scale_aft: true
+    interface_enabled: true
```
ttsuhasini, line 26: Where are these deviations used in the code? I can't see them being used in the test. Please point me to the line no.

sudhinj: Could you please let me know where is the bgp deviation. It mentions about ISIS and low scale is set to true. We need high scale isnt it.

yini101: > Where are these deviations used in the code? I can't see them being used in the test. Please point me to the line no.

This comes from `ts := isissession.MustNew(t).WithISIS()` where `WithISIS()` calls `addISISOC()` and uses it there. It's not very obvious, but without this, the opposite behavior would fail to configure on the device.

`addISISOC`: https://github.com/openconfig/featureprofiles/blob/dfca16c6e57449b71a0d82a7ac7cda6575457714/internal/isissession/isissession.go#L111

> Could you please let me know where is the bgp deviation. It mentions about ISIS and low scale is set to true. We need high scale isnt it.

The BGP deviation is going to be a separate PR for high scale which is coming very soon. This is the PR that allows this test to pass for Arista at minimum for low scale first.

feature/afts/otg_tests/afts_base/afts_base_test.go
```diff
@@ -187,7 +187,14 @@ func (tc *testCase) configureDUT(t *testing.T) error {
...
 	ts.ATEIntf1.Isis().Advanced().SetEnableHelloPadding(false)
-	ts.PushAndStart(t)
+
+	if err := ts.PushAndStart(t); err != nil {
+		return err
+	}
+	_, err = ts.AwaitAdjacency()
+	if err != nil {
+		return err
+	}
```
ElodinLaarz, line 197: Nit: Return more context in the err and in-line the `ts.AwaitAdjacency`.

E.g.
```
if _, err = ts.AwaitAdjacency(); err != nil {
    return fmt.Errorf("no IS-IS adjacency formed: %v", err)
}
```

ttsuhasini (approved):

ElodinLaarz (approved):

rohit-rp (approved):

PR #4440, opened 2025-07-25
add new FNT for chassis reboot with large config push

Swetha-haridasula:
This test is for covering b/418856494.


Swetha-haridasula:
Incorrect pull

PR #4445, opened 2025-07-25
Add test gnpsi-1 Sampling and Subscription Check

ijernoiu:
Readme location: https://github.com/openconfig/featureprofiles/blob/main/feature/gnpsi/otg_tests/sampling_test/README.md

Deviations identified for test gNPSI-1
gnpsi_oc_unsupported - issue id: 433989578


feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+			samplingRate: flow.SamplingRate,
+			size:         record.FrameLength,
+			packet:       record.Header,
+		}
+	}
+
+	for _, fc := range flowConfigs {
+		configureFlow(t, &top, &fc)
+		ate.OTG().PushConfig(t, top)
+		otg.StartProtocols(t)
```
ElodinLaarz, line 227: Since this function is called `verify...` I am surprised that we're doing configuration here? I would refactor this to keep the configuration separate from the verification steps. (Here and elsewhere.)

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+
+		sampleCount := 0
+		samplesWithFlows := 0
+		sflowPacketsToValidate := []sflowPacket{}
+		t.Logf("Starting traffic for %s", fc.name)
+		go func() {
+			otg.StartTraffic(t)
+			waitForTraffic(t, otg, fc.name, trafficTime)
+		}()
+		time.Sleep(2 * time.Second)
```
ElodinLaarz, line 237: Try not to in-line magic values? I am not even sure what this is doing here. Maybe it's trying to avoid a race condition? But we should be using wait groups.

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+		timeout := time.After(trafficTime)
+		continueLoop := true
```
ElodinLaarz, line 239: Give this a better name to describe the situation when the loop is cancelled.

There is almost certainly a cleaner way to wait here, as we could do something like `if time.Since(start) < trafficTime` (if things are given proper units) which seems a bit suspicious.

This would also allow us to remove the `select` statement which is contributing to the deep nesting here.

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+		otg.StartProtocols(t)
+
+		sampleCount := 0
+		samplesWithFlows := 0
+		sflowPacketsToValidate := []sflowPacket{}
+		t.Logf("Starting traffic for %s", fc.name)
+		go func() {
+			otg.StartTraffic(t)
+			waitForTraffic(t, otg, fc.name, trafficTime)
+		}()
```
ElodinLaarz, line 236: Why is this in a separate go routine?

Do we not want to block on this?

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+				} else {
+					t.Logf("Total SFlow packets: %d", len(sflowPacketsToValidate))
+					checkSFlowPackets(t, dut, sflowPacketsToValidate, fc)
+				}
+				continueLoop = false
+			default:
+				resp, err := stream.Recv()
+				if err != nil {
+					t.Errorf("[ERROR] Error receiving gNPSI sample: %v", err)
+					continue
```
ElodinLaarz, line 255: If we are error-ing here, do we need to continue?

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+				}
+				sampleCount++
+				t.Logf("Received gNPSI sample no %d:", sampleCount)
+				if len(resp.Packet) > 0 {
```
ElodinLaarz, line 259: The nesting in here is a bit wild-- swapping the order of some checks or refactoring out to another function can help make it cleaner to read.

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+								case layers.SFlowExtendedSwitchFlowRecord:
+									continue
+								case layers.SFlowExtendedRouterFlowRecord:
+									continue
+								default:
+									t.Logf("Unknown record type: %T, value: %+v", r, r)
+								}
+							}
+						}
+					} else {
```
ElodinLaarz, line 283: e.g. the "error" flow here can be swapped to

```
if len(sflow.FlowSamples) == 0 {
  t.Logf("No flow samples in this SFlow packet")
}
```

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
ElodinLaarz: More likely, after reading the logic below, we'd like to have a go routine processing the gNPSI notifications, whilst we wait for the flow to finish, and then when the flow is completed, we can then send a "finish" signal to the processor, wait for it to finish up (wait groups), and then continue our verification.

(e.g. call out to `checkSFlowPackets` outside any sort of loop in the main flow of the verification loop.)

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+		case ondatra.ARISTA:
+			if strings.HasPrefix(intfName, "Ethernet") {
+				num, _ := strings.CutPrefix(intfName, "Ethernet")
+				parts := strings.Split(num, "/")
+				if len(parts) == 2 {
+					slot, _ := strconv.Atoi(parts[0])
+					port, _ := strconv.Atoi(parts[1])
+					return uint32(slot*1000 + port)
+				}
+			}
```
ElodinLaarz, line 611: Don't in-line vendor deviations. (And also this vendor deviation is surprising-- but I'll ignore that for now...)

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+const (
+	sampleSize                        = 256
+	grpcPort                          = 6070
+	trafficTime                       = 30 * time.Second
+	IPv4                              = "IPv4"
+	IPv6                              = "IPv6"
+	samplingRate                      = 1000000
+	packetRate                        = 500000
+	defaultPacketsToSend       uint32 = 6000000
+	flowCountTolerancePct             = 0.5
```
ElodinLaarz, line 42: This seems surprisingly high? And also is not mentioned in the README. Seems suspicious.

I assume this tolerance is because the test is not actually written to behave properly. We should be able to account for where all these values have gone.  

(Making something up, but we could check `/sampling/sflow/collectors/collector/state/packets-sent` or similar, as an example-- I.e. "there are ways to not worry about 50% of samples being missing")

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+				num, _ := strings.CutPrefix(intfName, "Ethernet")
+				parts := strings.Split(num, "/")
+				if len(parts) == 2 {
+					slot, _ := strconv.Atoi(parts[0])
+					port, _ := strconv.Atoi(parts[1])
+					return uint32(slot*1000 + port)
+				}
+			}
+		}
+		return 0
```
ElodinLaarz, line 613: Also this just returns 0 for other vendors-- please write a test that behaves reasonably independent of the vendor present.

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+			t.Logf("SFlow Packet %d: Egress interface matches expected interface %d", index+1, egressIntf)
+		}
+
+		switch flowConfig.ipType {
+		case IPv4:
+			ipLayer := sflowPacket.packet.Layer(layers.LayerTypeIPv4)
+			if ipLayer != nil {
+				ipv4 := ipLayer.(*layers.IPv4)
+				t.Logf("Source IP: %s, Destination IP: %s", ipv4.SrcIP, ipv4.DstIP)
+				if ipv4.SrcIP.String() != atePortPair[0].IPv4 || ipv4.DstIP.String() != atePortPair[1].IPv4 {
```
ElodinLaarz, line 656: Avoid calls to `[0]` and `[1]` since there's no useful iteration here.

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+			t.Logf("SFlow Packet %d: Ingress interface matches expected interface %d", index+1, ingressIntf)
+		}
+
+		if sflowPacket.egressIntf != egressIntf {
+			t.Errorf("[ERROR] SFlow packet %d: Egress interface %d does not match expected interface %d", index+1, sflowPacket.egressIntf, egressIntf)
+		} else {
+			t.Logf("SFlow Packet %d: Egress interface matches expected interface %d", index+1, egressIntf)
+		}
+
+		switch flowConfig.ipType {
```
ElodinLaarz, line 650: These could be their own functions-- the nesting is pretty hard to follow.

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+	dp2 := dut.Port(t, "port2")
+	ingressIntf := processInterfaceNumber(dut, dp1.Name())
+	egressIntf := processInterfaceNumber(dut, dp2.Name())
+
+	adjustedSize := flowConfig.frameSize
+	if adjustedValues, found := adjustedFrameSizeMap[flowConfig.frameSize]; found {
+		adjustedSize = adjustedValues[flowConfig.ipType]
+	}
+
+	for index, sflowPacket := range sflowPackets {
```
ElodinLaarz, line 626: Optional: We could just drop all the `else` cases and just log the sflow packet overall? And then we get the value in the logs and if we don't have an error, the value is good.

So many if / else is messy as-is.

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+				} else {
+					t.Logf("Sflow Packet %d: IPv4 source and destination IP match expected values", index+1)
+				}
+
+			} else {
+				t.Errorf("[ERROR] No IPv4 layer found in packet")
+			}
+		case IPv6:
+			ipLayer := sflowPacket.packet.Layer(layers.LayerTypeIPv6)
+			if ipLayer != nil {
```
ElodinLaarz, line 668: Swap error indent flow.

```
if ipLayer == nil {
  // t.Error
  return
}
// good case
```

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+	egressIntf := processInterfaceNumber(dut, dp2.Name())
+
+	adjustedSize := flowConfig.frameSize
+	if adjustedValues, found := adjustedFrameSizeMap[flowConfig.frameSize]; found {
+		adjustedSize = adjustedValues[flowConfig.ipType]
+	}
+
+	for index, sflowPacket := range sflowPackets {
+		if sflowPacket.size != adjustedSize {
+			t.Errorf("[ERROR] SFlow packet size %d does not match expected frame size %d", sflowPacket.size, flowConfig.frameSize)
```
ElodinLaarz, line 628: `adjustedSize` instead of `frameSize`?

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+		}
+	}
+}
+
+func restartGnpsiService(t *testing.T, dut *ondatra.DUTDevice) {
+	switch dut.Vendor() {
+	case ondatra.ARISTA:
+		t.Log("Restarting gNPSI service on DUT")
+		cliTemplate := `
+management api gnpsi
```
Ankur19, line 574: Any restart logic if required in test should be done through gNOI protocol, which requires the process name.

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+security pki certificate generate self-signed %s key %s parameters common-name localhost
+!
+		
+management security
+ssl profile %s
+certificate %s key %s
+trust certificate %s
+tls versions 1.2 1.3
+!
+`, keyFile, certFile, keyFile, profileName, certFile, keyFile, certFile)
```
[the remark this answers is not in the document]
Ankur19: For more context, we do allow some cli within the test if it is necessary for the test, e.g. testing something on or off in the test that doesnot have any OpenConfig.

However, in your case, the Device Under Test (dut) is expected to already have a SSL profile configured and gNPSI enabled (may not be configured with the right parameters). So this config can be part of your binding setup.

topologies/binding/options.go
```diff
@@ -32,6 +32,7 @@ var (
 	p4rtPort    = flag.Int("p4rt_port", 9559, "default P4RT part")
 	ateGNMIPort = flag.Int("ate_gnmi_port", 50051, "default ATE gNMI port")
 	ateOTGPort  = flag.Int("ate_grpc_port", 40051, "default ATE OTG port")
+	gnpsiPort   = flag.Int("gnpsi_port", 6070, "default gNPSI port")
```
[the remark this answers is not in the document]
ijernoiu: The default value of the port is mentioned in the gNPSI EFT TOI document from Arista

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+
+	t.Logf("Configuring Loopback Interface")
+	configureLoopbackInterface(t, dut)
+
+	t.Logf("Configuring Interfaces")
+	configureDUTPort(t, dut, &dutPort1, dp1)
+	configureDUTPort(t, dut, &dutPort2, dp2)
+
+	t.Log("Configuring SSL Profile")
+	configureSslProfile(t, dut, false)
```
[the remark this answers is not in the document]
ijernoiu: Removed

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+			if data != "" {
+				os.WriteFile(certFile, []byte(data), 0644)
+			}
+
+			getKeyDataCommand := fmt.Sprintf("bash cat %s%s", keyPath, keyFile)
+			data = runCliCommand(t, dut, getKeyDataCommand)
+			if data != "" {
+				os.WriteFile(keyFile, []byte(data), 0644)
+			}
+		}
```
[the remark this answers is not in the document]
ijernoiu: This is a leftover from the first tries to establish the gnpsi server connection. I have removed the entire part

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	sampleSize                        = 256
+	grpcPort                          = 6070
+	trafficTime                       = 30 * time.Second
+	IPv4                              = "IPv4"
+	IPv6                              = "IPv6"
```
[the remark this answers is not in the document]
ijernoiu: There are multiple places where ipType is expected as a string value (flow configuration, cli workarounds, utils functions). The 2 constants are not tied to the sflow plugin in any way. I think that a better approach would be to have a constants package from which the tests would reuse what's needed

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,919 @@
...
+	"github.com/openconfig/ondatra/otg"
+	"github.com/openconfig/ygnmi/ygnmi"
+	"github.com/openconfig/ygot/ygot"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	sampleSize                        = 256
+	grpcPort                          = 6070
```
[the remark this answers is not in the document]
ijernoiu: Rework to get the port from binding, but code looks less readable that declaring a const for the port number:
svc := introspect.GNPSI
dialer := introspect.DUTDialer(t, dut, svc)
t.Logf("Using gNPSI port: %d", dialer.DevicePort)

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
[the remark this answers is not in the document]
ijernoiu: We are working on a split setup, so I cannot expect to have things pre-configured on the DUT as we are always running on a fresh config. If the SSL profile related config can be moved to a binding file specific for this test, I will try to rework it with this approach and cleanup the test

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
ijernoiu: The current approach was to have the wait for traffic to finish in the "background" and process the gNPSI notifications in the main routine.

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
ijernoiu: On the Arista setup used for test development, the SFlow packet ingress and egress interfaces have the format 1001 for Eth1/1, 2001 for Eth2/2 and so on. There is no other setup available to see how the gNPSI server behaves for a different vendor

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
ijernoiu: I see that there is a gnoi.KillProcess method that also handles process restart, but the gNPSI process should also be defined in the gnoi package and checked to see what the process name is for multiple vendors

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
ijernoiu: This function is configuring the tester flows (no dut configuration is applied here). Knowing that some test cases have multiple flow configs that are applied sequentially,  I will edit the run function for each test case so that it configures the flows outside and then calls the verify function on the existing tester configuration.

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
ijernoiu: This logic will be reworked by moving the traffic wait in the main routine, checking the gNPSI samples in a different go routine and once the traffic has finished (as opposed to when the timeout occurs) stop the separate routine and start processing received samples.

ijernoiu:
I have addressed the remaining comments:
1. Wait for samples on a separate go routine
2. Use GNOI to restart the GNPSI service
3. SSL profile configuration moved to binding file (using Arista specific CLI commands)

Updated run log can be found at: https://partnerissuetracker.corp.google.com/issues/415458482

Once gnmi support for configuring GNPSI is in place, I will implement the OC GNPSI configuration function.

ijernoiu:
Rebased on latest main changes. Bug for tracking GNPSI OC support: #433989578

feature/gnpsi/otg_tests/sampling_test/sampling_and_subscription_check_test.go
```diff
@@ -0,0 +1,750 @@
...
+				return uint32(n)
+			}
+		}
+	default:
+		if n, err := strconv.Atoi(intfName); err == nil {
+			return uint32(n)
+		}
+	}
+	return 0
+}
```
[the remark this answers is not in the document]
ElodinLaarz: (From our offline discussion.)

From reading the spec, it seems like the value "should be" the `ifIndex`? So, we can compare against that to find which interface this comes from.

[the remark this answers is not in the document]
ijernoiu: Checking SFlow Packet interface values against interface ifindex was successful so this function was replaced with ifindex check

ElodinLaarz:
```
Step #1 - "fp-presubmit": === RUN   TestSamplingAndSubscription
Step #1 - "fp-presubmit":     sampling_and_subscription_check_test.go:171: Configuring Loopback Interface
Step #1 - "fp-presubmit": I1029 14:23:55.665853   35879 knebind.go:372] Dialing service "gNMI" on DUT dut with options [0xc003ae9dc0 0xc003ae9dd0 0xc003ae9de0]
Step #1 - "fp-presubmit":     sampling_and_subscription_check_test.go:171: Configuring Interfaces
Step #1 - "fp-presubmit":     sampling_and_subscription_check_test.go:171: Got interface Ethernet9 ifIndex: 9
Step #1 - "fp-presubmit":     sampling_and_subscription_check_test.go:171: Got interface Ethernet8 ifIndex: 8
Step #1 - "fp-presubmit":     sampling_and_subscription_check_test.go:171: Configuring SFlow
Step #1 - "fp-presubmit":     sampling_and_subscription_check_test.go:171: Configuring GNPSI
Step #1 - "fp-presubmit": W1029 14:23:57.879919   35879 knebind.go:184] Service "gNPSI" has no known KNE name, trying "gNPSI"
Step #1 - "fp-presubmit":     sampling_and_subscription_check_test.go:171: Failed to retrieve DUT grpc dialer: service "gNPSI" not found on DUT "dut"
Step #1 - "fp-presubmit": --- FAIL: TestSamplingAndSubscription (2.21s)
Step #1 - "fp-presubmit": FAIL
```

Alas, Ondatra's KNE binding [doesn't know](https://github.com/openconfig/ondatra/blob/5174d60e8dfc654c2957a4025ab8a5c557a223a2/knebind/solver/solver.go#L59) how to find gNPSI.

@ijernoiu Can you open an issue in the Ondatra repo to add gNPSI support? (Doesn't block this PR...)

---

Also can you include your latest logs from a run of the test since you changed it to start using ifIndex? Code seems generally good.

ElodinLaarz (commented):
Generally LGTM. Just run the tests again to show the ifIndex mappings are working as expected.

The last thing about adding OpenConfig pushed gNPSI is tracked in partner issue 433989578.

internal/cfgplugins/gnpsi.go
```diff
@@ -0,0 +1,61 @@
...
+func ConfigureGNPSI(t *testing.T, dut *ondatra.DUTDevice, params *GNPSIParams) {
+	if deviations.GnpsiOcUnsupported(dut) {
+		configureGNPSIFromCLI(t, dut, params)
+	} else {
+		configureGNPSIFromOC(t, dut)
+	}
+}
+
+func configureGNPSIFromOC(t *testing.T, dut *ondatra.DUTDevice) {
+	// TODO : Implement GNPSI OC configuration when supported in OC
```
dplore:
```suggestion
	// TODO : Implement GNPSI OC configuration when supported in OC (ref: https://github.com/openconfig/public/pull/1385)
```

ijernoiu:
Updated test run logs can be found at: https://partnerissuetracker.corp.google.com/issues/415458482

PR #4449, opened 2025-07-28
README: add remove policy bound port test

rohit-rp:
For an interface that is bound to a PBF policy
*   Remove an interface bound to a policy
*   Add a new interface to the policy
*   Remove an interface from both a policy and the device


feature/policy_forwarding/otg_tests/add_remove_policy_bound_interface_test/README.md
```diff
@@ -0,0 +1,206 @@
+# PF-1.24: Add and remove interface bound to PBF
```
dplore, line 1: Could we leverage the existing [base_vrf_selection](https://github.com/openconfig/featureprofiles/tree/77f186259fc15381e4ce4d84ddf0d4b935de39ae/feature/policy_forwarding/policy_vrf_selection/otg_tests/base_vrf_selection) test and do one of:

1. Add  "remove " and "add" steps to the existing test
 
or

3. in this new README, refer to the base_vrf_selection test, reusing that same configuration, and then do the interface add/remove steps as already mentioned here?  

My goal is  to simplify the code/reuse existing logic

dplore: Ok, I see.  SGTM.

dplore (approved):

dplore:
Please also add this test to `testregistry.textproto`.

PR #4452, opened 2025-07-28
Fix testbed name for CNTR-2 tests

goabhinav:


feature/container/networking/tests/container_connectivity/metadata.textproto
```diff
@@ -4,4 +4,4 @@
 uuid: "dc9a9682-8ec1-4ca0-8f54-635d690bb488"
 plan_id: "CNTR-2"
 description: "Container network connectivity tests"
-testbed: TESTBED_DUT_ATE_2LINKS
+testbed: TESTBED_DUT_DUT_4LINKS
```
alshabib, line 7: Do we need four links? Or is that the minimum?

PR #4454, opened 2025-07-28
Refactor aft script to use config push in a common library

ElodinLaarz:
In preparation for writing the AFT Reboot script, which will use a lot of the same infrastructure, we factor out some logic to do the configuration into a common library to be called by the old and new scripts.

### DIFF
--- a/internal/afts/afts.go
+++ b/internal/afts/afts.go
@@ -1,0 +1,301 @@
+// Package afts provides helper functions to advertise routes to be verified in AFT.
+package afts
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/open-traffic-generator/snappi/gosnappi"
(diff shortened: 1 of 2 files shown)


internal/afts/afts.go
```diff
@@ -0,0 +1,311 @@
...
+		return fmt.Errorf("bgpV4Addr is nil")
+	}
+	if route.Dev == nil {
+		return fmt.Errorf("device is nil")
+	}
+	return nil
+}
+
+// ConfigureAdvertisedEBGPRoutes advertises the given routes over eBGP from the ATE on the given interfaces.
+func ConfigureAdvertisedEBGPRoutes(t *testing.T, route *BGPRoute) error {
```
dplore, line 65: bgp configuration should be in files within internal/cfgplugins

PR #4467, opened 2025-08-06
Update Ondatra to v0.10.1

jasdeep-hundal:
OTG YANG models are being updated to provided data for IS-IS Graceful Restart testing (https://github.com/open-traffic-generator/models-yang/commit/b1f53895fc16472538354af7e211ead5b07c1851). OTG GNMI test clients will need to be able to interpret these paths. This is captured in the latest Ondatra release (commit at https://github.com/openconfig/ondatra/commit/01f6057d3a6674db64f2830ff0224aad9fcd01f9), which we need to up date featureprofiles to use.


singhavnish2516:
Hi @jasdeep-hundal,
Can you please add a description/ reason for this change?

PR #4475, opened 2025-08-07
gNMI 1.6 FNT test for gNMI server running on default and non-default network instance

dbarney1106:
gNMI 1.6 FNT test for gNMI server running on default and non-default network instance

### DIFF
--- a/feature/gnmi/gnmi_ni_test/gnmi_ni_test.go
+++ b/feature/gnmi/gnmi_ni_test/gnmi_ni_test.go
@@ -1,0 +1,120 @@
+package gnmi_ni_test
+
+import (
+	"testing"
+
+	"github.com/openconfig/featureprofiles/internal/attrs"
+	"github.com/openconfig/featureprofiles/internal/deviations"
+	"github.com/openconfig/featureprofiles/internal/fptest"
+	"github.com/openconfig/ondatra"
+	"github.com/openconfig/ondatra/gnmi"
+	"github.com/openconfig/ondatra/gnmi/oc"
+)
+
+func TestMain(m *testing.M) {
+	fptest.RunTests(m)
+}
+
+const (
+	customVRFName = "customVRFName"
+)
+
+var (
+	dutPort1 = attrs.Attributes{
+		Desc:    "dutPort1",
+		IPv4:    "192.0.2.1",
+		IPv4Len: 30,
+		IPv6:    "2001:0db8::192:0:2:1",
+		IPv6Len: 126,
+	}
+
+	dutPort2 = attrs.Attributes{
+		Desc:    "dutPort2",
+		IPv4:    "192.0.2.5",
+		IPv4Len: 30,
+		IPv6:    "2001:0db8::192:0:2:5",
+		IPv6Len: 126,
+	}
+)
+
+func TestCodelab(t *testing.T) {
+	// configure DUT
+	dut := ondatra.DUT(t, "dut")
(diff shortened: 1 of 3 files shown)


feature/gnmi/gnmi_ni_test/gnmi_ni_test.go
```diff
@@ -0,0 +1,120 @@
...
+func ConfigureDUT(t *testing.T, dut *ondatra.DUTDevice) {
+
+	dp1 := dut.Port(t, "port1")
+	dp2 := dut.Port(t, "port2")
+
+	// Configure default network instance.
+	fptest.ConfigureDefaultNetworkInstance(t, dut)
+
+	// Configure gNMI server on default network instance.
+	fptest.CreateGNMIServer(t, dut, deviations.DefaultNetworkInstance(dut))
```
dplore, line 57: let's move this function to /internal/cfgplugins

Functions in cfgplugins should be called to construct and return a gnmi.BatchReplace object.  ([example](https://github.com/openconfig/featureprofiles/blob/bc105a443b44d862c70e91112708b5a339c71ae5/internal/cfgplugins/sflow.go#L52))   The BatchReplace object can be appended to by multiple cfgplugins until all the desired config is present.

Then code in the _test.go file should call batch.Set() to replace the portion of the configuration in the BatchReplace object.  ([example](https://github.com/openconfig/featureprofiles/blob/main/feature/sflow/otg_tests/sflow_base_test/sflow_base_test.go#L212-L215))

This allows the config snippets to be reused to compose configs according to the test requirements.  The ondatra call that touches the DUT (batch.Set) is what's really being tested (we are not testing the config generation), so the Set should be in the _test.go, not the configuration generation helper.

feature/gnmi/gnmi_ni_test/gnmi_ni_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// Deviations for vendors that require explicit interface to network instance assignment.
+	if deviations.ExplicitInterfaceInDefaultVRF(dut) {
+		fptest.AssignToNetworkInstance(t, dut, dp1.Name(), deviations.DefaultNetworkInstance(dut), 0)
+		fptest.AssignToNetworkInstance(t, dut, dp2.Name(), deviations.DefaultNetworkInstance(dut), 0)
+	}
+}
+
+// ConfigureAdditionalNetworkInstance configures a new network instance in DUT and changes the network instance of port2
+func ConfigureAdditionalNetworkInstance(t *testing.T, dut *ondatra.DUTDevice, ni string) {
```
dplore, line 71: as above, this should move to cfgplugins

internal/fptest/networkinstance.go
```diff
@@ -34,6 +34,15 @@ func ConfigureDefaultNetworkInstance(t testing.TB, d *ondatra.DUTDevice) {
 	})
 }
 
+// ConfigureCustomNetworkInstance configures a non-default network instance name and type.
+func ConfigureCustomNetworkInstance(t testing.TB, d *ondatra.DUTDevice, ni string) {
```
dplore, line 38: This looks good, but instead of adding to fptest, use cfgplugins

dplore (changes requested):

feature/gnmi/gnmi_ni_test/gnmi_ni_test.go
dbarney1106: batch.set added with helper functions

PR #4488, opened 2025-08-14
Add new policy-definitions deviation to RT-1.28

frasieroh:
Forthcoming changes will introduce a deviation to the policy-definitions container. Once this happens, this test will fail because the gNMI GET output will not unmarshal into the standard struct.


frasieroh:
Hi Darren and Ram, thanks for the feedback. Primarily I just wanted to get ahead of the conversation and point out this is _going_ to be a problem but I'm not at all picky about the specific fix. If there's a way to handle this in ondatra than we can do that. I'll try to pick this back up shortly.

PR #4495, opened 2025-08-19
Restore Juniper cPTX for cJunosEvolved use.

kmo7:
cJunosEvolved is the new specific example of cPTX.

* Restores the cPTX directory with the test topo files.

* Adds content to the README file related to above.


robshakir:
@rohit-rp @nikki-blr -- PTAL

kmo7:
> @kmo7 Please fix the proto validation error
If by this comment, you mean to add the ci/cd related changes back in, those were left out on purpose since Google no longer wishes to use cPTX (cJunosEvolved now) and had actually removed our code from featureprofiles here last year. I added it back in after discussion and agreement with Google. It is added in for the benefit of other people who are interested in using cJunosEvolved with Ondatra. Also, as these are just topology definition files, they don't affect the running code normally. They had been tested already when we used them last and I tested the ones we use again before submitting for review. Long answer, hoping it addresses the comment.

PR #4501, opened 2025-08-20
Fix directory names for zr tests

singhavnish2516:


dplore (commented):
Need to also update testregistry.textproto

dplore (approved):
I approve, but please update testregistry.textproto before merging.

nikki-blr (approved):

PR #4502, opened 2025-08-20
Public static binding implementation of Ondatra DialSSH API

kumaryoge:
This is an implementation of DialSSH API which was added to Ondatra binding (https://github.com/openconfig/ondatra/commit/7739ec5b05349c5798adc9b31270d2eed2def4e1) to enable tests to ssh into devices via username/password or public key authentication methods.


topologies/binding/binding.go
```diff
@@ -197,27 +197,63 @@ func (d *staticDUT) DialP4RT(ctx context.Context, opts ...grpc.DialOption) (p4pb
...
+	case binding.PasswordAuth:
+		config = &ssh.ClientConfig{
+			User: auth.User,
+			Auth: []ssh.AuthMethod{
+				ssh.Password(auth.Password),
+				ssh.KeyboardInteractive(sshInteractive(auth.Password)),
+			},
+		}
+	case binding.KeyAuth:
+		signer, err := ssh.ParsePrivateKey(auth.Key)
```
Ankur19, line 226: This will probably not work with our use case since we may use a key chain.

For our use case, can you do: `signer, err := ssh.Unmarshal(auth.Key, ssh.Signer)`?

topologies/binding/ssh_test.go
```diff
@@ -0,0 +1,249 @@
+// Copyright 2022 Google LLC
```
Ankur19, line 1: The license is outdated. Can you use addlicense tool?

https://github.com/google/addlicense

topologies/binding/ssh.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2022 Google LLC
```
Ankur19, line 1: License is outdated

PR #4503, opened 2025-08-20
TE_17.1 : vrf_policy_driven_te_test.go 

cprabha:
Hi,

1. For DECAP-DEFAULT FLOW,  we need to use dscpEncapNoMatch in both inner and outer HDR.  Instead if we use dscpEncapMatch packets will be sent over encap tunnel and will not send via DEFAULT VRF.

2. For IPV6 packets,  ECN uses last two bits of TOS field.  Hence ipv6 tos will be calculated by DSCP (left shift by 2)+ ECN . 
    Ex: If we send packets with 30 DSCP, and ECN = 11 (3) TOS =  30 << 2 + 3 = 123
         If we send packets with 30 DSCP, and ECN = 01 (1) TOS =  30 << 2 + 1 = 121

3.  IF the packet is marked as ECN 11, which means packet congestion experienced, in such cases we need to copy ECN 11 to outer/inner HDR in order communicate about congestion in the network .

4.  There is no TTL used in createFlow , hence added TTL value. We can use same for verification purposes. Otherwise expected TTL may vary.

Thanks,
Prabha


feature/gribi/otg_tests/vrf_policy_driven_te/vrf_policy_driven_te_test.go
```diff
@@ -1741,14 +1745,12 @@ func validateTrafficDecap(t *testing.T, captureFile *os.File, expectedInHdrDscp
 		} else {
 			testStats.IPv6CapturedPackets++
 			ipv6Packet, _ := ipv6Layer.(*layers.IPv6)
-			if actualDscp := uint32(ipv6Packet.TrafficClass); actualDscp != expectedInHdrDscp {
+			// In IPv6 packets, ECN uses last two bits of the TOS byte. And TOS field be calculated as DSCP<<2 (left shift by 2)+ ECN.
+			v6ExpectedInHdrDscp := expectedInHdrDscp<<2 + expectedInHdrEcn
+			if actualDscp := uint32(ipv6Packet.TrafficClass); actualDscp != v6ExpectedInHdrDscp {
```
dracazacula, line 1750: thank you for spotting this. the logic is now correct, but can we keep the structure with 2 separate checks, one for DSCP and another for ECN. The error message would be some much clear and it would be easier to spot where the error is.

same as how we check for IPv4: 
image

```
actualDscp = uint32(ipv6Packet.TrafficClass >> 2)
if actual != expectedDscp ....

actualEcn = uint32(ipv6Packet.TrafficClass & 0x03)
if actualEcn != expected ...
```

Thanks!

feature/gribi/otg_tests/vrf_policy_driven_te/vrf_policy_driven_te_test.go
```diff
@@ -1969,7 +1971,7 @@ func testGribiDecapMatchSrcProtoNoMatchDSCP(ctx context.Context, t *testing.T, d
 						isInnHdrV4:  true,
 						InnHdrSrcIP: atePort1.IPv4,
 						InnHdrDstIP: ipv4InnerDst,
-						inHdrDscp:   []uint32{dscpEncapA1},               // Different than outer DSCP
+						inHdrDscp:   []uint32{dscpEncapNoMatch},          // Different than outer DSCP
```
dracazacula, line 1974: Can the comments be deleted, as they are no longer relevant? Thank you!

feature/gribi/otg_tests/vrf_policy_driven_te/vrf_policy_driven_te_test.go
```diff
@@ -1980,16 +1982,16 @@ func testGribiDecapMatchSrcProtoNoMatchDSCP(ctx context.Context, t *testing.T, d
 						isInnHdrV4:    false,
 						InnHdrSrcIPv6: atePort1.IPv6,
 						InnHdrDstIPv6: ipv6InnerDst,
-						inHdrDscp:     []uint32{dscpEncapA1},               // Different than outer DSCP
+						inHdrDscp:     []uint32{dscpEncapNoMatch},          // Different than outer DSCP
```
dracazacula, line 1985: same comment as above for the comments. thank you!

dracazacula (changes requested):

feature/gribi/otg_tests/vrf_policy_driven_te/vrf_policy_driven_te_test.go
cprabha: updated thank you

feature/gribi/otg_tests/vrf_policy_driven_te/vrf_policy_driven_te_test.go
```diff
@@ -108,7 +108,7 @@ const (
 	tolerancePct             = 2
 	flowNegTest              = "flowNegTest"
 	ipv4InnerDst             = "138.0.11.8"
-	ipv6InnerDst             = "2001:db8::138:0:11:8"
+	ipv6InnerDst             = "2001:db8::138:0:11:3"
```
dracazacula, line 111: test-1 is checking the scenario where we want to reach the `DEFAULT` vrf, so modifying the IP to match the `ENCAP` rule would make the test deviate from its intent.

PR #4514, opened 2025-08-23
Fixing partially the hiba_authentication test

ram-mac:
Below are the changes
1.Using dir := t.TempDir() for creating temporary directory to store ssh keys/certificates.
2.Parsing this dir while creating hiba keys 	"credz.CreateHibaKeys(t, dut, dir)"
3.Passing this dir to all the relevant functions inside Credz.go


internal/security/credz/credz.go
```diff
@@ -279,7 +279,7 @@ func RotateAuthenticationArtifacts(t *testing.T, dut *ondatra.DUTDevice, keyDir,
 	var artifactContents []*cpb.ServerKeysRequest_AuthenticationArtifacts
 
 	if keyDir != "" {
-		data, err := os.ReadFile(fmt.Sprintf("%s/%s", keyDir, dutKey))
+		data, err := os.ReadFile(fmt.Sprintf("%s/%s", keyDir, dut.Name()))
```
dipchauh, line 282: Why is this changed to `dut.Name()`? I see the name `dut` still being used in many places to create ssh keys/certificates. This can create inconsistencies and cause the test to fail.
https://github.com/openconfig/featureprofiles/blob/634c13fb50263b4fd7b01ef82c462dfebcf680b1/feature/gnsi/credentialz/tests/host_certificates/host_certificates_test.go#L63
https://github.com/openconfig/featureprofiles/blob/634c13fb50263b4fd7b01ef82c462dfebcf680b1/feature/gnsi/credentialz/tests/host_certificates/host_certificates_test.go#L71
https://github.com/openconfig/featureprofiles/blob/634c13fb50263b4fd7b01ef82c462dfebcf680b1/internal/security/credz/credz.go#L430-L432

ram-mac: Without this change, the test was not able to find the certificate; With this change at least it will know which dut to read from.
The inconsistency will go away after i make some more changes, see PR4511

PR #4523, opened 2025-08-28
Add Test DP-1.17: Egress traffic DSCP rewrite

desaimg1:
Readme location:
https://github.com/openconfig/featureprofiles/blob/main/feature/qos/otg_tests/egress_traffic_classification_and_rewrite_test/README.md

Test is failing coz of this bug ID : [b/419618177](https://issuetracker.google.com/issues/419618177)

The log is attached here: https://partnerissuetracker.corp.google.com/issues/415458482

### DIFF
--- a/feature/qos/otg_tests/egress_traffic_classification_and_rewrite_test/egress_traffic_classification_and_rewrite_test.go
+++ b/feature/qos/otg_tests/egress_traffic_classification_and_rewrite_test/egress_traffic_classification_and_rewrite_test.go
@@ -1,0 +1,1081 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package egress_traffic_classification_and_rewrite_test
+
+import (
+	"net"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+	"github.com/google/gopacket/pcap"
+	"github.com/open-traffic-generator/snappi/gosnappi"
+	"github.com/openconfig/featureprofiles/internal/attrs"
+	"github.com/openconfig/featureprofiles/internal/cfgplugins"
+	"github.com/openconfig/featureprofiles/internal/deviations"
+	"github.com/openconfig/featureprofiles/internal/fptest"
+	"github.com/openconfig/featureprofiles/internal/helpers"
+	"github.com/openconfig/featureprofiles/internal/otgutils"
+	"github.com/openconfig/featureprofiles/internal/qoscfg"
+	"github.com/openconfig/ondatra"
+	"github.com/openconfig/ondatra/gnmi"
+	"github.com/openconfig/ondatra/gnmi/oc"
+	"github.com/openconfig/ondatra/netutil"
+	"github.com/openconfig/ygnmi/ygnmi"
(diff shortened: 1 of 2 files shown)


feature/qos/otg_tests/egress_traffic_classification_and_rewrite_test/egress_traffic_classification_and_rewrite_test.go
```diff
@@ -0,0 +1,1081 @@
...
+			af6.Enabled = ygot.Bool(true)
+		}
+	}
+	gnmi.Replace(t, dut, dutConfPath.Config(), niProto)
+
+}
+
+// Configures the given DUT interface.
+func configInterfaceDUT(p *ondatra.Port, a *attrs.Attributes, dut *ondatra.DUTDevice) *oc.Interface {
+	i := a.NewOCInterface(p.Name(), dut)
```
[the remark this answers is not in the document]
desaimg1: comments addressed

danameme (approved):

sezhang2 (approved):

sezhang2:
> > The qos config needs to use cfgplugins/qos.go.
> 
> Hi @dplore Thanks for reviewing it. we are using CLI after pushing partial supported config through OC and based on OC config we are adding the DSCP values which is specific to this testcase. In this case is it necessary to move the config to cfgpluggins.

Hi @desaimg1 Can you use a deviation for CLI config per @dplore 's comment and open a bug to track it? Then we can review/approve it. thanks!

desaimg1:
> > > The qos config needs to use cfgplugins/qos.go.
> > 
> > 
> > Hi @dplore Thanks for reviewing it. we are using CLI after pushing partial supported config through OC and based on OC config we are adding the DSCP values which is specific to this testcase. In this case is it necessary to move the config to cfgpluggins.
> 
> Hi @desaimg1 Can you use a deviation for CLI config per @dplore 's comment and open a bug to track it? Then we can review/approve it. thanks!

HI @sezhang2 I have already added deviations and bug details in the script. updated the script now as well..

internal/cfgplugins/qos.go
```diff
@@ -372,3 +375,53 @@ func GetPolicyCLICounters(t *testing.T, dut *ondatra.DUTDevice, policyName strin
...
 }
+
+func ConfigureQosDscpRemarkSpecific(t *testing.T, dut *ondatra.DUTDevice) {
+	if deviations.QosRemarkOCUnsupported(dut) {
+		switch dut.Vendor() {
+		case ondatra.ARISTA:
+			aristaConfigureRemarkIpv4(t, dut)
+			aristaConfigureRemarkIpv6(t, dut)
+		}
+	}
```
dplore, line 386: Please include the OC standard way to rewrite/remark DSCP.  The key path is:

[/qos/classifiers/classifier/terms/term/actions/remark/config/set-dscp](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-qos.html#qos-classifiers-classifier-terms-term-actions-remark-config-set-dscp)

desaimg1: it is included in qos.go file in line number 156.
		if !deviations.QosRemarkOCUnsupported(dut) {
			action.GetOrCreateRemark().SetDscp = ygot.Uint8(class.RemarkDscp)
		}

since it is aborting the script, we are adding it as another function.

desaimg1:
> The qos remarking function needs to specify the OC standard way to configure DSCP (in addition to the vendor deviation you discovered)

@dplore 
OC standard way of DSCP rewrite is already inluded in qos.go of cfgplugins file in line number 156.
if !deviations.QosRemarkOCUnsupported(dut) {
action.GetOrCreateRemark().SetDscp = ygot.Uint8(class.RemarkDscp)
}

since it is aborting the script, we are adding it as another function.

internal/cfgplugins/qos.go
dplore: ok, please add comments explaining this ordering of the configuration both inline where ConfigureQosDscpRemarkSpecific is called, and as a description for the function.

Also this should be a private function, so it should be lower case.

dplore (changes requested):

dplore (changes requested):

dplore:
Approved.  Please bring your branch up to date with main and this can be merged

PR #4528, opened 2025-09-01
gNMI-1.14 fix

youssefyamoutyy:
gNMI-1.14 failing with:

**error-message 1:** IPv6 Router Advertisement 'suppress' leaf is deprecated and cannot be used together with 'enable' leaf. Please use 'enable' leaf to manage IPv6 Router Advertisement.
**error-message 2:** metadata size is out of range: 1365340 ~The payload in the test ends up exceeding 1MB so we need to reduce the size https://github.com/openconfig/public/blob/master/release/models/extensions/openconfig-metadata.yang#L36


youssefyamoutyy:
closing pr

PR #4542, opened 2025-09-04
Add JUNIPER case to container lifecycle test

ragpreet:
This PR adds support for the JUNIPER platform in containerz_test.go and ignores build artifacts (cntrsrv binary and vendor/ directory) using a .gitignore file.


ragpreet:
will start over creating a branch

PR #4558, opened 2025-09-09
Fix RT-5.6: interface_loopback_aggregate

nsadhasivam:
As per [Readme](https://github.com/openconfig/featureprofiles/blob/611c0146efbe3c1ed89dc4de9f37e330baf68aa1/feature/interface/interface_loopback_aggregate/otg_tests/interface_loopback_aggregate/README.md?plain=1#L13) we need to verify the remote end is down. which is done at [L194](https://github.com/openconfig/featureprofiles/blob/611c0146efbe3c1ed89dc4de9f37e330baf68aa1/feature/interface/interface_loopback_aggregate/otg_tests/interface_loopback_aggregate/interface_loopback_aggregate_test.go#L194)

Removing the link down verification on otg as per b/442816178


AmrNJ (approved):
LGTM

rohit-rp (approved):

PR #4561, opened 2025-09-09
Use FT for unsupported Cisco paths in optics_power_bias_current_test

islandmoves:
Use Cisco FT for laser thresholds.
    
Note the refactoring required due to the technical limitations of WithFT detailed in https://github.com/openconfig/ygnmi/blob/29c5534d22a769d82e502535665cb4461d13e651/ygnmi/ygnmi.go#L336-L376
    
We refactor so that each leaf is in a separate gnmi.Lookup() call.
    
Also note the performance regression caused by these technical limitations: each gnmi.Lookup() used for a specific component internally subscribes to *all* components and then filters the result to only the component requested.
    
These changes may result in some test slowdown with or without WithFT applied. The impact from this is unmeasured at this time.


singhavnish2516:
Closing this as https://github.com/openconfig/featureprofiles/pull/4628 has already added the changes to the corresponding test: feature/platform/tests/optics_power_and_bias_current_test/optics_power_and_bias_current_test.go

PR #4571, opened 2025-09-10
Creating device thruput test readme

ram-mac:
Creating the device thruput test README


feature/system/throughput/otg_tests/dev_thruput/README.md
dplore, line 1: Folder should probably be feature/interface/benchmark

feature/system/throughput/otg_tests/dev_thruput/README.md
```diff
@@ -0,0 +1,180 @@
...
+        "name": "eth0"
+      }
+    ]
+  }
+}
+```
+
+## TestCase-1:
+### PERF-1.1 Test IPv4 traffic 400G throughput
+  *   Test traffic using Fixed frame size of 64 bytes
```
dplore, line 100: The README could be structured similar to how the code would likely be structured:
-  Define a list of flows
-  Run a subtest for each flow
- Define the pass/fail criteria for the test.  
   -  (which is common across all flows, no need to repeat this in the README text)
   - Make sure each criteria is precise.  "System health" is ambiguous.  But you could refer to HEALTH-1.1.

dplore, line 100: Need to also define number of packets (which IMO is preferred over duration) and a PPS rate.

PR #4575, opened 2025-09-10
Remove unused deviation gribi_encap_header_unsupported

tengyiG:


dplore (approved):

dplore:
I went ahead and merged this without approve from gribi owners as this change is a no-op

PR #4581, opened 2025-09-11
Update deviations

AmrNJ:
1. Update deviations in p4rt_election_test.go and metadata.textproto files. 
2. Update deviations for tunnel_acl_based_test


MarcCharlebois:
> Orthogonal: `p4rt_election_test` only needs a DUT. Can we please change the testbed for it in the `metadata.textproto`?

I don't agree with this point as there's clearly a port assigned in this test. This will only pass if `TESTBED_DUT` would have a port defined in the testbed.

AmrNJ:
A new topology to suit this test has been created with #4641.

PR #4590, opened 2025-09-12
Refactor sFlow configuration plugins and adjust test flows for KNE.

dplore:
Refactor sFlow configuration plugins and adjust test flows for KNE.

Use parameter structs for sflow cfgplugin This updates sflow to comply with the [updated styleguide for featureprofiles cfgplugins](https://github.com/openconfig/featureprofiles/pull/4476).
This change introduces parameter structs for `NewSFlowGlobalCfg` and `NewSFlowCollector` in `cfgplugins/sflow.go` to improve readability. The `sflow_base_test.go` is updated to use these new structs

Add flow parameters specific to KNE device models.  KNE devices have reduced packet forwarding performance.  The use of KNE with this test is helpful for development purposes.  But as the README indicates, the minimum target device form factor is fixed form factor hardware.

Fixed error in sflow sample validation The previous code overcounted sflow packets. Note that sflow packets are expected to contain 1 or more samples.

Fixed error where a duration of time was being used instead of a Fixed number of packets in the OTG call. This lead to variability in the test due to timing differences between runs. Now with a fixed number of packets configured on the OTG, we can match exactly how many packets are sent and perfectly detect packet and sample loss.


singhavnish2516:
Hi @dplore,
Please fix the static analysis check.

PR #4597, opened 2025-09-16
Update system_ntp_test.go

AmrNJ:
Update the code to have both ntp source address and ntp server in the same address family.


Swetha-haridasula (approved):
lgtm

feature/system/ntp/tests/system_ntp_test/system_ntp_test.go
```diff
@@ -37,7 +38,9 @@ var (
 	dutlo0Attrs = attrs.Attributes{
 		Desc:    "Loopback ip",
 		IPv4:    "203.0.113.1",
+		IPv6:    "2001:db8::1:1:1:1",
 		IPv4Len: 32,
+		IPv6Len: 128,
```
dplore, line 43: This value is not used in the test?  If not needed, it should be removed

feature/system/ntp/tests/system_ntp_test/system_ntp_test.go
```diff
@@ -133,7 +136,16 @@ func TestNtpServerConfigurability(t *testing.T) {
 				ntp.SetEnabled(true)
 				for _, address := range testCase.addresses {
 					server := ntp.GetOrCreateServer(address)
-					server.SetSourceAddress(dutlo0Attrs.IPv4)
+					switch dut.Vendor() {
+					case ondatra.ARISTA:
```
dplore, line 140: If this is a limitation for only one vendor, and if we require IPv6 device to IPv4 NTP server to be supported, then this is a deviation.

If we don't require IPv6 device to IPv4 NTP server and will instead always use the same address family, then we could make this the standard and don't need the switch dut.Vendor() statement.

feature/system/ntp/tests/system_ntp_test/system_ntp_test.go
dplore, line 1: I noticed the test still does not pass on ceos after this change.  Is that expected?

dplore (approved):

PR #4599, opened 2025-09-17
Fix RT-14.2.2 test case to reflect B4 production use case

prakashbadri-arista:
In Google transit, the ENCAP_TE_VRF will have a 0/0 route pointing to DEFAULT VRF to redirect any traffic that does not match on a specific prefix into default VRF. The current test validation where 0/0 points to a egress interface is not supported in EOS for some platforms and this does not reflect the actual production use case. I have changed the test to have the default route point to default VRF and validate that the packet is routed based on the BGP route in the default VRF.

This default route in encap VRF use case has already been discussed with nachiketas from Google.


gggsmith (changes requested):
Thank you @prakashbadri-arista!

gggsmith (approved):

PR #4604, opened 2025-09-18
Acctz-3.1  record_subscribe_non_grpc changes

arvbaska1:

- Add a flag "static_binding" to use it to fetch the SSH IP:Port info when run with testbeds using static binding file,

- Add below deviations
acctz_records_authn_field_unsupported

- Add role field https://github.com/openconfig/gnsi/blob/main/acctz/acctz.proto#L170 in https://github.com/openconfig/featureprofiles/blob/main/internal/security/acctz/acctz.go#L884, Right now its not set

-


internal/security/acctz/acctz.go
```diff
@@ -187,33 +201,54 @@ func getGrpcTarget(t *testing.T, dut *ondatra.DUTDevice, service introspect.Serv
 	return resolvedTarget.String()
 }
 
-func getSSHTarget(t *testing.T, dut *ondatra.DUTDevice) string {
```
ram-mac, line 190: With PR4571 there is no need to have getSSHTarget.
So, we are using it again in this test?

PR #4621, opened 2025-09-19
RT-1.4 BGP Graceful restart deviation

shpaulch:
Added Deviation for BGP Graceful restart for Peer group


proto/metadata.proto
```diff
@@ -961,6 +961,9 @@ message Metadata {
     // Cisco: b/429233045
     string ciscoxr_laser_ft = 331;
 
+    //Device that do not support BGP Graceful restart for Peer Group
```
ram-mac: @shpaulch - Can you please add bug for this deviation with the details

ram-mac (changes requested):

proto/metadata.proto
karthikeya-remilla: Opened deviation bug https://partnerissuetracker.corp.google.com/issues/468284935 for tracking

Issue #4611, opened 2025-09-19
credentialz3 - gotHostCertificateCreatedOn should be validated for nano Seconds


nkrishk:
This time unit  (gotHostCertificateCreatedOn) was changes in this PR https://github.com/openconfig/featureprofiles/pull/4583

old code: (verified gotHostCertificateCreatedOn is in nano seconds)
https://github.com/ram-mac/featureprofiles/blob/6b6d9968cf1a6f1af8d67db9fb750cb303407485/feature/gnsi/credentialz/tests/host_certificates/host_certificates_test.go#L109

new after PR code: (takes gotHostCertificateCreatedOn as seconds)
https://github.com/ram-mac/featureprofiles/blob/7de1f683804cfef55fb80c8d879e3323f58fc4fa/feature/gnsi/credentialz/tests/host_certificates/host_certificates_test.go#L106

this PR assumes this time is in seconds, but it has to be nano seconds as per below issue https://github.com/openconfig/gnsi/issues/241

so we have to revert this time related changes to validate it for nano seconds

PR #4622, opened 2025-09-21
Add cisco deviation credz 3, credz 5

nkrishk:
### credz5 : rotate_authentication_artifacts_list_unsupported

In RotateAuthenticationArtifacts, the public key and certificate data are appended to artifactContents as follows:

"auth_artifacts": [
  {
    "private_key": "..."
  },
  {
    "certificate": "..."
  }
]

The current code can be found at https://github.com/ram-mac/featureprofiles/blob/7de1f683804cfef55fb80c8d879e3323f58fc4fa/internal/security/credz/credz.go#L280-L297

The private_key is in one object, while the certificate is in another. Consequently, auth_artifacts[0] contains only the key, and auth_artifacts[1] contains only the certificate. The server cannot determine which certificate corresponds to which private key.

For Cisco, public key data and certificate data must be in the same JSON object within auth_artifacts, structured as follows:

One object containing both private_key and certificate.

```
"auth_artifacts": [
  {
    "private_key": "...",
    "certificate": "..."
  }
]
```

Instead of two separate appends for PrivateKey and Certificate (current implementation), we need one append that includes both, as shown below.

```
 		artifactContents = append(artifactContents, &cpb.ServerKeysRequest_AuthenticationArtifacts{
 			Certificate: data,
 			PrivateKey:  data1,
 		})
```

### credz 3: host_certificate_created_on_in_seconds

As per this issue https://github.com/openconfig/featureprofiles/issues/4611,

host_certificate_created_on should be in nanoseconds. Cisco does not currently support this, so we are adding a deviation.


nkrishk:
new PR raised to fix CLA issue
https://github.com/openconfig/featureprofiles/pull/4623

PR #4623, opened 2025-09-21
Add cisco deviation credz 3, credz 5

nkrishk:
### credz5 : rotate_authentication_artifacts_list_unsupported

In RotateAuthenticationArtifacts, the public key and certificate data are appended to artifactContents as follows:

"auth_artifacts": [
  {
    "private_key": "..."
  },
  {
    "certificate": "..."
  }
]

The current code can be found at https://github.com/ram-mac/featureprofiles/blob/7de1f683804cfef55fb80c8d879e3323f58fc4fa/internal/security/credz/credz.go#L280-L297

The private_key is in one object, while the certificate is in another. Consequently, auth_artifacts[0] contains only the key, and auth_artifacts[1] contains only the certificate. The server cannot determine which certificate corresponds to which private key.

For Cisco, public key data and certificate data must be in the same JSON object within auth_artifacts, structured as follows:

One object containing both private_key and certificate.

```
"auth_artifacts": [
  {
    "private_key": "...",
    "certificate": "..."
  }
]
```

Instead of two separate appends for PrivateKey and Certificate (current implementation), we need one append that includes both, as shown below.

```
 		artifactContents = append(artifactContents, &cpb.ServerKeysRequest_AuthenticationArtifacts{
 			Certificate: data,
 			PrivateKey:  data1,
 		})
```

### credz 3: host_certificate_created_on_in_seconds

As per this ignsi ssue https://github.com/openconfig/gnsi/issues/241,

host_certificate_created_on should be in nanoseconds. Cisco does not currently support this, so we are adding a deviation.

#### deviation file changes
For hiba_authentication_test,  (both both Certificate and PrivateKey together should be combined in artifactContents)
the deviation is in file  [internal/security/credz/credz.go](https://github.com/openconfig/featureprofiles/pull/4623/files#diff-ba2ba26e1a0afac90f7cb349eac0fa02bf77f2429cc4e855217a85b698dfca25)
if deviations.RotateAuthenticationArtifactsListUnsupported(dut) {

For host_certificate_test, (cisco currently supports HostCertificateCreatedOn in seconds instead of nano seconds)
the deviation is directly in test file -> feature/gnsi/credentialz/tests/host_certificates/host_certificates_test.go
https://github.com/openconfig/featureprofiles/pull/4623/files#diff-60263049d84a142d127c18a53353a28f0d0e1eb8a99069d29223bd1c39248d5dR107

`  if deviations.HostCertificateCreatedOnInSeconds(dut) {`


Chounoki (approved):

ksgireesha:
Hi @rohit-rp , could you please restart the check-approvals workflow on PR #4623? The Cisco team is blocked by it and they don't have the privileges to re-trigger it themselves

morrowc:
your description of the problem and reasoning for the solution is a bit overly opaque for me...

What is the actual problem?
What is the thing you are attempting to do here to fix that problem?
Why is this a thing that is special to a single vendor in the world of 'all vendors are openconfig vendors'?

ksgireesha:
Hi @morrowc, just to clarify, my comment above was just an administrative request to get the CI checks restarted for the team!

@nkrishk / @karthikeya-remilla  - As the author of this PR, could you please jump in and address Chris's questions regarding the technical justification for these Cisco deviations?

nkrishk:
This PR can be closed because the core `RotateAuthenticationArtifacts` change was subsequently implemented in #4762.
PR #4762 changed the helper globally so the private key and certificate are placed in the same AuthenticationArtifacts entry, removing the need for the Cisco-specific rotate_authentication_artifacts_list_unsupported deviation proposed here.
The related Credentialz test updates have also evolved through later PRs, so this PR is now outdated and no longer suitable for merging. Thanks to everyone who reviewed and contributed feedback.

PR #4626, opened 2025-09-22
fixing record_subscribe_full, record_subscribe_partial tests

ram-mac:
Adding deviations for Cisco as the GRIBI records are not getting created
Make changes to the way the records are read for both records_subscribe_full and records_subscribe_partial


mihirpitale-googler (approved):
lgtm

PR #4630, opened 2025-09-22
Add match_as_path_set_unsupported deviation

yini101:
Add deviation to skip `match-as-path-set` leaf under /routing-policy/policy-definitions/policy-definition/statements/statement/conditions/bgp-conditions/match-as-path-set

Apply deviation to affected tests:
* RT-1.24
* RT-7.3
* RT-7.4
* RT-7.11


yini101:
> @yini101 Please can you resolve the conflicts?

Done

PR #4631, opened 2025-09-22
Update testregistry.textproto

vvlakshmanamurthy:
Updated BMP pre and post policy tests

### DIFF
--- a/testregistry.textproto
+++ b/testregistry.textproto
@@ -178,6 +178,18 @@ test: {
   exec: " "
 }
 test: {
+  id: "BMP-2.7"
+  description: "BMP pre-policy test"
+  readme: ""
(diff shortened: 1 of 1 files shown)


testregistry.textproto
```diff
@@ -177,6 +177,18 @@ test: {
...
+  description: "BMP pre-policy test"
+  readme: ""
+  exec: " "
+}
+test: {
+  id: "BMP-2.8"
+  description: "BMP post-policy test"
+  readme: ""
+  exec: " "
+}
```
dplore, line 191: Id's should match the READMEs;  (BMP-1.2, BMP-1.3)

Also fill in the `readme:` URL

PR #4634, opened 2025-09-23
adding Nokia deviations and ZRP files to MVC800 transceiver tests

rezachit:


feature/platform/transceiver/tests/zrp_800_platform_paths_test/zrp_800_platform_paths_test.go
```diff
@@ -0,0 +1,881 @@
...
+					defer trStreams[p.Name()].Close()
+					defer hwPortStreams[p.Name()].Close()
+					defer interfaceStreams[p.Name()].Close()
+				}
+
+				for _, p := range dut.Ports() {
+					gnmi.Await(t, dut, gnmi.OC().Interface(p.Name()).OperStatus().State(), timeout, oc.Interface_OperStatus_UP)
+					awaitRxPowerStats(t, p, params, oc.Interface_OperStatus_UP)
+				}
+				time.Sleep(extraWaitTime) // Wait extra time for telemetry to be updated.
```
dplore, line 119: see earlier comment on sleep

internal/cfgplugins/interface.go
```diff
@@ -373,22 +374,28 @@ func NewInterfaceConfigAll(t *testing.T, dut *ondatra.DUTDevice, batch *gnmi.Set
...
 		Name:    ygot.String(p.Name()),
 		Type:    oc.IETFInterfaces_InterfaceType_ethernetCsmacd,
 		Enabled: ygot.Bool(params.Enabled),
 	}
-	if p.PMD() == ondatra.PMD800GBASEZR || p.PMD() == ondatra.PMD800GBASEZRP {
+	switch {
+	case deviations.PortSpeedDuplexModeUnsupportedForInterfaceConfig(dut):
+		// No port speed and duplex mode config for devices that do not support it.
+	case p.PMD() == ondatra.PMD400GBASEZR || p.PMD() == ondatra.PMD400GBASEZRP:
+		// No port speed and duplex mode config for 400GZR/400GZR Plus as it is not supported.
```
dplore, line 397: If lack of support for this is acceptable for this feature/test, this doesn't have to be a deviation.  Deviations should be added however if this is considered a bug that must be fixed.  (and allowing lack of this function temporarily is ok)

feature/platform/transceiver/tests/zrp_800_terminal_device_paths_test/zrp_800_terminal_device_paths_test.go
```diff
@@ -0,0 +1,685 @@
...
+					defer otnStreams[p.Name()].Close()
+					defer interfaceStreams[p.Name()].Close()
+				}
+
+				// Wait for streaming telemetry to report the channels as up.
+				for _, p := range dut.Ports() {
+					gnmi.Await(t, dut, gnmi.OC().Interface(p.Name()).OperStatus().State(), timeout, oc.Interface_OperStatus_UP)
+					awaitQValueStats(t, dut, p, params, oc.Interface_OperStatus_UP)
+				}
+				time.Sleep(extraWaitTime) // Wait extra time for telemetry to be updated.
```
dplore, line 110: see earlier comment on sleep

feature/platform/transceiver/tests/zrp_800_terminal_device_paths_test/zrp_800_terminal_device_paths_test.go
```diff
@@ -0,0 +1,685 @@
...
+		if !ok {
+			t.Fatalf("QValue stats are not as expected for %v after %v minutes.", p.Name(), timeout.Minutes())
+		}
+	default:
+		t.Fatalf("Unsupported oper status for %v: %v", p.Name(), operStatus)
+	}
+}
+
+// validateNextSample validates the stream data.
+func validateNextSample(t *testing.T, dut *ondatra.DUTDevice, p *ondatra.Port, params *cfgplugins.ConfigParameters, wantOperStatus oc.E_Interface_OperStatus, interfaceData *ygnmi.Value[*oc.Interface], otnChannelData, ethChannelData *ygnmi.Value[*oc.TerminalDevice_Channel], operationalMode uint16) {
```
dplore, line 193: since these `validate*` functions are cut and paste for each test, perhaps they should be moved into a common library.  Perhaps a new  `/internal/interfaces/interfaces.go` ?

feature/platform/transceiver/tests/zr_800_platform_paths_test/zr_800_platform_paths_test.go
```diff
@@ -114,41 +116,41 @@ func TestComponentPaths(t *testing.T) {
 					gnmi.Await(t, dut, gnmi.OC().Interface(p.Name()).OperStatus().State(), timeout, oc.Interface_OperStatus_UP)
 					awaitRxPowerStats(t, p, params, oc.Interface_OperStatus_UP)
 				}
-				time.Sleep(3 * samplingInterval) // Wait extra time for telemetry to be updated.
+				time.Sleep(extraWaitTime) // Wait extra time for telemetry to be updated.
```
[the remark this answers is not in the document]
rezachit: For all codes always the main and long pole PMs  for all ports are being watched through awaitX functions... this extra time is due to the fact the stream sample updated with a separate subscription and PM being available in the different subscription does not mean it will be available instantly for the one being used for validation, so at the time we decided to add this extra time to prevent this kind of scenarios.

internal/cfgplugins/interface.go
rezachit: this is related to the Nokia special behavior which we have to live with it. Arista does not have this deviation though that's why we have to keep it.

feature/platform/transceiver/tests/zrp_800_terminal_device_paths_test/zrp_800_terminal_device_paths_test.go
rezachit: I think we do not have a common place for a validation or anything related to the telemetry do you know where I can  move it to? and the function is not the same in all files it only will be similar to two files and the final goal if you remember the discussion we had 1 month ago will be consolidate into one files with multiple testbeds...

dplore: ah ok, if they need to be different for the different tests, then that is ok.  If only two tests have same validation then my opinion is it is not mandatory to use a common function.  3 or more, then yes, we should.

I will resolve this and if you like, you can move the common validation to a new file, `internal/interfaces/interfaces.go`. Otherwise you can leave it as is.

feature/platform/transceiver/tests/zr_800_platform_paths_test/zr_800_platform_paths_test.go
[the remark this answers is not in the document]
dplore: Ok, we discussed and it seems the issue is a little more complex.  The data is being accessed via a samplestream and there isn't some function to "Await" for samples currently.  

Perhaps a samplestream.NextAwait function could be introduced in a future PR that would alleviate the need to do a sleep like this.  Please add a TODO comment here so we can remember this next time we wonder why this sleep is here.  THanks!

dplore (approved):

PR #4664, opened 2025-10-01
Fixed issue with credz.SSHWithPassword

swatisp98:
============================================================
Why credz.SSHWithPassword can fail (DialSSH unimplemented):
============================================================
credz.SSHWithPassword is a helper in our test infra that wraps SSH access through the Ondatra/binding layer. It does not necessarily call Go's native ssh.Dial ? instead it asks the ondatra binding to provide an SSH connection.

============
Root cause:
============
Direct ssh.Dial: Always attempts a TCP connection to the given host:port. credz.SSHWithPassword: An abstraction that requires the DUT binding to implement SSH access ? it can be unimplemented for some KNE devices, causing immediate failure.

====================
Recommended action:
====================
Implement a simple fallback: Call credz.SSHWithPassword first (preferred), and if it returns an error (especially DialSSH unimplemented), fall back to ssh.Dial(sshAddr, ...)


swatipai30101998:
@ram-mac 

Since this is being tested on hardware and not a KNE device
Can you please share how does your binding and testbed files look like:

swatipai30101998:
@ram-mac 


Script is specific for KNE device qualification. It's not for hardware, Can you please qualify on KNE and let us know the results

path: feature/security/aaa/**kne_tests**/tls_authentication_over_grpc_test/tls_authentication_over_grpc_test.go

PR #4665, opened 2025-10-01
Acctz_3.1 test changes

arvbaska1:

- Add a flag "static_binding" to use it to fetch the SSH IP:Port info when run with testbeds using static binding file,

- Add below deviations

- Add role field https://github.com/openconfig/gnsi/blob/main/acctz/acctz.proto#L170 in https://github.com/openconfig/featureprofiles/blob/main/internal/security/acctz/acctz.go#L884, Right now its not set.

- Authn field is only populated for cisco for Login/Once/Enable records, not for type Operation, so set to nil for records in this test which are of type SESSION_STATUS_OPERATION.
https://github.com/openconfig/gnsi/blob/main/acctz/acctz.proto spec does'nt mandate it needs to set for session status operations.

- Added a different username/password specific to which fails authorization only, and modify existing failusername for fail authentication purpose.


ram-mac (changes requested):

internal/security/acctz/acctz.go
```diff
@@ -199,33 +220,56 @@ func SetupUsers(t *testing.T, dut *ondatra.DUTDevice, configureFailCliRole bool)
...
-	if err != nil {
-		t.Log("DUT does not support `Service` function, will attempt to resolve dut name field.")
+func getSSHTarget(t *testing.T, dut *ondatra.DUTDevice, staticBinding bool) string {
 
-		// Suppose ssh could be not 22 in some cases but don't think this is exposed by introspect.
-		dialTarget := fmt.Sprintf("%s:%d", dut.Name(), defaultSSHPort)
-		resolvedTarget, err := net.ResolveTCPAddr("tcp", dialTarget)
-		if err != nil {
-			t.Fatalf("Failed resolving ssh target %s", dialTarget)
+	if staticBinding {
```
ram-mac, line 225: @arvbaska1 - Is this fix specific to a testbed where static binding is happening? In case we skip this static binding will we be able to run the test still?

karthikeya-remilla:
Obsolete PR. Closing this.

PR #4674, opened 2025-10-03
RT-7.1 Panic issue fix

shpaulch:
In cases where either IPv4 address is nil or IPv6 address is nil, panic happens.
Nil check works only if both ipv4 and ipv6 address are nil.
Changed nil check to config interface if wither IPv4 or IPv6 address is nil


feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
```diff
@@ -147,7 +147,7 @@ func configureDUT(t *testing.T, dut *ondatra.DUTDevice) {
 	lo0 := gnmi.OC().Interface(loopbackIntfName).Subinterface(0)
 	ipv4Addrs := gnmi.LookupAll(t, dut, lo0.Ipv4().AddressAny().State())
 	ipv6Addrs := gnmi.LookupAll(t, dut, lo0.Ipv6().AddressAny().State())
-	if len(ipv4Addrs) == 0 && len(ipv6Addrs) == 0 {
+	if len(ipv4Addrs) == 0 || len(ipv6Addrs) == 0 {
```
gggsmith: Two quick thoughts here:

- Changing `&&` to `||` might inadvertently recreate the interface if it only has one IP family configured. Should we keep `&&` and fix the potential index out-of-bounds issue by updating the slice access logic in the `else` block instead? Let me know if I missed something!
-  This snippet is duplicated across several test files. Could we extract it into a common helper in `internal/helpers/helpers.go` (e.g., `GetOrCreateLoopback`)?

gggsmith (changes requested):

feature/bgp/policybase/otg_tests/default_policies_test/bgp_default_policies_test.go
shpaulch: Created a new fn in helpers.go to create loopback address.
Updated logic not to use && or ||

gggsmith (approved):

ram-mac (approved):

ram-mac (approved):

PR #4691, opened 2025-10-09
Add BGP scale test (RT-1.65) using OTG

Abdelfatah26:
Add BGP scale test (RT-1.65) using OTG.

This PR introduces a new BGP scale test suite (`bgp_scale_test.go`) for Feature Profiles, covering various scenarios:
-   **Steady State:** Tests 
-   **Session and Route Scale:** 
-   **BGP Events:** Tests include altering BGP attributes, withdrawing routes, and resetting BGP peers, with convergence measurements.
-   **MSS Configuration:** Includes a test case with MSS configured.

The test uses OTG for ATE configuration and traffic generation. It includes detailed configurations for interfaces, ISIS, and BGP on both DUT and ATE, along with telemetry verification for ISIS and BGP, and traffic validation.

New deviations `BgpGrHelperDisableUnsupported` and `BgpGracefulRestartUnderAfiSafiUnsupported` are added to `deviations.go` and used in the `metadata.textproto` for vendor-specific handling.


proto/metadata.proto
```diff
@@ -974,6 +974,12 @@ message Metadata {
     // Cisco: b/447513282
     bool syslog_oc_unsupported = 334;
 
+    // Device does not support to disable bgp GR-HELPER.
```
ram-mac: @Abdelfatah26 - You need to raise a bug with the vendor for this deviation and that bug id in the comment section. Also is this required across all vendors ?

proto/metadata.proto
```diff
@@ -974,6 +974,12 @@ message Metadata {
...
+    bool bgp_gr_helper_disable_unsupported = 335;
+
+    // Device does not support bgp GR-RESTART under AFI/SAFI.
```
ram-mac: Same as above.

proto/metadata.proto
Abdelfatah26: updated. Thank you

proto/metadata.proto
Abdelfatah26: updated. Thank you

feature/bgp/otg_tests/bgp_scale_test/bgp_scale_test.go
```diff
@@ -0,0 +1,2211 @@
...
+		}
+		if tc.verifyTraffic != nil {
+			t.Run(tc.name+" Verify Traffic", func(t *testing.T) {
+				t.Logf("Verify Traffic")
+				for _, v := range tc.verifyTraffic {
+					v(t, ate, top)
+				}
+			})
+		}
+		if tc.testCleanup != nil {
```
ram-mac: @Abdelfatah26 - I think this can be moved into a new function which can be called to cleanup based on the requirement.

feature/bgp/otg_tests/bgp_scale_test/bgp_scale_test.go
```diff
@@ -0,0 +1,2211 @@
...
+// intAbs returns the absolute value of an integer.
+func intAbs(n int) int {
+	if n < 0 {
+		return -n
+	}
+	return n
+}
+
+// atePort3IPv4 returns ip addresses starting 50.2.%d.2 for every vlanID.
+func atePort3IPv4(vlan int) string {
```
ram-mac: @Abdelfatah26 - I think there is a function in the bgp.go util to return the IP using net utility.
Can we leverage that function instead of statically assigning the IP's in the 3rd octet. Applicable for both IPv4, IPv6.
Also, VLAN assignment should be dynamically picked up based on a start number and being assigned as you progress in the test?

feature/bgp/otg_tests/bgp_scale_test/bgp_scale_test.go
Abdelfatah26: Thanks Ram for your comment.
As discussed, this a testCase attrib, the used functions is already defined as part of bgp helpers

ram-mac (approved):

feature/bgp/otg_tests/bgp_scale_test/bgp_scale_test.go
```diff
@@ -0,0 +1,2261 @@
...
+
+func vlanID(i int) int {
+	return int(dutPort4.index)*10 + i
+}
+
+func TestMain(m *testing.M) {
+	fptest.RunTests(m)
+}
+
+func TestBook(t *testing.T) {
```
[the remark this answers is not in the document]
Abdelfatah26: removed. thank you

PR #4708, opened 2025-10-16
Transceiver 13

fmolinar:


hmod2001:
Problem
The zr_low_power_mode_test was failing on Cisco devices (8200, 8808) because the test unconditionally tried to validate the manufacturing date (mfg-date) field from transceiver components, but Cisco devices don't support this OpenConfig leaf.

Added deviation configuration and Fixed test code to conditionally create mfgDate stream only for devices that support it:

Removed unconditional creation of streamMfgDate
Moved mfgDate stream creation inside deviation check: if !deviations.GetTransceiverMfgDateUnsupported(dut)
Added similar conditional handling for description field

Verified test now passes on Cisco 8200 and 8808 devices

[log-t13.txt](https://github.com/user-attachments/files/22957702/log-t13.txt)

AmrNJ (changes requested):
Add streamDescription back to the test.

AmrNJ (approved):
LGTM

fmolinar:
this is running off 5207

PR #4703, opened 2025-10-15
Implement AFT Atomic Test

ElodinLaarz:
This pull request implements the AFT Atomic test, as well as updating the `README.md` documentation for readability, improving the test objectives, procedures, and expected outcomes. The changes improve the explanation of how the `atomic` flag should behave in gNMI notifications during network churn events, and restructure the document for readability and completeness.

**Test objectives and expected behavior:**

* Expanded the summary to clearly state that the test verifies the correct setting of the `atomic` flag in gNMI notifications for AFT paths, specifying when it should be `true` (initial sync and link recovery) and `false` (deletes during link down events).

**Test procedure and verification steps:**

* Detailed the configuration variables, their defaults, and the steps for IS-IS and eBGP setup, including expected notification behaviors for initial sync and link events.
* Clarified verification steps for scenarios involving link down/up events, specifying expected notification types, presence of the `atomic` flag, and route states in the DUT’s RIB and FIB.

**Documentation structure and completeness:**

* Reorganized sections for improved readability, using clearer headings and bullet points.
* Moved and expanded the OpenConfig path and RPC coverage section to the end of the document, listing all relevant state paths and RPCs used in the test.

PR #4750, opened 2025-10-25
Implement BGP neighbor configuration.

ElodinLaarz:
Verified in https://github.com/openconfig/featureprofiles/pull/4703#issuecomment-3446861126.

===

This pull request introduces several new helper functions and refactors BGP neighbor configuration logic in `internal/cfgplugins/bgp.go`. The main changes focus on handling vendor-specific deviations, improving multipath configuration, and simplifying neighbor creation for BGP testing. These updates enhance code modularity and make deviation handling more explicit.

**Deviation handling and configuration improvements:**

* Added `HandleMaxPrefixesDeviation`, which updates neighbor max prefixes only if the `BGPMissingOCMaxPrefixesConfiguration` deviation is set, with logic specific to Arista devices.
* Implemented `handleMultipathDeviation` to manage multipath configuration at the AFI/SAFI level, including vendor-specific deviation logic for Juniper devices.

**Neighbor creation and validation enhancements:**

* Introduced `CreateBGPNeighbors`, a helper to create BGP neighbors, peer groups, and global configuration, enforcing that all neighbors have the same local and peer AS numbers.
* Added `sameAS` utility to verify that all provided neighbors share the same AS numbers, used by `CreateBGPNeighbors`.

**Refactoring and consistency:**

* Refactored `ConfigureBGPNeighbor` to use the new parameter name `routerID` for consistency, replacing `routerId`. (F8362


ttsuhasini (approved):

internal/cfgplugins/bgp.go
```diff
@@ -638,15 +638,145 @@ func DeviationAristaBGPNeighborMaxPrefixes(t *testing.T, dut *ondatra.DUTDevice,
 	}
 }
 
-func ConfigureBGPNeighbor(t *testing.T, dut *ondatra.DUTDevice, ni *oc.NetworkInstance, routerId, peerAddress string, routerAS, peerAS uint32, ipType string, sendReceivePaths bool) {
+// HandleMaxPrefixesDeviation updates neighbor max prefixes only if deviation BGPMissingOCMaxPrefixesConfiguration
+// is set.
+func HandleMaxPrefixesDeviation(t *testing.T, dut *ondatra.DUTDevice, neighbors []*BgpNeighbor) {
```
dplore: shouldn't this be a private function and embedded into `CreateBGPNeighbors` like `handleMultipathDeviation`

ElodinLaarz: I was avoiding that beacause `DeviationAristaBGPNeighborMaxPrefixes` does a Set request directly (and is used outside of this library). Will fix `DeviationAristaBGPNeighborMaxPrefixes` in a refactoring PR for the external caller, but for now we can just make it private.

internal/cfgplugins/bgp.go
```diff
@@ -638,15 +638,145 @@ func DeviationAristaBGPNeighborMaxPrefixes(t *testing.T, dut *ondatra.DUTDevice,
...
+				SetPeerGroup(peerGrpNameV6)
+			neighbor.GetOrCreateAfiSafi(oc.BgpTypes_AFI_SAFI_TYPE_IPV6_UNICAST).
+				SetEnabled(true)
+		}
+	}
+	return protocol, nil
+}
+
+// ConfigureBGPNeighbor configures a BGP neighbor.
+func ConfigureBGPNeighbor(t *testing.T, dut *ondatra.DUTDevice, ni *oc.NetworkInstance, routerID, peerAddress string, routerAS, peerAS uint32, ipType string, sendReceivePaths bool) {
```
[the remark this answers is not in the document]
dplore: Oh, hrm, GitHub diff shows it is new, but I do see it.  Nevermind then, you don't need to fix it.  :)

internal/cfgplugins/bgp.go
```diff
@@ -638,15 +638,172 @@ func DeviationAristaBGPNeighborMaxPrefixes(t *testing.T, dut *ondatra.DUTDevice,
...
+// CreateBGPNeighbors creates BGP neighbors for the given router ID, peer group names, and
+// neighbors. The global AS and router ID are set to the AS and router ID of the first neighbor,
+// assuming that all neighbors provided have the same local AS and the same peer AS.
+// The sb is updated with the BGP neighbors configuration.
+func CreateBGPNeighbors(t *testing.T, dut *ondatra.DUTDevice, sb *gnmi.SetBatch, cfg BGPNeighborsConfig) error {
+	if len(cfg.Nbrs) == 0 {
+		t.Logf("No BGP neighbors found for router ID: %s, peer group name v4: %s, peer group name v6: %s", cfg.RouterID, cfg.PeerGrpNameV4, cfg.PeerGrpNameV6)
+		return nil
+	}
+	if !sameAS(cfg.Nbrs) {
```
ram-mac: This check is not required all the times; Make this check optional.
There are usecases where one can define a local as=200 and remote as=300 for a ISP where the actual ISP as=100
So in such cases, this check becomes a showstopper for testing.

internal/cfgplugins/bgp.go
```diff
@@ -638,15 +638,172 @@ func DeviationAristaBGPNeighborMaxPrefixes(t *testing.T, dut *ondatra.DUTDevice,
...
+// is set. TODO: Add config created in DeviationAristaBGPNeighborMaxPrefixes to the SetBatch.
+func handleMaxPrefixesDeviation(t *testing.T, dut *ondatra.DUTDevice, _ *gnmi.SetBatch, cfg BGPNeighborsConfig) error {
+	t.Helper()
+	if !deviations.BGPMissingOCMaxPrefixesConfiguration(dut) {
+		return nil
+	}
+	switch dut.Vendor() {
+	case ondatra.ARISTA:
+		for _, nbr := range cfg.Nbrs {
+			DeviationAristaBGPNeighborMaxPrefixes(t, dut, nbr.Neighborip, 0)
```
ram-mac: @ElodinLaarz - AFAIK, the maximum prefix per neighbor is applicable for Cisco too. There is a base minimum scale for prefixes that are allowed (AFAIR it used to be 250k routes IPv4/IPv6). So it is worth adding this for Cisco vendor too.
Currently 'DeviationAristaBGPNeighborMaxPrefixes' is modeled specifically for Arista and if we add case for Cisco too then this utility needs to be modified and make general for all vendors.

internal/cfgplugins/bgp.go
ElodinLaarz: In this case, we're configuring all the neighbors under a single Peer Group (and so the ASN of the peer group can only take on one value).

internal/cfgplugins/bgp.go
paramasivamn1: AFT Base test is passing for Cisco-8808 with 24.3.30

PR #4751, opened 2025-10-25
Add single topology deviation handling for ISIS.

ElodinLaarz:
Verified in https://github.com/openconfig/featureprofiles/pull/4703#issuecomment-3446861126.

===

This pull request adds a new helper function to handle a specific ISIS protocol deviation for single topology requirements. The main change is the introduction of logic to ensure IPv6 multi-topology settings match those of IPv4 for Cisco devices.

ISIS protocol deviation handling:

* Added the `HandleSingleTopologyDeviation` function in `internal/cfgplugins/isis.go` to set the IPv6 multi-topology AFI/SAFI to IPv4 values for Cisco devices when the single topology deviation is required. This helps maintain consistent ISIS configuration across vendors and test environments.

### DIFF
--- a/.gemini/config.yaml
+++ b/.gemini/config.yaml
@@ -1,0 +1,14 @@
+# See reference at https://developers.google.com/gemini-code-assist/docs/customize-gemini-behavior-github#config.yaml-schema
+have_fun: false
+memory_config:
+  disabled: false
+code_review:
+  disable: false
+  comment_severity_threshold: MEDIUM
+  max_review_comments: -1
+  pull_request_opened:
+    help: false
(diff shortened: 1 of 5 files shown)


singhavnish2516 (changes requested):

dplore:
Hi, I see [5 tests which implement the relevant deviation](https://github.com/search?q=repo%3Aopenconfig%2Ffeatureprofiles+ISISSingleTopologyRequired&type=code), but none are using the NewISIS cfgplugin.  Can we show at least one use of this function which passes on Cisco DUT?

PR #4755, opened 2025-10-27
gNMI-1.18_fix

manan-patel:


dbarney1106 (approved):
Cisco 8000e is passing without deviation, LGTM

PR #4760, opened 2025-10-28
Use single stream for entire test

ttsuhasini:
Improved test efficiency by creating and reusing a single AFTStreamSession across test steps.


singhavnish2516 (approved):
nit: Comments should be complete sentences.

PR #4765, opened 2025-10-30
Use index 1 for creating subinterface instead of index 0.

singhavnish2516:


feature/interface/otg_tests/telemetry_interface_last_change_test/README.md
```diff
@@ -22,7 +22,7 @@ changes.
 1.  **Configure Ethernet Interface**:
     *   Select a port on the DUT.
     *   Configure it as an Ethernet interface.
-    *   Create a subinterface with index 0.
+    *   Create a subinterface with index 1.
```
ram-mac: @singhavnish2516 - why do we change the index to 1; generally index starts from 0 for vendors too?

dplore: Vendors should support subinterface `0`.  If this is not working on a given NOS it is a bug which could be mitigated by a deviation.

Background:
OC requires a subinterface for IP address configuration.  Some NOS do not define a subinterface for the IP address configuration when it applies to the physical port (ie: Arista EOS, Cisco XR).  Other NOS (ie: JunOS, SRLinux) always use subinterfaces for the IP address configuration, just like OC.  

The OC subinterface index of `0` is required to map to the physical interface for those vendors (Cisco, Arista) that don't use a subinterface.

FYI, this reference doc covers many other more complex interface configuration references, although ironically does not have details on the simple, single subinterface case: https://github.com/openconfig/public/blob/master/doc/network_instance_examples.md

singhavnish2516: For Arista and Cisco, as the subinterface index 0 is required to map to the physical interface, the OC Path `/interfaces/interface/subinterfaces/subinterface/state/
last-change` does not change. Hence, the change. We want to validate `/interfaces/interface/subinterfaces/subinterface/state/
last-change` changes with enable/disable, hence, any non-zero index seemed to be a better choice.

singhavnish2516: Yes, this doesn't seem to be working for Juniper. Please suggest if you think we should need a deviation for this?

singhavnish2516:
Closing this PR for now. Will re-open if required.

PR #4781, opened 2025-11-05
DP-1.12: Adding Queue management profile binding to the output interface

charantejag504:
DP-1.12: Adding Queue management profile binding to the output interface


sezhang2 (approved):
LGTM. It is a very small change.

dplore (approved):

PR #4782, opened 2025-11-06
fixing fabric redundancy test

ram-mac:
Fixing the fabric redudancy test with below changes
1. adding deviation interface_enable=true for Arista
2. adding some change sfor the configure DUT to configure the interface type and enabling the interface particularly for the fabric_redundancy subtest


AmrNJ (approved):
LGTM

PR #4786, opened 2025-11-07
gNMI-1.6 fix

manan-patel:


feature/gnmi/tests/gnmi_ni_test/metadata.textproto
```diff
@@ -19,7 +19,7 @@ platform_exceptions: {
     vendor: CISCO
   }
   deviations: {
-    interface_enabled: true
+    skip_grpc_validation_in_default_vrf: true
```
dbarney1106: Skip grpc validation in default? Seems more like we are tring to skip the configuration.

CreateGNMIServer uses batchUpdate which in OC, this should not change between old and new grpc sever syntax.

dbarney1106 (commented):
I will close this pull request and submit a new one with modifications based on bootstrap config test

PR #4794, opened 2025-11-11
Fix dscp remark enabled by default

ijernoiu:
Fixing NewQoSClassifierConfiguration function in qos cfgutils.
Previously, the remark action was set even if the tests using this function did not require any dscp remarking.

### DIFF
--- a/feature/policy_forwarding/vrf_selection/otg_tests/vrf_selection/metadata.textproto
+++ b/feature/policy_forwarding/vrf_selection/otg_tests/vrf_selection/metadata.textproto
@@ -4,7 +4,7 @@
 uuid:  "db187faa-0f6d-44f8-86d2-832df9c285fc"
 plan_id:  "PF-1.6"
 description:  "Policy based VRF selection for IPV4/IPV6"
-testbed:  TESTBED_DUT_ATE_2LINKS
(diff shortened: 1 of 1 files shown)


internal/cfgplugins/qos.go
```diff
@@ -147,15 +147,16 @@ func NewQoSClassifierConfiguration(t *testing.T, dut *ondatra.DUTDevice, q *oc.Q
...
 				condition.GetOrCreateIpv4().SetDscpSet(class.DscpSet)
-			} else {
+			case oc.Qos_Classifier_Type_IPV6:
 				condition.GetOrCreateIpv6().SetDscpSet(class.DscpSet)
 			}
 		}
 
 		// DSCP remark configuration is not supported. Adding external static configuration after QoS OC configuration
-		if !deviations.QosRemarkOCUnsupported(dut) {
+		if class.RemarkDscp != 0 && !deviations.QosRemarkOCUnsupported(dut) {
```
ram-mac: @ijernoiu - How do we handle the case when the condition is not met, lets say the RemarkDscp != 0 && no deviation is present, what is the value you would set.

ram-mac (approved):
Comments Addressed. Looks good now

PR #4795, opened 2025-11-11
Refactor MED validation logic in flow control tests

karthikeya-remilla:
for cisco, since bgp_set_med_action_unsupported = true, the MED value verification for prefixes should be handled accordingly. 

This applies to the subtest "Configure eBGP increase MED Import Export Policy". 

verified that test passes locally.


ram-mac:
@karthikeya-remilla  - Have validated this internally for Cisco and it is passing. Now need to check if it breaks other vendor or not. Will keep you posted.

PR #4797, opened 2025-11-11
Add FeatureProfileStatus to NOSImageProfile

dplore:
Added supported_tests field to NOSImageProfile and created FeatureProfilesTests message.

The goal is each software release of a NOS can provide a textproto of this data showing which featureprofiles tests have run and their expected results.

### DIFF
--- a/feature/container/containerz/tests/container_lifecycle/containerz_test.go
+++ b/feature/container/containerz/tests/container_lifecycle/containerz_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/openconfig/featureprofiles/internal/containerztest"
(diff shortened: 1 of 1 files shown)


proto/nosimage.proto
```diff
@@ -54,4 +57,26 @@ message NOSImageProfile {
...
+
+  // supported_tests is a list of featureprofiles metadata.plan_ids which the
+  // NOS software supports (verified as passing)
+  repeated FeatureProfileStatus featureprofile_status = 8;
+
+}
+
+message FeatureProfileStatus {
+  // The unique identifier for a featureprofiles test plan.  This string must match
+  // the Metadata plain_id defined for the referenced test.
```
kapilrajput-g: Minor typo fix: "plain_id" should be "plan_id" to match the field name and the Metadata definition

PR #4801, opened 2025-11-12
fixing timing issue for RT-1.64: BGP Import/Export Policy (Control plane only) Functional Test Case

desaimg1:
Fixing timing issue in the existing merged script. Please find the latest log at https://partnerissuetracker.corp.google.com/issues/415458482


ram-mac (approved):
The changes looks good and the test is passing with the changes. Will have to monitor the test for any flakiness.

PR #4812, opened 2025-11-14
[ MTU-1.5 ]Fix PMTU handling test to handle rate-limited packets

prakashbadri-arista:
The test was failing because rate-limited packets are not accounted
  for in drop counters, causing mismatches between expected and actual
  drop counts.

  Changes:
  - Add initializeBaselineDropCounters() to capture baseline drop counts before running tests, allowing delta-based validation across multiple test flows
  - Relax drop counter validation to check for any increase (> 0) instead of exact packet count matching with tolerance
  - Add detailed per-component logging in verifyPacketProcessingAggregateDrops and verifyFragmentTotalDrops for better debugging
  - Rename currentPacketProcessingAggregateDrops to newPacketProcessingAggregateDrops and currentFragmentTotalDropsCount to newFragmentTotalDropsCount for clarity

### DIFF
--- a/feature/mtu/otg_tests/pmtu_handing/pmtu_handing_test.go
+++ b/feature/mtu/otg_tests/pmtu_handing/pmtu_handing_test.go
@@ -407,24 +407,62 @@ func createFlowAndVerifyTraffic(t *testing.T, td testData, tt testDefinition, wa
 	return outPkts
 }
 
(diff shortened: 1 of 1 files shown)


ram-mac (approved):
Approving.. The issue seen needs to be followed up with Ixia

PR #4818, opened 2025-11-17
Add deviation for explicit BGP RIB streaming config

akalluru1:
RT-1.30 verifies the BGP RIB state streamed through following YANG path: /network-instances/network-instance/protocols/protocol/bgp/rib

Arista devices need explicit BGP RIB streaming config to send data for the above YANG path.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -22,13 +22,7 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version-file: 'go.mod'
-    - name: Cache
-      uses: actions/cache@v4
-      with:
-        path: |
-          ~/go/pkg/mod
-          ~/.cache/go-build
-        key: ${{ github.job }}-${{ runner.os }}-go-build-${{ hashFiles('**/go.sum') }}
+        cache: true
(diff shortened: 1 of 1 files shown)


proto/metadata.proto
```diff
@@ -1044,7 +1044,10 @@ message Metadata {
     // Device does not support adding remote-syslog config under non-default VRF
     // Cisco: https://partnerissuetracker.corp.google.com/u/0/issues/459659437
     bool syslog_non_default_vrf_unsupported = 351;
-    
+
+    // Device that requires explicit config to support BGP RIB streaming
```
ram-mac: @akalluru1 - Can you please create a bug for the deviation for us to track it

ram-mac:
@akalluru1 - Also are we sure this deviation is required only for just the nested_policy test? How about the chained_policy_test?

akalluru1:
By mistake discarded the changes in https://github.com/akalluru1/featureprofiles/tree/main which resulted in closing this PR. I've addressed above comments from @ram-mac & created a new PR - https://github.com/openconfig/featureprofiles/pull/4955

PR #4842, opened 2025-11-22
Changes to GNMI-1.6 to pass with bootstrap config

danielbarney:
PR passing for Cisco using following as GRPC bootstrap config

grpc
 vrf default
 dscp cs4
 port 57400
 gnmi
 !
 p4rt
 !
 gnpsi
 !
 gribi
 !
 server gnxi-default
  vrf default
  port 9339
  services GNMI
  services GNOI
  services GNSI
 !
 server gnxi-legacy-default
  vrf default
  port 57400
  services GNMI
  services GNOI
  services GNSI
 !
 max-streams 128
 max-streams-per-user 128
 address-family dual
 max-request-total 256
 max-request-per-user 32
!

### DIFF
--- a/feature/gnmi/tests/gnmi_ni_test/gnmi_ni_test.go
+++ b/feature/gnmi/tests/gnmi_ni_test/gnmi_ni_test.go
@@ -100,6 +100,7 @@ func ConfigureAdditionalNetworkInstance(batch *gnmi.SetBatch, t *testing.T, dut 
 }
 
 func ValidateNetworkInstance(t *testing.T, dut *ondatra.DUTDevice) {
+
 	t.Log("\nValidating network instance after configuration\n")
 	// Get all network instances.
 	netInstanceList := gnmi.GetAll(t, dut, gnmi.OC().NetworkInstanceAny().State())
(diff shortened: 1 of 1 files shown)


feature/gnmi/tests/gnmi_ni_test/gnmi_ni_test.go
```diff
@@ -55,58 +57,90 @@ func TestCodelab(t *testing.T) {
...
 	for _, gnmiServer := range gnmiServerList {
-		if gnmiServer.GetName() == deviations.DefaultNetworkInstance(dut) {
-			defaultInstanceState := gnmi.Get(t, dut, gnmi.OC().System().GrpcServer(deviations.DefaultNetworkInstance(dut)).State())
+		t.Logf("----------------------------------------------------")
+		t.Logf("gNMI server name: %v", gnmiServer.GetName())
+		t.Logf("----------------------------------------------------")
+		t.Logf("gNMI server network-instance: %v", gnmiServer.GetNetworkInstance())
+		t.Logf("gNMI server port: %v", gnmiServer.GetPort())
+		t.Logf("gNMI server enable: %v", gnmiServer.GetEnable())
+		t.Logf("gnmi server services: %v", gnmiServer.GetServices())
```
dplore: nit: Instead of logging these things, the code could simply verify the expected values and throw a t.Error or t.Fatal is not found.    (Because we don't expect people to be reading logs to determine test pass/fail or get FYI information).   We'd rather see the results of a call to `t.Run()` with some useful test name.  That can be parsed by go test framework/tools.

dplore (commented):
Regarding the comment about the bootstrap configuration.  All featureprofiles tests expect that basic g* services are enabled before running the tests.  Adding additional g* services however should be part of the config done in the test.

The bootstrap you mentioned seems to do both?  But I think the  bootstrap should only configure some "bootstrap" gnmi service.

dplore:
> Regarding the comment about the bootstrap configuration. All featureprofiles tests expect that basic g* services are enabled before running the tests. Adding additional g* services however should be part of the config done in the test.
> 
> The bootstrap you mentioned seems to do both? But I think the bootstrap should only configure some "bootstrap" gnmi service.

Oh, now I see, the example bootstrap configuration only enables gnmi on the default network-instance (vrf).  That is fine.

feature/gnmi/tests/gnmi_ni_test/gnmi_ni_test.go
dbarney1106: Removed.

dplore (approved):

PR #4860, opened 2025-11-26
RT-4.11 aft scale summary fix

youssefyamoutyy:
1. Brief description and need for this PR

`TestBGP` in the AFT scale summary test suite was consistently timing out with "context deadline exceeded" errors when verifying AFT entries. The test would fail even though routes were successfully being installed on the device as shown below.

`gnmic -a source-ip --skip-verify -u cisco -p cisco sub --path "/network-instances/network-instance[name=default]/afts/aft-summaries/ipv4-unicast/protocols/protocol[origin-protocol=BGP]/state/counters/aft-entries" --stream-mode SAMPLE --sample-interval 5s --encoding json_ietf`

{
  "source": "source-ip",
  "subscription-name": "default-1763957491",
  "timestamp": 1763956505853183804,
  "time": "2025-11-23T22:55:05.853183804-05:00",
  "prefix": "openconfig:",
  "updates": [
    {
      "Path": "network-instances/network-instance[name=DEFAULT]/afts/aft-summaries/ipv4-unicast/protocols/protocol[origin-protocol=BGP]",
      "values": {
        "network-instances/network-instance/afts/aft-summaries/ipv4-unicast/protocols/protocol": {
          "state": {
            "counters": {
              "aft-entries": "1000"
            }
          }
        }
      }
    }
  ]
}

## Root Cause Analysis

The issue was in the use of `gnmi.Await()` for verifying AFT entry counts. The `gnmi.Await()` function uses gNMI subscriptions that wait for **value changes/updates** to match a target value.

gnmi.Await() establishes a gNMI subscription to a path and waits to receive an update event where the value changes to the match target. If the value is **already at the target when the subscription start**, no update event is triggered and the subscription times out waiting for an update that never arrives.

### How it fails this test:
- Test flow: Configure → Push config → Start protocols → Wait for BGP → **Call verification**
- By the time `VerifyDUT()` is called, the 1000 routes from both BGP and ISIS had already been installed and AFT counters were already at 1000 (the target value defined in the test).
- `gnmi.Await()` in the meantime is waiting for an update event from 0→1000 that never happens and the result is a timeout after 1 minute

2. Proposed changes

Replaced `gnmi.Await()` with `gnmi.Get()` queries that directly poll and compare the current AFT entry counts. This aligns with the test [README](https://wwwin-github.cisco.com/B4Test/featureprofiles/tree/main/feature/afts/summary/otg_tests/scale_aft_summary) where we need to validate total number of entries of AFT for IPv4 and IPv6.

- Replaced 4 `gnmi.Await()` calls with 4 `gnmi.Get()` queries
- Each query directly compares the current AFT entry count to `RouteCount` (1000)


feature/afts/summary/otg_tests/scale_aft_summary/route_test.go
```diff
@@ -519,32 +520,35 @@ func configureBGPv6Routes(peer gosnappi.BgpV6Peer, ipv6 string, name string, pre
...
+		ygnmi.WithSampleInterval(interval),
+	)
+}
 
+func VerifyDUT(t *testing.T, dut *ondatra.DUTDevice) {
+	verifyAftEntries := func(desc string, query ygnmi.SingletonQuery[uint64]) {
+		t.Helper()
+		val, _ := gnmi.Watch(t, gnmiOptsForSample(t, dut, 1*time.Second), query, time.Minute, func(v *ygnmi.Value[uint64]) bool {
+			return v.IsPresent()
+		}).Await(t)
```
ashishpawar-google, line 536: Using `return v.IsPresent()` as the `gnmi.Watch` predicate causes `Await` to exit on the very first sample received from the DUT, even if the value is `0` or partial scale routes that haven't finished converging yet. 
Because `IsPresent()` is immediately true on sample 1, `Watch` only waits ~1 second instead of utilizing the 1-minute convergence window. The subsequent equality check then evaluates only once and fails immediately if route programming is still in progress.
Please update the predicate to check for the target route count so it continues polling until convergence (or timeout)

ashishpawar-google (changes requested):
Please rebase / squash the commits before merging to drop commits 1 and 5 so the commit history remains clean and only includes the AFT scale summary test changes.

feature/afts/summary/otg_tests/scale_aft_summary/route_test.go
ashishpawar-google, line 536: Discarding the error from `.Await(t)` with `val, _ := ...` and `got, _ := val.Val()` means that if the watch times out or the stream drops, `got` defaults to `0` and prints an assertion error (`got: 0, want: 100000`) rather than clearly reporting a gNMI timeout/RPC error. We should check and report `err` returned by `Await(t)`

PR #4861, opened 2025-11-26
Added default_network_instance for vendor CIENA in metadata file

prince-codec:
deviations added for Vendor Ciena to use "default" as default_network_instance whenever get or set by the testcase


danielbarney (approved):

dpelton-ciena:
> LGTM for platform, what is the plan to remove this deviation? Is ciena plan to use "DEFAULT" instead of "default"?

Thanks for the feedback @dbarney1106.

Our current code uses "default", and we are investigating the work required to transition to use "DEFAULT".

PR #4870, opened 2025-12-02
Update DUT initialization for test PF-1.2; Remove deviation check for counters

ijernoiu:
Update DUT initialization profile used by test PF-1.2.
Remove deviation check for OC PolicyForwarding counters


ram-mac:
Validated the PR 4870 and the test still failing in Google environment. Please see attached the fail logs for reference in bug 498146099

PR #4874, opened 2025-12-03
Container2 test changes.

arvbaska1:
Made below changes to the test:

- in dial gNMI case use gNMI IANA port 9339
- Add a sample gRIBI entry for gRIBI Get response to work


alshabib:
Please fix the checker error.

PR #4903, opened 2025-12-09
Addressed BGP route policy issue and fixed

keysightgems:
Attached logs here https://partnerissuetracker.corp.google.com/issues/415458482


ram-mac (changes requested):
Please make changes in the comment above.

ram-mac (approved):
Validated the test and it is passing. So approving it

PR #4910, opened 2025-12-10
[RT-1.24] Adding default policy

rolandphung:
Set default import policy to REJECT_ROUTE when it's eBGP. Also extend the time it takes for wait for the prefixes to be installed.


AmrNJ (approved):
LGTM.

rohit-rp (approved):

feature/bgp/otg_tests/bgp_2byte_4byte_asn_policy_test/bgp_2byte_4byte_asn_policy_test.go
AmrNJ: Please increase the timeout to 30s.

PR #2066, opened 2023-08-28
Configure NI name whenever configuring type

kjahed:
Leaf ref validation on the router side requires list keys to be configured. This PR changes all tests that configure the default NI _type_ to also configure it's _name_. Note that we use update instead of replace to avoid deleting any configuration under the NI done outside the test (e..g, static routes).

PR #4932, opened 2025-12-15
Cisco: remove obsolete deviations for 2 FNTs

karthikeya-remilla:
The leafref changes should not be under a deviation for FP-1.1 and P4RT-1.3, as these changes are enforced by XR as per OC/Yang spec. 

Similar change was done for OC NI config in the past : https://github.com/openconfig/featureprofiles/pull/2066

Note: Removed PowerDisableEnableLeafRefValidation from deviations.go which is no longer used. But removing the deviation power_disable_enable_leaf_ref_validation from metadata.proto and metadata.pb.go will cause huge churn due to metadata field renumbering, so I'm leaving that as is.


karthikeya-remilla:
@AmrNJ could you please review this and close?

ksgireesha:
Hi @karthikeya-remilla , Please fix the Static Analysis and Static Analysis (ARM64) erorrs

karthikeya-remilla:
Added latest patch to fix SA errors. They should be resolved now. 

Waiting for checks to complete.

karthikeya-remilla:
@rohit-rp @ksgireesha since the checks are now completed and green, could we please continue with merging this?

PR #4934, opened 2025-12-15
Enable skip_tcp_negotiated_mss_check for BGP TCP MSS path MTU test

mukherso:
**Files changed**

- (M) feature/bgp/otg_tests/bgp_tcp_mss_path_mtu/metadata.textproto

**Change Scope**

Updated the featureprofiles BGP TCP MSS path MTU test behavior to account for the configured TCP MSS value being reported in transport/state, rather than the negotiated MSS value.

The tcp-mss leaf is defined only under the transport configuration in the OpenConfig public model, and the same value is reflected in the corresponding state container for both peer-groups and neighbors. As a result, the effective configured MSS value is being reported in transport/state, not the negotiated MSS derived from TCP session establishment :

```
container transport {  
  description  
    "Transport session parameters for the BGP neighbor";  
  container config {  
    description  
      "Configuration parameters relating to the transport  
      session(s) used for the BGP neighbor";  
    uses bgp-common-neighbor-group-transport-config;  
  }  
  container state {  
    config false;  
    description  
      "State information relating to the transport session(s)  
      used for the BGP neighbor";  
    uses bgp-common-neighbor-group-transport-config;  
    uses bgp-neighbor-transport-state;  
  }  
}
```

However, the following tests:

https://github.com/openconfig/featureprofiles/blob/main/feature/bgp/otg_tests/bgp_tcp_mss_path_mtu/bgp_tcp_mss_path_mtu_test.go#L382

https://github.com/openconfig/featureprofiles/blob/main/feature/bgp/otg_tests/bgp_tcp_mss_path_mtu/bgp_tcp_mss_path_mtu_test.go#L482

currently expects the negotiated MSS value when querying: `/bgp/neighbors/neighbor/transport/tcp-mss/state
`

To align the test with OpenConfig implementation, the Arista-specific deviation `skip_tcp_negotiated_mss_check` is enabled. This allows the test suite to pass while preserving correct validation of TCP MSS configuration semantics.


navaneethyv (approved):

yini101:
Hello @navaneethyv and @AmrNJ, I see the PR was approved 2 weeks. Is there anything we need to do to unblock and merge the PR? Thanks.

PR #4936, opened 2025-12-16
RT-1.31 bgp_3level_nested_policy_test.go

cprabha:
Hi,

Below are the modifications done in RT-1.31.

1.) PERMIT ALL route policy is not required, we can enable default accept policy, this will only allow BGP routes. 
     Explicit PERMIT ALL route policy will allow non bgp routes as well.

2.) Added clean up policy after each test , in case if any of subtest fails, it does not retain policy and that will cause failure in next sub testcase.

3.) Enabled one prefix in invert route filter policy so make sure invert works fine for required prefix.

4.) In each of nested policy we need ACCEPT (this is similar to RT-7.11, please check RT-7.11)

5.) By default, Ebgp does not accept routes , default value is REJECT. Hence explicit ACCEPT is needed to import routes.

Thanks,
Prabha


ram-mac:
Validated the changes in Juniper and it is passing. Checking if this change will break any other vendors.

ram-mac:
Validated the changes for Arista and Juniper. Tests are passing and hence approving the PR

ram-mac (approved):

PR #4983, opened 2026-01-12
Update DUT initialization for test PF-1.8

ijernoiu:


danameme (approved):

ram-mac:
@ijernoiu - I tried validating this today in my setup and it is failing to apply the hwinit configs. Attached the logs to the bug
https://partnerissuetracker.corp.google.com/issues/415458482#comment171

```
[Error] the key size of feature 'qos mac' is larger than the configured primary key size limit of feature sharing the same TCAM bank
```

ram-mac (changes requested):

PR #4986, opened 2026-01-13
Create README.md

vvlakshmanamurthy:
Egress static mpls lsp test


ram-mac (approved):

ram-mac:
@vvlakshmanamurthy - One request.. Can you please keep the same name for file and directory? so instead of creating a new README.md [feature/mpls/otg_tests/Egress_static_mpls_lsp/README.md], we modify the existing one which is [feature/mpls/otg_tests/static_lsp_test/README.md] That way we do not have to move the test into the new directory. I will just modify the test to meet the requirements?

PR #4989, opened 2026-01-13
added subnet mask deviation for TE-9.3 FNT

manan-patel:


AmrNJ:
@manan-patel Please address Gemini comments and provide description.

PR #4991, opened 2026-01-15
gNMI-10: Skipping TestLagMacAddress when the base config is not present

priyaranjannanda:
Changes to Skip TestLagMacAddress 
- When the base config doesn't have LACP related config


rohit-rp (commented):
I need some context here, specifically why would the base config not have LAG Config. IMO the base config should be fixed and necessary config should be included. Maybe we can perform a check to validate is LAG config is present (validate whatever is expected from the base config and fail the test rather than skip it)

priyaranjannanda:
> I need some context here, specifically why would the base config not have LAG Config. IMO the base config should be fixed and necessary config should be included. Maybe we can perform a check to validate is LAG config is present (validate whatever is expected from the base config and fail the test rather than skip it)

This is LACP config and the test case 'TestLagMacAddress' checks for Lacp interface state on this test case. 

It is already skipped in TestLacpMember' with the args.LACPBaseConfigPresent and this commit extend the same for 'TestLagMacAddress' 

https://github.com/openconfig/featureprofiles/blob/e915a0a807b15f15dae224c493b998dd08410e50/feature/gnmi/otg_tests/telemetry_basic_check_test/telemetry_basic_check_test.go#L699

rohit-rp (approved):

PR #5019, opened 2026-01-27
updating route count

cprabha:
Hi ,

I have updated vendor specific route count supported on new platform.

Thanks,
Prabha


feature/gribi/otg_tests/fib_failed_due_to_hw_res_exhaust_test/fib_failed_due_to_hw_res_exhaust_test.go
```diff
@@ -110,6 +111,11 @@ var (
 	fibPassedDstRoute      string
 	fibFailedDstRoute      string
 	fibFailedDstRouteInHex string
+	routeCountModelList    = []string{"PTX10002-36QDD", "PTX12008"}
+
+	vendorUpdatedSpecRoutecount = map[ondatra.Vendor]uint32{
+		ondatra.JUNIPER: 3750000,
```
gggsmith: Just a quick suggestion on the execution: could we move the hardware model check out of the Go logic and into a deviation?

If we use platform_exceptions with a hardware_model_regex in metadata.textproto, the framework will handle the DUT matching for us. There is a similar example we can follow in feature/gnmi/otg_tests/telemetry_basic_check_test/metadata.textproto.

gggsmith (changes requested):
Thank you for the PR @cprabha !

feature/gribi/otg_tests/fib_failed_due_to_hw_res_exhaust_test/fib_failed_due_to_hw_res_exhaust_test.go
cprabha: @gggsmith we are adding deviations for scenarios where we have an alternate OC path or CLI. In this case, the platform supports higher FIB route counts and therefore needs a larger value to fully exhaust the FIB. As a result, applying a deviation may not be suitable, as it could incorrectly suggest that this functionality is unsupported on the platform.

gggsmith: @cprabha I think there's a slight terminology mismatch. In FNT are the standard mechanism for handling  platform-specific variance, including parameter tuning like route counts.

Using platform_exceptions with a hardware_model_regex in the metadata is the idiomatic way to provide these constants to the test logic. This keeps the Go code clean and vendor-neutral, e.g.:
```
  // 1. Get the default vendor route count. 
  // For Cisco/Arista/Juniper-default, this is 2,500,000. For Nokia, this is 2,600,000.
  routeCount := vendorSpecRoutecount[args.dut.Vendor()] 
  
  // 2. Query the deviation.
  // If the DUT is NOT Juniper PTX10002-36QDD or PTX12008, customCount is 0 (no override).
  // If the DUT IS one of those Juniper models, customCount is 3,750,000.
  if customCount := deviations.RouteCountToMaxoutFib(args.dut); customCount != 0 {
      routeCount = customCount // Only executes for the target Juniper models
  }
```
and metadata:
```
platform_exceptions: {
  platform: {
    vendor: JUNIPER
    hardware_model_regex: "PTX10002-36QDD|PTX12008"
  }
  deviations: {
    route_count_to_maxout_fib: 3750000
  }
}
```

gggsmith (changes requested):

feature/gribi/otg_tests/fib_failed_due_to_hw_res_exhaust_test/fib_failed_due_to_hw_res_exhaust_test.go
cprabha: Sure, let me add these

cprabha: updated please check

feature/gribi/otg_tests/fib_failed_due_to_hw_res_exhaust_test/fib_failed_due_to_hw_res_exhaust_test.go
```diff
@@ -462,17 +462,24 @@ func configureDUT(t *testing.T, dut *ondatra.DUTDevice) {
...
 
-	if _, ok := vendorSpecRoutecount[args.dut.Vendor()]; !ok {
+	routeCount, ok := vendorSpecRoutecount[args.dut.Vendor()]
+
+	if !ok {
 		t.Fatalf("Please provide BGP route count for vendor to maxout FIB %v in var vendorSpecRoutecount ", args.dut.Vendor())
 	}
+
+	if args.dut.RawAPIs().BindingDUT().HardwareModel() == "PTX10002-36QDD" || args.dut.RawAPIs().BindingDUT().HardwareModel() == "PTX12008" {
+		routeCount = deviations.MaxOutFIBRouteCount(args.dut)
```
gggsmith: Thank you! Could we add `max_out_fib_route_count` for the other vendors here as well? Having this parameter across all vendors would allow us to remove the hardware model matching logic from the tests entirely.

gggsmith (changes requested):

feature/gribi/otg_tests/fib_failed_due_to_hw_res_exhaust_test/fib_failed_due_to_hw_res_exhaust_test.go
cprabha: @gggsmith updated for other vendors, please check

gggsmith (approved):
Thank you @cprabha !

rohit-rp (approved):

PR #5022, opened 2026-01-28
Acctz-4.2: Allow larger gRPC responses by increasing MaxCallRecvMsgSize

karthikeya-remilla:
Increase payload size and adjust gRPC message size limit in record payload truncation test. 

Summary:
This change increases the gRPC MaxCallRecvMsgSize limit to 45 MB. By default, gRPC clients reject responses larger than the configured maximum size, which was causing failures when handling larger payloads.

Motivation:
Some service calls were returning responses that exceeded the default message size limit.
Raising the limit ensures the client can safely handle larger responses from the server.

Changes:
Added grpc.MaxCallRecvMsgSize(11000000) option when initializing the gRPC client.
This sets the maximum receive message size to ~45 MB.

Impact:
Positive: Prevents client errors for large responses.
Neutral: No changes for normal-sized responses.


ram-mac:
This test is already passing for all vendors and now this change is not required or relevant. If required we can revisit later

PR #5023, opened 2026-01-28
Fix authZ Rotate: Add Recv() call after FinalizeRotation

youssefyamoutyy:
### 1. Brief description and need for this PR

The Rotate() function was not properly consuming the finalize response from the gRPC server, leaving the rotation stream in an incomplete state. This caused subsequent rotate operations to fail with 'rotation is already in progress' error in for ```feature/gnsi/authz/tests/authz/authz1_4_test.go```

### 2. Proposed changes

Add a Recv() call after sending the FinalizeRotation request to properly
complete the rotation handshake and allow subsequent operations


Chounoki (approved):

ram-mac (approved):
LGTM

PR #5071, opened 2026-02-10
CNTR-3: Supervisor failover test

goabhinav:


feature/container/failover/failover_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+		// containerztest.DeployAndStart handles pushing the image and starting the container.
+		// We use it instead of containerztest.Setup because it does not return a cleanup
+		// function, which is ideal for a failover test where we want to verify persistence.
+		if err := containerztest.DeployAndStart(ctx, t, cli, opts); err != nil {
+			t.Fatalf("Failed to deploy and start container: %v", err)
+		}
+		if err := verifyVolumeExists(ctx, cli, volName); err != nil {
+			t.Fatalf("Volume not found after creation: %v", err)
+		}
```
f-sommerauer: would consider moving this up after create volume (slightly easier to read)

feature/container/failover/failover_test.go
```diff
@@ -0,0 +1,207 @@
...
+	})
+
+	t.Run("VerifyRecovery", func(t *testing.T) {
+		t.Log("Waiting for DUT to reconnect...")
+		// Allow some time for the switchover to initiate and the connection to drop.
+		time.Sleep(switchoverWait)
+
+		// Refresh clients after reconnection.
+		cli = containerztest.Client(t, dut)
```
MarcCharlebois: Validate that standby supervisor is not the same as previously obtained before attempting reconnection. We could end up reconnecting to the same supervisor here.

youssefyamoutyy:
verifyImageExists is missing from the implementation. 
Also switchover is a simple RPC call causing some tests to fail,  maybe it should be changed it to await the active RP's SwitchoverReady telemetry, then issues the RPC?

youssefyamoutyy:
The DUT's containerz Remove* operations return codes.Unknown even when they succeed. That will depend on the vendor right? 

The cleanup functions were checking only for codes.NotFound with status.Code(err) != codes.NotFound, so the codes.Unknown (which is supposed to be success) error was being treated as a failure.  In other terms, cleanup logged t.Errorf (marking test as failed) even though the removal actually worked.

MarcCharlebois:
I agree with @youssefyamoutyy here. I'm working on a patch to fix this.
I'm currently testing it, but in short it'll include a function to ensure supervisors are synchronized before each switchover as well as before the cold reboot (to ensure the last switchover fully syncs before attempting to reboot - as I've also noted some issues around that area).

PR #5086, opened 2026-02-13
Test semantic validation

chitadi:
Two test files were modified to ensure that the OC paths which were supposed to be covered by the test, as specified by the ReadME, are actually covered. Essentially a semantic check has been performed.  

In optics_power_and_bias_current_test.go, the following paths were semantically validated through the changes:
/component/transceiver/thresholds/threshold/state/laser-bias-current-lower
/component/transceiver/thresholds/threshold/state/laser-bias-current-upper

In telemetry_interface_packet_counter_test.go, the following paths were semantically validated through the changes:
/interfaces/interface/state/cpu
/interfaces/interface/state/management


AmrNJ (approved):
LGTM.

PR #5092, opened 2026-02-17
Fix config issue for interfaces not configured correctly in TE-17.1

AishwaryaSKulakrni:
- Populated interface-ref with base interface and subinterface index when interface name contains a dot notation.
- Added check for InterfaceRefConfigUnsupported deviation to handle devices that don't support interface-ref config.
- Ensured ISIS interface configuration properly parses interface names and sub-interface indices.


AmrNJ:
@karthikeya-remilla I think we can close this PR since we have #5560 ?

karthikeya-remilla:
This PR is obsoleted by https://github.com/openconfig/featureprofiles/pull/5560 so we can close this.

PR #5182, opened 2026-02-27
Fix flaky QoS output queue counter validation in DP-1.4

aks03dev:
1. Brief description and need for this PR
- The QoS output queue counters test (TestQoSCounters) was failing intermittently because the counter validation logic was too strict. It required every consecutive pair of 30-second gNMI SAMPLE updates to show a different value. When two consecutive samples were equal (e.g., both zero or both the same value), the test failed even though the DUT was forwarding traffic correctly.

- This could have happened because with a 30-second sample interval, the first sample can arrive before the pipeline has populated the counter, and the next sample can still reflect the same value.

- The test’s goal is to verify that QoS output queue counters are live and incrementing during active traffic over a 300-second window, not that every consecutive sample pair differs. The previous logic caused false failures when the pipeline had not yet updated between two samples.

2. Proposed changes
- Replace the requirement that every consecutive 30-second sample pair must differ with a check that at least one increment is observed across all samples in the 300-second window. This keeps the intent (counters are live and incrementing) while tolerating the initial lull

- Increase the sleep before starting the gNMI SAMPLE subscription from the arbitrary 5 seconds to a standard 30 seconds. This gives the counter pipeline more time to complete its first cycle after traffic starts, reducing the chance of early zero sample 

Update: As per the review comment, the 300s interval would weaken the test condition considering latency requirement, changing the measuring interval to 45s

### DIFF
--- a/feature/qos/otg_tests/qos_output_queue_counters_test/qos_output_queue_counters_test.go
+++ b/feature/qos/otg_tests/qos_output_queue_counters_test/qos_output_queue_counters_test.go
@@ -231,10 +231,10 @@ func TestQoSCounters(t *testing.T) {
 	t.Logf("Running traffic 1 on DUT interfaces: %s => %s ", dp1.Name(), dp2.Name())
 	t.Logf("Sending traffic flows: \n%v\n\n", trafficFlows)
 	ate.OTG().StartTraffic(t)
-	time.Sleep(5 * time.Second)
-	outputQosPerSecoundCounterOK := validateoutputQosPerSecoundCounter(t, dut, dp1, dp2, trafficFlows)
+	time.Sleep(30 * time.Second)
+	outputQosPerSecondCounterOK := validateOutputQosPerSecondCounter(t, dut, dp1, dp2, trafficFlows)
 	ate.OTG().StopTraffic(t)
-	if !outputQosPerSecoundCounterOK {
+	if !outputQosPerSecondCounterOK {
 		t.Errorf("Output QoS per second counter is not updated correctly")
 	}
 	time.Sleep(30 * time.Second)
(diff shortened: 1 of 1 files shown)


sezhang2 (changes requested):
It is fine to wait longer time to check counters.  300s is too long and may miss the issue with high latency. A bug should be filed if the QoS counters do not increment every 30 seconds.

Do we know the longest interval for QoS counters to be updated?

aks03dev:
>It is fine to wait longer time to check counters. 300s is too long and may miss the issue with high latency. A bug should be filed if the QoS counters do not increment every 30 seconds.
Do we know the longest interval for QoS counters to be updated?

No, we have not measured that, is there time period that you suggest after which we should be checking for the increment?

sezhang2:
> > It is fine to wait longer time to check counters. 300s is too long and may miss the issue with high latency. A bug should be filed if the QoS counters do not increment every 30 seconds.
> > Do we know the longest interval for QoS counters to be updated?
> 
> No, we have not measured that, is there time period that you suggest after which we should be checking for the increment?

I suggest we check every 35s or 45s to allow a little bit delay.

aks03dev:
> > > It is fine to wait longer time to check counters. 300s is too long and may miss the issue with high latency. A bug should be filed if the QoS counters do not increment every 30 seconds.
> > > Do we know the longest interval for QoS counters to be updated?
> > 
> > 
> > No, we have not measured that, is there time period that you suggest after which we should be checking for the increment?
> 
> I suggest we check every 35s or 45s to allow a little bit delay.

Sure, will choose 45s and make the change in this PR

sezhang2 (approved):

PR #5183, opened 2026-02-27
Fix LAG operational status for disabled member ports on Juniper devices

priyaranjannanda:
Changes for Juniper Devices 
- when LAG member port is disabled, the LAG oper status goes to LOWER_LAYER_DOWN instead of DOWN


priyaranjannanda:
Changes are merged through https://github.com/openconfig/featureprofiles/pull/5343/changes

PR #5185, opened 2026-02-28
Add ICMPv6 layer with echo payload creation

priyaranjannanda:
- Added creation of an ICMPv6 layer and set an echo payload


AmrNJ:
This change is already merged.

PR #5188, opened 2026-03-03
URPF-1.1: Updated the TCAM for invalid IPv6 test validation.

keysightgems:
**Readme Location:** https://github.com/openconfig/featureprofiles/blob/main/feature/urpf/otg_tests/urpf_vrf_test/README.md

Find the attached logs here: https://partnerissuetracker.corp.google.com/u/1/issues/415458482


internal/cfgplugins/staticnexthop.go
```diff
@@ -400,11 +401,17 @@ func NextHopGroupConfigForIpOverUdp(t *testing.T, dut *ondatra.DUTDevice, params
 					tunnelDst += fmt.Sprintf("entry %d tunnel-destination %s \n", i, addr)
 				}
 				cli = fmt.Sprintf(`
+					qos rewrite ipv4-over-udp inner dscp disabled
+					qos rewrite ipv6-over-udp inner dscp disabled
```
nupkanoi, line 405: These commands are not needed anymore

keysightgems: Yes, those commands are no longer required for this test. However, they are still needed for other tests, so we can't remove them unconditionally. I think removing them may affect those tests. If they are not required for this specific test, we can add another parameter to control whether those commands are applied. Please let me know what you think, and I will update the code accordingly.

PR #5194, opened 2026-03-04
gnmi.Replace call added

gauriwalinjkar91:
Adding gnmi.Replace to push configuration to DUT


rohit-rp (commented):
Can you include why this change is needed in the PR description please, any bug reference or context is helpful (if it exists)?

gauriwalinjkar91:
> Can you include why this change is needed in the PR description please, any bug reference or context is helpful (if it exists)?

Interface configuration required by the test was not being applied. This pull req adds explicit gNMI Set call to push the generated interface configurations to the DUT. No other bug reference is applicable/available.

rohit-rp (approved):

PR #5205, opened 2026-03-09
Fix GNMI-1.18 for Arista

prakashbadri-arista:
The fix updates the regex for identifying IC components in Arista devices so that the enttries for per fap components such as "SwitchChip2/0.1", "SwitchChip3/1.4" are skipped.


ram-mac (approved):

jmunusa (approved):
LGTM

PR #5218, opened 2026-03-11
acctz-3.1: increase ssh defer sleep for result consistency

dipchauh:
Increase the defer sleep time to 6s to ensure ssh commands complete before the connection closes and prevent intermittent test failures.
Also added check to not skip shell records


ram-mac (approved):

trathod1:
Conflict Resolved.

PR #5248, opened 2026-03-22
gnoi-3.1 complete chassis reboot fix

hmod2001:
gNOI-2.1 (Packet Link Qualification)
Packets dropped tolerance: The GetPacketsDropped() check was asserting a strict want 0, causing the member-link (aggregate) PLQ subtests to fail with exactly 1 dropped packet across all runs and both link speeds (100G and 400G). This is a consistent off-by-one in IOS-XR's aggregate PLQ packet counter, not real packet loss (1 drop out of ~24.9M packets = 0.000004%). The check is updated to use the existing tolerance variable (0.0001%) already defined in the function for sent/received mismatch checks, making the dropped packet check consistent with the rest of the result validation.

gNOI-3.1 (Complete chassis reboot)
The post-reboot component count comparison was failing on Cisco 8808 because certain fabric card components present pre-reboot were absent post-reboot.
Proposed Changes
Filter DISABLED components from pre/post reboot count comparison: On the 8808, fabric card NPU components (0/FC{0,2,3,7}-NPU0, 0/FC{0,2,3,7}-NPU1) are reported with DISABLED oper-status pre-reboot but are absent from the telemetry tree entirely post-reboot. This caused a persistent count mismatch of 6 components that never resolved within the 15-minute timeout. Add a filteredCompCount helper that excludes DISABLED and UNSET components from the count, and use it in place of the raw GetAll length comparison. The debug diff logging retains its existing UNSET-only filter and is unchanged.


ksgireesha:
@hmod2001  - It looks like there are a few issues blocking this PR from merging:

1) The `PR Approval Workflow` check is failing.
2) Multiple checks are stuck pending. Could you please merge the latest changes from `main` to clear the `Check for outdated files` pending state and re-trigger the CI?
3) Please address the open review comment from Gemini Code Assist regarding the magic number timeout value (`60*time.Second`) and officially mark the thread as resolved in the UI.

karthikeya-remilla:
@ksgireesha PR is now ready for review and merge. Please proceed.

ksgireesha:
@rohit-rp, Please merge this PR

PR #5265, opened 2026-03-26
Revert "Added missing entries in test registry"

kapilrajput-g:
Reverts openconfig/featureprofiles#5260


kapilrajput-g:
Its revered because some new entries of tests in test registry does not follow required structure.

PR #5266, opened 2026-03-26
feat: Double GUE Decapsulation (PF-1.26)

mellowdrifter:
Replacing PR 5081 with a clean, rebased version.

### DIFF
--- a/feature/policy_forwarding/otg_tests/double_gue_decap/README.md
+++ b/feature/policy_forwarding/otg_tests/double_gue_decap/README.md
@@ -1,0 +1,214 @@
+# PF-1.26: Double GUEv1 Decapsulation for Overlay Probing
+
+## Objective
+
+This test validates the router's (DUT) capability to perform double-decapsulation using a soft-loopback mechanism in the DUT. It ensures that overlay probe traffic, dual-encapsulated with GUE headers, can be correctly decapsulated in two stages:
+
+* Stage 1: Decapsulation of the outermost GUE header (H1) and redirection to a loopback interface via LPM lookup.
+* Stage 2: Recirculation through the soft-loop and decapsulation of the middle GUE header (H2), exposing the inner payload (H3) for final forwarding.
+
+The test specifically verifies this behavior for both IPv4 and IPv6 middle headers (H2) `[Outer_IP_header][UDP][Middle_IP_header][UDP][Inner_IP_header][PAYLOAD]`
+
+## Topology
+
+* ATE Port 1: Connected to DUT Port 1 (Ingress). Senders of double-encapsulated probe traffic.
+* ATE Port 2: Connected to DUT Port 2 (Egress). Receiver of final decapsulated traffic.
+* DUT Port 3: Configured in soft-loop mode. This port serves as the recirculation point for the double-decap process.
+
+ ```mermaid
+graph LR;
+A[ATE:Port1] --Ingress (Double Encap)--> B[DUT];
+B --Egress (Decapped)--> C[ATE:Port2];
+```
+
+## Procedure
+
+### DUT Configuration
+* Decap Aggregates: Configure DUT:Port3 with an aggregate IPv4 /28 and IPv6 /60 block address.
+* Soft-Loop: Enable software loopback on DUT Port 3.
+* Forwarding Rules:
+ - Configure dual decap policies on the DUT to facilitate the decapsulation of Header 1 and Header 2 (both UDP 6080). The decap policies for Header 2 (IPv4 or IPv6 | UDP) and Header 1 (IPv6|UDP) must target specified aggregates: for Header 2, utilize the IPv4 /28 and IPv6 /60 blocks assigned to DUT Port 3; for Header 1, use a /60 subnet of the destination address.
+- The destination IP for Header 2 must resolve to the IP address assigned to the soft-loop interface (DUT Port 3) post decapsulation of Header 1 to facilitate recirculation.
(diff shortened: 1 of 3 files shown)


sachendras (approved):

feature/policy_forwarding/otg_tests/double_gue_decap/README.md
```diff
@@ -0,0 +1,212 @@
...
+| **Flow Type #2 (IPv6/IPv6/IPv6)** | Outer IP | ATE-P1-IP | DECAP-DST-OUTER | 6080 | 35 | 70 |
+| | Middle IP | IPV6-MID-SRC | DECAP-DST-INNER | 6080 | 32 | 60 |
+| | Inner IP | IPV6-SRC-HOST | IPV6-DST-HOST | N/A | 20 | 50 |
+| **Flow Type #3 (IPv6/IPv4/IPv4)** | Outer IP | ATE-P1-IP | DECAP-DST-OUTER | 6080 | 35 | 70 |
+| | Middle IP | IPV4-MID-SRC | DECAP-DST-INNER | 6085 | 32 | 60 |
+| | Inner IP | IPV4-SRC-HOST | IPV4-DST-HOST | N/A | 20 | 50 |
+
+### Traffic Generation:
+
+* Flow #1 (IPv4 H2): ATE Port 1 sends a packet with [H1: GUE][H2: IPv4 GUE][H3: Payload].
```
ram-mac, line 50: @mellowdrifter - The traffic should also include a scenario to specify the packet size. Double-encapsulated packets are prone to exceeding MTU limits. Tests should verify that the DUT correctly handles (or drops/reports) packets that exceed the soft-loopback interface's MTU during the recirculation stage.

feature/policy_forwarding/otg_tests/double_gue_decap/README.md
```diff
@@ -0,0 +1,212 @@
...
+    * 100% packet loss on ATE Port 2.
+
+#### PF-1.26.5: Negative - Middle Header no IPv6 destination available
+* Configure Flow Type #2 with an IPv6 inner header that has an unreachable destination IP.
+* Initiate traffic.
+* **Verification**:
+    * DUT decapsulates the outer header.
+    * DUT decapsulates the middle header.
+    * DUT performs LPM on the inner destination `IPV6-DST-HOST` and drops the packet due to no route.
+    * 100% packet loss on ATE Port 2.
```
ram-mac, line 102: @mellowdrifter - We need to add another scenario to verify behavior if the route lookup for H3 fails to leak back into the DEFAULT VRF. We can add an event to bring down the recirculation interface administratively.

feature/policy_forwarding/otg_tests/double_gue_decap/README.md
```diff
@@ -0,0 +1,212 @@
...
+
```
ram-mac, line 103: @mellowdrifter - Another scenario is that the decapsulating nodes can still perform load balancing (ECMP via hashing) correctly. So, after the first decapsulation is done, the DUT can use the headers of H2 for hashing when forwarding to the loopback or the final egress interface.

ram-mac (changes requested):

feature/policy_forwarding/otg_tests/double_gue_decap/README.md
mellowdrifter: Can we be more specific here? If a packet comes into a DUT, it's already traversed the devices using the maximum link MTU. i.e. as long as the max interface MTU and loopback MTU are the same, this won't happen?

I know in a lab we can force even bigger packets over an interface using test gear, but we're not doing that in prod

feature/policy_forwarding/otg_tests/double_gue_decap/README.md
mellowdrifter: I don't understand this comment. What do you mean by 'route lookup for H3 fails to leak back into the default vrf' ?

different vrfs are design specific. at least in our design, all interfaces are in the same default vrf, there are no other vrfs

feature/policy_forwarding/otg_tests/double_gue_decap/README.md
mellowdrifter: is this a specific scenario? I'm not sure if we're asking the vendors to take h2 into account here or not?

feature/policy_forwarding/otg_tests/double_gue_decap/README.md
sachendras: @ram-mac , imo too, Soft-loop MTU must be the same as the interface MTU (default say 1500) 
Also, payload in this case is not expected to be more than 60B in production and in total the largest packet size with several GUE headers shouldn't be more than 512B max. So the default must function well. 
In other words, lets not add scenarios that will not be used in production inflating the feature requirements.

feature/policy_forwarding/otg_tests/double_gue_decap/README.md
sachendras: The intent of FNT is double decap functionality. While ECMP hashing is a requirement for Production traffic, hashing verification as such must be a separate FNT (which I think exists).

feature/policy_forwarding/otg_tests/double_gue_decap/README.md
ram-mac: ACK

feature/policy_forwarding/otg_tests/double_gue_decap/README.md
ram-mac: ACK

ram-mac (approved):

PR #5272, opened 2026-03-27
ate_tests: add deviation for gNOI KillProcess

brianneville:
add a new deviation which when set, allows for gNOI KillProcess RPCs to be executed for cases where the YANG
/system/processes/process[pid=<PID>]/state/name value does not line up with the name that KillProcess expects when attempting to kill the process.


ram-mac (commented):
Added some suggestions

PR #5277, opened 2026-03-31
Fixing acctz tests record_subscribe_full, record_subscribe_partial, record_subscribe_non_grpc

ram-mac:
Add a flag "static_binding" to use it to fetch the SSH IP:Port info when run with testbeds using static binding file,

Add below deviations

Add role field https://github.com/openconfig/gnsi/blob/main/acctz/acctz.proto#L170 in https://github.com/openconfig/featureprofiles/blob/main/internal/security/acctz/acctz.go#L884, Right now its not set.

Authn field is only populated for cisco for Login/Once/Enable records, not for type Operation, so set to nil for records in this test which are of type SESSION_STATUS_OPERATION.
https://github.com/openconfig/gnsi/blob/main/acctz/acctz.proto spec does'nt mandate it needs to set for session status operations.

Added a different username/password specific to which fails authorization only, and modify existing failusername for fail authentication purpose.

Added Juniper specific role configuration


mihirpitale-googler (approved):
lgtm

feature/gnsi/acctz/tests/record_subscribe_full/record_subscribe_full_test.go
```diff
@@ -100,7 +100,7 @@ func TestAccountzRecordSubscribeFull(t *testing.T) {
 		path := r.GetGrpcService().GetRpcName()
 		id := r.GetSessionInfo().GetUser().GetIdentity()
 		// Skip if the path is not in the list of paths to be tested or if the id is not a success or fail username.
-		if !slices.Contains(acctz.TestPaths, path) || !slices.Contains([]string{acctz.SuccessUsername, acctz.FailUsername}, id) {
+		if !slices.Contains(acctz.TestPaths, path) || !slices.Contains([]string{acctz.SuccessUsername, acctz.FailAuthenticateUsername}, id) {
```
morrowc: this is getting complicated, you PROBABLY want to make this a function isSkip() and actually do the logic in a more clean manner.

PR #5294, opened 2026-04-03
TRANSCEIVER-13.1 Add distinct performance monitoring interval

jnelson-sj:
Use a distinct performance monitoring interval to avoid overwriting the global samplingInterval variable.


jnelson-sj:
@rezachit

navaneethyv:
IS this an arista spec ific change 
?

PR #5297, opened 2026-04-03
Adding conditional passing of metadata to devices

ram-mac:
Arista devices expect metadata to have username, password even in mTLS scenario. 
So, this change modifies GetRequestMetadata to conditionally omit the "password" from the returned metadata when using mTLS with Arista devices. 
This change ensures that Arista devices, which extract the username from the certificate in mTLS mode, do not fail due to an unexpected password in the metadata. 
The vendor is now passed to PassCred to enable this vendor-specific logic.

### DIFF
--- a/topologies/binding/binding.go
+++ b/topologies/binding/binding.go
@@ -605,7 +605,7 @@ func dialOpts(bopts *bindpb.Options) ([]grpc.DialOption, error) {
 		opts = append(opts, grpc.WithTransportCredentials(tlsConfig))
 	}
 	if bopts.Username != "" {
-		c := &creds{bopts.Username, bopts.Password, !bopts.Insecure}
+		c := &creds{bopts.Username, bopts.Password, opb.Device_ARISTA, !bopts.secure}
 		opts = append(opts, grpc.WithPerRPCCredentials(c))
 	}
 	if bopts.MaxRecvMsgSize != 0 {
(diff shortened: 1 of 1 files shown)


singhavnish2516 (changes requested):

topologies/binding/binding.go
```diff
@@ -699,30 +668,38 @@ func loadCertificates(bopts *bindpb.Options) (*x509.CertPool, tls.Certificate, e
...
-		username = c.username
+
+	// ONLY add username if it wasn't already in the context metadata
+	if !userOverride {
+		returnMap["username"] = user
 	}
-	if password == "" {
-		password = c.password
+
+	// Arista Quirk, Omit password if using mTLS.
```
mudhireddy, line 697: This seems to contradict the description of the PR. Description says "Arista devices expect metadata to have username, password even in mTLS scenario."

In mTLS scenario Arista expects metadata to not contain username and password.

mudhireddy:
@ram-mac are you planning to merge this PR soon to remove duplicate usernames in the metadata?     ACCTZ-1.1, ACCTZ-2.1 fail with duplicate metadata username error and this is the proposed fix. 

" acctz.go:729: Got expected error fetching capabilities with bad creds, error: rpc error: code = InvalidArgument desc = only one username should be provided, got [failauthuser failauthuser]
    acctz.go:770: Error fetching capabilities, error: rpc error: code = InvalidArgument desc = only one username should be provided, got [acctztestuser acctztestuser]"

mudhireddy:
I created a PR with an alternative proposal for the same issue, can you please review it https://github.com/openconfig/featureprofiles/pull/5700

PR #5305, opened 2026-04-06
Singhavnish/aft base test get request

singhavnish2516:
* Add a temp test to validate AFT Get Request (Subscribe Once)


paramasivamn1 (commented):
Hi Avnish, Can you add the key requirements for this changes and also our observation so that reviewers will get the context and behavior. so that they can provide the possible optimization or other  inputs.

PR #5311, opened 2026-04-07
[RT-2.11] Watch AFT entries instead of Get

yini101:
Replace gnmi.Get with gnmi.Watch (2-minute timeout) for IPv4 and IPv6 AFT entry checks in Route_checks. After disabling IS-IS passive and awaiting adjacency UP, the dut requires additional time to settle before the AFT entry is visible via gNMI.


navaneethyv (approved):

navaneethyv:
@rohit-rp  this is good to go, please approve and merge.

PR #5314, opened 2026-04-07
Fix package naming causing copybara failures

RishabhAgarwal-2001:


ram-mac (approved):
Its just a naming convention change from "setupservice" to "setup_service"

LGTM

PR #5319, opened 2026-04-07
RT-7.11: Add Arista deviation and explicit reject in called policy

jjw-a-0617:
1. Add bgp_rib_streaming_config_required deviation for Arista.
2. Add explicit REJECT_ROUTE to match_community_regex so routes not matching the regex community are rejected by the called policy rather than falling off the end with implicit result.


AmrNJ (approved):
LGTM.

rohit-rp (approved):

PR #5337, opened 2026-04-13
Added missing entries in test registry and also fixing some discrepancies in test plan id

kapilrajput-g:
…


AmrNJ (approved):

dplore:
Please add run data.

Also because the README for `gnmi_subscriptionlist_test` was touched, it brought it into scope canonical OC which this older file is missing.    This canonical OC is not very important for that readme,  so I am ok bypassing this and keeping this PR scoped to just updating testregistry.textproto

dplore (approved):

PR #5346, opened 2026-04-16
updated deviations for juniper in record_subscribe_non_grpc

Eeramma:
added deviations for shell accounting.  shell commands are not accounted for juniper added deviations accordingly. Channel id is not supported in acctz record.


ram-mac:
@Eeramma - There are some static failures due to formating errors. Please take a look and also there are conflicts.

ram-mac (approved):
LGTM

PR #5356, opened 2026-04-17
Fabric card count check

mnp-juniper:
Issue : Instead of Fabric card count check ,code was checking for LC count.
Fix:corrected to check for fabric card count


AmrNJ (approved):
LGTM.

PR #5364, opened 2026-04-20
gNOI-5.1: Remove std_dev check from ping test

raviv-arista:
The test pings loopback address where ping RTTs are extremely tight (30-50 us). The variance between packets is typically < 500 ns, which is < 0.0005 ms. Linux ping (iputils) reports mdev rounded to 3 decimal places in ms, so this rounds to mdev = 0.000 ms, and the gNOI server reports std_dev = 0 ns.

The test expectation for ping std_dev > 1 is not correct. Removed std_dev check itself as std_dev will always be greater than or equal to 0.


navaneethyv:
what does the README say ? 
[raviv-arista](https://github.com/raviv-arista)

can you check and update that too?

navaneethyv (approved):

raviv-arista:
> what does the README say ? [raviv-arista](https://github.com/raviv-arista)
> 
> can you check and update that too?

Hi Navaneeth, README.md does not need any updates.

PR #5370, opened 2026-04-21
Adding reboot test covering testing bugs

rohit-rp:
Coverage for:
http://b/418856223	ISIS database entries going missing after a routing process restart
http://b/286086308	Verify chassis reboots with P4RT active 
http://b/399612422	SSH connection to the MFF chassis dropping after a reboot

### DIFF
--- a/feature/system/system_base_test/tests/reboot_test/README.md
+++ b/feature/system/system_base_test/tests/reboot_test/README.md
@@ -12,7 +12,7 @@
 
 ## Procedure
 
-#### Initial Setup:
+### Test environment setup:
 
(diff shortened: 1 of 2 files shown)


feature/system/system_base_test/tests/reboot_test/README.md
```diff
@@ -0,0 +1,147 @@
+# System-2.1: DUT Reboot Test
+
+## Summary
+
+- Configure BGP, IS-IS and redistribution
```
navaneethyv, line 5: The original bug was a restart of ISIS service and not reboot. Could you move it to actual restart case ?

PR #5393, opened 2026-04-28
Update system_ntp_test.go

AmrNJ:
Update the logic of default VRF since it was recently changed with PR#4679 causing regressions for test case.

### DIFF
--- a/feature/system/ntp/tests/system_ntp_test/system_ntp_test.go
+++ b/feature/system/ntp/tests/system_ntp_test/system_ntp_test.go
@@ -49,6 +49,17 @@ var (
 
 // TestNtpServerConfigurability validates that NTP servers can be configured on the DUT.
 func TestNtpServerConfigurability(t *testing.T) {
+	dut := ondatra.DUT(t, "dut")
+	loopbackIntfName := netutil.LoopbackInterface(t, dut, loopbackIntf[dut.Vendor()])
+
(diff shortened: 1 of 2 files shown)


feature/system/ntp/tests/system_ntp_test/system_ntp_test.go
```diff
@@ -84,7 +84,7 @@ func TestNtpServerConfigurability(t *testing.T) {
 	dut := ondatra.DUT(t, "dut")
 	loopbackIntfName := netutil.LoopbackInterface(t, dut, loopbackIntf[dut.Vendor()])
 	for _, testCase := range testCases {
-		if testCase.vrf != "" {
+		if testCase.vrf != "DEFAULT" {
```
ram-mac: @AmrNJ - Can we use deviations.defaultnetworkinstance instead of hardcoding the vrf name?

ram-mac (changes requested):

feature/system/ntp/tests/system_ntp_test/system_ntp_test.go
AmrNJ: Updated the code to use deviation instead of hardcoded instance name.

ram-mac (approved):

rohit-rp (approved):

PR #5406, opened 2026-05-01
[Health-1.1] Update TestComponentStatus

rolandphung:
As per the gNOI spec, Check is required to be called in order to Get the cached value.
See https://github.com/openconfig/gnoi/blob/main/healthz/healthz.proto#L57

Lots of empty values are checked if they exist, however since the default value is `false`, the value may not exist in the `Get`. The test should only check the value and not the existance for the component to be not empty.

Also spell out `empty` in variables to be clearer.


AmrNJ (approved):
LGTM.

rohit-rp (approved):

PR #5408, opened 2026-05-01
[Health-1.1] Devaition for secondary controller card expectations

rolandphung:
EOS does not support streaming telemetry information about the secondary controller card. This means we would bypass the controller card checks when the redundant role is secondary. Deviations
secondary_controller_card_cpu_utilization_unsupported and secondary_controller_card_memory_utilization_unsupported are created to manage this behavior.

EOS also doesn't have linecard CPUs and memory so enabling the linecard_cpu_utilization_unsupported and
linecard_memory_utilization_unsupported are required.


AmrNJ (approved):
LGTM.

rohit-rp (approved):

PR #5409, opened 2026-05-04
[gNSI][TBRotate] Add a base framework for rotating trustbundles.

morrowc:
Add a base set of tests for the rotate-trustbundle process.

Ideally we can expand these once we have some that are working :)
Robotic change here... btw.

PR #5410, opened 2026-05-04
[gNSI][ServerCertRot] Server cert rotation tests.

morrowc:
Add a basic test framework for server cert rotations.


feature/gnsi/certz/tests/trust_bundle_rotation/trust_bundle_rotation_test.go
morrowc: This file got here because this branch is a branch off the gnsiTests branch :( derp.
but.. the only thing changing in this branch is the prior file.

PR #5409 has this file.

morrowc:
this should have come out of a branch from main, not from the feature gnsiTests branch :(

PR #5421, opened 2026-05-05
gNMI-1.25: deviation subinterface_0_state_unsupported for Cisco

karthikeya-remilla:
Context of the sub-interface indexing issue: https://partnerissuetracker.corp.google.com/issues/502419972
Deviation tracking bug: https://partnerissuetracker.corp.google.com/u/0/issues/509766094

Per Google's preference, we are reusing the existing deviation subinterface_0_state_unsupported instead of creating a new one for index=1 which would also require readme and test changes.


AmrNJ (approved):
LGTM.

rohit-rp (approved):

PR #5426, opened 2026-05-05
Adding additional delay between two check request

Sagrohi:
As per the recent commit to this file, an additional checkreq RPC call has been introduced alongside the existing one. According to the Juniper implementation guidelines, there must be a minimum delay of 120 seconds between two RPC calls; therefore, an extra delay of 120 seconds has been added.


Sagrohi:
@googlebot I signed it!

feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go
```diff
@@ -145,7 +145,7 @@ func TestCopyingDebugFiles(t *testing.T) {
 		t.Logf("Healthz Get attempt %d failed: %v", i+1, err)
 		time.Sleep(30 * time.Second)
 	}
-
+	time.Sleep(120 * time.Second)
```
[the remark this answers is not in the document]
Sagrohi: As per the recent commit to this file, an additional checkreq RPC call has been introduced alongside the existing one. According to the Juniper implementation guidelines, there must be a minimum delay of 120 seconds between two RPC calls; therefore, an extra delay of 120 seconds has been added.

feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go
```diff
@@ -145,7 +145,10 @@ func TestCopyingDebugFiles(t *testing.T) {
 		t.Logf("Healthz Get attempt %d failed: %v", i+1, err)
 		time.Sleep(30 * time.Second)
 	}
-
+	if dut.Vendor() == ondatra.JUNIPER {
+		t.Logf("Waiting 120s between Healthz RPC calls for Juniper...")
+		time.Sleep(120 * time.Second)
```
ram-mac, line 150: @Sagrohi - Instead of using static Sleep of 120 seconds, Can we use gnmi.Watch or gnmi.await functions to address this wait time?

ram-mac (changes requested):

feature/gnoi/system/tests/copying_debug_files_test/copying_debug_files_test.go
Sagrohi: The 120-second delay is a Juniper implementation-level rate limiting constraint between two consecutive gNOI  healthz check RPC calls — it is not a state transition we can observe via gNMI telemetry. (gnmi.Watch) or (gnmi.await) require a meaningful OC path to poll against a target value. Since no such path exists to indicate "device is ready for the next CheckReq", a static (time.Sleep) is the only applicable approach here.

ram-mac: Okay

ram-mac (approved):

PR #5428, opened 2026-05-06
Refactor code and add txPowerReadingError value

priyaranjannanda:
The changes include 

- Refactoring the code for validation check
- Adding txPowerReadingError value for consistent validation
- Updated the constant values for inactive optical channel power 
-


alokmtri-g:
@priyaranjannanda ,
could you fix Gemini review comments , which are valid for Optical power validation.

PR #5441, opened 2026-05-11
PF-1.22: fixes for ipv4_guev1_decap_and_hashing_test

davinsingh-arista:
Flow-9 and Flow-10 were failing because they happened to be load balancing with a distribution of 200,000/800,000. This distribution depends on the random load balancing polynomial on the device, and since today the test only sends packets with 10 UDP source ports, there is not enough entropy. It would be better to increase the number of UDP src ports to even the distribution and get the test to pass more consistently.

I've attached the log from a passing test run:
[pf-1_22.txt](https://github.com/user-attachments/files/27603406/pf-1_22.txt)


mohli10:
I have rerun the tests(fptest virtual), and they seem to be failing. Please recheck this and make sure they are passing.

PR #5442, opened 2026-05-11
Add PQC base system test

marcushines:


mihirpitale-googler (approved):
lgtm

feature/system/system_base_test/tests/system_g_protocol_test/system_pqc_protocol_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"google.golang.org/grpc/status"
+
+	gpb "github.com/openconfig/gnmi/proto/gnmi"
+	spb "github.com/openconfig/gnoi/system"
+	authzpb "github.com/openconfig/gnsi/authz"
+	gribipb "github.com/openconfig/gribi/v1/proto/service"
+	p4rtpb "github.com/p4lang/p4runtime/go/p4/v1"
+)
+
+func dialPQCConn(t *testing.T, dut *ondatra.DUTDevice, svc introspect.Service, wantPort uint32, nonStandardPort bool) *grpc.ClientConn {
```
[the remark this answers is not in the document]
marcushines: this isn't correct 

func TestMain(m *testing.M) {
	fptest.RunTests(m)
} 

is in the the other source file

feature/system/system_base_test/tests/system_g_protocol_test/system_pqc_protocol_test.go
```diff
@@ -0,0 +1,157 @@
...
+	dialer := introspect.DUTDialer(t, dut, svc)
+	t.Logf("Dialing %s on %s (Port: %d) with PQC", svc, dialer.DialTarget, dialer.DevicePort)
+	if !nonStandardPort {
+		if dialer.DevicePort != int(wantPort) {
+			t.Fatalf("DUT is not listening on standard port for %q: got %d, want %d", svc, dialer.DevicePort, wantPort)
+		}
+	}
+
+	// Custom TLS config for PQC
+	// 0x6399 is the code point for X25519Kyber768Draft00
```
brianneville: for Go specifically, the `0x6399` codepoint was only introduced temporarily (and it was never exported from the tls package).

it was added in https://go-review.googlesource.com/c/go/+/586656 in go1.23 then removed+replaced in go1.24 https://go-review.googlesource.com/c/go/+/630775 which was when Go got its first real PQC support - see the release notes here https://go.dev/doc/go1.24#cryptotlspkgcryptotls. 

Since this featureprofiles repo is on go1.25, you could use `tls.X25519MLKEM768` here, and this would be compatible with servers which have progressed beyond go 1.23 as well (and should remain forward-compatible since i doubt that the tls.X25519MLKEM768 is going to ever be replaced in the go stdlib)

marcushines: thanks will update

feature/system/system_base_test/tests/system_g_protocol_test/README.md
```diff
@@ -30,6 +30,11 @@ used for mTLS.
 | service-1.3 | gNSI client  | gNSI authz Get works |
 | service-1.4 | gRIBI client | gRIBI Get works |
 | service-1.5 | p4rt client  | P4RT Capabilities works |
+| service-1.6 | gNMI PQC client   | gNMI Get works  |
+| service-1.7 | gNOI PQC client   | gNOI system Time works |
+| service-1.8 | gNSI PQC client   | gNSI authz Get works |
+| service-1.9 | gRIBI PQC client   | gRIBI Get works |
+| service-1.10 | p4rt PQC client   | P4RT Capabilities works |
```
robshakir: For the description here, can we call out that these are specifically for PQC-algorithms or just differentiate them somehow from the ones that above?

feature/system/system_base_test/tests/system_g_protocol_test/README.md
```diff
@@ -30,6 +30,11 @@ used for mTLS.
 | service-1.3 | gNSI client  | gNSI authz Get works |
 | service-1.4 | gRIBI client | gRIBI Get works |
 | service-1.5 | p4rt client  | P4RT Capabilities works |
+| service-1.6 | gNMI PQC client   | gNMI Get works  |
```
robshakir: Can you add these to the `testregistry.textproto` at the root of the repo please?

feature/system/system_base_test/tests/system_g_protocol_test/README.md
```diff
@@ -38,6 +43,8 @@ used for mTLS.
 
 ## OpenConfig Path and RPC Coverage
 
+## OpenConfig Path and RPC Coverage
```
robshakir: Do we need this heading to be repeated?

feature/system/system_base_test/tests/system_g_protocol_test/README.md
```diff
@@ -30,6 +30,11 @@ used for mTLS.
 | service-1.3 | gNSI client  | gNSI authz Get works |
 | service-1.4 | gRIBI client | gRIBI Get works |
 | service-1.5 | p4rt client  | P4RT Capabilities works |
+| service-1.6 | gNMI PQC client   | gNMI Get works  |
+| service-1.7 | gNOI PQC client   | gNOI system Time works |
+| service-1.8 | gNSI PQC client   | gNSI authz Get works |
+| service-1.9 | gRIBI PQC client   | gRIBI Get works |
+| service-1.10 | p4rt PQC client   | P4RT Capabilities works |
```
robshakir, line 43: It'd be useful to have some verbiage in the README that describes what the PQC tests do differently to the base tests.

feature/system/system_base_test/tests/system_g_protocol_test/system_pqc_protocol_test.go
```diff
@@ -0,0 +1,156 @@
...
+	if !nonStandardPort {
+		if dialer.DevicePort != int(wantPort) {
+			t.Fatalf("DUT is not listening on standard port for %q: got %d, want %d", svc, dialer.DevicePort, wantPort)
+		}
+	}
+
+	// Custom TLS config for PQC
+	tlsConfig := &tls.Config{
+		InsecureSkipVerify: true,
+		CurvePreferences:   []tls.CurveID{tls.X25519MLKEM768, tls.X25519, tls.CurveP256},
```
robshakir: Did you consider just adding a parameter to the existing `dialConn` in `system_g_protocol_test.go` that takes the `CurvePreferences` or the `tls.Config` so that we don't duplicate the whole fn?

robshakir: If you did this, you could probably make it such that you just have a table-driven test for each of the existing `TestXXX` functions with the different parameters.

marcushines: so since these will become the base test for all new systems i felt this was a better seperation
addtionally this uses the new GRPC new client which is the preferred method going forward so in followups we should acutally migrate test to use this new API

feature/system/system_base_test/tests/system_g_protocol_test/README.md
marcushines: i guess these must have predated that file as i don't see the others in there ?

robshakir: I think so, but all tests are meant to be registered in there. Thanks for fixing.

feature/system/system_base_test/tests/system_g_protocol_test/README.md
robshakir: Did you push the change here?

robshakir (approved):

robshakir (approved):

PR #5451, opened 2026-05-12
gnmi-3 union_replace test

dplore:
The tests defined here will cover the basic operation of `union_replace` to
add, remove and change configuration and the scenarios defined in [gNMI union_replace section 5.3](google.com/url?q=https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-union_replace.md%2353-union-behavior-options&sa=D&source=docs&ust=1778631692989790&usg=AOvVaw0vo1CKDv7mnCB5jeUMD9zb).
The goal is to catch bugs related to add/remove/change actions and avoid the
complexity of attempting to test all combinations of features and their
configuration items.

In depth configuration scenarios for the full landscape of configuration data
will not be covered in tests of the union_replace feature.  Instead, in depth
configuration will be covered within the tests for the given feature. For
example, detailed coverage of BGP configuration with OC and CLI will be covered
in the [featureprofiles BGP feature folder](https://github.com/openconfig/featureprofiles/tree/main/feature/bgp).


feature/gnmi/otg_tests/union_replace_test/union_replace_test.go
```diff
@@ -0,0 +1,577 @@
...
+	"github.com/openconfig/featureprofiles/internal/attrs/attrs"
+	"github.com/openconfig/featureprofiles/internal/deviations"
+	"github.com/openconfig/featureprofiles/internal/fptest"
+	"github.com/openconfig/featureprofiles/internal/helpers"
+
+	"github.com/openconfig/ygnmi/ygnmi"
+	"github.com/openconfig/ondatra/gnmi"
+	"github.com/openconfig/ondatra/gnmi/oc"
+	"github.com/openconfig/featureprofiles/ondatra"
+	"github.com/openconfig/featureprofiles/ondatra"
```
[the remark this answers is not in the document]
dplore: fixed

dplore:
`vet: tests/union_replace_test/union_replace_test.go:246:51: sb1.String undefined (type *"github.com/openconfig/ondatra/gnmi".SetBatch has no field or method String)` will be fixed when https://github.com/openconfig/ygnmi/pull/184 merges and new release of ondatra is created.

feature/gnmi/tests/union_replace_test/README.md
```diff
@@ -2,181 +2,290 @@
...
-In depth tests for use of union_replace should reside within the features being
-tested.
+Steps
 
-## Testbed type
+* Get baseline configuration
+* Push configuration baseline + A to the DUT
+* Get configuration A.1
+* Verify A.1 == baseline + A
+* Push configuration baseline + A to the DUT
```
gprateekjn: Are we repeating the operations or did these lines get repeated here? If repeating why?

feature/gnmi/tests/union_replace_test/README.md
```diff
@@ -2,181 +2,290 @@
...
-Generate a configuration D.2 with a port-speed mismatch in the CLI which should be accepted and applied by the DUT.
+## gNMI-3.5 - union_replace move configuration
+In some scenarios it is observed that moving a configuration from one interface
+to another can trigger bugs. Particularly if there is some conflicting element
+in the configuration such as an IP address.  This test moves a an IP address
+from interface 1 to interface 2 using union replace.
 
-### gNMI-3.7 - union_replace rejected with error in CLI with OC
-Verify a DUT rejects and rolls back a gnmi.Set union_replace with an invalid configuration in origin CLI.  Verify the original configuration is preserved.  
+### gnmi-3.5.1 move IP address between interfaces using OC
```
gprateekjn: Steps to recreate this testcase is missing.

feature/gnmi/tests/union_replace_test/README.md
```diff
@@ -2,181 +2,290 @@
...
+
+* Get configuration E.1 from DUT.
+* Generate a configuration E.2  which includes the overlapping configuration.
+* Push the configuration E.1 + E.2 to the DUT.
+* Confirm the DUT rejects the gnmi.Set with a gRPC error code of
+  INVALID_ARGUMENT. Log the contents of the optional gRPC error string.
+* Get configuration E.3 from DUT.
+* Verify the configuration is as expected
+  * For option 1 NOS, verify E.1 == E.3  (the configuration is unchanged).
+  * For option 2 NOS, verify E.2 == E.3 (the CLI config is updated, the OC
```
gprateekjn: Shouldnt this be E.1+E.2 == E.3 ?

feature/gnmi/tests/union_replace_test/README.md
```diff
@@ -2,181 +2,290 @@
...
-Perform the steps where a configuration D.2 where the port-speed and breakout set using OC.
+* Identify an interface without a transceiver module installed.
+* Get configuration D.1 from DUT
+* Generate a configuration D.2 , including a port-speed and breakout mode for
+the interface without a transceiver.
+* Push the configuration to the DUT Verify the gnmi.Set is accepted
+* Get configuration D.3
+* Verify  D.2 == D.3 configuration.  That is, verify the “config” leaves for
+breakout mode and port speed are set to the target values.
+* Verify the state for the interface is oper-state DOWN.
```
gprateekjn: Is this the expectation across vendors or is it a Arista specific behavior?

gprateekjn (commented):
added some high level comments.

feature/gnmi/tests/union_replace_test/README.md
```diff
@@ -2,181 +2,290 @@
...
+The goal is to catch bugs related to add/remove/change actions and avoid the
+complexity of attempting to test all combinations of features and their
+configuration items.
+
+In depth configuration scenarios for the full landscape of configuration data
+will not be covered in tests of the union_replace feature.  Instead, in depth
+configuration will be covered within the tests for the given feature. For
+example, detailed coverage of BGP configuration with OC and CLI will be covered
+in the [featureprofiles BGP feature
+folder](https://github.com/openconfig/featureprofiles/tree/main/feature/bgp).
```
robshakir: Are there plans to add such tests?

feature/gnmi/tests/union_replace_test/README.md
```diff
@@ -2,181 +2,290 @@
...
+
+### Baseline DUT management configuration
+
+`union_replace` causes the entire device configuration to be overwritten.  This
+is a problem for featureprofiles because there is an assumption that a
+“baseline configuration” is present on the DUT, allowing its management
+interface(s) to be reachable to the ondatra test runner. This baseline
+configuration varies per environment (eg: at vendors and at Google) and is not
```
robshakir:
```suggestion
configuration varies per environment and is not
```

feature/gnmi/tests/union_replace_test/README.md
```diff
@@ -2,181 +2,290 @@
...
+specified by the featureprofiles code.  A featureprpfiles test with
+`union_replace` must have or gain knowledge of the “baseline configuration” so
+it can be safely replaced without disrupting management connectivity.
+
+This test uses ssh to to obtain the baseline configuration in CLI format and
+merge that with the union_replace configuration generated by the test.
```
robshakir: Is this something that can be found from the ONDATRA binding? It seems like the binding must know what configuration it pushed.

feature/gnmi/tests/union_replace_test/README.md
dplore: This is repeating intentionally.  It's a very simple test to ensure idempotency.  

ie:  Let's push the same configuration we received and verify that nothing has changed. 

Added some text to clarify that this is intentional

feature/gnmi/tests/union_replace_test/README.md
dplore: added!

feature/gnmi/tests/union_replace_test/README.md
dplore: E.1 (the original config) should equal E.3.  This is because E.2 contains configure triggering an error condition which the DUT should reject, leaving the original E.1 configuration intact.   While we require the error, we also go further and get the configuration after the error to confirm that the latest (E.3) configuration is the same as the original, baseline configuration (E.1).

feature/gnmi/tests/union_replace_test/README.md
dplore: No, the work to refactor existing tests to use `union_replace` is not scoped yet but will be.

feature/gnmi/tests/union_replace_test/README.md
dplore: Done

feature/gnmi/tests/union_replace_test/README.md
dplore: At the moment, rather than extracting a copy of the running-config via CLI/gNMI to store it as a string inside a Go variable, ONDATRA delegates the definition of the "original config" to the testbed reservation implementation.   ONDATRA only implements a [.Push(t)](https://github.com/openconfig/ondatra/blob/975b1bf75c21a1480931ad248e70511153575c8f/config/vendor_test.go#L151) function at this time which triggers the  binding to re-apply whatever baseline was present when the device reservation session started.  I searched around a bit and see there is no`config.Get` or similar method to access the DUT's baseline config that an ONDATRA binding is expected to implement.     

I guess we could add a `config.Get` function into the ONDATRA binding to expose the DUT baseline config?    That could be a future PR I think?   Some, perhaps rambling considerations:
- This would introduce a new expectation on the binding which I wonder if it would cause more work for the community?  I'm not sure what the value would be versus getting the configuration from the DUT in the test itself?    It would increase freedom for test environments to support different ways to set/get the baseline configuration in this test.  Is abstracting out ssh into the binding vs. into a featureprofiles library  important for this PR?  
- Is it a good idea to invoke the binding from within a test?   Should the framework underlying functions like [config.Push](https://github.com/openconfig/ondatra/blob/975b1bf75c21a1480931ad248e70511153575c8f/config/vendor_test.go#L151) (and some future, config.Get) remain hidden from the tests using the framework?

dplore:
@gprateekjn  resolved all your comments
@robshakir ready for your response

feature/gnmi/tests/union_replace_test/README.md
```diff
@@ -2,181 +2,315 @@
...
+unvalidated config to a DUT, but it seems unlikely to reveal bugs.
+
+For the case where the config passes validation but contains a semantic error,
+a test case could be configuring an interface to use a QoS queue that doesn’t
+exist.  In this case, the queue name is a string, not a leafref.  See
+qos/interfaces/interface/input/queues/queue/config/name.    (I think this leaf
+is a string and not a leafref because a device may expose queues which are not
+explicitly configured)
 
 Steps
```
gprateekjn: Missed the sub heading to this test. Should it be 3.7.1?

feature/gnmi/tests/union_replace_test/README.md
```diff
@@ -2,181 +2,315 @@
...
 
+* Get configuration E.1 from DUT.
+* Generate a configuration E.2  which includes invalid configuration (see sub
+  tests).
+* Push the configuration E.1 + E.2 to the DUT.
+* Confirm the DUT rejects the gnmi.Set.
+* Get configuration E.3 from DUT.
+* Verify E.1 == E.3  (the configuration is unchanged).
+
+### gnmi-3.7.1 reference validation error in OC
```
gprateekjn: reference validation error in CLI test is missing

feature/gnmi/tests/union_replace_test/README.md
gprateekjn: I am referring to Option 2, where device will accept E.1 + E.2 which should be equal to E.3 (NOS which accepts overlapping config)

gprateekjn (approved):
Added few comments, rest LGTM

feature/gnmi/tests/union_replace_test/README.md
dplore: These steps are common to the two below subheadings, 3.7.1 and 3.7.2.  Added text to clarify

feature/gnmi/tests/union_replace_test/README.md
dplore: Oh I see, I have corrected this.   (verify E.1 + E.2 == E.3)

feature/gnmi/tests/union_replace_test/README.md
dplore: revised to use invalid interface instead of BGP policy to simplify this case.  Added OC and CLI validation cases.

dplore: Thanks for catching this!

feature/gnmi/tests/union_replace_test/README.md
dplore: All implementations should behave this way.

dplore (commented):
@gprateekjn comments resolved

@robshakir comments responded to and marked resolved, but feel free to reopen if the response is not 'good enough'

feature/gnmi/tests/union_replace_test/README.md
dplore: Overall I think this is a good question, but I don't think it should block this PR.  If you feel this is important enough we can open an issue to make a change to ondatra.

feature/gnmi/tests/union_replace_test/README.md
dplore: Let's sync offline on the urgency of this and how to track it.

gprateekjn (approved):

feature/gnmi/tests/union_replace_test/README.md
dplore: I dug a little deeper into this.  In the implementation I used ondatra's RawAPIS().CLI() RunCommand interface which is defined in the [ondatra binding.](https://github.com/openconfig/ondatra/blob/975b1bf75c21a1480931ad248e70511153575c8f/binding/binding.go#L236-L241)  

So the binding can implement this however is needed for a given test environment.    I've updated the README to remove the reference to ssh.

rohit-rp (approved):

PR #5455, opened 2026-05-13
Implementation of Cisco Drop Counters Test

RishabhAgarwal-2001:


feature/platform/integrated_circuit/otg_tests/pipeline_counters_drop_test/pipeline_counters_drop_test.go
```diff
@@ -0,0 +1,423 @@
...
+					found = true
+					t.Logf("Counter %s (matched by substring %q) found with value: %d", k, dropReason, v)
+					break
+				}
+			}
+			if !found {
+				t.Errorf("Counter %s not found in telemetry", dropReason)
+			}
+		}
+	})
```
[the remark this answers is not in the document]
RishabhAgarwal-2001: We have already discussed and it is not possible to simulate these drops in FNT setup.

cc: @ram-mac

feature/platform/integrated_circuit/otg_tests/pipeline_counters_drop_test/pipeline_counters_drop_test.go
```diff
@@ -0,0 +1,423 @@
...
```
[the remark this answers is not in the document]
RishabhAgarwal-2001: We have tried several things to simulate this drop but none of them worked -
* Sending traffic to an unknown host
* Using disconnected VLANs to simulate lookup failure
Thus we are going with a simple telemetry check.

feature/platform/integrated_circuit/otg_tests/pipeline_counters_drop_test/metadata.textproto
```diff
@@ -6,3 +6,12 @@ plan_id: "gNMI-1.29"
...
 testbed: TESTBED_DUT_ATE_2LINKS
 tags: TAGS_TRANSIT
+
+platform_exceptions {
+  platform {
+    vendor: CISCO
+  }
+  deviations {
+    ciscoxr_vendordrop_ft: "ciscoxr-vendordrops-ft"
+  }
```
dplore, line 16: this is not the right pattern for deviations, see https://github.com/openconfig/featureprofiles/tree/main/internal/deviations#readme

feature/platform/integrated_circuit/otg_tests/pipeline_counters_drop_test/pipeline_counters_drop_test.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+
+	if len(nativePaths) == 0 {
+		t.Fatalf("No native paths found for functional translator %q", ftName)
+	}
+
+	var allUpdates []*gnmipb.Update
+
+	for _, nativePath := range nativePaths {
+		t.Logf("Initiating SubscribeRequest (ONCE) for native path: %v", nativePath)
```
dplore, line 169: why not use the ondatra function for this?  (gnmi.Get)

RishabhAgarwal-2001: These paths are still not available in the OC model and hence I am forced to take this route.

rohit-rp (changes requested):
Address gemini comments plz

rohit-rp:
@RishabhAgarwal-2001 add pass logs to b/512764000

feature/platform/integrated_circuit/otg_tests/pipeline_counters_drop_test/metadata.textproto
RishabhAgarwal-2001: Hi @dplore - I think we are using this deviation for the first use case described in the README - `Deviations may be created to use alternate OC or use CLI instead of OC to achieve the operational intent described in the README.`

These OC paths are still not supported by the vendor natively and hence we are using a deviation. With deviation added we are using translator to read these paths. Once these paths are natively supported by the vendor (and hence there is no longer a deviation), vendors can remove this deviation and test should continue to pass. 

Let me know if you see any holes in my understanding of the situation. Thanks!

RishabhAgarwal-2001:
> @RishabhAgarwal-2001 add pass logs to b/512764000

Addressed Gemini comments and have also added pass logs. 

Additionally, I have added comments in the test implementation to ensure the FPT-Gap analysis check does not fail

rohit-rp (approved):

PR #5464, opened 2026-05-14
credz: Force ED25519 key type for HIBA key generation

nkrishk:
Force ED25519 key type when generating HIBA CA, user, and host keys
via `hiba-ca.sh` in `createHibaKeysGen`.

Without an explicit `-t ed25519` argument, `hiba-ca.sh` delegates to
`ssh-keygen` which may default to a different key type (e.g., RSA).
This causes verification failures since the credentialz test
infrastructure expects ED25519 keys.

Changes:
- Pass `-t ed25519` to all three `hiba-ca.sh -c` invocations (CA,
  user, and host key generation) in `createHibaKeysGen`.


Chounoki (approved):

ksgireesha:
Hi @rohit-rp, could you please restart the check-approvals workflow on PR #[5464](https://github.com/openconfig/featureprofiles/pull/5464)? The Cisco team is blocked by it and they don't have the privileges to re-trigger it themselves

PR #5465, opened 2026-05-14
P200 | Health-1.1 fix

suredhar:
Skip the test on "No Cisco line card CPU base components found," because the P200 Mustang is a fixed box without line cards.


karthikeya-remilla:
Closing this PR as we are in the process of raising a more enhanced check for CPU components for this scenario.

PR #5466, opened 2026-05-14
[PF-1.4] Test fixes for ip_guev1_static_decap_subnet_range_test.go

davinsingh-arista:
[PF-1.4] Fix test by left shifting DSCP value correctly, verifying differently for pass through traffic

Problems:
1. GUE_Decapsulation_of_inner_IPv4_traffic_over_DECAP_subnet_range passes with the correct dscp value, because the we set and expect the correct dscp in the flow (innerIpHeader.Priority().Dscp().Phb().SetValue(innerDscpValue)). However, for IPV6, we set the TC rather than the DSCP (innerIpHeader.TrafficClass().SetValue), which requires left shifting the DSCP by 2. The IPV6 test is not left shifting by 2 when creating the flow, so we are getting the following error:
    ip_guev1_static_decap_subnet_range_test.go:622: ERROR: IPv6 DSCP and TTL value not preserved after Decap. Expected : DSCP - 32 , TTL - 49 Got : DSCP - 8 , TTL - 49

2. The last 2 tests for pass through (no decap should occur) are failing with the following error: ip_guev1_static_decap_subnet_range_test.go:626: ERROR: Could not find packet with matching inner source IP (198.51.100.1 or 2001:DB8:1::1) in capture 
The pass through test is calling verifyCaptureDscpTtlValue, which expects a decapsulated packet. We should not call verifyCaptureDscpTtlValue because pass through tests do not perform any decapsulation.

Fix:
1. Left shift the DSCP by 2 to correctly obtain the TC for creating the IPV6 flow
2. Rather than calling verifyCaptureDscpTtlValue for pass through tests, maybe we could call a new function verifyPassThroughOuterHeader to verify that no decapsulation has occurred.
4. Today, the number of packets is determined by the variable packetPerSecond. It might be more readable to have a variable fixedPackets for the number of packets being sent.

Passing run: 
[pf-1_4.log](https://github.com/user-attachments/files/27771373/pf-1_4.log)

### DIFF
--- a/feature/policy_forwarding/decapsulation/otg_tests/ip_guev1_static_decap_subnet_range/ip_guev1_static_decap_subnet_range_test.go
+++ b/feature/policy_forwarding/decapsulation/otg_tests/ip_guev1_static_decap_subnet_range/ip_guev1_static_decap_subnet_range_test.go
@@ -69,6 +69,7 @@ const (
 	innerTTL        = uint32(50)
 	outerTTL        = uint32(70)
 	srcPortValue    = 30000
+	fixedPackets    = 10000
 )
 
 var (
@@ -120,6 +121,7 @@ type testCase struct {
 	trafficDestIp     string
 	trafficShouldPass bool
 	verifyCounters    bool
+	isPassThrough     bool
 }
 
 func TestIpGueStaticDecapsulation(t *testing.T) {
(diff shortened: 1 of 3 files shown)


davinsingh-arista:
Duplicate of #5437

Issue #5445, opened 2026-05-12
FNT Gap Analysis is broken.


morrowc:
**Describe the bug**

```
Error: Gemini analysis failed: gemini api returned non-ok status 404: {
  "error": {
    "code": 404,
    "message": "models/gemini-2.5-pro is not found for API version v1beta, or is not supported for generateContent. Call ListModels to see the list of available models and their supported methods.",
    "status": "NOT_FOUND"
  }
}
```

looks like some gemini bits need an update?
I attempted to fix this with just swapping:
`const geminiAPIURL = "https://generativelanguage.googleapis.com/v1beta/models/%s:generateContent?key=%s"`
with:
`const geminiAPIURL = "https://generativelanguage.googleapis.com/v1/models/%s:generateContent?key=%s"`

that appears to require some JSON changes in the reply catching.. nope, not today satan!

morrowc:
https://github.com/openconfig/featureprofiles/pull/new/Issue5445 <- PR5446

morrowc:
Maybe fixed now.

morrowc:
This is still broken

dplore:
@singhavnish2516

morrowc:
https://github.com/openconfig/featureprofiles/pull/5450 - out to avnish/darren.

(the robot suggested we flip to the v1alpha version.. sure, maybe? :)

morrowc:
and that PR has some confligts.. so.. meh.
Anyways.. ideally avnish can find a solution :)

morrowc:
Hey there folks @chitadi has sent a few PR for this and it doesn't appear we're anywhere close to fixed?

https://github.com/openconfig/featureprofiles/pull/5409 still shows the failure:

```
Error: Gemini analysis failed: http request failed: Post "https://generativelanguage.googleapis.com/v1beta/models/gemini-3.1-pro-preview:generateContent?key=***": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
Error: Gemini analysis failed: gemini api returned non-ok status 503: {
  "error": {
    "code": 503,
    "message": "This model is currently experiencing high demand. Spikes in demand are usually temporary. Please try again later.",
    "status": "UNAVAILABLE"
  }
}
```

perhaps this is a momentary failure?

morrowc:
```
[...]

PR #5467, opened 2026-05-14
Update tools/gap-analyzer Gemini model from '3.1-pro-preview' to '3-flash-preview'

chitadi:
3.1-pro-preview has high demand due to which the tool would exit with a failure. We have decided to move to a lighter model with less demand as a replacement


singhavnish2516 (approved):

dplore (approved):

dplore:
This may fix  #5445

PR #5468, opened 2026-05-14
 TestID-16.4: Fixes for export policy, drain policy, and prefix validation

nakulnayak:
This PR adds a few fixes for export policy, drain policy, and prefix validation in the gRIBI to BGP redistribution fpt.
 - Apply BGP export policy to TEST_VRF instead of DEFAULT network instance, so the policy is attached to the actual BGP neighbor
 - Use gnmi.Replace instead of gnmi.Update when prepending drain policy to the export chain, since Update merges with the existing leaf-list and preserves the original policy ordering
 - Match validated prefix against the actual gRIBI route (198.51.100.1/32) instead of the aggregate prefix-set (198.51.100.0/26). This is similar to what is done on lines 641-643 in the function  "TestGRIBIBGPRedistribution".

### DIFF
--- a/feature/gribi/otg_tests/gribi_to_bgp_redistribution/gribi_to_bgp_redistribution_test.go
+++ b/feature/gribi/otg_tests/gribi_to_bgp_redistribution/gribi_to_bgp_redistribution_test.go
@@ -245,10 +245,10 @@ func configureRoutingPolicies(t *testing.T, dut *ondatra.DUTDevice) {
 	// ----------------------------
 	// ATTACH EXPORT POLICY TO BGP
 	// ----------------------------
-	policy := root.GetOrCreateNetworkInstance(deviations.DefaultNetworkInstance(dut)).GetOrCreateProtocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP").GetOrCreateBgp().GetOrCreateNeighbor(atePort2.IPv4).GetOrCreateAfiSafi(oc.BgpTypes_AFI_SAFI_TYPE_IPV4_UNICAST).GetOrCreateApplyPolicy()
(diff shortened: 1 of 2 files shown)


feature/gribi/otg_tests/gribi_to_bgp_redistribution/gribi_to_bgp_redistribution_test.go
```diff
@@ -511,7 +511,10 @@ func validateRoutingPolicyV4(t *testing.T, dut *ondatra.DUTDevice, ate *ondatra.
 	for _, bgpPrefix := range bgpPrefixes {
 		t.Logf("Received prefix %s/%d", bgpPrefix.GetAddress(), bgpPrefix.GetPrefixLength())
 
-		if bgpPrefix.Address != nil && bgpPrefix.GetAddress() == v4RoutePrefix && bgpPrefix.PrefixLength != nil && bgpPrefix.GetPrefixLength() == v4RoutePrefixLen {
+		parts := strings.Split(routePrefix, "/")
+		prefixAddr := parts[0]
+		prefixLen, _ := strconv.Atoi(parts[1])
```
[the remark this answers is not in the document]
nakulnayak: agreed, done

nakulnayak:
Attached is a passing test log
[TestID-16.4-passingLog.txt](https://github.com/user-attachments/files/27978682/TestID-16.4-passingLog.txt)

singhh07 (approved):

feature/gribi/otg_tests/gribi_to_bgp_redistribution/metadata.textproto
```diff
@@ -14,7 +14,7 @@ platform_exceptions: {
     default_network_instance: "default"
     omit_l2_mtu: true
     interface_config_vrf_before_address: true
-    table_connections_unsupported: true
+    table_connections_unsupported: false
```
AmrNJ: Why is the deviation not removed?

nakulnayak: Thanks for reviewing. I'd missed this, just fixed it @AmrNJ

PR #5476, opened 2026-05-15
ACCTZ-7.1: gNSI.acctz.v1 (Accounting) Test Accounting Authentication Failure - Multi-transaction

mihaitomaro:
Readme Location: https://github.com/openconfig/featureprofiles/blob/main/feature/gnsi/acctz/tests/AccountingAuthenFailMulti/README.md

Attached logs here: https://partnerissuetracker.corp.google.com/u/2/issues/513266214


feature/gnsi/acctz/tests/AccountingAuthenFailMulti/accounting_authen_fail_multi_test.go
```diff
@@ -0,0 +1,486 @@
...
+
+		// Verify ip_proto is 6 (TCP) for SSH.
+		if ipProto := sessionInfo.GetIpProto(); ipProto != 6 {
+			t.Errorf("ip_proto is %d, want 6 (TCP) for SSH: %s", ipProto, prettyPrint(record))
+		}
+
+		// Verify local_address and local_port. On Arista, local = DUT (server), remote = client.
+		if la := sessionInfo.GetLocalAddress(); la != "" {
+			if la != expectedRemoteAddr {
+				t.Logf("local_address is %q (DUT), expected SSH target %q: platform-dependent interpretation", la, expectedRemoteAddr)
```
sam-homa: should this be t.Errorf() instead?

feature/gnsi/acctz/tests/AccountingAuthenFailMulti/accounting_authen_fail_multi_test.go
```diff
@@ -0,0 +1,486 @@
...
+			}
+		} else {
+			t.Logf("local_address not populated by DUT (platform-dependent)")
```
sam-homa: same here and the ones below, should they all be Logf?

feature/gnsi/acctz/tests/AccountingAuthenFailMulti/accounting_authen_fail_multi_test.go
```diff
@@ -0,0 +1,486 @@
...
+	t.Logf("Found %d authentication failure records (expected %d scenarios)", len(gotRecords), len(connMetas))
+	for i, r := range gotRecords {
+		t.Logf("  Record %d: user=%q, authn_type=%v, authn_status=%v, cause=%q",
+			i, r.GetSessionInfo().GetUser().GetIdentity(),
+			r.GetSessionInfo().GetAuthn().GetType(),
+			r.GetSessionInfo().GetAuthn().GetStatus(),
+			r.GetSessionInfo().GetAuthn().GetCause())
+	}
+
+	// Verify each authentication failure record.
```
sam-homa, line 452: this verification logic is too large and difficult to read. it'd be easier to read if you could split to multiple verify...() functions

internal/security/acctz/acctz.go
sam-homa, line 459: Is it common in this directory to use underscore for types or other IDs?

it's not a Golang best practice

sam-homa, line 518: Is this commented function serving a purpose? or should we clean it up?

feature/gnsi/acctz/tests/AccountingAuthenFailMulti/accounting_authen_fail_multi_test.go
mihaitomaro: Correct. fixed this in a commit.

feature/gnsi/acctz/tests/AccountingAuthenFailMulti/accounting_authen_fail_multi_test.go
mihaitomaro: Correct. fixed this in a commit.

feature/gnsi/acctz/tests/AccountingAuthenFailMulti/accounting_authen_fail_multi_test.go
mihaitomaro: acctz: ACCTZ-7.1: refactor record verification into helper functions

Extract inline verification loop into three helper functions:
verifySessionAuthn (session status, authn type/status/cause, user identity, and timestamp),
verifyNetworkFields (ip_proto, local/remote address and port, channel_id),
verifyServiceRequest (CmdService/GrpcService type and field constraints).

internal/security/acctz/acctz.go
mihaitomaro: Same here

feature/gnsi/acctz/tests/AccountingAuthenFailMulti/accounting_authen_fail_multi_test.go
```diff
@@ -0,0 +1,469 @@
...
+// invalidCredentials defines a set of invalid credentials to use for authentication failure tests.
+type invalidCredentials struct {
+	description string
+	username    string
+	password    string
+	authnType   acctzpb.AuthnDetail_AuthnType
+}
+
+// InvalidCredentialSets returns the set of invalid password credential combinations to test.
+func InvalidCredentialSets() []invalidCredentials {
```
Chounoki: Prefer to return `[]*invalidCredentials` to avoid unnecessary copying.

feature/gnsi/acctz/tests/AccountingAuthenFailMulti/accounting_authen_fail_multi_test.go
```diff
@@ -0,0 +1,469 @@
...
+			t.Errorf("CmdService cmd_args should be omitted for auth failure record: %s", prettyPrint(record))
+		}
+		if authz := sr.CmdService.GetAuthz(); authz != nil && authz.GetStatus() != acctzpb.AuthzDetail_AUTHZ_STATUS_UNSPECIFIED {
+			t.Errorf("CmdService authz should be omitted for auth failure record (status=%v): %s", authz.GetStatus(), prettyPrint(record))
+		}
+	case *acctzpb.RecordResponse_GrpcService:
+		if sr.GrpcService.GetServiceType() == acctzpb.GrpcService_GRPC_SERVICE_TYPE_UNSPECIFIED {
+			t.Errorf("GrpcService service_type is unspecified for record: %s", prettyPrint(record))
+		}
+		t.Errorf("Got GrpcService record for SSH auth failure (unexpected): %s", prettyPrint(record))
```
Chounoki: Not sure whether this error is unconditional.

feature/gnsi/acctz/tests/AccountingAuthenFailMulti/accounting_authen_fail_multi_test.go
mihaitomaro: Thanks for the suggestion. I looked at switching to a slice of pointers, but in this case I prefer keeping a value slice for clearer value semantics and to avoid introducing pointer/nil handling where we do not need shared mutation.
To address the copying concern in the range loop, I updated iteration to index-based access so we avoid per-iteration struct copies while keeping the existing API and behavior.

feature/gnsi/acctz/tests/AccountingAuthenFailMulti/accounting_authen_fail_multi_test.go
mihaitomaro: Thanks for calling this out. You are right that the previous logic could emit two errors for the same GrpcService record path. I updated the check to keep GrpcService unconditionally unexpected for this SSH-auth-failure test, but consolidated reporting into a single primary error that includes service_type. This keeps the intent unchanged and avoids duplicate/noisy failures. I left the authz omission check in place as an independent validation.

internal/security/acctz/acctz.go
mihaitomaro: This naming convention already exists throughout the codebase, so I didn't introduce it in this change. Renaming these identifiers could impact other tests or require updates across multiple files. If the recommendation is to remove underscores and align with the Go style guide, I can make that change here and update all corresponding references as well. Please let me know if you'd like me to proceed.

mihaitomaro: This commented code already existed in the file, so I didn't remove it as part of this change. As far as I can tell, removing it shouldn't impact any other tests. If you'd prefer to clean it up, I can remove it in this change.

feature/gnsi/acctz/tests/AccountingAuthenFailMulti/accounting_authen_fail_multi_test.go
Chounoki: I guess there is some misunderstanding here.

1. Since you always return a valid slice in this function, there will never be a concern of nil pointer.
2. The copying happens whenever you call this `InvalidCredentialSets()` function, not just during the iteration.

I think in this case, it is perfectly safe to change the return to `[]*invalidCredentials`, while no other code modification is needed.

I.e.:
```
	return []*invalidCredentials{
		&invalidCredentials{
			description: "empty username",
			username:    "",
			password:    "somepassword",
			authnType:   acctzpb.AuthnDetail_AUTHN_TYPE_PASSWORD,
		},
		......
		......
```

mihaitomaro: I updated InvalidCredentialSets() to return []*invalidCredentials and adjusted usage accordingly so the loop consumes pointer elements directly.

Chounoki (approved):

ksgireesha:
Hi @AmrNJ  - Friendly ping! As agreed in the tracking bug ([b/514874974](https://b.corp.google.com/issues/514874974)), the plan is to merge this PR as-is and push the `acctz.go` fixes in a separate follow-up PR.

Could you please provide the final code owner approval for featureprofiles-approvers so we can get this merged? 

Thanks

morrowc (approved):

ksgireesha:
@rohit-rp , review validate, approve and merge this PR please.

rohit-rp (approved):

PR #5484, opened 2026-05-18
Add test enrollz-1

ijernoiu:
Issues identified: https://partnerissuetracker.corp.google.com/issues/514403340
Logs location: https://partnerissuetracker.corp.google.com/issues/514436481


topologies/atedut_4_enrollz.binding
```diff
@@ -0,0 +1,126 @@
...
+    cli: "transport gnmi default"
+    cli: "service enrollz"
+    cli: "service attestz"
+    cli: "bash /usr/bin/tpm2_readPubPpk > /tmp/ppk_pub.pem"
+  }
+ 
+  gnmi {
+    target: "dut-hostname:6030"
+    mutual_tls: true
+    trust_bundle_file: "path-to-test-file/enrollz.crt"
```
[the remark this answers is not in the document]
ijernoiu: The binding file attached in the PR is just for reference, real values must be set when running the test

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
```diff
@@ -0,0 +1,1469 @@
...
+	gnmiTarget := introspect.DUTDialer(t, dut, introspect.GNMI).DialTarget
+	return grpc.NewClient(gnmiTarget,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithPerRPCCredentials(insecureRPCCreds{username: username, password: password}),
+	)
+}
+
+func restoreRunningConfig(t *testing.T, dut *ondatra.DUTDevice, fileName string) {
+	t.Helper()
+	conn, err := insecureGNMIClient(t, dut)
```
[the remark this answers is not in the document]
ijernoiu: Refactored function and moved to cfgplugins

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
```diff
@@ -0,0 +1,1376 @@
...
+func (d *hmacEnrollDeps) GetIakCert(ctx context.Context, req *epb.GetIakCertRequest) (*epb.GetIakCertResponse, error) {
+	return d.enrollzC.GetIakCert(ctx, req)
+}
+
+func (d *hmacEnrollDeps) RotateOIakCert(ctx context.Context, req *epb.RotateOIakCertRequest) (*epb.RotateOIakCertResponse, error) {
+	return d.enrollzC.RotateOIakCert(ctx, req)
+}
+
+func (d *hmacEnrollDeps) RotateAIKCert(ctx context.Context, opts ...grpc.CallOption) (epb.TpmEnrollzService_RotateAIKCertClient, error) {
+	return d.enrollzC.RotateAIKCert(ctx, opts...)
```
yun2kim: This RPC is for TPM1.2 so TPM2.0 test shouldn't depend on this RPC.

topologies/atedut_4_enrollz.binding
```diff
@@ -0,0 +1,126 @@
...
+# DUT and the ATE, unless overridden by a specific device or protocol.
+options {
+  username: "username"
+  password: "password"
+}
+
+duts {
+  id: "dut"
+  name: "dut-hostname"  # Change this to the device hostname.
+  vendor: ARISTA
```
yun2kim: To keep our test suite generic, it would be great if we could avoid exposing vendor-specific data in this test case. Could you help verify that no vendor-specific information is included in the PR changes? 
I also noticed the port number `6030` in the target below (unsure if this is also specific to Arista). Is there a particular dependency or reason we need to include these specific details?

yun2kim (commented):
Submitting an initial review about *.binding file. Since the change is large, I'll submit another review if needed after reviewing the file that contained HMAC flow test.

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
```diff
@@ -0,0 +1,1376 @@
...
+		{
+			"enrollz-1.3-InvalidClientCertificate",
+			"Enrollz service presents an invalid client certificate",
+			func(t *testing.T) error {
+				badClient, err := enrollzClientWithTLS(t, dut, invalidTLSConfig(t))
+				if err != nil {
+					t.Logf("connection rejected at dial time (expected): %v", err)
+					return nil
+				}
+				_, err = badClient.GetControlCardVendorID(ctx, &epb.GetControlCardVendorIDRequest{
```
yun2kim, line 314: Why calling `GetControlCardVendorID` with a bad client instead of directly notifying that expected error didn't happen?

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
```diff
@@ -0,0 +1,1376 @@
...
+				})
+				return requireRejectedRequest(t, err)
+			},
+		},
+		{
+			"enrollz-1.20-UnsupportedKeyTemplate",
+			"GetIdevidCsrRequest with unsupported KeyTemplate → STATUS_UNSUPPORTED",
+			func(t *testing.T) error {
+				resp, err := enrollzC.GetIdevidCsr(ctx, &epb.GetIdevidCsrRequest{
+					ControlCardSelection: activeCard(),
```
yun2kim, line 546: Nit: Can you add the valid Key as well in the request (e.g., `Key: epb.Key_KEY_EK`)? To show that everything else is correct except this invalid key template. Similar in other places too.

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
```diff
@@ -0,0 +1,1376 @@
...
+			},
+		},
+		{
+			"enrollz-1.21-InvalidControlCard",
+			"RotateOIakCert with invalid control card → rollback",
+			func(t *testing.T) error {
+				_, err := enrollzC.RotateOIakCert(ctx, &epb.RotateOIakCertRequest{
+					Updates: []*epb.ControlCardCertUpdate{{
+						ControlCardSelection: &cpb.ControlCardSelection{ControlCardId: &cpb.ControlCardSelection_Slot{Slot: "invalid-slot-xyz"}},
+						OiakCert:             "fake", OidevidCert: "fake",
```
yun2kim: Nit: There should be a newline between the two fields. Below cases have the same format.

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
```diff
@@ -0,0 +1,1376 @@
...
+				_, err := enrollzC.RotateOIakCert(ctx, &epb.RotateOIakCertRequest{
+					Updates: []*epb.ControlCardCertUpdate{{
+						ControlCardSelection: activeCard(),
+						OiakCert:             "fake", OidevidCert: "fake",
+					}},
+				})
+				if err := requireRejectedRequest(t, err); err != nil {
+					return err
+				}
+				return verifyCerts(t, ctx, dut, enrollzC, nil, baseline.hasStandby, &baseline)
```
yun2kim: Can you help me understand this part? Why still calls `verifyCerts` for this case? I first thought it's because of the rollback so the previous case makes sense, but looks like it's used in non-rollback cases so maybe not?

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
```diff
@@ -0,0 +1,1376 @@
...
+					Updates:      []*epb.ControlCardCertUpdate{{ControlCardSelection: activeCard(), OiakCert: oiak, OidevidCert: oidevid}},
+					SslProfileId: defaultSSLProfile,
+				})
+				if err := requireRejectedRequest(t, err); err != nil {
+					return err
+				}
+				return verifyCerts(t, ctx, dut, enrollzC, nil, baseline.hasStandby, &baseline)
+			},
+		},
+		{
```
yun2kim, line 699: _Please correct me if I'm missing something:_
It appears this successful test case uses the `oIAKOnlyForStandbyCAClient` to achieve a scenario where the **active** control card is issued both oIAK and oIDevID certificates, but the **standby** control card is only issued an oIAK. The enrollment process aims to provision both oIAK and oIDevID to all control cards, including the standby. 

Or, is this a failure case?

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
```diff
@@ -0,0 +1,1376 @@
...
+				}
+				if err := verifyCerts(t, ctx, dut, enrollzC, ownerCA, hasStandby, nil); err != nil {
+					return err
+				}
+				t.Log("Post-reboot certificates verified")
+				return nil
+			},
+		},
+		{
+			"enrollz-1.31-RebootAfterGetIdevidCsrAndReEnroll",
```
yun2kim, line 754: Can you add the same idempotency check for multiple-cards enrollment case as well?

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
ijernoiu: This function is defined in order to implement EnrollSwitchWithHMACChallengeInfraDeps interface which is used in the enrollment logic. It is not used by the test workflow

topologies/atedut_4_enrollz.binding
ijernoiu: The binding file was submitted for reference, especially because of the test prerequisites like the TLS configuration. There is no dependency on specific ports or other setup related details.

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
ijernoiu: The grpc client does not perform a check of the certificates at initialization time and no error is returned. The validation can be observed when sending the  first RPCs through the client

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
ijernoiu: For the test cases where the enrollment request is expected to be rejected the verifyCerts call is intended to check that the certificates were not modified during the operation even if the error is returned. For successful enrollment test cases the function checks the validity of the certificates. I can split verifyCerts into 2 separate functions if it can improve code readability.

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
ijernoiu: This is a success test case in which the DUT should accept an enrollment request with a missing oIDevID cert for the standby control card. This type of request was simulated using the oIAKOnlyForStandbyCAClient which returns an empty oIDevID certificate. The full enrollment with both oIAK and oIDevID for all cards is part of subtests 1.1 and 1.2

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
ijernoiu: Can you give some more details?  There is already a verifyCerts call to check successful enrollment cert validity

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
yun2kim: Okay, but is there any reason why are we keeping this even though it is not used as well as this RPC is not for TPM 2.0?

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
yun2kim: Can you explain more about this part? My understanding of gRPC mTLS is that certificate validation is a fundamental part of the TLS handshake, which occurs during the initial grpc.Dial or connection establishment. If there's an issue with the TLS configuration (e.g., invalid configuration, untrusted certificates), the Dial call itself should fail, preventing a connection from being established in the first place. I guess maybe the reason the test case can't catch the error during connection establishment is because `InsecureSkipVerify` is set true?

topologies/atedut_4_enrollz.binding
yun2kim: Yes but can you redact the vendor information? Even the name of the vendor as well. We prefer not to put any vendor information in the open-source testing.

internal/cfgplugins/dut_initialize.go
```diff
@@ -1357,3 +1358,74 @@ func EnableHardwareCounters(t *testing.T, dut *ondatra.DUTDevice, feature string
 		t.Fatalf("Unsupported vendor: %v", dut.Vendor())
 	}
 }
+
+// BackUpConfig saves the current running configuration from the DUT into the specified file on local flash storage.
+func BackUpConfig(t *testing.T, dut *ondatra.DUTDevice, fileName string) {
+	t.Helper()
+	switch dut.Vendor() {
+	case ondatra.ARISTA:
```
yun2kim, line 1513: Here and line 1387, I see `case ondatra.ARISTA` which I believe it is due to the .binding file having `ARISTA` as the value for `vendor` field. Can you remove this and make it generic?

yun2kim (commented):
initial follow up

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
yun2kim: Yes can you make the change? Thanks

topologies/atedut_4_enrollz.binding
ijernoiu: Information cleaned-up in commit #5aebe50

internal/cfgplugins/dut_initialize.go
ijernoiu: In the reboot testcases (1.30 and 1.31) I need to persist the DUT config after rebooting (certificates are persistent without any action from the test). This code was added in the cfgplugins package and uses cli commands which are vendor specific to persist and restore running config, especially the SSL profile part needed to re-establish TLS. This backup is not possible through gnmi get/set.

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
ijernoiu: The client is constructed using the grpc.NewClient function which creates an idle grpc channel with no I/O operations at creation time. The connection is initiated when first sending RPCs through the client or when manually calling Connect. The test is expecting the TLS error to occur when sending the GetControlCardVendorID RPC which should also connect and validate the client. I can add an intermediary step to manually connect and wait for client state changes before the RPC is sent (it should go into error state once the certs are validates)

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
ijernoiu: The function must exist in order to implement the biz.EnrollzDeviceClient interface (embedded deep in EnrollSwitchWithHMACChallengeInfraDeps struct) and have the hmacEnrollDeps struct used in the enrollment workflow. I have edited the body of the function to return error instead of further delegating the call to the enrollz client to be more clear about the purpose of the function

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
yun2kim: Thank you for explaining about it. I don't think you have to make changes. Looks like the current way is recommended if using `grpc.NewClient` in gRPC-Go (v1.60+) which is the latest.

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
yun2kim: I see, I understood now. Can you just then slightly change the test case name for this? `MixedCardEnrollment` is a bit confusing. Maybe like `~ActiveFullStandbyOIakOnly`

feature/security/attestz/tests/enrollz_tpm20_hmac/enrollz_tpm20_hmac_test.go
yun2kim: Sry for late update on this thread, I completely missed it. 

Actually, I re-read this part again and noticed `enrollSels` covers both **no stand by card** and **both active and standby cards** cases. So resolving this thread.

topologies/atedut_4_enrollz.binding
```diff
@@ -0,0 +1,125 @@
...
+# DUT and the ATE, unless overridden by a specific device or protocol.
+options {
+  username: "username"
+  password: "password"
+}
+
+duts {
+  id: "dut"
+  name: "dut-hostname"  # Change this to the device hostname.
+
```
yun2kim, line 18: Nit: I think I wasn't clear enough here. Can you add like `vendor: "VENDOR" # Specify the hardware vendor of the DUT`? Redacting a specific vendor name (e.g., Arista) but to still give some impression that vendor can be specified here.

internal/cfgplugins/dut_initialize.go
yun2kim: Verified that this is acceptable. Thanks for the explanation here as well.

topologies/atedut_4_enrollz.binding
ijernoiu: From what I've seen in the other sample binding files in the topologies folder, no one mentions the vendor field.

yun2kim:
LGTM. Thanks,

betuls (approved):

ksgireesha:
@rohit-rp  - The branch is now fully synced and all checks are passing. Since we already have the Security approval from Betul and the technical LGTM from Yuni, could you please provide the final code owner approval and merge PR #[5484](https://github.com/openconfig/featureprofiles/pull/5484)?

rohit-rp (approved):

Chounoki (approved):

PR #3313, opened 2024-07-19
CERTZ2.1 server_certificates

priyacj:
CERTZ2.1 server_certificates testcase
CERTZ2.2 server_certificates testcase

PR #5489, opened 2026-05-19
Certz2.1 updated the timeout and changed the error response as fatal

priyacj:
Prior commit : https://github.com/openconfig/featureprofiles/pull/3313 (ref)


priyacj:
closing the merge request as its available on https://github.com/openconfig/featureprofiles/blob/main/feature/gnsi/certz/tests/server_certificates/server_certificates_test.go

PR #5500, opened 2026-05-22
Add watch for ACL counter updates

ijernoiu:
Add watch for ACL counter matched packets instead of one-time check


ram-mac (approved):
Validated the test with these changes and the test is passing; Testlog is available in bug 498147637

ram-mac:
@ijernoiu - Have validated and test is now passing. Please resolve issues in this PR, so it can be merged ASAP

ijernoiu:
Merged latest changes

PR #5502, opened 2026-05-25
Increase HTTP client timeout to 300 seconds

chitadi:
Ensure any delay from the Gemini API, which is currently seeing extremely high traffic so requests might get queued, is accounted for.


singhavnish2516 (approved):
LGTM

PR #5520, opened 2026-05-28
ACL-1.2 Add Strata platform support to acl_update_test

deva-arista:
acl_update_test for vendor "ARISTA" was set for "sand" platform. Strata platform (e.g.
DCS-7050CX4M-48D8-F) is slower in updating ACL and requires a higher maxDroppedPackets value.

This change adds the hardware_model_regex to differentiate between different ARISTA platforms.
* (M) feature/acl/otg_tests/acl_update_test/acl_update_test.go
  - Use effectiveMaxDroppedPackets based on maxDroppedPackets and dut.Model()
* (M) internal/cfgplugins/dut_initialize.go
  - Skip TCAM profile initialization for strata dut model


deva-arista:
b/493609997

PR #5525, opened 2026-06-01
Otg gnpsi-1 update for Cisco run

ijernoiu:


proto/metadata.proto
```diff
@@ -1361,6 +1361,8 @@ message Metadata {
     // https://partnerissuetracker.corp.google.com/issues/497757203
     bool dhcp_relay_oc_unsupported = 431;
 
+    bool missing_sflow_gnpsi_export = 432;
```
ram-mac: @ijernoiu - We need to have a bug reference with Cisco for this deviation and update the PR accordingly.

internal/deviations/deviations.go
```diff
@@ -2226,3 +2226,7 @@ func GnoiBgpGracefulRestartUnsupported(dut *ondatra.DUTDevice) bool {
 func DhcpRelayOcUnsupported(dut *ondatra.DUTDevice) bool {
 	return lookupDUTDeviations(dut).GetDhcpRelayOcUnsupported()
 }
+
+func MissingSflowGnpsiExport(dut *ondatra.DUTDevice) bool {
```
ram-mac, line 2260: @ijernoiu - We need to have a bug reference with Cisco for this deviation and update the PR accordingly.
Also we need to add the details in the [deviations.textproto](https://github.com/open-traffic-generator/featureprofiles/blob/main/internal/deviations/deviations.textproto) for the same.

ram-mac (changes requested):

ijernoiu:
Updated the deviation details

ram-mac:
@ijernoiu - I have validated and the test is not passing. I have attached the logs to the bug 510677651 for reference.

PR #5526, opened 2026-06-01
Add test acctz-9.1

ijernoiu:
Logs at: https://partnerissuetracker.corp.google.com/issues/517964850


internal/security/acctz/acctz.go
```diff
@@ -1622,6 +1639,130 @@ func SendShellCommand(t *testing.T, dut *ondatra.DUTDevice, staticBinding bool)
...
+func enableAccountingStartStop(t *testing.T, dut *ondatra.DUTDevice) {
+	var cliCommand string
+	switch dut.Vendor() {
+	case ondatra.CISCO:
+		cliCommand = "aaa accounting commands default start-stop local"
+	case ondatra.ARISTA:
+		cliCommand = "aaa accounting commands all default start-stop logging"
+	}
+	helpers.GnmiCLIConfig(t, dut, cliCommand)
+}
```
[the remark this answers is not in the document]
ijernoiu: The vendor switches in the acctz helper are needed as the service configuration is very specific to each vendor. They do not represent actual deviations

feature/gnsi/acctz/tests/AccountingPrivEscalation/accounting_priv_escalation_test.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+
+		if !resp.record.Timestamp.AsTime().After(startTime) {
+			t.Logf("Skipping record: timestamp %v not after start time %v", resp.record.Timestamp.AsTime(), startTime)
+			continue
+		}
+
+		foundUser := false
+		userIdentity := resp.record.GetSessionInfo().GetUser().GetIdentity()
+		for _, rec := range records {
```
Chounoki: I don't know the details about `records`, but you are searching through the entire records for user, yet later you do the `cmp.Diff` against `records[recordIdx]` only.

Doesn't that mean the user must present in `records[recordIdx]` instead of other indices?

feature/gnsi/acctz/tests/AccountingPrivEscalation/accounting_priv_escalation_test.go
```diff
@@ -0,0 +1,303 @@
...
+		if err == nil && level > 0 {
+			return level
+		}
+		return 0
+	}
+}
+
+func verifyReportedString(t *testing.T, field, got, want string) {
+	t.Helper()
+	if got == "" {
```
Chounoki: Do we also need to check `want`?
Right now as long as `got` is blank, it is treated as success (log only) regardless of `want`.

feature/gnsi/acctz/tests/AccountingPrivEscalation/accounting_priv_escalation_test.go
```diff
@@ -0,0 +1,303 @@
...
+		return
+	}
+	if want != "" && got != want {
+		t.Errorf("%s mismatch: got %q, want %q", field, got, want)
+	}
+}
+
+func verifyReportedUint32(t *testing.T, field string, got, want uint32) {
+	t.Helper()
+	if got == 0 {
```
Chounoki: Same as above, `want` is ignored and returns as success.

feature/gnsi/acctz/tests/AccountingPrivEscalation/accounting_priv_escalation_test.go
ijernoiu: Added error when got is empty and want is not empty

feature/gnsi/acctz/tests/AccountingPrivEscalation/accounting_priv_escalation_test.go
ijernoiu: Removed loop through expected records, check only if received record corresponds to the expected record at recordIdx

feature/gnsi/acctz/tests/AccountingPrivEscalation/accounting_priv_escalation_test.go
Chounoki: I mean even after your update, it still returns success, just with a logging.
Shouldn't we report error (i.e. t.Errorf()) instead of return when got == "" && want != ""?

feature/gnsi/acctz/tests/AccountingPrivEscalation/accounting_priv_escalation_test.go
Chounoki: Same here, maybe we should call t.Errorf() when got == 0 && want != 0.
Is there any reason why we are letting such case pass through?

feature/gnsi/acctz/tests/AccountingPrivEscalation/accounting_priv_escalation_test.go
ijernoiu: Please check commit 9bcf424, the code is failing the test if got is empty and want is not empty

```
if got == "" {
	if want != "" {
		t.Errorf("%s is not populated; want %q", field, want)
	} else {
		t.Logf("%s is not populated", field)
	}
	return
}
```

Is there a problem with the approach?

Chounoki: > the code is failing the test if got is empty and want is not empty

I think we should probably check why it fails (i.e why got is empty yet want is not empty) instead of just letting it pass, because it could actually be a bug of the code, since "got" should always be equal to "want".

W can't blindly assume the code is always correct and twist the test checking logic to fit the code.

ijernoiu: I have simplified the validation as want will never be empty (the expected response is manually constructed and validated previously). If there is an error with the received record it is checked before in the test

feature/gnsi/acctz/tests/AccountingPrivEscalation/accounting_priv_escalation_test.go
ijernoiu: Same approach as for the string validation (from the comments above)

Chounoki (approved):

ksgireesha:
@AmrNJ  - Friendly ping! Could you please review the discrepancy noted in the tracking bug and advise on how we should proceed?

@ijernoiu  - It looks like the `Test Registry Sync Check` is pending/stuck. Could you please check on this so we are clear to merge once the final approval is provided? Thanks!

ksgireesha:
@rohit-rp - The branch is now fully synced and all 19 CI checks are passing. Since we already have the Security approval from @Chounoki , could you please provide the final code owner approval (featureprofiles-approvers) and merge PR #[5526](https://github.com/openconfig/featureprofiles/pull/5526)?

rohit-rp (approved):

PR #5529, opened 2026-06-02
Update fpcFromPort function to use gNMI lookup

priyaranjannanda:
- Updated the fpcFromPort function to use gNMI lookup for component details instead of regex pattern.


ram-mac (approved):
Validated the PR and the test is passing; see b/530109126 for details.

PR #5532, opened 2026-06-03
add deviation to skip authz verification of gnoi ping rpc

mudhireddy:
policy based gNSI Authz tests rotate the policy and does the verification by sending Probe request for various rpcs as well as executing the actual rpcs. In case of GnoiSystemPing, a bug in parsing zone identifiers in IPV6 address is causing a failure and that ultimately fails Authz tests. Add a deviation to skip that specific rpc in Authz verification. Once the bug is fixed and delivered this can go away


navaneethyv:
Team, requesting you to close the conflicts so that we can move ahead.

navaneethyv (changes requested):
Conflicting changes, please fix and also the changes.

PR #5541, opened 2026-06-07
gNMI-1.21: Fix BGP max-prefix and use chassis-level utilization for A…

prakashbadri-arista:
…rista

Two issues caused consistent test failures on Arista DUTs:

1. BGP peer stuck in permanent Idle after hitting the default max-routes limit (256K) from a prior test run. Fixed by setting max-prefixes to 0 (no limit) using the Arista-specific augmented path via DeviationAristaBGPNeighborMaxPrefixes, following the same pattern as AFT-1.1.

2. Utilization was read from per-IC component paths, which includes fabric ASICs that have no FIB resources and always report 0%. Changed Arista to use chassis-level aggregation across all Routing/Resource* and Routing2/Resource* entries under Component("Chassis").Chassis(). Resources that don't exist on a given platform are skipped via gnmi.Lookup.


AmrNJ:
@prakashbadri-arista Please fix Gemini comments.

PR #5557, opened 2026-06-11
ACL-1.2: Refactor verifyACLCounters for eventual consistency

rolandphung:
Change verifyACLCounters to wait for the expected counters to align instead of sampling the first counter that was successful. The device needs some time to complete updating the counter.

Also add `SkipACLCountersVerificationDuringUpdate` deviation to `verifyFlowStatistics` during the make before break update test case.


rolandphung:
Bump @rohit-rp and @alokmtri-g. This change shouldn't be too big, and it will help pass rates.

PR #5523, opened 2026-05-29
Adding test for AFT-6.2

Steubert:
**README:** https://github.com/openconfig/featureprofiles/blob/main/feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering_dualstack/README.md

**Issues:** 
https://partnerissuetracker.corp.google.com/u/1/issues/514565554
https://partnerissuetracker.corp.google.com/u/1/issues/517809756

**Logs:** https://partnerissuetracker.corp.google.com/u/1/issues/517838946

PR #5217, opened 2026-03-11
Add AFT filter README

ElodinLaarz:
Create initial READMEs for testing out the AFTs prefix filtering feature.

[OpenConfig PR for model](https://github.com/openconfig/public/pull/1441)

This pull request adds new test plans and metadata for AFT prefix filtering, expanding coverage to dual-stack (IPv4/IPv6) filtering and dynamic updates to prefix sets. The new documentation describes test setups, validation steps, and OpenConfig model coverage for these scenarios.

**New test plans and documentation:**

* Added a test plan for AFT Prefix Filtering Dual-Stack, including a detailed `README.md` that covers independent and concurrent application of IPv4 and IPv6 prefix filters, validation steps, and proposed OpenConfig model extensions.
* Added a test plan for AFT Prefix Filtering Dynamic Updates, with a `README.md` describing tests for dynamic addition and removal of prefixes from active sets and expected gNMI notifications, as well as OpenConfig path and RPC coverage.

**Metadata additions:**

* Added `metadata.textproto` files for the new test plans, specifying UUIDs, plan IDs, descriptions, and required testbed types for AFT Prefix Filtering (single-stack), Dual-Stack, and Dynamic Updates. [[1]](diffhunk://#diff-f9b9f9f8c12229555842d696b6d02a283e21ed1d4ef6d379494825b955e65bf6R1-R7) [[2]](diffhunk://#diff-649ab57807a587350a0910895e63badeff7f4df91952b80d32c5f8f022d7f4f5R1-R7) [[3]](diffhunk://#diff-d023aa9626632e4b0777c83f73a21a50ea9f26ce3880796e43395f6c5d0913fdR1-R7)

PR #5580, opened 2026-06-16
Adding test for AFT-6.1

Steubert:
**README:** https://github.com/open-traffic-generator/featureprofiles/blob/main/feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/README.md

Test  https://partnerissuetracker.corp.google.com/u/1/issues/517809756

**Logs:** https://partnerissuetracker.corp.google.com/u/1/issues/524449184


feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+
+func TestMain(m *testing.M) {
+	fptest.RunTests(m)
+}
+
+// TestAFTPrefixFiltering tests the behavior of the AFT global filter with various policies and prefix sets.
+func TestAFTPrefixFiltering(t *testing.T) {
+	dut, ate := ondatra.DUT(t, "dut"), ondatra.ATE(t, "ate")
+	configureDUT(t, dut)
+	configureATE(t, ate)
```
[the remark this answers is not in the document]
Steubert: As per https://partnerissuetracker.corp.google.com/u/1/issues/517809756 global filter is not supported having this deviation and the test will fail now as it's a known issue.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+// buildOCPath builds a path for the given OpenConfig path.
+func buildOCPath(path string) *gpb.Path {
+	return &gpb.Path{Origin: "openconfig", Elem: []*gpb.PathElem{{Name: path}}}
+}
+
+// setFilter configures the AFT global filter with the given ipv4/ipv6 policy
+// names and returns any error from the underlying Set RPC. On platforms where
+// AftsGlobalFilterPolicyOCUnsupported is true, this is a no-op that returns a
+// nil response and a nil error.
+func setFilter(t *testing.T, dut *ondatra.DUTDevice, _ globalFilter) (*gpb.SetResponse, error) {
```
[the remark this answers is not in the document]
Steubert: If I make it f then will get static analysis error as it's a unused variable. We have have change it only when we are uncommenting the OC.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	if p.excludedPfx != "" {
+		_, err := fix.listenUntil(t, aftcache.InitialSyncStoppingCondition(t, fix.dut, map[string]bool{p.excludedPfx: true}, fix.nhMapIPv4(), fix.nhMapIPv6()))
+		if err != nil {
+			t.Errorf("listenUntil: %v", err)
+		}
+	}
+}
+
+// testNonExistentPolicy tests the behavior of the filter when a non-existent policy is applied.
+func testNonExistentPolicy(t *testing.T, fix *testFixture) {
```
romeyod: As mentioned in issue 491765154#comment10, Config reference validation checks/cases described in AFT-6.1.2 & AFT-6.1.3 https://github.com/openconfig/featureprofiles/pull/5217/changes#diff-5ff28fc61b19c81414e0951ee6391d08a2de87f8a629d1214fe26d0e1a224f7e will not be supported in Arista Phase 1 implementation.

Maybe we can add a new deviation for these cases

proto/metadata.proto
```diff
@@ -1378,6 +1378,11 @@ message Metadata {
     // rather than at the integrated-circuit component level.
     // Arista: https://partnerissuetracker.corp.google.com/issues/523026741
     bool use_chassis_aggregate_utilization = 435;
+
+    // Devices that does not support afts global filter policy
```
romeyod: I see 436 is already assigned in latest code
https://github.com/openconfig/featureprofiles/blob/8d1ef2a0372d77e2feecd66b61e9defa887334f9/proto/metadata.proto#L1383

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+		{name: "AFT-6.1.6-DenyActionExclusionList", fn: testDenyAction},
+		{name: "AFT-6.1.7-NonPrefixSetMatchCriteria", fn: testNonPrefixSetMatch},
+	}
+
+	for _, tc := range testTable {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			gnmiClient, err := dut.RawAPIs().BindingDUT().DialGNMI(t.Context())
+			if err != nil {
+				t.Errorf("DialGNMI: %v", err)
```
romeyod: Should this be t.Fatalf?

internal/deviations/deviations.go
```diff
@@ -2254,3 +2254,9 @@ func InterfaceIDFormatRequiredForPolicyForwarding(dut *ondatra.DUTDevice) bool {
 func UseChassisAggregateUtilization(dut *ondatra.DUTDevice) bool {
 	return lookupDUTDeviations(dut).GetUseChassisAggregateUtilization()
 }
+
+// AftsGlobalFilterPolicyOCUnsupported returns true if "/network-instances/network-instance/afts/global-filter-policy" OC path is not supported.
+// Arista: b/514565554
```
romeyod, line 2318: I see multiple PRs are adding `AftsGlobalFilterPolicyOCUnsupported` as well as some duplicate code blocks/helper functions. Maybe we need to consolidate and coordinate the changes?
https://github.com/openconfig/featureprofiles/pull/5573
https://github.com/openconfig/featureprofiles/pull/5523
https://github.com/openconfig/featureprofiles/pull/5522

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
+// Copyright 2024 Google LLC
```
romeyod: nit: 2026

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	entry := ps.GetOrCreatePrefix(ipPrefix, maskRange)
+	entry.IpPrefix, entry.MasklengthRange = ygot.String(ipPrefix), ygot.String(maskRange)
+	gnmi.Replace(t, dut, gnmi.OC().RoutingPolicy().DefinedSets().PrefixSet(name).Config(), ps)
+}
+
+// installAcceptAllPolicy installs a policy that accepts all routes.
+func installAcceptAllPolicy(t *testing.T, dut *ondatra.DUTDevice, name string) {
+	t.Helper()
+	pd := &oc.RoutingPolicy_PolicyDefinition{Name: ygot.String(name)}
+	stmt, _ := pd.AppendNewStatement("10")
```
romeyod: Error handling missing

romeyod (commented):
Assuming setFilter/deleteFilter cannot be tested until we have a vendor implementation

paramasivamn1:
Kindly update the base profile based on latest readme(https://github.com/openconfig/featureprofiles/pull/5606)

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	}
+	errs := verifyState(aft, []string{pfx19851100024, pfx2030113028}, false)
+	errs = append(errs, verifyState(aft, []string{pfx10064024}, true)...)
+	if len(errs) > 0 {
+		t.Errorf("verification failed: %v", errs)
+	}
+}
+
+// testNonPrefixSetMatch tests the behavior of the filter when matching on non-prefix sets.
+func testNonPrefixSetMatch(t *testing.T, fix *testFixture) {
```
akalluru1: This test case won't be supported on ARISTA device in the 1st phase so please add a deviation, thanks!

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	}
+	return errs
+}
+
+func TestMain(m *testing.M) {
+	fptest.RunTests(m)
+}
+
+// TestAFTPrefixFiltering tests the behavior of the AFT global filter with various policies and prefix sets.
+func TestAFTPrefixFiltering(t *testing.T) {
```
akalluru1: General pattern observed in various FNT tests was to define the helper & individual testcase functions before the main Test function. So, please consider following the same pattern

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+		}
+		deleteStaticRoute(t, fix.dut, p.matchingNewPfx)
+		_, err = fix.waitPrefixesAbsent(t, map[string]bool{p.matchingNewPfx: true})
+		if err != nil {
+			t.Errorf("waitPrefixesAbsent: %v", err)
+		}
+	}
+	if p.noMatchNewPfx != "" {
+		addStaticRoute(t, fix.dut, p.noMatchNewPfx, nhIPv4)
+		time.Sleep(noUpdateWindow)
```
akalluru1: 3 seconds may be too small. 30 seconds should be enough time for the static route to get processed, installed in the FIB but get denied from AFT streaming

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+		filterPath := buildFilterPath(deviations.DefaultNetworkInstance(dut))
+		atomicDeleteReq := &gpb.SetRequest{
+			Delete: []*gpb.Path{filterPath, delPath},
+		}
+		if _, err := gnmiClient.Set(context.Background(), atomicDeleteReq); err != nil {
+			t.Errorf("atomic delete: %v", err)
+		}
+	}
+
+	_, err = fix.syncAndGet(t, map[string]bool{pfx10064024: true})
```
akalluru1: As part of the 1st phase If the policy definition is not present while AFT filter is referring to it then all the routes will be rejected by the AFT filter since it is a case of non-existent policy.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	}
+	errs := verifyState(aft, p.wantPresent, true)
+	errs = append(errs, verifyState(aft, p.wantAbsent, false)...)
+	if len(errs) > 0 {
+		t.Errorf("verification failed: %v", errs)
+	}
+
+	if p.matchingNewPfx != "" {
+		addStaticRoute(t, fix.dut, p.matchingNewPfx, nhIPv4)
+		_, err := fix.listenUntil(t, aftcache.InitialSyncStoppingCondition(t, fix.dut, map[string]bool{p.matchingNewPfx: true}, fix.nhMapIPv4(), fix.nhMapIPv6()))
```
akalluru1: `fix.syncAndGet` can be used here too?

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	}
+	configureRoutingPolicies(t, dut)
+	if _, err := setFilter(t, dut, globalFilter{ipv4Policy: policyPrefixSetA}); err != nil {
+		t.Errorf("setFilter(%+v): %v", globalFilter{ipv4Policy: policyPrefixSetA}, err)
+	}
+	aft, err := fix.syncAndGet(t, map[string]bool{pfx19851100024: true, pfx2030113028: true})
+	if err != nil {
+		t.Errorf("syncAndGet: %v", err)
+	}
+	errs := verifyState(aft, []string{pfx19851100024, pfx2030113028}, true)
```
akalluru1: `syncAndGet` will not return until the prefixes are present or times out. If it times out then `err` would be non nil. Is it required to call `verifyState` again with the expected prefixes?

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	if len(errs) > 0 {
+		t.Errorf("verification failed: %v", errs)
+	}
+
+	deleteFilter(t, dut)
+	gnmi.Delete(t, dut, policyPath)
+	_, err = fix.listenUntil(t, aftcache.InitialSyncStoppingCondition(t, fix.dut, map[string]bool{pfx10064024: true}, fix.nhMapIPv4(), fix.nhMapIPv6()))
```
akalluru1: `syncAndGet` can be used.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	if err != nil {
+		t.Errorf("listenUntil: %v", err)
+	}
+	configureRoutingPolicies(t, dut)
```
akalluru1: Is this needed? If so, verification is missing after this config step.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+func testChangePrefixSet(t *testing.T, fix *testFixture) {
+	dut := fix.dut
+	if _, err := setFilter(t, dut, globalFilter{ipv4Policy: policyPrefixSetA}); err != nil {
+		t.Errorf("setFilter(%+v): %v", globalFilter{ipv4Policy: policyPrefixSetA}, err)
+	}
+	aft, err := fix.syncAndGet(t, map[string]bool{pfx19851100024: true, pfx2030113028: true})
+	if err != nil {
+		t.Errorf("syncAndGet: %v", err)
+	}
+	errs := verifyState(aft, []string{pfx19851100024, pfx2030113028}, true)
```
akalluru1: same comment as earlier

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
akalluru1: I guess this is to keep all the policies in the same starting state for each testcase if that is the intention why not call it at line 222 or 223

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	installPolicy(t, dut, policyPrefixSetA, []policyStmt{{seq: "10", prefixSet: psB, result: oc.RoutingPolicy_PolicyResultType_ACCEPT_ROUTE}})
+	aftAfter, err := fix.listenUntil(t, aftcache.DeletionStoppingCondition(t, dut, map[string]bool{pfx19851100024: true, pfx2030113028: true}))
+	if err != nil {
+		t.Errorf("listenUntil: %v", err)
+	}
+	errs = verifyState(aftAfter, []string{pfx19851100024, pfx2030113028, pfx2001DB8264, pfx2001DB821128}, false)
+	if len(errs) > 0 {
+		t.Errorf("verification failed: %v", errs)
+	}
+	configureRoutingPolicies(t, dut)
```
akalluru1: same comment as earlier

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	return &gpb.Path{Origin: "openconfig", Elem: []*gpb.PathElem{{Name: path}}}
+}
+
+// setFilter configures the AFT global filter with the given ipv4/ipv6 policy
+// names and returns any error from the underlying Set RPC. On platforms where
+// AftsGlobalFilterPolicyOCUnsupported is true, this is a no-op that returns a
+// nil response and a nil error.
+func setFilter(t *testing.T, dut *ondatra.DUTDevice, _ globalFilter) (*gpb.SetResponse, error) {
+	t.Helper()
+	if deviations.AftsGlobalFilterPolicyOCUnsupported(dut) {
```
akalluru1: If this deviation is True there is no point in running this test at all so shouldn't be better to skip at the beginning of `TestAFTPrefixFiltering` function itself.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	//		}
+	//		gnmi.Update(t, dut, gnmi.OC().NetworkInstance(deviations.DefaultNetworkInstance(dut)).Afts().GlobalFilter().Config(), gf)
+	return nil, nil
+}
+
+// deleteFilter deletes the AFT global filter. On platforms where
+// AftsGlobalFilterPolicyOCUnsupported is true, this is a no-op.
+func deleteFilter(t *testing.T, dut *ondatra.DUTDevice) {
+	t.Helper()
+	if deviations.AftsGlobalFilterPolicyOCUnsupported(dut) {
```
akalluru1: same comment as earlier

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	t.Helper()
+	installExactPrefixSet(t, dut, psA, []string{pfx19851100024, pfx2030113028, pfx1985110132})
+	installExactPrefixSet(t, dut, psB, []string{pfx2001DB8264, pfx2001DB821128})
+	installRangePrefixSet(t, dut, psVRFA, pfx100641024, "24..32")
+	installRangePrefixSet(t, dut, psSubnet, pfx20301130024, "25..32")
+	installRangePrefixSet(t, dut, psSubnetV6, pfx2001DB8364, "65..128")
+	installAcceptAllPolicy(t, dut, policyMatchAll)
+	installPolicy(t, dut, policyPrefixSetA, []policyStmt{{seq: "10", prefixSet: psA, result: oc.RoutingPolicy_PolicyResultType_ACCEPT_ROUTE}})
+	installPolicy(t, dut, policyPrefixSetB, []policyStmt{{seq: "10", prefixSet: psB, result: oc.RoutingPolicy_PolicyResultType_ACCEPT_ROUTE}})
+	installPolicy(t, dut, policyPrefixSetVRFA, []policyStmt{{seq: "10", prefixSet: psVRFA, result: oc.RoutingPolicy_PolicyResultType_ACCEPT_ROUTE}})
```
akalluru1: `policyPrefixSetVRFA` is not used anywhere in this test so should this be removed? If removed then `psVRFA` should also be removed.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	"google.golang.org/grpc/status"
+)
+
+const (
+	aftConvergenceTime   = 120 * time.Second
+	noUpdateWindow       = 3 * time.Second
+	policyMatchAll       = "POLICY-MATCH-ALL"
+	policyPrefixSetA     = "POLICY-PREFIX-SET-A"
+	policyPrefixSetB     = "POLICY-PREFIX-SET-B"
+	policyPrefixSetVRFA  = "POLICY-PREFIX-SET-VRF-A"
```
akalluru1: `VRFA` is not configured in this test so this is not required.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,653 @@
...
+	policySubnet         = "POLICY-SUBNET"
+	policySubnetV6       = "POLICY-SUBNET-V6"
+	policyMultiStmt      = "POLICY-MULTI-STMT"
+	policyDenyPrefixSetA = "POLICY-DENY-PREFIX-SET-A"
+	policyTagMatch       = "POLICY-TAG-MATCH"
+	policyNotYetExist    = "POLICY-DOES-NOT-YET-EXIST"
+	psA                  = "PREFIX-SET-A"
+	psB                  = "PREFIX-SET-B"
+	psVRFA               = "PREFIX-SET-VRF-A"
```
akalluru1: `VRFA` is not configured in this test so this is not required.

akalluru1 (commented):
Please resolve the addressed comments.

internal/cfgplugins/policyforwarding.go
```diff
@@ -1463,7 +1463,7 @@ func ConfigureTrafficPolicyACL(t *testing.T, dut *ondatra.DUTDevice, params ACLT
 }
 
 // ConfigureVRFSelectionPolicyOC configures vrf_selection_policy_c on DUT port1.
-func ConfigureVRFSelectionPolicyOC(t *testing.T, dut *ondatra.DUTDevice, vrfBatch *gnmi.SetBatch) {
+func ConfigureVRFSelectionPolicyOC(t *testing.T, dut *ondatra.DUTDevice, encapVRFs []string) {
```
akalluru1, line 1466: comment from Claude:
```
  ConfigureVRFSelectionPolicyOC signature change breaks callers
  policyforwarding.go:1466 — Signature changed from (t, dut, vrfBatch *gnmi.SetBatch) to (t, dut, encapVRFs []string),
  but gribifullscale.go:373 still passes a *gnmi.SetBatch. This is a type mismatch that breaks the existing caller.
```

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,947 @@
...
+	cfgplugins.AddPrefixSetPolicy(t, rp, cfgplugins.PrefixSetPolicyParams{PolicyName: policyMultiStmt, StatementNames: []string{defaultStatementName, secondStatementName}, PrefixSetNames: []string{v4PfxSetA, subnetPfxSetV4}, MatchPrefixSet: true, PolicyResult: oc.RoutingPolicy_PolicyResultType_ACCEPT_ROUTE})
+	cfgplugins.AddPrefixSetPolicy(t, rp, cfgplugins.PrefixSetPolicyParams{PolicyName: policyDenyPfxSetA, StatementNames: []string{defaultStatementName, secondStatementName}, PrefixSetNames: []string{v4PfxSetA, ""}, MatchPrefixSet: true, PrefixDeny: true, PolicyResult: oc.RoutingPolicy_PolicyResultType_ACCEPT_ROUTE})
+	cfgplugins.AddPrefixSetPolicy(t, rp, cfgplugins.PrefixSetPolicyParams{PolicyName: policyTagMatch, StatementNames: []string{tagStatementName}, MatchPrefixSet: true, SetTag: true, PolicyResult: oc.RoutingPolicy_PolicyResultType_ACCEPT_ROUTE})
+	gnmi.BatchReplace(batch, gnmi.OC().RoutingPolicy().Config(), rp)
+	batch.Set(t, dut)
+}
+
+// addSingleStaticRoute adds one static route to the given network instance
+func addSingleStaticRoute(t *testing.T, dut *ondatra.DUTDevice, niName, prefix, index, nextHop string) error {
```
akalluru1: comment from Claude:
```
  addSingleStaticRoute always returns nil

  Test line 103-109: Returns error but never returns a non-nil error. The ondatra batch.Set(t, dut) will t.Fatal internally on failure, so the callers' if err != nil checks (e.g. lines 406, 452, 486) are dead code.
```

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,947 @@
...
+	t.Helper()
+	batch := &gnmi.SetBatch{}
+	cfgplugins.ConfigureStaticRoute(t, dut, batch, cfgplugins.ConfigureStaticRouteParams{NetworkInstance: niName, Prefix: prefix, Index: index, NextHop: nextHop})
+	batch.Set(t, dut)
+	return nil
+}
+
+// removeStaticRoute deletes a static route (keyed by prefix) from the given network instance
+func removeStaticRoute(t *testing.T, dut *ondatra.DUTDevice, niName, prefix string) error {
```
akalluru1: similar comment as earlier

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,947 @@
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// verifyNextHopsRetained confirms the next-hop-group and next-hops backing a
+// prefix are still present in the filtered AFT.
+func verifyNextHopsRetained(aft *aftcache.AFTData, prefix string) error {
```
akalluru1: nit: `verifyNexthopGroupsAndNexthopsRetained`?

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,947 @@
...
+
+	wantPrefixes := make(map[string]bool)
+	for _, p := range tc.matchPrefixes {
+		wantPrefixes[p] = true
+	}
+
+	t.Logf("%s - Initial Synchronization", tc.name)
+
+	collector := aftcache.NewAFTStreamSession(ctx, t, aftpf.GnmiClientSession(t, dut, aftpf.PrefixesParams{Ctx: ctx}), dut)
+	collector.ListenUntilPreUpdateHook(context.Background(), t, subscriptionWait, []aftcache.NotificationHook{aftcache.VerifyAtomicFlagHook(t)}, aftcache.InitialSyncStoppingCondition(t, dut, wantPrefixes, map[string]bool{atePort1.IPv4: true}, map[string]bool{atePort1.IPv6: true}))
```
akalluru1: nit: very long line, split into multiple lines and please fix other places too

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,947 @@
...
+		} else {
+			dynV6NH = map[string]bool{atePort1.IPv6: true}
+		}
+
+		if err := addSingleStaticRoute(t, dut, ni, tc.dynamicAddPrefix, fmt.Sprintf("%d", staticRouteIndex+600+iterIdx), tc.nextHop); err != nil {
+			t.Errorf("%s: failed to add dynamic prefix %s: %v", tc.name, tc.dynamicAddPrefix, err)
+		} else {
+			// Keep watching the same streaming session so the dynamic add is
+			// observed as an incremental notification on the existing stream.
+			aftpf.RunCollector(t, aftpf.RunCollectorParams{Ctx: context.Background(), Collector: collector, Stop: aftcache.InitialSyncStoppingCondition(t, dut, map[string]bool{tc.dynamicAddPrefix: true}, dynV4NH, dynV6NH), Timeout: subscriptionWait})
```
akalluru1: please create a common verification function that can be reused across all the testcases in this function instead of repeating similar pattern.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,947 @@
...
+
+	if err := addSingleStaticRoute(t, dut, ni, matchPrefix, fmt.Sprintf("%d", staticRouteIndex+700), atePort1.IPv4); err != nil {
+		t.Errorf("Failed to add match prefix %s: %v", matchPrefix, err)
+		return
+	}
+
+	cfgplugins.ConfigureGlobalFilterPolicies(t, dut, cfgplugins.ConfigureGlobalFilterPoliciesParams{V4Policy: policyNotYetExist, V6Policy: "", VRFName: ni})
+
+	wantPrefixes := map[string]bool{matchPrefix: true}
+	collector := aftcache.NewAFTStreamSession(ctx, t, aftpf.GnmiClientSession(t, dut, aftpf.PrefixesParams{Ctx: ctx}), dut)
```
akalluru1, line 750: Try to reuse the common verification function here too

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,947 @@
...
+
+	aftpf.AFTFilterAwaitBGPConvergence(t, dut, ni)
+
+	tests := []struct {
+		name string
+		test func(t *testing.T, dut *ondatra.DUTDevice)
+	}{
+		{
+			name: "aft-6.1.1-testPrefixSetPolicySubscription",
+			test: testPrefixSetPolicySubscription,
```
akalluru1: definition of this & the next test are above `TestMain` but rest of them are below this. Let's put all of them above `TestMain` or before instead of splitting them

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,947 @@
...
+	// Same session observes the prefix-set swap as streamed deletions.
+	aftpf.RunCollector(t, aftpf.RunCollectorParams{Ctx: context.Background(), Collector: collector, Stop: aftcache.DeletionStoppingCondition(t, dut, map[string]bool{"198.51.100.0/24": true}), Timeout: subscriptionWait})
+	swapAFT, err := collector.ToAFT(t, dut)
+	if err != nil {
+		t.Errorf("ToAFT failed after swap: %v", err)
+	} else {
+		aftpf.VerifyPrefixesAbsent(t, aftpf.PrefixesParams{InfoAFT: swapAFT, Prefixes: []string{"198.51.100.0/24", "203.0.113.0/28"}})
+	}
+
+	time.Sleep(notificationWaitTime)
```
akalluru1, line 900: what is the rationale behind the `Sleep`?

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,947 @@
...
+	cfgplugins.IsIPv6InterfaceARPresolved(t, ate, cfgplugins.AddressFamilyParams{InterfaceNames: interfaceNamesList})
+
+	aftpf.AFTFilterAwaitBGPConvergence(t, dut, ni)
+
+	tests := []struct {
+		name string
+		test func(t *testing.T, dut *ondatra.DUTDevice)
+	}{
+		{
+			name: "aft-6.1.1-testPrefixSetPolicySubscription",
```
akalluru1: Only this testcase has v4 & v6 verification but not the rest of them what's the reason?

internal/aft_prefix_filtering/aft_prefix_filtering_service.go
```diff
@@ -0,0 +1,560 @@
...
+	c := aftcache.NewAFTStreamSession(cfg.Context, t, cfg.Client, dut)
+	if *debugNotifications {
+		c.WithDebug()
+		t.Log("DEBUG MODE ENABLED: Recording all gNMI notifications to memory.")
+	}
+	return c
+}
+
+// AFTFilterConfigureDUT configures the DUT with basic port1/port2 interfaces in the DEFAULT network instance.
+func AFTFilterConfigureDUT(t *testing.T, dut *ondatra.DUTDevice) *gnmi.SetBatch {
```
akalluru1: what's the reason start this & other functions with `AFTFilter`? They're in a different package that provides enough isolation required.

internal/deviations/deviations.go
```diff
@@ -1910,6 +1910,12 @@ func InterfaceEthernetInblockErrorsUnsupported(dut *ondatra.DUTDevice) bool {
 	return lookupDUTDeviations(dut).GetInterfaceEthernetInblockErrorsUnsupported()
 }
 
+// WithIPAddressUnsupported returns true when an indirect next-hop (direct interface IP) with forwarding viable is used, since this is not supported.
+// Nokia b/428883444
+func ForwardingViableFailoverWithIndirectNHUnsupported(dut *ondatra.DUTDevice) bool {
```
akalluru1, line 1915: Is this relevant to the AFT test? seems like sync or some branch parenting problem?

internal/telemetry/aftcache/aft_cache.go
```diff
@@ -1131,3 +1131,139 @@ func writeNotifications(t *testing.T, notifications []*gnmipb.SubscribeResponse,
...
+	NotificationWait time.Duration
+}
+
+// WaitForUpdateDeleteNotification returns a PeriodicHook that waits until:
+//
+//   - an UPDATE notification is received for updatePrefix
+//   - a DELETE notification is received for deletePrefix
+//
+// The hook returns true only after both notifications have been observed.
+func WaitForUpdateDeleteNotification(t *testing.T, cfg NotificationExpectation) PeriodicHook {
```
akalluru1: Isn't it enough to use this function in various scenarios what is the reason for having the 2 subsequent functions?

internal/cfgplugins/policyforwarding.go
```diff
@@ -1561,3 +1561,128 @@ func ConfigureCLIDecapVRFMode(t *testing.T, dut *ondatra.DUTDevice) {
 	t.Log("Enabling next-hop decapsulation VRF mode")
 	helpers.GnmiCLIConfig(t, dut, cliConfig)
 }
+
+// PrefixSetPolicyParams has the parameters to configure prefix set policy.
```
akalluru1: These doesn't belong to this file since `policyforwarding.go` is for `/network-instances/network-instance/policy-forwarding` container but this & below functions are related to `routing-policy` module so  please add a different file for these new common APIs. I missed this in other reviews please let Mohan also know about it.

paramasivamn1 (commented):
steubert, Can you also include default route with exact match as one of the term in the filter for both IPv4 and ipv6.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
Steubert: I'm skipping the test at the beginning itself for DUT when the afts_global_filter_policy_oc_unsupported deviation is enabled

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
Steubert: Made it as t.Fatalf

proto/metadata.proto
Steubert: Updated according to latest metadata.proto. Before merging it will show conflits. At that time we can update the latest number.

internal/deviations/deviations.go
Steubert: New library added consolidating common function for all these tests

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
Steubert: Skipping the test at the beginning

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
Steubert: The swap only produces delete notifications for PREFIX-SET-A's members; PREFIX-SET-B's members were never installed in the AFT, so there is no further positive event to wait on. Wait out a full notification window and re-check the snapshot to catch any spurious re-add that a bug might otherwise introduce after the deletion settles.

romeyod:
Did we generate and update `networkinstance-0.go` in repo ondatra/gnmi/oc/networkinstance  ?

Steubert:
> Did we generate and update `networkinstance-0.go` in repo ondatra/gnmi/oc/networkinstance ?

We've not generated it. As of now we've commented out the OC part that uses global-filter. Our team have not updated ondatra package till now.

paramasivamn1 (approved):

ksgireesha:
Hi @romeyod , @akalluru1 - We have LGTM from @paramasivamn1 . Please review and approve this PR

akalluru1 (approved):
changes look good to me.

ksgireesha:
Hi @romeyod and @rohit-rp , Please review, and approve the PR.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
romeyod: This needs to have a seperate deviation imo as we will be setting afts_global_filter_policy_oc_unsupported to False in Phase 1

romeyod:
https://github.com/openconfig/featureprofiles/pull/5580#discussion_r3482881255 
We need to close out on this one

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
Steubert: @romeyod Need a clarification. Currently, DUT do not support the global filter feature. Even if we set afts_global_filter_policy_oc_unsupported to False, the DUT will still reject the OC path. In that case, there doesn't seem to be any value in proceeding with the test, right? If so, why would we need a new deviation? Please correct me if my understanding is incorrect.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,1098 @@
...
+			nmIdx := fmt.Sprintf("%d", staticRouteIndex+650+iterIdx)
+			addSingleStaticRoute(t, dut, ni, tc.dynamicNonMatchPrefix, nmIdx, tc.nextHop)
+
+			// Drain the existing stream and confirm the non-matching prefix
+			// never surfaces; it must stay absent from the filtered view.
+			aftpf.CollectAndVerify(t, dut, aftpf.RunCollectorParams{
+				Ctx:       context.Background(),
+				Collector: collector,
+				Stop:      aftcache.DeletionStoppingCondition(t, dut, map[string]bool{tc.dynamicNonMatchPrefix: true}),
+				Timeout:   notificationWaitTime,
```
akalluru1: `notificationWaitTime` (30s) is incorrect because periodic interval in `ListenUntil` is 2 mins (https://github.com/openconfig/featureprofiles/blob/main/internal/telemetry/aftcache/aft_cache.go#L622C35-L622C51). `subscriptionWait` should be used here instead so that periodic hook gets a chance to run at least once. Romeyo found this issue after testing with AFT filter config through OC.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
romeyod: I discussed this with @akalluru1. We can add a new deviation later. 

> In that case, there doesn't seem to be any value in proceeding with the test, right? If so, why would we need a new deviation?

To answer this, once Arista or any vendor starts supporting this feature, they selectively add deviations for  certain test cases that they do not support. This way the supported test cases are run and unsupported ones are skipped.

romeyod (approved):

ksgireesha:
Hi @rohit-rp / @navaneethyv - We have 3 LGTMs to this PR. Please review, validate, approve and merge this PR as [openconfig/featureprofiles-approvers](https://github.com/orgs/openconfig/teams/featureprofiles-approvers)

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,1098 @@
...
+	// }
+	// if v6Policy != "" {
+	// 	gnmi.BatchUpdate(batch, gf.Ipv6Policy().Config(), v6Policy)
+	// }
+	// // Requires a non-fatal Set variant to inspect the returned status code
+	// // against wantCode instead of failing the test on RPC error.
+	// return nil
+	return fmt.Errorf("aft global filter policy is expected to be supported on %s, "+
+		"but no OpenConfig implementation is available", dut.Vendor())
+}
```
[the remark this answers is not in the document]
rohit-rp: please address this

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/afts_prefix_filtering_test.go
```diff
@@ -0,0 +1,1098 @@
...
+	// once the bindings are regenerated against openconfig/public >= 3.3.0.
+	// batch := &gnmi.SetBatch{}
+	// gnmi.BatchDelete(batch, gnmi.OC().NetworkInstance(niName).Afts().GlobalFilter().Ipv4Policy().Config())
+	// gnmi.BatchDelete(batch, gnmi.OC().NetworkInstance(niName).Afts().GlobalFilter().Ipv6Policy().Config())
+	// gnmi.BatchDelete(batch, gnmi.OC().RoutingPolicy().PolicyDefinition(policyName).Config())
+	// batch.Set(t, dut)
+	// return nil
+	return fmt.Errorf("aft global filter deletion is expected to be supported on %s, "+
+		"but no OpenConfig implementation is available", dut.Vendor())
+}
```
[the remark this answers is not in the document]
rohit-rp: please address this

rohit-rp:
please address all open comments

Steubert:
All comments are addressed

rohit-rp (approved):

PR #5581, opened 2026-06-16
Adding large_gnmi_set_and_reboot readme

rohit-rp:
This is to address: b/463393615#comment2

- The config database was locked after a switch reboot triggered via CLI, preventing configuration commits.
- This was caused by an incomplete configuration change operation during the reboot.
- The `mgd-init` application struggled to initialize the configuration database due to a high interface scale, exceeding its timeout.


AmrNJ (approved):
LGTM.

PR #5620, opened 2026-06-25
RT-1.103: Add test requirement for BGP dynamic neighbors

ccie57960:
Scope
  - Verifies establishment of iBGP sessions via dynamic peering (listen range).
  - Validates consistent Router-ID usage across dynamically established sessions.
  - Ensures independent session maintenance even when different peers share the same Router-ID (Transport Separation).
  - Validates secure dynamic peering with MD5 authentication.


feature/bgp/otg_tests/bgp_dynamic_neighbors/README.md
```diff
@@ -0,0 +1,154 @@
...
+### Configuration
+
+1.  Configure DUT with a global router-id.
+2.  Configure a peer-group for dynamic neighbors (e.g., `RM_DYNAMIC_PEERS`).
+3.  Configure dynamic neighbor prefixes (listen range, e.g., `2001:db8:1::/64`) associated with the peer-group.
+4.  Configure MD5 authentication on the peer-group.
+
+### Tests
+
+#### RT-1.103.1 - Dynamic Peering and Router-ID Validation
```
ram-mac, line 41: @ccie57960 - The passive listener must only accept BGP TCP connections from a specific ASN associated with that prefix range. 
The current test only verify succesful session establishment and lack a negative verification step. 
Add a step to attempt starting a BGP session from an IP in the listen range but using incorrect/Different ASN and verify if the DUT rejects the connection or that it fails to transition to Established state.

ram-mac (changes requested):

feature/bgp/otg_tests/bgp_dynamic_neighbors/README.md
```diff
@@ -0,0 +1,154 @@
...
+*   **Step 3**: Verify that sessions transition to `ESTABLISHED`.
+*   **Step 4**: Attempt to initiate a session from ATE with an incorrect password or no password.
+*   **Step 5**: Verify that the session FAILS to establish.
+
+#### RT-1.103.4 - Duplicate ATE Router-ID (Transport Separation)
+
+*   **Step 1**: Establish a BGP session from ATE Peer 1 (Source IP `2001:db8:1::1`) with a specific Router-ID (e.g., `192.0.2.200`).
+*   **Step 2**: Initiate a BGP session from ATE Peer 2 (Source IP `2001:db8:1::2`) using the **same** Router-ID (`192.0.2.200`).
+*   **Step 3**: Verify that **both** sessions transition to `ESTABLISHED` and are maintained independently, relying on transport address separation.
+
```
ram-mac, line 70: @ccie57960 - A routing scenario where the exact same prefix (e.g., 10.0.0.0/8) is simultaneously advertised from ATE to DUT and vice versa needs to be covered. BGP best-path rules, a router cannot install or advertise the same prefix in both directions simultaneously and pick the best path based on the decision process. Maybe we can cover one case as below.

#### RT-1.103.5 - Overlapping Prefix Best-Path Selection and Reconvergence

Verifies that the DUT correctly evaluates BGP best-path parameters (e.g., MED) and executes seamless route reconvergence when the same prefix is received concurrently from both the core and customer-facing interfaces.

ram-mac (changes requested):

ram-mac (approved):

PR #5640, opened 2026-07-01
gNOI-2.1 800G support

manan-patel:


rohit-rp:
@manan-patel Please address review comments and add description of the change, specifically the timers change, how do you know it works for other vendors also?

PR #5649, opened 2026-07-01
gNMI-1.10 Added PTX10002-36QDD for skipping Fabric check

priyaranjannanda:
- Added PTX10002-36QDD for skipping Fabric check


navaneethyv (commented):
please take care of the comment

PR #5652, opened 2026-07-02
Adding new test for Port mirroring coverage in FNT

ram-mac:
Add new test for coverage of Port Mirroring 
- Adding README.md file which outlines the test procedure for local packet mirroring (to FILE action) across physical interfaces and LAG member ports in both ingress and egress directions
- Adding metadata.textproto which would outlines the TestID and other details


AmrNJ (approved):
LGTM.

PR #5673, opened 2026-07-06
gNMI-1.18: Fix Arista IC pattern to exclude FAP core sub-components

mrigank-arista:
The Arista icPattern regex ^SwitchChip matches both parent FAP components (SwitchChip3/0) and core sub-components (SwitchChip3/0.0). Core sub-components do not support backplane-facing-capacity telemetry because fabric bandwidth is a per-FAP property, not per-core. Tighten the regex to ^SwitchChip\d+/\d+$ to match only parent FAPs.

Also remove duplicate OpenConfig Path and RPC Coverage section from README.md.


AmrNJ (approved):
LGTM.

PR #5679, opened 2026-07-09
RT-1.28, RT-2.12 : Deviation: skip_isis_set_metric_style_type for Juniper

sudhirakondi:
Junos devices do not currently support setting the below config path:

/routing-policy/.../actions/openconfig-isis-policy:isis-actions/config/set-metric-style-type

RT-1.28 and RT-2.12 are being enabled with Ondatra deviation skip_isis_set_metric_style_type to skip the configuration of this leaf for Juniper devices.


AmrNJ (approved):
LGTM.

PR #5689, opened 2026-07-10
Remove duplicate test_infra_id flag from AUTHZ-1.4 test

nkrishk:
### Summary
Removes the locally defined -test_infra_id flag from authz1_4_test.go and uses the shared fptest.TestInfraID flag from internal/fptest/config.go (added in #5544).

### Problem
authz1_4_test.go defined its own flag:
`testInfraID = flag.String("test_infra_id", "*/cafyauto", ...)`

internal/fptest/config.go already defines the same flag:
`TestInfraID = flag.String("test_infra_id", "*/cafyauto", "SPIFFE-ID used by test infra clients")`

Registering the same flag name twice in one test binary causes a panic at init time (flag redefined: test_infra_id) when both packages are linked.


MarcCharlebois:
https://github.com/openconfig/featureprofiles/pull/5544#issuecomment-4754741658 @ram-mac  I've mentioned this one before, can you PTAL?

ram-mac (approved):

PR #5700, opened 2026-07-15
Fix duplicate metadata issue in rpc calls

mudhireddy:
Problem being fixed: gRPC credential overrides were being passed via
metadata.NewOutgoingContext / metadata.AppendToOutgoingContext, which caused
duplicate metadata entries — the binding's PerRPCCredentials already sets
username/password metadata, so the outgoing context metadata was being appended
on top of it.

Solution (2 files):

1. topologies/binding/binding.go — Replaces the metadata-based credential
   override mechanism with a context.Value-based approach:
   - Adds a CredsOverride struct and ContextWithCreds() helper that stores
     credentials in the context via a typed key (credsOverrideKey{}).
   - Changes creds.GetRequestMetadata() to check for a CredsOverride in the
     context instead of reading from gRPC outgoing metadata. Defaults to the
     binding-level credentials if no override is present.
   - Removes the google.golang.org/grpc/metadata import.

2. internal/security/acctz/acctz.go — Updates all acctz test RPC senders
   (SendGnmiRPCs, SendGnoiRPCs, SendGnsiRPCs, SendGribiRPCs, SendP4rtRPCs)
   to use the new approach:
   - Replaces all metadata.NewOutgoingContext/metadata.AppendToOutgoingContext
     calls with fpbinding.ContextWithCreds().
   - Passes context.Background() (instead of the creds-laden context) to Dial*
     methods, since credentials now flow through GetRequestMetadata at RPC time
     rather than at connection time.
   - Removes the now-unused getMetadataKeys() helper function.
   - For gNOI and gNSI (non-Nokia), switches from the convenience
     dut.RawAPIs().GNOI(t)/GNSI(t) to
     dut.RawAPIs().BindingDUT().DialGNOI()/DialGNSI() to have direct control
     over the connection context.


mudhireddy:
PR https://github.com/openconfig/featureprofiles/pull/5891 replaces this

PR #5714, opened 2026-07-19
TE-18.3: Updated the code as per the readme changes and validated

keysightgems:
**Readme Location:** https://github.com/openconfig/featureprofiles/blob/main/feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/README.md

**gribigo library issue:** https://github.com/openconfig/gribigo/issues/282 (The test was successfully validated using local changes to fluent.go.) The static check failed because of this issue. It will pass once the issue is fixed.

**Attached logs here:** https://partnerissuetracker.corp.google.com/u/2/issues/536257627


tengyiG:
You need to resolve the go/static analysis: looks like you are calling fluent.UDPV4EncapHeader, which does not exist in pinned gribigo v0.1.1. The PR relies on an uncommitted local dependency change. Static
     Analysis fails with undefined: fluent.UDPV4EncapHeader

feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/mpls_in_udp_scale_test.go
```diff
@@ -814,6 +930,185 @@ func expectedMPLSinUDPMultiOpResults(t *testing.T, vrfs []string, nextHopID, nex
...
+								WithDSCP(uint64(outerDSCP)),
+						),
+				)
+				nhID++
+			}
+		}
+		//------------------------------------------------
+		// NHG
+		//------------------------------------------------
+		for i := 0; i < nhsPerNHG; i++ {
```
tengyiG, line 1114: This is 20,480 updates to 20 keys, rather than 20 NHGs with 1,024 members. Depending on the target, repeated entries may overwrite one another or be rejected; they do not portably represent the required ECMP groups.

feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/mpls_in_udp_scale_test.go
```diff
@@ -1048,45 +1392,65 @@ func configureVRFProfiles(ctx context.Context, t *testing.T, ateConfig gosnappi.
...
 		v6Entries, vrfV6PfxMap := buildProfile4IPv6Routes(t, dut, totalNHs/2, nhsPerNHG, baseIPv6, vrfs, nextHopGroupID)
 		entries = append(entries, v4Entries...)
 		entries = append(entries, v6Entries...)
 		// === Expected OpResults ===
 		wantAdds, wantDels = expectedProfile4MPLSinUDPOpResults(t, nextHopID, nextHopGroupID, totalNHGs, totalPrefixes, nhsPerNHG, baseIPv4, baseIPv6, vrfs)
-		flows = append(flows, append(fa4.createFlow(t, ateConfig, "ipv4", fmt.Sprintf("ip4mpls_p%d", profile), dstmac, fa4.srcPort[0], outerDSCP, startVLANPort1, vrfV4PfxMap, false, tDstLists), fa6.createFlow(t, ateConfig, "ipv6", fmt.Sprintf("ip6mpls_p%d", profile), dstmac, fa6.srcPort[0], outerDSCP, startVLANPort1, vrfV6PfxMap, false, tDstLists)...)...)
+		flows = append(flows, append(fa4.createFlow(t, ateConfig, "ipv4", fmt.Sprintf("ip4mpls_p%d", profile), dstmac, fa4.srcPort[0], outerDSCP, startVLANPort1, vrfV4PfxMap, false, tV6DstLists), fa6.createFlow(t, ateConfig, "ipv6", fmt.Sprintf("ip6mpls_p%d", profile), dstmac, fa6.srcPort[0], outerDSCP, startVLANPort1, vrfV6PfxMap, false, tV6DstLists)...)...)
+	case profileMultiVRFIPv4ECMP:
```
tengyiG, line 1438: CreateVRFs returns DEFAULT first, but profiles 6-8 slice from index zero

tengyiG: Consequently, routes are installed in DEFAULT, contrary to the README requirement that prefix entries reside in non-default NIs

feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/mpls_in_udp_scale_test.go
```diff
@@ -1564,39 +1949,85 @@ func (fa *flowAttr) createFlow(t *testing.T, cfg gosnappi.Config, flowType, name
...
+				}
+				for _, count := range skewPfxCounts {
+					for i := 0; i < count; i++ {
+						nonDefaultPfx1V4Lists = append(nonDefaultPfx1V4Lists, dutV4AddrPfx1)
+						nonDefVlanIDs = append(nonDefVlanIDs, vlanID)
+					}
+					vlanID++
+				}
+			} else {
+				total := 0
```
tengyiG, line 2053: VRF 1 is assigned to subinterface 1/VLAN 101, while the first non-default flow starts at VLAN 100

feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/mpls_in_udp_scale_test.go
tengyiG, line 1904: This only checks a historical 50/50 split across physical ports 3 and 4. It does not verify distribution across 8 or 1,024 encapsulating next hops.

Packet capture independently allow-lists all sources, destinations, and labels without validating their NHG/VRF association, so the malformed Profile 8 groups could still satisfy the test.

feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/mpls_in_udp_scale_test.go
```diff
@@ -814,6 +930,185 @@ func expectedMPLSinUDPMultiOpResults(t *testing.T, vrfs []string, nextHopID, nex
...
+			for nhsIdx := 0; nhsIdx < nhsPerNHG; nhsIdx++ {
+				if flowType == "ipv6" {
+					entries = append(entries,
+						fluent.NextHopEntry().
+							WithNetworkInstance(deviations.DefaultNetworkInstance(dut)).
+							WithIndex(nhVal).
+							AddEncapHeader(
+								fluent.MPLSEncapHeader().WithLabels(currentLabel),
+								fluent.UDPV6EncapHeader().
+									WithSrcIP(srcIPs[nhsIdx]).
```
tengyiG: This resets the source index for every NHG, so only eight outer source addresses are used, rather than the documented 32 rotated sources

feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/mpls_in_udp_scale_test.go
```diff
@@ -740,8 +755,109 @@ func buildProfile4IPv6Routes(t *testing.T, dut *ondatra.DUTDevice, totalNHGs, ro
...
+	totalNHGs := totalNHs / routesPerNHG
+	// NHGs are distributed evenly across the configured VRFs.
+	vrfCount := len(vrfs)
+	if totalNHGs < vrfCount {
+		t.Fatalf("invalid configuration: totalNHGs (%d) must be >= VRFs (%d)", totalNHGs, vrfCount)
+	}
+	baseNHGsPerVRF := totalNHGs / vrfCount
+	extraNHGs := totalNHGs % vrfCount
+	totalPrefixes := totalNHs
+	if flowType == "ipv6" {
```
tengyiG, line 801: This creates globally unique prefixes instead of the required inner-IP reuse across VRFs.

keysightgems:
> You need to resolve the go/static analysis: looks like you are calling fluent.UDPV4EncapHeader, which does not exist in pinned gribigo v0.1.1. The PR relies on an uncommitted local dependency change. Static
> Analysis fails with undefined: fluent.UDPV4EncapHeader

Yes, this is already a known issue. I have mentioned it in the PR comments and have raised issue separately. Once the required change is merged into the pinned gribigo dependency, the static analysis failure (undefined: fluent.UDPV4EncapHeader) will be resolved.

ksgireesha:
Hi @keysightgems,

Just a gentle reminder to address the pending review comments on this PR. Once you have pushed the fixes, please re-request a review so we can get this approved and merged!

feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/mpls_in_udp_scale_test.go
keysightgems: Currently, we don't have support to configure Next Hops (NHs) and Next Hop Groups (NHGs) in non-default VRFs. If we exclude the DEFAULT network instance, the NHGs are not resolved through the MoUDP tunnels, so the profile does not function correctly. Because of this limitation, we cannot remove DEFAULT until non-default VRF support is available for NH/NHG programming. If there is an alternative approach to achieve the same behavior while keeping the prefix entries in non-default network instances, please let me know and I will update the implementation accordingly.

feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/mpls_in_udp_scale_test.go
keysightgems: Same as above and as of now we can keep it VLAN 100.

feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/mpls_in_udp_scale_test.go
keysightgems: I investigated this further and found that the issue was due to the tunnel IP mapping rather than the use of the DEFAULT network instance. After correcting the tunnel IP mapping, the profile works as expected even without the DEFAULT network instance.

I have updated the implementation accordingly, and the prefix entries can remain in the non-default network instances as intended.

feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/mpls_in_udp_scale_test.go
keysightgems: Fixed

feature/gribi/mpls_in_udp/otg_tests/mpls_in_udp_scale/mpls_in_udp_scale_test.go
keysightgems: OpenConfig currently does not support retrieving the encrypted NHG ID with the NextHopGroupName. Enable validation of traffic distribution across the next-hops within a group once support is available.

keysightgems:
Addressed all the comments and committed the code.

keysightgems:
The static failure will be fixed once we update to the new version of gribigo after the PR https://github.com/openconfig/gribigo/pull/285 is merged.

PR #5770, opened 2026-07-28
Stabilizing code: Add watch loop before querying counter values

AmrNJ:
Added stabilizations - Add watch loop with gnmi Await before querying counter values, updated traffic duration. All the tests were passing in the first attempt (except encap_decap_scale - more underlying issues)

### DIFF
--- a/feature/policy_forwarding/vrf_selection/otg_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
+++ b/feature/policy_forwarding/vrf_selection/otg_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
@@ -30,8 +30,8 @@ import (
 	"github.com/openconfig/ondatra/gnmi"
 	"github.com/openconfig/ondatra/gnmi/oc"
 	"github.com/openconfig/testt"
-	"github.com/openconfig/ygot/ygot"
 	"github.com/openconfig/ygnmi/ygnmi"
+	"github.com/openconfig/ygot/ygot"
 )
 
 const (
@@ -520,14 +520,14 @@ func testTrafficFlows(t *testing.T, ate *ondatra.ATEDevice, top gosnappi.Config,
 	for _, flow := range flows {
 		t.Run(flow.Name(), func(t *testing.T) {
 			t.Logf("*** Verifying %v traffic on OTG ... ", flow.Name())
-			
+
 			if _, ok := gnmi.Watch(t, ate.OTG(), gnmi.OTG().Flow(flow.Name()).Transmit().State(), 15*time.Second, func(val *ygnmi.Value[bool]) bool {
 				transmitState, present := val.Val()
 				return present && !transmitState
 			}).Await(t); !ok {
 				t.Fatalf("Timeout waiting for flow %s to stop transmitting", flow.Name())
 			}
-			
+
 			outPktsRaw := gnmi.Get(t, ate.OTG(), gnmi.OTG().Flow(flow.Name()).Counters().OutPkts().State())
 			if outPktsRaw == 0 {
 				t.Fatalf("OutPkts == 0, want >0.")
--- a/feature/policy_forwarding/vrf_selection/otg_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
+++ b/feature/policy_forwarding/vrf_selection/otg_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
@@ -520,14 +520,18 @@ func testTrafficFlows(t *testing.T, ate *ondatra.ATEDevice, top gosnappi.Config,
 	for _, flow := range flows {
 		t.Run(flow.Name(), func(t *testing.T) {
 			t.Logf("*** Verifying %v traffic on OTG ... ", flow.Name())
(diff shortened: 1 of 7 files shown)


feature/system/control_plane_traffic/otg_tests/default_copp_test/default_copp_test.go
```diff
@@ -385,7 +385,42 @@ func (ce *commonEntities) testCoppSystemHelper(t *testing.T, tc *coppSystemTestc
...
+		if !tc.increasedDropCount {
+			break // If we expect NO drops and they increased, break
+		}
+
+		if time.Since(start) > timeout {
+			t.Logf("Timeout reached waiting for telemetry updates.")
+			break
+		}
+
+		time.Sleep(pollInterval)
```
navaneethyv, line 422: why 5 and not 1 ?

navaneethyv: Also please add a description as to why these changes are made and what is the expected outcome.

feature/system/control_plane_traffic/otg_tests/default_copp_test/default_copp_test.go
```diff
@@ -385,7 +385,42 @@ func (ce *commonEntities) testCoppSystemHelper(t *testing.T, tc *coppSystemTestc
 	finalCounters := gnmi.Get(t, ce.dut, gnmi.OC().Interface(ce.dut.Port(t, dutIncomingPort).Name()).Counters().State())
 	finalInPkts := finalCounters.GetInPkts()
 	t.Logf("Final incoming packets: %v", finalInPkts)
-	finalPktCounts := ce.getDroppedPktCounts(t, tc.counters)
+
+	var finalPktCounts []float64
```
navaneethyv, line 389: can this be ?
```
countersPath := gnmi.OC().Interface(ce.dut.Port(t, dutIncomingPort).Name()).Counters().State()

_, ok := gnmi.Watch(t, ce.dut, countersPath, 45*time.Second, func(val *ygnmi.Value[*oc.Interface_Counters]) bool {
	counters, present := val.Val()
	if !present {
		return false
	}
	
	// Fetch your current drops based on the telemetry update
	// (You'll need to adapt getDroppedPktCounts to read from the 'counters' object rather than doing a gnmi.Get)
	finalPktCounts := extractDroppedPktCounts(counters, tc.counters)

	for idx := range tc.counters {
		if tc.increasedDropCount {
			if finalPktCounts[idx] <= initialPktCounts[idx] {
				return false // Keep watching until it increases
			}
		} else {
			if finalPktCounts[idx] != initialPktCounts[idx] {
				// We expected no drops, but they increased. 
				// We can return true here to break the watch, and let the 
				// downstream test logic fail the test.
				return true 
			}
		}
	}
	
	return true // All conditions satisfied, stop watching
}).Await(t)

if !ok && tc.increasedDropCount {
	t.Errorf("Timeout reached waiting for drops to increase on counters: %v", tc.counters)
}
```

feature/system/control_plane_traffic/otg_tests/default_copp_test/default_copp_test.go
AmrNJ: We do not have a definitive number but 1s would be very frequent polling.

navaneethyv:
/gemini review

to check if we are getting those await comments from gemini

feature/system/control_plane_traffic/otg_tests/default_copp_test/default_copp_test.go
navaneethyv: ack. I think we can convert the whole thing into await.  Let us see if that helps.

feature/system/control_plane_traffic/otg_tests/default_copp_test/default_copp_test.go
AmrNJ: Needs an additional helper function extractDroppedPktCounts, I will locally verify and then commit the code.

feature/policy_forwarding/vrf_selection/otg_tests/protocol_dscp_rules_for_vrf_selection_test/protocol_dscp_rules_for_vrf_selection_test.go
```diff
@@ -518,12 +520,31 @@ func testTrafficFlows(t *testing.T, ate *ondatra.ATEDevice, top gosnappi.Config,
...
+				}).Await(t)
+				if ok {
+					inPktsRaw, _ := inPktsVal.Val()
+					inPkts = float32(inPktsRaw)
+				} else {
+					inPkts = float32(gnmi.Get(t, ate.OTG(), gnmi.OTG().Flow(flow.Name()).Counters().InPkts().State()))
+				}
+			} else {
+				inPkts = float32(gnmi.Get(t, ate.OTG(), gnmi.OTG().Flow(flow.Name()).Counters().InPkts().State()))
+			}
```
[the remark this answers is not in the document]
AmrNJ: Updated with code suggestion.

feature/system/control_plane_traffic/otg_tests/default_copp_test/default_copp_test.go
AmrNJ: Vendor native paths are not streaming with gnmi.Watch. I have asked for clarification from vendor.

AmrNJ: Added a TODO until the behavior is clarified

navaneethyv (approved):

rohit-rp (approved):

PR #5784, opened 2026-07-29
Fix metadata textproto and test registry

RishabhAgarwal-2001:


navaneethyv (approved):

navaneethyv:
Merging this without interface approvers as this is presubmit fix

PR #5799, opened 2026-07-29
Addition of a Shadow Go Static Check which is 75% faster than the existing Check Saving 35mins per run

RishabhAgarwal-2001:
We are adding a new check that does the exact same thing as the Go check but faster. Currently running it as a shadow with the original check for few weeks before maturing it to replace the existing check. 

image


.github/workflows/go-experimental-arm.yml
```diff
@@ -0,0 +1,141 @@
+name: Go (ARM64 Experimental)
```
navaneethyv, line 1: what will be the name of this?

RishabhAgarwal-2001: This is the name of the new check. I have explicitly written experimental to ensure people know that it can fail. 

Will replace it with the name of the original check once we mature it to replace the existing check.

RishabhAgarwal-2001: You can see this name in the checks list below.

.github/workflows/go-experimental-arm.yml
```diff
@@ -0,0 +1,141 @@
...
+  contents: read
+
+env:
+  XDG_CACHE_HOME: /mnt/cache
+  GOMODCACHE: /mnt/cache/go-mod
+
+jobs:
+  build:
+    name: Build (ARM64)
+    runs-on: ubuntu-24.04-arm
```
kapilrajput-g, line 24: By adding specific image version , it will require a maintenance step to keep updating to required new image.

RishabhAgarwal-2001: Currently, GitHub does not publish generic latest tag as it does for `x64` based runners. But since this is coming directly from GitHub, any updates or vulnerabilities in future would be easy to fix.

.github/workflows/go-experimental-arm.yml
```diff
@@ -0,0 +1,141 @@
...
+        done
+    - name: Get revive
+      run: go install github.com/mgechev/revive@v1.3.4
+    - name: Run revive
+      run: revive ./...
+    - name: Get staticcheck
+      run: go install honnef.co/go/tools/cmd/staticcheck@latest
+    - name: Cache staticcheck
+      uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
+      with:
```
kapilrajput-g, line 135: GitHub Actions shares a cache storage limit (typically 10GB) across the entire repository. If this experimental workflow uploads a new, heavy staticcheck cache archive on every single commit, it will rapidly fill up that 10GB quota. GitHub will respond by aggressively evicting older caches. This means your experimental workflow could actively delete the caches used by your required workflows (like go.yml), slowing down CI for the whole team.

Possible  Fix: Change the key to use something stable, like a hash of your go.mod file, so it only uploads a new cache when dependencies change.

.github/workflows/go-experimental-arm.yml
kapilrajput-g, line 1: To Monitor during experimental phase: Since you removed GOGC=30, watch the GitHub Actions logs for OOM (Out of Memory) kills. If the runner crashes silently or the static analysis steps fail randomly, memory spikes are the likely culprit.

kapilrajput-g (changes requested):

.github/workflows/go-experimental-arm.yml
RishabhAgarwal-2001: Yes right. I will keep this in head while tracking the shadow run over the weeks.

kapilrajput-g (approved):
Look good to me. As we discussed this check will run in trial mode and we will evaluate for some time for its correctness and stability of run.

PR #5827, opened 2026-08-05
Fix division by zero panics

AmrNJ:
Add a check if txPackets is 0 before dividing by Txpacket which will cause panic failures.


AmrNJ:
Gemini  review comments have been addressed. Remove duplicate calls in all applicable tests.

navaneethyv (approved):

rohit-rp (approved):

PR #5837, opened 2026-08-06
[Optimization] Improve DSCP Transparency test by reducing the number of flows

RishabhAgarwal-2001:
Changes in this PR are to reduce the number of flows used by the DSCP Transparency test. In its current state, the test used 64 flows which consumed a lot of resource on OTG causing it to take a very long time. In our runs, we noticed test running for around 2 hours and ultimately timing out!

With these changes, we use 7 flows on OTG, without impacting the efficacy or leniency of the test. This resulted in a gain of around 80% in performance with total runtime now down to around 25 minutes. 

README has also been updated to reflect the new reality of the implementation.


navaneethyv:
@RishabhAgarwal-2001 can you take care of canonical OC ?

sezhang2 (approved):

PR #5838, opened 2026-08-06
Update config.go

alokmtri-g:
Changes are made to make device selection flexible based on provided device label.

### DIFF
--- a/tools/ci-trigger/pr.go
+++ b/tools/ci-trigger/pr.go
@@ -160,7 +160,7 @@ func (p *pullRequest) createBuild(ctx context.Context, buildClient *cloudbuild.S
 		}
 	physicalDeviceLoop:
 		for _, physicalDevice := range p.Physical {
-			if physicalDevice.Type == d {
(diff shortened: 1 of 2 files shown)


tools/ci-trigger/pr.go
```diff
@@ -160,7 +160,7 @@ func (p *pullRequest) createBuild(ctx context.Context, buildClient *cloudbuild.S
 		}
 	physicalDeviceLoop:
 		for _, physicalDevice := range p.Physical {
-			if physicalDevice.Type == d {
+		        if physicalDevice.Type.Vendor == d.Vendor && (d.HardwareModel == "" || physicalDevice.Type.HardwareModel == d.HardwareModel) {
```
navaneethyv: why is this needed ?

alokmtri-g: This ensures that if a HardwareModel is not specified, the process does not fail. Instead, it gracefully defaults to verifying just the device vendor.

PR #5844, opened 2026-08-06
feature/sflow: fix KNE sFlow source-address to match capture filter

gauriwalinjkar91:
Align sflowCfgv{4,6}KNE.SrcAddr* with dutlo0Attrs (loopback), matching
the physical config path. Only the KNE code path changes; hardware
behavior is unchanged.

### DIFF
--- a/feature/sflow/otg_tests/sflow_base_test/sflow_base_test.go
+++ b/feature/sflow/otg_tests/sflow_base_test/sflow_base_test.go
@@ -79,14 +79,14 @@ var (
 	sflowCfgv4KNE = &cfgplugins.SFlowGlobalParams{
 		Ni:              mgmtVRF,
 		IntfName:        "port1",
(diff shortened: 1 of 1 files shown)


feature/sflow/otg_tests/sflow_base_test/sflow_base_test.go
```diff
@@ -79,14 +79,14 @@ var (
 	sflowCfgv4KNE = &cfgplugins.SFlowGlobalParams{
 		Ni:              mgmtVRF,
 		IntfName:        "port1",
-		SrcAddrV4:       "192.0.2.1",
+		SrcAddrV4:       "203.0.113.1",
```
navaneethyv, line 82: Thanks,

But why are these specific IP addresses important in this test cases and why are they needed? Can you add some comments here ?

PR #5851, opened 2026-08-08
telemetry: add min, max, and avg laser bias current OC paths to ZR and ZRP READMEs

singhavnish2516:
Add missing /avg, /min, and /max optical-channel laser-bias-current telemetry state paths to zr_laser_bias_current_test and zrp_laser_bias_current_test README files to match the Go test assertions.


ahsaanyousaf:
For laser bias current we do not require min max avg values and only rely on the instant values today. We do not need to add these until we have a path to enable these with our WBB vendors. Thanks

PR #5865, opened 2026-08-10
Add static route resiliency test

dharanitharanr-cell:
Add static route resiliency test file.


kapilrajput-g:
/fptest physical

Test failed on Cisco and Arista because lack of required setup avaiklability.

PR #5894, opened 2026-08-12
AFT-6.1/AFT-6.2 - Added support for other vendor 

Steubert:
AFT-6.1 & AFT-6.2 - Added support for other vendor and updated global filter policy path to use raw gnmi path as OC path is not available on latest ondatra

README: 
https://github.com/open-traffic-generator/featureprofiles/blob/main/feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering/README.md
https://github.com/openconfig/featureprofiles/blob/main/feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering_dualstack/README.md

Logs: 
https://partnerissuetracker.corp.google.com/u/1/issues/524449184
https://partnerissuetracker.corp.google.com/u/1/issues/517838946

Issue: https://partnerissuetracker.corp.google.com/u/1/issues/545820595


internal/aft_prefix_filtering/aft_prefix_filtering_service.go
```diff
@@ -349,10 +336,11 @@ type bgpNeighborSpec struct {
 // It applies the supplied router ID and single-AFI neighbors with the accept-all import/export policy.
 func configureBGPInstance(t *testing.T, ni *oc.NetworkInstance, routerID string, neighbors []bgpNeighborSpec) {
 	t.Helper()
+	dut := ondatra.DUT(t, "dut")
```
ashishpawar-google: Both callers of configureBGPInstance (ConfigureBGP and ConfigureScaleBGP) already receive dut *ondatra.DUTDevice as a parameter. Having the helper call ondatra.DUT(t, "dut") creates an implicit dependency on a hardcoded testbed reservation ID ("dut"), breaks test isolation, and makes testing multi-DUT or custom topologies impossible.

internal/aft_prefix_filtering/aft_prefix_filtering_service.go
```diff
@@ -349,10 +336,11 @@ type bgpNeighborSpec struct {
 // It applies the supplied router ID and single-AFI neighbors with the accept-all import/export policy.
 func configureBGPInstance(t *testing.T, ni *oc.NetworkInstance, routerID string, neighbors []bgpNeighborSpec) {
```
ashishpawar-google: Avoid Hardcoded Device Lookup in Helper Functions.

internal/aft_prefix_filtering/aft_prefix_filtering_service.go
```diff
@@ -540,6 +534,25 @@ func DialGNMI(t *testing.T, dut *ondatra.DUTDevice) (gpb.GNMIClient, error) {
 	return gnmiClient, gnmiErr
 }
 
+// globalFilterPolicyPath builds the raw gNMI path for an AFT global-filter
+// policy leaf ("ipv4-policy" or "ipv6-policy") on the given network-instance.
+// The afts/global-filter augment (openconfig-aft-global-filter.yang, added in
+// openconfig/public >= 3.3.0) is not yet present in the ondatra `oc` bindings,
+// so the path is constructed manually.
```
ashishpawar-google, line 540: While raw gNMI is the correct unblocking mechanism until Ondatra regenerates its oc package against openconfig/public >= 3.3.0, add a formal TODO comment so this tech debt is easily trackable when Ondatra updates.

ashishpawar-google (changes requested):

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering_dualstack/metadata.textproto
```diff
@@ -16,5 +16,18 @@ platform_exceptions: {
     afts_global_filter_policy_oc_unsupported: true
     bgp_missing_oc_max_prefixes_configuration: true
     link_local_instead_of_nh: true
+    interface_enabled: true
+    default_bgp_instance_name: "BGP"
```
romeyod, line 20: Dont we need to add this in AFT-6.1 and others?

romeyod (changes requested):
Dont we need to add defaultV4Route in baseIPv4Prefixes and defaultV6Route in baseIPv6Prefixes? same for other AFT-6.XX tests

Steubert:
> Dont we need to add defaultV4Route in baseIPv4Prefixes and defaultV6Route in baseIPv6Prefixes? same for other AFT-6.XX tests

We can't add the default routes to baseIPv4Prefixes/baseIPv6Prefixes because ConfigureBaseRoutes installs each prefix as a static route with a known ATE next-hop — but the default route's next-hop isn't known to the test. That's why the default routes are treated as a precondition and validated via CheckForDefaultRoutes instead of being configured.

The default routes are still added to the global filter, though — they're members of PREFIX-SET-A/PREFIX-SET-B (via pfxSetAMembers/pfxSetBMembers), which back POLICY-PREFIX-SET-A/POLICY-PREFIX-SET-B attached to afts/global-filter. They're also included in v4MatchPrefixes/v6MatchPrefixes (and thus in the expected wantPrefixes for InitialSyncStoppingCondition), so filter matching is still verified against them. Same applies to the other AFT-6.x tests.

feature/afts/filtered_streaming/otg_tests/afts_prefix_filtering_dualstack/metadata.textproto
Steubert: > Dont we need to add this in AFT-6.1 and others?

I'm checking this. Will update the script accordingly

Steubert: Added platform exceptions for the tests that uses the respective deviations

romeyod:
> > Dont we need to add defaultV4Route in baseIPv4Prefixes and defaultV6Route in baseIPv6Prefixes? same for other AFT-6.XX tests
> 
> We can't add the default routes to baseIPv4Prefixes/baseIPv6Prefixes because ConfigureBaseRoutes installs each prefix as a static route with a known ATE next-hop — but the default route's next-hop isn't known to the test. That's why the default routes are treated as a precondition and validated via CheckForDefaultRoutes instead of being configured.
> 
> The default routes are still added to the global filter, though — they're members of PREFIX-SET-A/PREFIX-SET-B (via pfxSetAMembers/pfxSetBMembers), which back POLICY-PREFIX-SET-A/POLICY-PREFIX-SET-B attached to afts/global-filter. They're also included in v4MatchPrefixes/v6MatchPrefixes (and thus in the expected wantPrefixes for InitialSyncStoppingCondition), so filter matching is still verified against them. Same applies to the other AFT-6.x tests.

Maybe I am missing something
Without the change I mentioned, the test configures only the non-default static routes. But it still:

  1. Includes 0.0.0.0/0 and ::/0 in PREFIX-SET-A / PREFIX-SET-B.
  2. Applies those prefix-set policies as the AFT global filters.
  3. Includes those defaults in matchPrefixes / wantPrefixes.
  4. Waits for InitialSyncStoppingCondition to receive every expected prefix.

  So with no independently preconfigured default routes, no default AFT entries exist to
  stream. The collector sees the other allowed routes, but remains short by one expected
  default prefix

Steubert:
> Maybe I am missing something Without the change I mentioned, the test configures only the non-default static routes. But it still:
> 
> 1. Includes 0

Updated the script to conditionally add either 0.0.0.0/0 or ::/0 to the global filter. The collector now expects default routes only when they already exist. We are avoiding configuring default routes in the tests because the gateway IP is unknown, and we do not want to risk losing management connectivity due to an incorrect gateway IP configuration.

PR #5895, opened 2026-08-12
Optimize one_sp_queue_traffic_test by replacing static sleep 

mansi0803:
[Test Optimization] Optimize one_sp_queue_traffic_test by replacing static sleep with telemetry watch

- Removed the redundant 30-second static sleep after stopping traffic.
- Added ExpectedTrafficLoss helper to dynamically watch OTG flow packet counters with gnmi.Watch.
- Updated imports to include otgtelemetry and ondatra otg packages.

### DIFF
--- a/feature/qos/otg_tests/one_sp_queue_traffic_test/one_sp_queue_traffic_test.go
+++ b/feature/qos/otg_tests/one_sp_queue_traffic_test/one_sp_queue_traffic_test.go
@@ -701,26 +701,21 @@ func TestOneSPQueueTraffic(t *testing.T) {
 			ate.OTG().StartTraffic(t)
 			time.Sleep(30 * time.Second)
 			ate.OTG().StopTraffic(t)
-			time.Sleep(30 * time.Second)
 
 			otgutils.LogFlowMetrics(t, ate.OTG(), top)
 			for trafficID, data := range trafficFlows {
+				expectedLossPct := 100.0 - data.expectedThroughputPct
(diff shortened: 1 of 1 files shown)


feature/qos/otg_tests/one_sp_queue_traffic_test/one_sp_queue_traffic_test.go
```diff
@@ -744,6 +742,49 @@ func TestOneSPQueueTraffic(t *testing.T) {
 	}
 }
 
+// ExpectedTrafficLoss watches an OTG flow until the loss percentage is within the expected range or times out, then asserts bounds.
+func ExpectedTrafficLoss(t testing.TB, otg *otg.OTG, flowName string, minLossPct, maxLossPct float32) {
```
AmrNJ: You can just call in the function from otgutils.

mansi0803: I tried using otgutils.GetFlowLossPct, but under the hood GetFlowStats waits for rxPackets == txPkts. Since this test expects drops on oversubscribed queues, it ends up hitting the full 30s timeout on every flow and adds ~5 mins to the run. But in our function ExpectedTrafficLoss, I avoids that by checking the expected loss range directly which helps in reducing the run time to around 5-6 min per sku

navaneethyv (approved):

rohit-rp (approved):

PR #5898, opened 2026-08-13
acctz-2.1: set gRIBI,P4RT aaa role based authorization unsupported deviation to true

mudhireddy:
set aaa role based authz unsupported deviation for gRIBI, P4RT for ACCTZ-2.1 test as well similar to the ACCTZ-1.1 fix


AmrNJ (approved):
LGTM.

PR #5923, opened 2026-08-18
Adding BGP Drain Policy Test

rohit-rp:
# RT-7.12: BGP Drain using Route Policy check b/540868197 for more info

* **Objective:** Test draining a BGP peering by applying an outbound policy to stop prefix advertisements.  
  * **Interfaces Used:** gNMI, BGP, Traffic Generator  
  * **Expected Outcome Summary:**  
    * Initially, prefixes are advertised and learned.  
    * Apply a "deny all" outbound route policy to the BGP neighbor via gNMI.  
    * The DUT withdraws the prefixes from the peer.  
    * Traffic to those prefixes from the peer's perspective now uses alternative paths or stops.  
    * Removing the policy restores advertisements and traffic.  
  * **Existing Coverage:**  
    * **None**

## 1\. Scale Enhancements

* **Prefix Scale:** Increase the number of advertised prefixes in the environment setup from 100 to a more robust scale, such as 10,000 for both IPv4 and IPv6, to better stress the control plane during the policy application.  
* **Peer Group Scale (Optional but recommended):** Consider adding a note or future test case for applying the drain policy at the BGP `peer-group` level to validate scaling when draining multiple peers simultaneously.

## 2\. Traffic Validation

To prevent flaky tests and false positives, discrete traffic steps should be replaced with continuous traffic streams. **Enhancements:**

* **Continuous Traffic:** Start continuous, background IPv4 and IPv6 traffic streams in RT-7.12.1.  
* **RT-7.12.2 Validation:** Instead of stopping/starting traffic, validate that while traffic is running, the application of the policy causes the Rx rate to drop to 0 at ATE port 2 (100% loss). Additionally, verify if the ATE captures BGP UPDATE (Withdraw) messages for the drained prefixes to confirm DUT behavior.  
* **RT-7.12.3 Validation:** Validate that removing the policy causes the Rx rate at ATE port 2 to recover to match the Tx rate (0% loss) without restarting the traffic streams.

## 3\. High-Priority Negative/Edge Test Cases

* **Proposed Subtest RT-7.12.4: AFI/SAFI Specific Policy Application**  
  * *Scenario:* Apply the "DENY\_ALL" export policy only to the `IPV4_UNICAST` afi-safi, leaving `IPV6_UNICAST` untouched.  
  * *Validation:* Verify IPv4 traffic drops to 100% loss, while IPv6 traffic continues with 0% loss. This validates granular policy control.  
* **Proposed Subtest RT-7.12.5: Application of Non-Existent Policy**  
  * *Scenario:* Attempt to apply an export policy named "NON\_EXISTENT\_POLICY" (which has not been defined in `/routing-policy/policy-definitions`) to the BGP neighbor.  
  * *Validation:* Verify that the gNMI Set operation fails with an appropriate error (e.g., Invalid Argument) and that the BGP session and existing routing state remain unaffected.


AmrNJ (approved):
LGTM.

RishabhAgarwal-2001 (approved):

feature/bgp/policybase/otg_tests/bgp_drain_policy_test/README.md
```diff
@@ -0,0 +1,221 @@
...
+  /network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/afi-safis/afi-safi/state/prefixes/received:
+  /network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/afi-safis/afi-safi/state/prefixes/sent:
+
+rpcs:
+  gnmi:
+    gNMI.Set:
+      union_replace: true
+      delete: true
+    gNMI.Get:
+      subscribes: true
```
kapilrajput-g: Seems Minor metadata typo here(Please confirm if this is not the case): it looks like `subscribes: true` is accidentally listed under `gNMI.Get`. Since subscriptions are handled by `gNMI.Subscribe` (which you already have correctly listed just below this), Please check if need to remove `subscribes: true` line from the `gNMI.Get` block to avoid any confusion.

rohit-rp: fixed

kapilrajput-g (approved):

navaneethyv:
@RishabhAgarwal-2001  what are your thoughts on the time addition due to this ? 

@ksgireesha  can you please ensure that the system tests have a marker for Rishabh to help review too ?

ksgireesha:
> @RishabhAgarwal-2001 what are your thoughts on the time addition due to this ?
> 
> @ksgireesha can you please ensure that the system tests have a marker for Rishabh to help review too ?

Hi @navaneethyv, Sure. I will ensure we apply the `systemTest` label to these PRs and tag @RishabhAgarwal-2001 as a reviewer so they are easy for him to track and review.

PR #5925, opened 2026-08-19
Remove static sleep from the LLDP Packetout Test

RishabhAgarwal-2001:


AmrNJ (approved):
LGTM

feature/p4rt/otg_tests/lldp_packetout_test/lldp_packetout_test.go
```diff
@@ -164,7 +165,14 @@ func testPacketOut(ctx context.Context, t *testing.T, args *testArgs) {
...
 
 			// Wait for ate stats to be populated
-			time.Sleep(60 * time.Second)
+			if test.expectPass {
+				gnmi.Watch(t, args.ate.OTG(), gnmi.OTG().Port(port).Counters().InFrames().State(), time.Minute, func(val *ygnmi.Value[uint64]) bool {
+					count, present := val.Val()
+					return present && count >= counter0 && (count-counter0 >= uint64(float64(packetCount)*0.95))
+				}).Await(t)
+			} else {
+				time.Sleep(10 * time.Second)
```
kapilrajput-g, line 174: Great improvement! Replacing the static sleep with gnmi.Watch will save a lot of execution time on successful runs.

I have one concern regarding the else block for the negative case (test.expectPass == false). The static sleep was reduced from 60 seconds to 10 seconds. While this speeds up the negative test, it increases the risk of a false positive if the device happens to errantly forward the packets after a delay (e.g., 15 seconds). Is there a specific reason we are confident that 10 seconds is a sufficient upper bound to wait for a packet drop? If not, we might want to keep the sleep at 60s for the else case to maintain the rigor of the negative test.

RishabhAgarwal-2001: Right now the test does not have any case for `expectPass == false`. Thus, this should not be an issue

kapilrajput-g (approved):

PR #5929, opened 2026-08-19
system: Restore initial DUT hostname and domain name in hostname_test

praveen-puli:
In TestHostname and TestDomainName, gnmi.Delete was previously called without restoring the DUT's initial hostname or domain name in a cleanup hook. On Arista EOS devices, deleting the hostname reverts the CLI prompt to "localhost#". When subsequent tests in a group reservation execute CLI commands (e.g. "show version" in Ondatra setup), the CLI regex matcher hung waiting for "<dut_name>#", causing a 630s timeout, aborted sessions, and orphaned Caelum config locks on the DUT.

This change captures the initial hostname and domain name (falling back to dut.Name()) and uses t.Cleanup to reliably restore them when the test completes.


AmrNJ (approved):
LGTM.

PR #5941, opened 2026-08-21
Bursty Traffic failure of AF4 queue

rangababuKS:
https://partnerissuetracker.corp.google.com/issues/548329845

on every iteration of the test the config is pushed before stopping the protocols.
As the issue seen in the above-mentioned tracker is intermittent, adding the Stop protocols and sleep before proceeding to next test iteration.


feature/qos/otg_tests/bursty_traffic_test/bursty_traffic_test.go
```diff
@@ -486,6 +486,8 @@ func TestBurstyTraffic(t *testing.T) {
 				}
 
 			}
+			ate.OTG().StopProtocols(t)
+			time.Sleep(30 * time.Second)
```
praveen-puli: It is not recommended to use hardcoded sleeps in tests. Is there an event-based waiting mechanism we can use instead?

praveen-puli (changes requested):

feature/qos/otg_tests/bursty_traffic_test/bursty_traffic_test.go
rangababuKS: I see some of the existing code has hard coded wait. that is when I had added it.
As far as I know event-based (something like hooks) waiting is not available. But using get_states we can do some poll-based waiting that too on ipv4_neighbors.

octpetre: Instead of sleep you can implement a reverse waitForARP where you make sure that the link-layer-address is missing / gone since upon stop_protocols that address is removed.

rangababuKS: Understood.
I am going to propose new changes, right now the configuration of flows and start and stop protocols are happening on every test iteration, instead I would move the config to top and only do start and stop traffic on each test. That will help test optimization. Also will add gnmi.Await for the wait instead of hardcoded sleeps

praveen-puli: Since we already have a sleep in this test, I think we can proceed with this in the current PR and raise a separate one to fix all other sleeps.

CC: @AmrNJ

praveen-puli (approved):

feature/qos/otg_tests/bursty_traffic_test/bursty_traffic_test.go
AmrNJ: We are no longer adding static sleeps instead you can consider adding stabilization functions like otgutils.ExpectedTrafficLoss or cfgplugins.VerifyRoutes before starting traffic.

AmrNJ: @rangababuKS Your suggested changes are good. Please proceed.

praveen-puli: Thanks @AmrNJ for sharing the best practice. @rangababuKS please do the changes accordingly!

rangababuKS:
@praveen-puli @AmrNJ 
I have updated the code with the proposed changes. please review it.
In our Arista testbed I am seeing the below error with respect to DUT (even before test script modification), let me know if I am missing something.
**"bursty_traffic_test.go:419: TransmitPkts count for queue "Ethernet4/1/1" on interface "NC1" not available within 1m0s"**

I have tested the code with B2B and with DUT in between, the config and start traffic and stop traffic on every test iteration is working fine.

below path isn't returning Tx and Rx packets
` gnmic -a 10.36.111.158:6030 -u admin -p admin --insecure get --path /qos/interfaces/interface[interface-id=Ethernet4/1/1]/output/queues/queue[name=NC1]
[
  {
    "source": "10.36.111.158:6030",
    "timestamp": 1787656629164982081,
    "time": "2026-08-25T11:17:09.164982081Z",
    "updates": [
      {
        "Path": "qos/interfaces/interface[interface-id=Ethernet4/1/1]/output/queues/queue[name=NC1]",
        "values": {
          "qos/interfaces/interface/output/queues/queue": {
            "openconfig-qos:config": {
              "name": "NC1",
              "queue-management-profile": ""
            },
            "openconfig-qos:name": "NC1"
          }
        }
      }
    ]
  }
]
`

rangababuKS:
@AmrNJ Could you please run the updated script in Google env before approving for the Merge

navaneethyv:
@rangababuKS is this tested ?

navaneethyv (approved):

praveen-puli (approved):
LGTM

rangababuKS:
> @rangababuKS is this tested ?

@navaneethyv Tested in 2 ways 1 is B2B where I had disabled the dut config and tests only Ixia code. With DUT I was seeing the below error where fetching the stats of DUT gives value not found, Which was mentioned in the pervious conversation on this PR.

`bursty_traffic_test.go:472: TransmitPkts count for queue "NC1" on interface "Ethernet3/1/1" failed to reach expected value
    bursty_traffic_test.go:474: Get(t) on dut(10.36.111.158) at /qos/interfaces/interface[interface-id=Ethernet3/1/1]/output/queues/queue[name=NC1]/state/transmit-pkts: path origin:"openconfig" elem:{name:"qos"} elem:{name:"interfaces"} elem:{name:"interface" key:{key:"interface-id" value:"Ethernet3/1/1"}} elem:{name:"output"} elem:{name:"queues"} elem:{name:"queue" key:{key:"name" value:"NC1"}} elem:{name:"state"} elem:{name:"transmit-pkts"}: value not present`
 

I had also cross checked with the qos_basic_test where the OC stats check was skipped via deviation. But from the logs provide in the bug [](https://partnerissuetracker.corp.google.com/issues/548329845) its contradicting as there is no deviation set. Also I had matched the Arista image version from the logs provided in bug.

Please validate from your end lets me know

PR #5977, opened 2026-09-02
Skip applying decap policy on interface when oc unsupported

ijernoiu:


internal/cfgplugins/policyforwarding.go
```diff
@@ -451,7 +451,6 @@ func InterfacePolicyForwardingConfig(t *testing.T, dut *ondatra.DUTDevice, a *at
 		}
 	} else {
 		ApplyPolicyToInterfaceOC(t, pf, params.InterfaceID, params.AppliedPolicyName)
```
AmrNJ, line 453: Remove this file as there are no changes.

feature/policy_forwarding/otg_tests/mpls_gre_udp_decap_test/mpls_gre_udp_decap_test.go
```diff
@@ -317,6 +317,9 @@ func configureDUT(t *testing.T, dut *ondatra.DUTDevice) {
 }
 
 func configureInputPolicy(t *testing.T, dut *ondatra.DUTDevice, tc testCase) {
+	if deviations.InterfacePolicyForwardingOCUnsupported(dut) {
```
AmrNJ, line 320: I noticed that there is no issue tracker with Arista for this behaviour. Please raise one and tag it in deviations.go file.

PR #6002, opened 2026-09-08
Add prefix-length state path to gnmi_set_test README.md

singhavnish2516:
Update the OpenConfig Path Coverage section in `feature/gnmi/tests/gnmi_set_test/README.md` to document the IPv4 subinterface prefix-length state path verified in `verifyInterface` (`gnmi_set_test.go`).

OpenConfig paths covered:
- `/interfaces/interface/subinterfaces/subinterface/ipv4/addresses/address/state/prefix-length`

### DIFF
--- a/feature/gnmi/tests/gnmi_set_test/README.md
+++ b/feature/gnmi/tests/gnmi_set_test/README.md
@@ -203,6 +203,7 @@ This test checks that the static protocol name is usable.
 paths:
   /interfaces/interface/name:
   /interfaces/interface/subinterfaces/subinterface/ipv4/addresses/address/ip:
+  /interfaces/interface/subinterfaces/subinterface/ipv4/addresses/address/state/prefix-length:
   /network-instances/network-instance/name:
 rpcs:
   gnmi:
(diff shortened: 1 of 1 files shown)


feature/gnmi/tests/gnmi_set_test/README.md
```diff
@@ -197,12 +197,84 @@ This test checks that the static protocol name is usable.
...
+}
+```
+
 ## OpenConfig Path and RPC Coverage
 
 ```yaml
 paths:
   /interfaces/interface/name:
   /interfaces/interface/subinterfaces/subinterface/ipv4/addresses/address/ip:
+  /interfaces/interface/subinterfaces/subinterface/ipv4/addresses/address/state/prefix-length:
```
kapilrajput-g, line 277: One minor request for the `## OpenConfig Path and RPC Coverage` section: you added the `state` path for `prefix-length`, but since this is a gNMI Set test, we should also list the `config` path to fully reflect the test's coverage.

Could you add the `config` path?

singhavnish2516: Ideally, we don't want to pollute the PR as the program that we are working on is related to Telemetry. The Canonical OC section had to be added to fix the failing pre-submit check. I hope that's ok as we don't want to claim support for non-telemetry paths which we don't own.

kapilrajput-g (approved):

PR #6021, opened 2026-09-10
updating platform details

cprabha:
Hi,

Added vendor specific details in metadata.textproto.


navaneethyv:
can you clarify what is this change and why is this needed ?

navaneethyv (changes requested):
please clarify why is this needed?

PR #6036, opened 2026-09-14
Polarization FNT

aaronmillisor:
The intention of this PR is to quantify the behavior of a series of devices processing the same flow. 

Hashing algorithms such as CRC are linear such that given the same input values and the same number of outputs the hasher result distribution will be the same.  

Absent a cryptographic hashing algorithm that can decorrelate using a simple seed, mitigation of this is typically handled using a variety of hasher knobs such as modifying polynomials, or altering the input parameters used for hashing on sequential devices.  This test quantifies the efficacy of the DUT hashing configuration in the context of sequential evaluation.

Each vendor may expose the management of these configuration knobs differently, this PR initially implements the test using perturbation of the device knobs using the router-id (Loopback IP address) which is what Cisco's IOS XR uses for automatic handling of hasher configurations. Other vendors may use a different mechanism for hasher perturbation (manual, chassis mac address, etc), so the specific per-vendor hook is pluggable.

The procedure for the test at a high level is:
1. Pass a random set of flows through a recursive, non-recursive, and LAG forwarding stage.
2. Measure that the ratio of traffic on each of the output interfaces is within tolerance based upon an ideal prediction
3. Capture the flows on a single one of the LAG members via OTG, discard all other flows
4. Modify the hasher configuration to behave as if the DUT is a 'different' router
5. Repeat the flows captured in step 3 through the DUT
6. Repeat to desired amount

This has been tested with both real hardware as well as using ixia-c in a virtual environment. The main challenge with executing this test is related to execution time. Two things cause this test to take awhile.

- The maximum number of flows that you can manually configure to send via OTG
- The maximum size of a PCAP you can receive via OTG
- Since we are constantly splitting the flows into a fraction of the original, even with a large initial burst of distinct flows the number of flows available to the hasher reduces to statistical noise after several iterations. 

If the OTG limitations above were improved the test execution could go much faster and could also lead to longer meaningful simulated device chains.

If the procedure sketched out in this PR is agreed to, the tests can be easily expanded to account for additional address families and encapsulations.


feature/gribi/otg_tests/polarization-ipv4/metadata.textproto
```diff
@@ -0,0 +1,16 @@
+# proto-file: github.com/openconfig/featureprofiles/proto/metadata.proto
+# proto-message: Metadata
+
+uuid: "07341742-d9d9-4f32-b881-9c5b43083bd2"
+plan_id: "LB-1.1"
+description: "ECMP Hashing Polarization Detection"
```
gggsmith, line 6: nit: wECMP

feature/gribi/otg_tests/polarization-ipv4/polarization_test.go
```diff
@@ -0,0 +1,986 @@
...
+	}
+	lag2.GetOrCreateAggregation().LagType = oc.IfAggregate_AggregationType_STATIC
+	s2 := lag2.GetOrCreateSubinterface(0)
+	s2.Index = ygot.Uint32(0)
+	a2v4 := s2.GetOrCreateIpv4().GetOrCreateAddress("198.19.2.1")
+	a2v4.Ip = ygot.String("198.19.2.1")
+	a2v4.PrefixLength = ygot.Uint8(plen24)
+	s2.GetOrCreateIpv4().GetOrCreateNeighbor("198.19.2.2").LinkLayerAddress = ygot.String("02:11:22:33:44:55")
+	s2.GetOrCreateIpv4().GetOrCreateNeighbor("198.19.2.3").LinkLayerAddress = ygot.String("03:11:22:33:44:55")
+	s2.GetOrCreateIpv4().GetOrCreateNeighbor("198.19.2.24").LinkLayerAddress = ygot.String("03:11:22:33:44:55")
```
gggsmith, line 807: Is that a mistake by chance - is MAC address supposed to be duplicate for  `198.19.2.3` and `198.19.2.24` ?

feature/gribi/otg_tests/polarization-ipv4/polarization_test.go
```diff
@@ -0,0 +1,986 @@
...
+	// tolerancePct is the acceptable deviation from expected distribution.
+	// Set high enough to accommodate natural hash variance on small replay sets.
+	// expressed as a percentage of total packets.
+	tolerancePct = 2
+
+	// lossTolerancePct bounds end-to-end traffic loss. The gRIBI forwarding tree
+	// spans both LAGs, so any meaningful loss means the DUT dropped forwarded
+	// traffic on one of ports2-5. Kept tight because LAG/ECMP forwarding of the
+	// installed routes is expected to be lossless.
+	lossTolerancePct = 0.5
```
gggsmith, line 70: For strict tolerance I think we should go lower to 0.1-0.2%

feature/gribi/otg_tests/polarization_ipv4/README.md
```diff
@@ -0,0 +1,168 @@
...
+    |
+    +--[1/9]--> NH 1012 --> 203.0.113.2 --> NHG 3000 (weight 1:1)
+                 +--[1/2]--> NH 1602 --> LAG2 (port4, port5)
+                 +--[1/2]--> NH 1603 --> LAG2 (port4, port5)
+    ```
+
+    All NHG weights are defined as named constants in the test for easy
+    tuning. The expected per-port distribution is computed automatically
+    from these weights. With the default weights, port2 receives ~39.5%
+    of traffic.
```
gggsmith, line 56: I might have missed some discussions here, but why are port 1, port3-4 left in a blind spot (I presume verification of polarization state for them should be straightforward to implement)?

gggsmith (commented):
Thank you for implementing this, @aaronmillisor! 

I will wait for review from singhh07 (as he is working on hashing verification) before approving.

feature/gribi/otg_tests/polarization_ipv4/polarization_test.go
```diff
@@ -0,0 +1,986 @@
...
+	}
+	lag2.GetOrCreateAggregation().LagType = oc.IfAggregate_AggregationType_STATIC
+	s2 := lag2.GetOrCreateSubinterface(0)
+	s2.Index = ygot.Uint32(0)
+	a2v4 := s2.GetOrCreateIpv4().GetOrCreateAddress("198.19.2.1")
+	a2v4.Ip = ygot.String("198.19.2.1")
+	a2v4.PrefixLength = ygot.Uint8(plen24)
+	s2.GetOrCreateIpv4().GetOrCreateNeighbor("198.19.2.2").LinkLayerAddress = ygot.String("02:11:22:33:44:55")
+	s2.GetOrCreateIpv4().GetOrCreateNeighbor("198.19.2.3").LinkLayerAddress = ygot.String("03:11:22:33:44:55")
+	s2.GetOrCreateIpv4().GetOrCreateNeighbor("198.19.2.24").LinkLayerAddress = ygot.String("03:11:22:33:44:55")
```
aaronmillisor: I did this to avoid having to deal with ARP/ND. All I'm doing is catching the packets on one of the ports so arp is unnecessary and I just nailed it up. It seems like the preference may be for using dynamic arp/nd which I can migrate to and eliminate the static configuration altogether.

feature/gribi/otg_tests/polarization_ipv4/polarization_test.go
```diff
@@ -0,0 +1,986 @@
...
+	// tolerancePct is the acceptable deviation from expected distribution.
+	// Set high enough to accommodate natural hash variance on small replay sets.
+	// expressed as a percentage of total packets.
+	tolerancePct = 2
+
+	// lossTolerancePct bounds end-to-end traffic loss. The gRIBI forwarding tree
+	// spans both LAGs, so any meaningful loss means the DUT dropped forwarded
+	// traffic on one of ports2-5. Kept tight because LAG/ECMP forwarding of the
+	// installed routes is expected to be lossless.
+	lossTolerancePct = 0.5
```
aaronmillisor: Hashing algorithms are not likely to pass at 0.1 or 0.2% unless you have a huge number of flows at each iteration. If you mean 1% or 2% that is possible but still will require many more flows. 5% is more practical in terms of real implementations.

feature/gribi/otg_tests/polarization_ipv4/README.md
aaronmillisor: Yes it looks a little weird, it's due to an attempt to minimize the number of interfaces necessary to execute the test. The flows are not symmetrical and I am 'dumping' multiple branches of the flows at various levels to the other sink ports. This makes measuring for polarization impossible because it is commingled except for the one measured port.  If we are willing to burn a larger number of ports on the TGEN, or else use a two-DUT configuration (to make the ports in use easier to source than a TGEN) we could probably measure on all the ports for polarization (still would only want to capture on one port thought).

Here's a picture in case it helps visualize.

```
ATE port1
    |
    v
 198.51.100.0/24
    |
    v
 NHG 101  (8:1)
    |
    |-- 8/9 (~88.9%) --> NH 1011 --> 203.0.113.1 --> NHG 2010 (8:1)
    |                        |
    |                        |-- 8/9 of 88.9% = ~79.0% --> NH 1501 --> LAG1
    |                        |                                    |
    |                        |                          +---------+---------+
    |                        |                          |                   |
    |                        |                        port2              port3
    |                        |                       ~39.5%             ~39.5%
    |                        |                    (capture+assert)     (same path,
    |                        |                                         not asserted)
    |                        |
    |                        `-- 1/9 of 88.9% = ~9.9% --> NH 1601 ----+
    |                                                                  |
    `-- 1/9 (~11.1%) --> NH 1012 --> 203.0.113.2 --> NHG 3000 (1:1)    |
                                 |                                     |
                                 |-- 1/2 of 11.1% = ~5.6% --> NH 1602 -+
                                 |                                     |
                                 `-- 1/2 of 11.1% = ~5.6% --> NH 1603 -+
                                                                       |
                                                                       v
                                                              LAG2 (mixed bucket)
                                                          three recursive NHs
                                                          land on the same bundle
                                                          (~9.9% + ~5.6% + ~5.6%)
                                                                       |
                                                             +---------+---------+
                                                             |                   |
                                                           port4               port5
                                                          ~10.5%              ~10.5%

```

feature/gribi/otg_tests/polarization_ipv4/README.md
```diff
@@ -0,0 +1,168 @@
+# LB-1.1: Load Balancing Hashing Polarization Detection
```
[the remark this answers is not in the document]
aaronmillisor: Fixed.

feature/gribi/otg_tests/polarization_ipv4/polarization_test.go
aaronmillisor: Actually I misread the PR comment, this is about loss tolerance not imbalance tolerance. I can modify this.

singhh07: even for imbalance, it shouldn't be more than 2%.

PR #6051, opened 2026-09-16
Enable Arista hierarchical FIB in BGP and ISIS scale tests

mageshkasi-byte:
Call cfgplugins.FeatureHierarchicalFIB from the BGP scale test (RT-1.65) and the shared ISIS scale library used by the IS-IS multi-adjacency scale test (RT-2.18), so both tests exercise the same hierarchical FEC resolution used in production.

Without hierarchical FEC resolution the DUT resolves next hops through a flat FEC hierarchy and exhausts the level-2 FEC table under next-hop-group churn at scale. The profile is gated inside NewDUTHardwareInit, which returns an empty string for non-Arista vendors and for cEOS, so behavior on other platforms is unchanged.


singhavnish2516 (approved):
Go approval

rszarecki (approved):

dplore (approved):

PR #6065, opened 2026-09-19
Fixed get failure handling at reboot delay boundary

mnp-juniper:
Issue :
In the test case reboot with delay the script is expecting the current date-time value to be greater than previous date-time value.But at the at the boundary of the robot delay the get fails and that erroneously assigned  a value 
0001-01-01 00:00:00 +0000 UTC  to the current date-time and the comparison 
"the current date-time value to be greater than the previous date-time captured” fails that results in script failure .

The fix:
  Handle the get failure at the reboot delay boundary with a log message and break the loop to proceed with the  the reboot bypassing the the current date-time and previous date-time comparison .


feature/gnoi/system/tests/complete_chassis_reboot/complete_chassis_reboot_test.go
```diff
@@ -153,18 +153,27 @@ func TestChassisReboot(t *testing.T) {
...
-					if err != nil && time.Since(start).Seconds() < rebootDelay.Seconds() {
-						t.Fatalf("Failed parsing current-datetime: %s.", err)
+					// A failed Get can legitimately happen right at the delay boundary if the
+					// DUT has already started rebooting. Only treat it as a hard failure if it
+					// happens well before the delay has elapsed; otherwise stop validating
+					// reachability here and let the post-reboot polling loop below take over.
+					if errMsg != nil {
+						// Allow a 5-second tolerance before the reboot delay to account for
+						// slight timing variations or early reboot initiation by the DUT.
+						if time.Since(start) < rebootDelay-5*time.Second {
```
navaneethyv: what is the rationale here?

